PR #2, opened 2021-09-02
Replaces this repo with new-snarkyjs from monorepo

bkase:
This is a port from the monorepo with a GitHub action for CI

For reviewers -- you can focus on the "adds github action" commit, the other commits are just removing files and adding files from the monorepo


jasongitmail (commented):
Some thoughts:
- `on: [push, pull_request]` would allow this to run when a PR is proposed too
- Should we run tests in this action too?
- Now that this isn't on the monorepo, Yarn isn't needed. What do you think about using NPM for consistency? It's NBD either way. But since NPM caught up & exceeded Yarn in speed, I'm not sure if there is any advantage to remaining with Yarn.

bkase:
We should run tests when they pass (but currently they don't 😄 , this should be fixed when @mrmr1993 handles the node side)

jasongitmail (approved):

.github/workflows/build-action.yml
```diff
@@ -14,7 +14,6 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '14'
-        cache: 'yarn'
     - run: npm install
```
jasongitmail, line 17: We should use `npm ci` instead of `npm install`. It will install the versions shown in `package-lock.json` (for consistency) & will be faster.

PR #7, opened 2021-10-04
Restructure Project Repo

MartinMinkov:
**Description**
This PR restructures the project directory to use sub-packages for both `snarkyjs-web` and `snarkyjs-node`. The following structure looks like this:
```
.
└── packages
    ├── snarky-web
    │   ├── snarky.js
    |   ├── chrome_bindings/
    |   ├── lib/
    │   ├── node_modules
    │   └── package.json
    ├── snarky-node
    │   ├── snarky.js
    |   ├── node_bindings/
    |   ├── lib/
    │   ├── node_modules
    │   └── package.json
...
```
A symlink is used to replicate files that are needed in both packages. This allows us to have separate packages for releasing different NPM modules. Later on, we can incorporate `lerna` to automate this process. 

**Tested**
Manually tested each package to have correct behavior


jasongitmail:
1. I'd suggest using `index.ts` or `index.js` for each of the entry filenames (instead of snarky.js), to be intuitive & not mix what we're referring to as Snarky.
2. I'd probably add an `src` directory in each containing `index.ts`, `lib`, & the bindings.
3. But my main concern is this may not work on Windows due to sym links. We should ensure Windows support is maintained.

matheus23:
Hi :wave: 

Very random comment on a foreign pull request, I know. I'm watching this repo due to personal interest.
Have you considered making a single package that is node & web compatible at the same time by using [`export map`](https://nodejs.org/api/packages.html#packages_package_entry_points)s (& see [webpack docs](https://webpack.js.org/guides/package-exports/)), similar to e.g. how [js-ipfs](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-utils/package.json#L89-L96) is doing this?

If you have, but just need more flexibility / better bundler support / anything, please ignore me :) I don't want to run around and annoy people with whatabouts.

MartinMinkov:
> Hi wave
> 
> Very random comment on a foreign pull request, I know. I'm watching this repo due to personal interest. Have you considered making a single package that is node & web compatible at the same time by using [`export map`](https://nodejs.org/api/packages.html#packages_package_entry_points)s (& see [webpack docs](https://webpack.js.org/guides/package-exports/)), similar to e.g. how [js-ipfs](https://github.com/ipfs/js-ipfs/blob/master/packages/ipfs-core-utils/package.json#L89-L96) is doing this?
> 
> If you have, but just need more flexibility / better bundler support / anything, please ignore me :) I don't want to run around and annoy people with whatabouts.

Hi Philipp! I have honestly been trying to do what you describe for a while but haven't got it to work successfully. I will look at the resources you linked and try a different approach as you suggested. Thank you for your comment!

PR #8, opened 2021-10-12
Build SnarkyJS for Node and Web under one package

MartinMinkov:
**Description**: 
This PR structures SnarkyJS node and chrome bindings to be unified under one project `src/` directory. To do this, we specify a dynamic import in `src/snarky.js` which exports off a window object in case of web or off the node bindings in case of node.

For both web and node, there will be a corresponding `web` and `node` build folder ready for distribution. The idea is that the users of web will import the npm package as `import from "@o1labs/snarkyjs/web"` and node users will import with `require("@o1labs/snarkyjs/node")`

CI was also updated to build both node and web in addition to running jest tests on the node side.

**Usage**:
See package.json to run the different build steps for node and web.

**Testing**:
Manual testing has been done by running `npm link` and installing globally. Tested for node only so far.
```
.
└── src
    ├── chrome_bindings
    ├── node_bindings
    ├── lib
    ├── examples
    snarky.js
    index.ts
...
```

### DIFF
--- a/src/snarky.js
+++ b/src/snarky.js
@@ -18,9 +18,9 @@ export const shutdown = () => {
 
 (async () => {
   if (typeof window !== 'undefined' && typeof window.document !== 'undefined') {
-    // Since we import off the window object, we don't actually use this import for exporting.
-    // The reason we import We require the web bindings just to get it included in the webpack output bundle
-    const _snarky = require('./chrome_bindings/snarky_js_chrome.bc.js');
+    // Since we export off the window object, we don't actually use this import for exporting.
+    // The reason we require the web bindings here is just to get it included in the webpack output bundle.
+    const _ = require('./chrome_bindings/snarky_js_chrome.bc.js');
     Field = window.__snarky.Field;
     Bool = window.__snarky.Bool;
     Circuit = window.__snarky.Circuit;


.prettierrc.js
```diff
@@ -1,3 +1,6 @@
 module.exports = {
-  ...require('gts/.prettierrc.json')
-}
+  trailingComma: "es5",
+  tabWidth: 2,
+  semi: true,
+  singleQuote: true,
+};
```
bkase, line 6: We’re going to need to bike shed these at some point, haha. But now is not the time

scripts/publish.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+npm run prepublish:both
+if [ -d "./dist" ] 
+then
+    cp package.json ./dist
+		cd ./dist
```
jasongitmail, line 6: slight formatting clean up

src/chrome_bindings/plonk_init.js
```diff
@@ -1,15 +1,47 @@
...
-    document.body.appendChild(newScript);
-    newScript.addEventListener('load', () => {
-      var newScript = document.createElement('script');
-      newScript.src = './snarkyjs_chrome.js'; // This is the bundled snarkyjs code -- see webpack.config.js on how the bundle is generated
-      document.body.appendChild(newScript);
+import init, * as plonk_wasm from '/plonk_wasm.js';
+import { override_bindings } from '/worker_run.js';
+
+export const initSnarkyJS = async (bundle) => {
+  var plonk_wasm_init = await init();
```
jasongitmail, line 5: Small nit to prefer using `let` instead of `var`, throughout this file. Would feel more modern

PR #13, opened 2021-10-26
[WIP] Rewrite joo code-gen modules to es6

nholland94:
Still a WIP. Needs more testing. In particular, the chrome bindings almost certainly don't work atm, but it shouldn't be too much more effort to make this work for those.


mitschabaude:
probably won't use this, closing for now

PR #21, opened 2021-11-11
Fix nodejs, version 2

mitschabaude:
Changes from the old fix-nodejs branch:
- Update snarky_js_bindings
- Use a top-level await for nodejs, to make sure consumers can't mess up
- Update bindings built with a clean tree
- Bump MINA_COMMIT file

Further changes on top of that:

- Fix node js interface:
  * Add separate separate entry-point for node ES modules
  * The ESM version top-level-awaits snarkyReady, the CJS version doesnt
  * This means that downstream ESM consumers can just import and immediately use everything, while CJS consumers must await snarkyReady themselves
  * In addition, ESM consumers can now use named imports -> not possible before -> enables typescript/ESM consumers

- Create simpler webpack config for node with smaller output
  * Introduces conditional imports of .node.js files
  * Fixes duplication of chrome bindings in node output and vice versa (previously caused by import of both bindings being present in the same file)
  * Moves to unminified build for node that doesn't bundle node dependencies (because we want dependencies to be shareable w/ other packages)
  * Fix other minor issues to ensure that npm pack produces a working package

- Adapt jest test to work with the new set-up
  * Note that jest must wait for snarkyReady, because it is CJS-only
  * It works now by importing from /dist and awaiting promise before doing anything
  * Food for thought: A different testing framework that plays well with ESM would enable more readable / maintainable tests

- Introduce conditional exports to enable imports from "@o1labs/snarkyjs" without subpaths


MartinMinkov (approved):
This is awesome! Did a manual QA of the bundles by running `npm link` and importing in a test repo. Everything checks out!

PR #22, opened 2021-11-12
Reduce bundle size by minifying

mitschabaude:
DONT MERGE YET, depends on PR #21

This just ships smaller jsoo files (25MB -> 5.5MB) for both node and chrome, produced by `esbuild --minify`


MartinMinkov (approved):
LGTM!

PR #23, opened 2021-11-12
clean interface for web

mitschabaude:
Make `index.ts` the actual entry-point of the web version instead of requiring to import `initSnarky()`.

Web version for now is ESM only and uses top level await so we can import and use everything directly.

The current idea is that consumers write their SmartContract using normal imports and the build command creates a
smartContract.js file that makes sure to lazy-load the resulting bundle and expose async methods like .prove() that
internally await lazy loading to finish.

With this setup, further optimizations like running *all* smart contract logic in a Web Worker would probably be simple.


mitschabaude:
Btw this PR depends on #21, #22 and I also still want set up tests that run in the browser

mitschabaude:
## Clean web interface

This is ready to review, with some minor hacks remaining that can be ironed out later

* Achieves easy portability of the web version by inlining Worker, Wasm and JSOO files into a single bundle
* To make that happen, switch from webpack to a multistep build script using esbuild
* Web version should now be just importable and usable in any project with standard web build tooling  
  *(tested in a simple tsc + esbuild project, TODO test with webpack, create-react-app, next, vite, rollup, parcel)*
* As a byproduct of inlining Web Workers, **snarkyjs now works in Firefox**! (Before, we used ESM import/export syntax in Workers which FF didn't support, now every Worker is bundled and therefore doesn't need imports)

### Minor hack:

Some of our Web Workers have the peculiar feature that they have to be able to spawn new versions of themselves. In the old ESM version, this was done by fetching the new Worker code from `import.meta.url`, by [this code](https://github.com/GoogleChromeLabs/wasm-bindgen-rayon/blob/main/src/workerHelpers.no-bundler.js). However, this (apart from being ESM syntax which breaks Firefox) comes with the assumption that the Worker is served under the same domain, which destroys portability of the lib.

Creating the same functionality with inline Worker is an interesting puzzle. It implies the Worker has to be [a quine](https://en.wikipedia.org/wiki/Quine_(computing)) - a computer program which can output its own source code (and in this case, also launch a Worker with that source code). I didn't quite make that work yet. So, as an intermediary hack, Workers spawn themselves by fetching their code from a CDN proxying this Github repo. I expect to remove this ugliness soon in another PR.

mitschabaude:
I have one more caveat to add

At the moment, there is no entry-point configured in the export map for CJS + browser consumers. I don't know whether bundlers use this case, but I fear some may do / trip up over the fact that the browser version is ESM only. Maybe I should test and change that before merging

mitschabaude:
Added another commit which fixes importing the node version in nextjs

mitschabaude:
> I have one more caveat to add
> 
> At the moment, there is no entry-point configured in the export map for CJS + browser consumers. I don't know whether bundlers use this case, but I fear some may do / trip up over the fact that the browser version is ESM only. Maybe I should test and change that before merging

I have tested it now with create-react-app and next. Both end up in the ESM entrypoints. Next manages to use our node version when building / doing SSR, and the browser version when running on the client, and both seems to work. This also implies a raw Webpack config can be made working since Next uses Webpack. These were actually the tools I was worried most about so I'm now confident this should be merged.

MartinMinkov (approved):

PR #24, opened 2021-11-18
Fix: add missing types to web build

mitschabaude:


mitschabaude:
Sorry for this PR - fixes #23 which was missing an important `d.ts` file in dist after build

MartinMinkov (approved):

PR #27, opened 2021-11-19
Async loading: Export proxy classes that work before snarky loads

mitschabaude:
Instead of exporting classes like Circuit, Field, etc directly, we construct "proxies" that are immediately available and look like the original classes.
﻿This simplifies using the node version of snarkyjs in CJS mode, because users don't have to await snarkyReady before accessing these classes.

Specifically, users can now declare their circuit at the top level:

```js
let snarky = require('@o1labs/snarkyjs');
let { Circuit, Poseidon, circuitMain, public_ } = snarky;

class Main extends Circuit {
  @circuitMain
  static main(preimage: Field, @public_ hash: Field) {
    Poseidon.hash([preimage]).assertEquals(hash);
  }
}
```

And then they only have to await snarkyReady when they actually want to *use* this class:

```js
let { snarkyReady } = snarky;

(async () => {
  await snarkyReady;

  let kp = Main.generateKeypair();
})
```

This currently does not affect browser / node-ESM consumers -- they don't need to use snarkyReady at all.
But it provides a potential mechanism for non-blocking loading of snarkyjs which we could also adopt in the ESM version - TBD!

Right now, this makes writing jest tests much more friendly -- see the updated tests.

Commits:
- export synchronously available proxy versions of all classes in node
- refactor to function and add hopefully helpful error messages
- make browser version more compatible with node ver -> for frameworks like next which may try to execute the same code in different environments
- simplify jest test, thanks to new proxying
- add another test


src/snarky-class-spec.json
```diff
@@ -0,0 +1,316 @@
+[
```
bkase, line 1: Is there an easy way we can avoid checking this in since it’s auto-generated? Make generating this the first step of the build or something. If it’s too much effort, we can do it later though.

bkase (approved):
This is awesome!

bkase:
As shared in chat, I’d prefer the `ready` name too

jasongitmail:
This PR looks fantastic! Exciting for the ease of use in testing.

Preference for `isReady` to indicate it is not returning a value.

MartinMinkov:
My preference is to use `isReady` for the naming here. My reasoning for such is that it's easier to infer for the user that it's a true/false situation in which snarky is loaded. If I was looking at the API with a fresh set of eyes, `ready` can mean different things vs the _is_ prefix that communicates it's ready or not just on a glance.

src/snarky-class-spec.json
mitschabaude: Yeah that should work!

mitschabaude: At least until I get to make the bode build fast enough that this becomes a noticeable bottleneck ^^

mitschabaude: Updated the PR!

PR #33, opened 2021-11-25
Finish refactor of web build

mitschabaude:
This is a follow-up of #23 which removes some problems that remained in the web build:

- We no longer maintain a manually-rewritten version of `plonk_wasm.js`, instead the original file is rewritten automatically during the build.
- We no longer need worker's src to be fetched from Github, instead now *everything* is created from a single bundle.
- That bundle no longer contains duplicates of code portions like the wasm bytecode.

The old strategy, which lead to duplicated code, was to include the source of our workers in the bundle as a *string*. The new strategy is to wrap all modules that are needed in a worker in *functions*. When starting a worker, we first dynamically create its source code by calling `Function.toString()` on the functions that make up the worker's code, and concatenate those. We also have a system for declaring dependencies of a function so that we can recursively find all the functions that need to be in the worker's code.


MartinMinkov (approved):
LGTM!

PR #34, opened 2021-11-29
Move node build to esbuild

mitschabaude:
...because the webpack build was awfully slow.

The new build uses typescript to handle decorators (with metadata) and create the d.ts files, and esbuild to bundle the tsc output into a CJS-compatible single file.

### Could this be even faster?
The build time is dominated by `tsc` (>1 sec), esbuild only takes a few ms. Sadly, there is no good way to get rid of the tsc invocation, because esbuild supports neither decorator metadata nor emitting d.ts files. There is also [SWC](https://swc.rs/) which does support decorator metadata, but also can't emit the d.ts files.


MartinMinkov (approved):
LGTM!

PR #36, opened 2021-12-01
Feature/workshop examples

mimoo:


README.md
```diff
@@ -33,5 +33,5 @@ npm run test
 ```bash
 # Make sure to have proper NPM credentials before publishing.
 # To get credentials, use `npm login`
-./scripts/publish.sh
```
mimoo, line 36: how is this going to work if we don't build dist and publish that :o

src/lib/party.ts
```diff
@@ -73,6 +73,15 @@ export type WithHash<T, H> = {
   hash: H;
 };
 
+/**
+ * One specific permission value.
+ *
+ * A [[ Perm ]] tells one specific permission for our snapp how it should behave
```
mimoo, line 79: Following the official examples of typedoc, they would write ``[[`Perm`]]``

README.md
mitschabaude: `npm publish` automatically runs the "prepublishOnly" script, which builds dist
And /dist is included in the files

IMO it's much simpler to publish from the root. Then all the path references in package.json ("export", "types", "main") work the same in development as in the published package. (The old publish.sh script didn't handle that, just broke the references)

mimoo: awesome, thanks for the explanation :)

.github/workflows/doc.yml
```diff
@@ -2,7 +2,7 @@ name: Build SnarkyJS
 on:
   push:
     branches:
-      - main
+      - feature/workshop-examples
```
mitschabaude, line 5: should revert this before merging

package.json
```diff
@@ -1,68 +1,63 @@
 {
   "name": "@o1labs/snarkyjs",
```
mitschabaude, line 2: Q: should we change package name to "snarkyjs"? I prefer the shorter version (has also been used already in some of the workshop/bootcamp example repos)

src/chrome_bindings/worker_run.js
```diff
@@ -17,61 +17,59 @@ export default function workerRun() {
     return {
       caml_pasta_fp_plonk_index_create: {
         args: [
-          plonk_wasm.WasmPastaFpPlonkGateVector,
+          plonk_wasm.WasmFpGateVector,
           undefined /* number */,
```
mitschabaude, line 21: comment for other reviewers: these are changes related to making kimchi work

src/lib/construction.ts
```diff
@@ -0,0 +1,182 @@
+import { Field, Proof } from '../snarky';
```
mitschabaude, line 1: this file is not used anywhere in the code - delete it? or needed later? @imeckler

src/lib/int.ts
```diff
@@ -0,0 +1,392 @@
...
+
+    let q = Circuit.witness(
+      Field,
+      () => new Field((BigInt(x.toString()) / BigInt(y_.toString())).toString())
+    );
+
+    q.rangeCheckHelper(UInt64.NUM_BITS).assertEquals(q);
+
+    // TODO: Could be a bit more efficient
+    let r = x.sub(q.mul(y_)).seal();
```
mitschabaude, line 71: what does `seal()` do here?

src/lib/int.ts
```diff
@@ -0,0 +1,392 @@
...
+    return new Sgn(Field.one);
+  }
+  static get Neg() {
+    return new Sgn(Field.one.neg());
+  }
+}
+
+export class Int64 {
+  // In the range [-2^63, 2^63 - 1]
+  @prop value: Field;
```
mitschabaude, line 309: `Int64` doesn't really seem implemented yet - it looks like it's just a wrapper around a Field, without anything constraining it to 64 bits.
This makes me a bit uneasy because we use `Int64` as the type for the delta in transactions

src/lib/circuit_value.ts
```diff
@@ -5,29 +5,36 @@ type Constructor<T> = { new (...args: any[]): T };
 
 export type Tuple<A, _N extends number> = Array<A>;
```
mitschabaude, line 6: This `Tuple` type isn't used anywhere, I'd generally prefer to remove exports that are neither used nor documented

mitschabaude:
General comment:

Currently we export most things by writing the export next to the definition, like
`export function aFunction() {`

IMO it's easier to understand the purpose of a file by listing its exports at the very top (after the imports), like
`export {aFunction, anotherThing, aThirdThing};`

Would you be OK if I gradually change this throughout the codebase? @bkase @jasongitmail @MartinMinkov @imeckler

src/snarky.d.ts
```diff
@@ -167,24 +447,24 @@ export class Bool {
...
 }
 
 export interface CircuitMain<W, P> {
   snarkyWitnessTyp: AsFieldElements<W>;
   snarkyPublicTyp: AsFieldElements<P>;
-  snarkyMain: (W, P) => void;
+  snarkyMain: (w: W, p: P) => void;
 }
 
 export class Circuit {
```
mitschabaude, line 470: Since Circuit has only static methods, I think it shouldn't be a class. It should just be a module/namespace - object with functions on it

src/snarky.d.ts
```diff
@@ -357,10 +640,138 @@ export const Poseidon: {
...
+  getAccount(publicKey: { g: Group }): Account | null;
+}
+
 /* TODO: Figure out types for these. */
-export const ofFieldElements: (x: any[], y: any[]) => any[];
-export const toFieldElements: (x: any[], y: any[]) => any[];
-export const sizeInFieldElements: (x: any[]) => number;
+export const ofFields: (x: any[], y: any[]) => any[];
+export const toFields: (x: any[], y: any[]) => any[];
+export const sizeInFields: (x: any[]) => number;
```
mitschabaude, line 775: These three exports and the next two don't really exist. Can I remove them?

src/snarky.d.ts
```diff
@@ -257,11 +536,15 @@ export class Circuit {
 
   static verify(publicInput: any[], vk: VerificationKey, pi: Proof): boolean;
 
-  static toFieldElements<A>(A): Field[];
+  static toFields<A>(a: A): Field[];
+
+  static inProver(): boolean;
+
+  static inCheckedComputation(): boolean;
```
mitschabaude, line 543: Just curious - what's the difference between `inProver` and `inCheckedComputation`?

src/snarky.d.ts
```diff
@@ -98,60 +290,148 @@ export class Field {
...
+   * Returns true if this [[`Bool`]] is equal to `y`.
+   * @param y a [[`Bool`]].
+   */
   equals(y: Bool | boolean): Bool;
+
+  /**
+   * Returns true if this [[`Bool`]] is true.
+   */
+  // TODO: that seems useless :D?
   isTrue(): Bool;
```
mitschabaude, line 397: can we remove `isTrue` and `isFalse`? :smile:

src/snarky.d.ts
```diff
@@ -98,60 +290,148 @@ export class Field {
...
   static toField(x: Bool | boolean): Field;
 
   static Unsafe: {
+    /**
+     * Converts a [[`Field`]] into a [[`Bool`]]. This is a **dangerous** operation
+     * as it assumes that the field element is either 1 or 0
+     * (which might not be true).
+     * @param x a [[`Field`]]
+     */
     ofField(x: Field | number | string | boolean): Bool;
```
mitschabaude, line 435: Instead of creating this `Unsafe` thing, it seems better to just have a function that behaves like
`ofField: (x: Field) => x.equals(1)`

src/lib/merkle_stack.ts
```diff
@@ -0,0 +1,54 @@
+import { AsFieldElements, Circuit, Field, Poseidon } from '../snarky';
```
mitschabaude, line 1: didn't review this because WIP

src/lib/mina.ts
```diff
@@ -1,214 +1,306 @@
...
+
+    const senderPubkey = sender.toPublicKey();
+    const txn = result
+      .then(() => getAccount(senderPubkey))
+      .then((senderAccount) => {
+        if (currentTransaction === undefined) {
+          throw new Error('Transaction is undefined');
+        }
+
+        const otherParties: Array<SnarkyParty> = currentTransaction.parties.map(
```
mitschabaude, line 181: I think we generally shouldn't use type annotations in assignments, except where necessary. TS infers these types, just like Ocaml. The benefits of type inference are that
* it is more flexible than prescribing a type
* less types have to be imported between files, leading to simpler code

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+    vestingPeriod: GlobalSlot,
+    vestingIncrement: Amount
+  ) {
+    this.initialMinimumBalance = initialMinimumBalance;
+    this.cliffTime = cliffTime;
+    this.cliffAmount = cliffAmount;
+    this.vestingPeriod = vestingPeriod;
+    this.vestingIncrement = vestingIncrement;
+  }
+}
```
mitschabaude, line 43: Classes like these, which don't have instance methods, should just be `type`s instead. That removes the need for a constructor and makes the code less painful. (This observation applies throughout this file)

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+export type Amount = UInt64;
+export const Amount = UInt64;
+export type Balance = UInt64;
+export const Balance = UInt64;
+export type Fee = UInt64;
+export const Fee = UInt64;
+export type GlobalSlot = UInt32;
+export const GlobalSlot = UInt32;
+export type SignedAmount = Int64;
+export const SignedAmount = Int64;
```
mitschabaude, line 16: I'm not a fan of having many names for the same types / classes - that just adds complexity IMO, our brain can't remember so many names

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+  }
+}
+
+export class ProtocolStatePredicate {
+  snarkedLedgerHash_: OrIgnore<Field>;
+  snarkedNextAvailableToken: ClosedInterval<UInt64>;
+  timestamp: ClosedInterval<UInt64>;
+  blockchainLength: ClosedInterval<UInt32>;
+  minWindowDensity: ClosedInterval<UInt32>;
+  lastVrfOutput_: OrIgnore<Field>;
```
mitschabaude, line 556: why do some of these properties have an underscore at the end?

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+ * A [[ Perm ]] tells one specific permission for our snapp how it should behave
+ * when presented with requested modifications.
+ *
+ * Use static factory methods on this class to use a specific behavior. See
+ * documentation on those methods to learn more.
+ */
+export class Perm {
+  constant: Bool;
+  signatureNecessary: Bool;
+  signatureSufficient: Bool;
```
mitschabaude, line 86: It would be less confusing to me if "signatureSufficient" would instead be called "proofNecessary", and the Bool negated (if that's accurate)

mitschabaude: What exactly does `constant` mean? I thought whether no one has the permission, but below it's also used for the case where everyone has the permission..

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+    this.setSnappUri = setSnappUri;
+    this.editRollupState = editRollupState;
+    this.setTokenSymbol = setTokenSymbol;
+  }
+}
+
+/* TODO: How should we handle "String"s, should we bridge them from OCaml? */
+export class String_ extends CircuitValue {}
+
+export class TokenSymbol extends CircuitValue {
```
mitschabaude, line 277: more unimplemented exports that I'd like to remove

src/lib/party.ts
```diff
@@ -0,0 +1,784 @@
...
+
+  get update(): Update {
+    return this.body.update;
+  }
+
+  get publicKey(): PublicKey {
+    return this.body.publicKey;
+  }
+
+  static createUnsigned(publicKey: PublicKey): Party<void> {
```
mitschabaude, line 743: There's usually no reason to type-annotate function return values. TS infers these types for us!

src/build/build.mjs
```diff
@@ -18,16 +29,12 @@ async function buildAndImport(srcPath) {
     outfile,
     target: 'esnext',
```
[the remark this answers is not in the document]
mitschabaude: Good idea!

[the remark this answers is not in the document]
mitschabaude: Ah sorry that doesn't work, I forgot. We need esnext here because otherwise esbuild complains about top-level await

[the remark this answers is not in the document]
mitschabaude: And this is the build script for ./run, where we definitely want top-level await

src/index.ts
```diff
@@ -2,3 +2,27 @@ export * from './snarky';
...
+export * as DataStore from './lib/data_store';
+export * from './lib/snapp';
+export * from './lib/optional';
+export * from './lib/proof_system';
+export * from './lib/party';
+
+import * as util from 'util';
+import * as Snarky from './snarky';
+
+/*
```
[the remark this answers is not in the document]
mitschabaude: Already removed them :)

.github/workflows/doc.yml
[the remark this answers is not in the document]
mitschabaude: Fixed!

package.json
```diff
@@ -1,68 +1,63 @@
...
+    "snarky-run": "src/build/run.mjs"
+  },
   "engines": {
-    "node": ">=16"
+    "node": ">=16.4.0"
   },
   "scripts": {
     "type-check": "tsc --noEmit",
     "build": "node src/build/buildNode.mjs",
     "build:server": "npm run build",
```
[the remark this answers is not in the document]
mitschabaude: Done!

package.json
```diff
@@ -1,68 +1,63 @@
...
-    "start:web": "./scripts/build-and-run-web-esbuild.sh",
+    "start:web": "./scripts/build-and-run-web.sh",
```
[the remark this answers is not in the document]
mitschabaude: Agreed, I removed the entire script folder now (start:web / exec)

src/lib/int.ts
```diff
@@ -0,0 +1,391 @@
...
+
+    let r_ = new UInt64(r);
+    let q_ = new UInt64(q);
+
+    r_.assertLt(new UInt64(y_));
+
+    return [q_, r_];
+  }
+
+  /** Integer division.
```
jasongitmail, line 81: small nit: The comment style is unusual. Usually the `/**` line doesn't contain comment text https://typedoc.org/guides/doccomments/

src/lib/int.ts
```diff
@@ -0,0 +1,391 @@
...
+    return new UInt32(z);
+  }
+
+  sub(y: UInt32 | number): UInt32 {
+    let z = this.value.sub(argToField('UInt32.sub', y));
+    z.rangeCheckHelper(UInt32.NUM_BITS).assertEquals(z);
+    return new UInt32(z);
+  }
+
+  lte(y: UInt32): Bool {
```
jasongitmail, line 251: iirc we had decided to use the long-form version of these names (e.g. `lessThanOrEqual`), similar to Jest, so that IDE fuzzy matching will work for either situation (i.e. a developer could type `lessTh...` or `lte` and still get the IDE suggestion)

src/lib/int.ts
```diff
@@ -0,0 +1,391 @@
...
+  gt(y: UInt32): Bool {
+    return y.lt(this);
+  }
+
+  assertGt(y: UInt32) {
+    y.assertLt(this);
+  }
+}
+
+class Sgn extends CircuitValue {
```
jasongitmail, line 284: Is there a reason we don't use Sign here? Too likely to be confused with signature generation?

src/lib/int.ts
```diff
@@ -0,0 +1,391 @@
...
+
+  add(y: Int64 | UInt32 | UInt64) {
+    return new Int64(this.value.add(y.value));
+  }
+
+  sub(y: Int64 | UInt32 | UInt64) {
+    return new Int64(this.value.sub(y.value));
+  }
+
+  repr(): { magnitude: Field; isPos: Sgn } {
```
jasongitmail, line 372: What is this?

src/lib/party.ts
bkase: I disagree, the types become nice documentation. Ie. If I see an arbitrary UInt64, is it an Amount or a Fee?  It would be better to add more type-safety beyond aliases so we don't mix these up https://basarat.gitbook.io/typescript/main-1/nominaltyping could be a good approach, but maybe it's too hacky?

jasongitmail:
General comment. Are builds are output to `web` & `server` directories. I like that naming convention better than `chrome` and `node` unless we're intended to maintain separate code for different runtimes, it'd make sense to stick with the generic names and list which runtimes we support in the README.

src/build/buildWeb.mjs
```diff
@@ -103,16 +113,20 @@ async function copy(copyMap) {
...
 
 function execPromise(cmd) {
-  return new Promise((r) =>
-    exec(cmd, (_, stdout) => {
-      r(stdout);
+  return new Promise((res, rej) =>
+    exec(cmd, (err, stdout) => {
+      if (err) {
+        console.log(stdout);
+        rej(err);
```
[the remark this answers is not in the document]
mitschabaude: Oh you're right - a control flow gotcha! Thanks. While [not strictly needed](https://stackoverflow.com/questions/32536049/do-i-need-to-return-after-early-resolve-reject) here because a rejected promise can't change back to resolved - so the resolve below is a no-op - it is good practice to return, to match the intended control flow.

src/build/run.mjs
```diff
@@ -1,22 +1,23 @@
...
   main,
   default: runDefault,
   keypair: keyPair,
+  keep,
 } = minimist(process.argv.slice(2));
 
 if (!filePath) {
   console.log(`Usage:
-./run [file]`);
+npx snarky-run [file]`);
```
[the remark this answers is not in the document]
mitschabaude: Ugly, but intended because it creates a string with a line break in it

src/examples/simple_snapp.ts
```diff
@@ -0,0 +1,63 @@
...
+  UInt64,
+  PrivateKey,
+  SmartContract,
+  Mina,
+  Party,
+  isReady,
+} from '@o1labs/snarkyjs';
+
+class SimpleSnapp extends SmartContract {
+  @state(Field) x = State<Field>();
```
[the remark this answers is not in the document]
mitschabaude: And change the examples (at least the ones in snapp CLI)

src/lib/int.ts
mitschabaude: Yes! Will try to find time this or next week, by now I can handle changing the Ocaml-side API

src/lib/int.ts
mitschabaude: @imeckler

src/examples/ex02_root.ts
```diff
@@ -15,9 +24,24 @@ class Main extends Circuit {
   }
 }
 
+console.log('generating keypair...');
```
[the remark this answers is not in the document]
mitschabaude: To be honest - I run this example file often for testing, and I like them because the whole operation takes more than 10 seconds and this waiting time is less confusing/annoying if you see what is currently happening. We could remove the timings

src/examples/wip.ts
```diff
@@ -0,0 +1,576 @@
...
+    .wait();
+
+  // .catch(e => console.log('error', e));
+  const a = await Mina.getAccount(snappPubkey);
+
+  console.log('final state value', a.snapp.appState[0].toString());
+}
+
+/*
+function mainold() {
```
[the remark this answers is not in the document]
mitschabaude: It currently doesn't execute

src/lib/party.ts
mitschabaude: Hm, interesting take @ documentation. I want to defend my position a bit (even if I'm perfectly fine with leaving the code as it is)

In my mind, variable names, function arguments and object properties should already all have descriptive names, so a descriptive name of the accompanying type doesn't add value and brings other drawbacks

For example, say we want to interact with a function that has the signature
`function(accountCreationFee: UInt64)`
Then we'll be happy, because we know that we're passing in a fee, and we also know how to create that fee - by instantiating a `UInt64`, which is a type we use all the time in snarkyJS so we probably did that a couple of times or have code to see how it is done. We also have a mental model already for what a `UInt64` is under the hood.

On the other hand, if we see this signature:
`function(accountCreationFee: Fee)`
Then the "fee" information is redundant, and we probably need to search for a while to figure out how to create such a `Fee` from the `number` we probably have. Also it may be frustrating that we don't understand what a `Fee` created from a number even *is*, so we'll have to settle for a shallower understanding of our program than in the example above.

I think this example can be generalized to basically any interaction with any type

mitschabaude:
> General comment: Our builds are output to `web` & `server` directories. I prefer that naming convention over than `chrome` and `node` unless we're intended to maintain separate code for different runtimes, it'd make sense to stick with the generic names and list which runtimes we support in the README.

I agree, at least with `chrome` -> `web`, but I'd rather make it `server` -> `node` because the node code is specifically targeted at the node runtime, not so much at a server environment. (And the web code is targeted at browser *runtimes* more than at being run on a website). If we at one point were to add deno compatibility, there's a small likelihood that we could reuse the web code, because the deno runtime tries to be close to web standards - but we could definitely not use the node code (and then the "server" label would become wrong).

I'll change the bindings directories' names after I changed them in the mina repo to keep copying simple, but will wait for that until everything is merged together there because merging changed directories is a pain

src/lib/data_store.ts
```diff
@@ -0,0 +1,198 @@
+import { AccumulatorMembershipProof, Index } from './merkle_proof';
```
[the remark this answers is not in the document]
mitschabaude: I'd say yes, let's remove them!

src/lib/party.ts
mitschabaude: I'm also concerned about the namespace pollution this causes. "Minimal API surface area" is one of the very few programming paradigms that I deeply believe in. party.ts has _26 exports_

src/snarky.d.ts
```diff
@@ -357,10 +640,138 @@ export const Poseidon: {
...
+
+export interface EpochDataPredicate_ {
+  ledger: EpochLedgerPredicate_;
+  seed: OrIgnore_<Field>;
+  startCheckpoint: OrIgnore_<Field>;
+  lockCheckpoint: OrIgnore_<Field>;
+  epochLength: ClosedInterval_<UInt32_>;
+}
+
+export interface ProtocolStatePredicate_ {
```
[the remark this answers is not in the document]
mitschabaude: I went ahead and removed them

src/snarky.d.ts
mitschabaude: Removed

src/lib/party.ts
mitschabaude: Removed

src/examples/simple_snapp.ts
```diff
@@ -0,0 +1,63 @@
...
+await isReady;
+
+const Local = Mina.LocalBlockchain();
+Mina.setActiveInstance(Local);
+
+const account1 = Local.testAccounts[0].privateKey;
+const account2 = Local.testAccounts[1].privateKey;
+
+const snappPrivkey = PrivateKey.random();
+const snappPubkey = snappPrivkey.toPublicKey();
```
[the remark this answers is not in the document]
mitschabaude: Addressed!

src/lib/int.ts
mitschabaude: Addressed!

src/build/buildWeb.mjs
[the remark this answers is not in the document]
mitschabaude: Addressed

src/lib/proof_system.ts
```diff
@@ -0,0 +1,22 @@
+import { Proof } from '../snarky';
+
+export function proofSystem(target: any): any {}
```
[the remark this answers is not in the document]
mitschabaude: leaving this in for the moment, because we'll work on this soon

src/lib/data_store.ts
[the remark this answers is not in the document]
mitschabaude: I removed these files from snarkyjs now, but moved them into the wip rollup example so we can easily iterate on them there, and move back into snarkyjs when finished

package.json
mitschabaude: Awesome, changed it everywhere

src/lib/snapp.ts
```diff
@@ -0,0 +1,327 @@
...
+    // }
+
+    if (key === '_states' || key === '_layout') {
+      throw new Error('Property names _states and _layout reserved.');
+    }
+
+    if (SnappClass._states == undefined) {
+      SnappClass._states = [];
+      let layout: Map<string, { offset: number; length: number }>;
+      SnappClass._layout = () => {
```
mitschabaude, line 160: I wonder why we are setting `_states` and `_layout` dynamically rather than keeping it simple and having these as class fields on SmartContract. We could even have private class fields with the `#` syntax!

mitschabaude: Aaah nvm. Of course it has to be dynamic since it's a property of the specific class that extends SmartContract

src/examples/wip.ts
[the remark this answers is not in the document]
mitschabaude: Changed the file to the original logic with the original main

src/lib/snapp.ts
```diff
@@ -0,0 +1,347 @@
...
+          p = Mina.getAccount(addr).then((a) => {
+            const xs: Field[] = [];
+            for (let i = 0; i < r.length; ++i) {
+              xs.push(a.snapp.appState[r.offset + i]);
+            }
+            return Circuit.witness(Circuit.array(Field, r.length), () => xs);
+          });
+        } else {
+          const res = Circuit.witness(Circuit.array(Field, r.length), () => {
+            throw 'unimplemented';
```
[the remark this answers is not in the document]
mitschabaude: Should be fine, I'll just change the code to look less as if it was unfinished :)

src/lib/int.ts
mitschabaude: Made an issue

PR #38, opened 2021-12-14
Split constructor and deploy (formerly init)

mitschabaude:
Make `SmartContract` constructor support instantiating already deployed snapp.

Move the init logic into a new method `.deploy()` that should be implemented by every child class, like that:

```ts
  @state(Field) x: State<Field>;

  constructor(address: PublicKey) {
    super(address);
  }

  deploy(initialBalance: UInt64, x: Field) {
    super.deploy();
    this.balance.addInPlace(initialBalance);
    this.x = State.init(x);
  }
}
```

TBD: With the current code, only assigning `this.x = ...` in `deploy` and not in the constructor works (thanks to `@state` decorator magic). IMO, this is also logically what should happen - the state is either initialized
(when deploying) or fetched from the blockchain (by clients which refer to an already deployed smart contract). So it shouldn't be set in the constructor.

TBD: Initially we wanted `@deploy` to be a decorator, for consistency. In the proposed API, I just can't find a use case for such a decorator. IMO, the snapp creator should explicitly call
`smartContract.deploy()` in his transaction, so it's not clear to me what a `@deploy` decorator should do. 

EDIT: This was tested to work with the sudoku example on [this branch](https://github.com/mitschabaude/snarkyjs-sudoku/tree/separate-deploy), which I think covers all relevant cases, but it'd be good to double-check the logic @imeckler

### DIFF
--- a/src/lib/snapp.ts
+++ b/src/lib/snapp.ts
@@ -203,7 +203,7 @@ export abstract class SmartContract {
     // this.state = [];
   }
 
-  init(initialBalance: UInt64 | UInt32) {
+  deploy(...args: any[]) {
     try {
       this.executionState().party.body.update.verificationKey.set = Bool(true);
     } catch (_error) {
--- a/src/lib/snapp.ts
+++ b/src/lib/snapp.ts
@@ -1,11 +1,4 @@
-import {
-  Circuit,
-  Field,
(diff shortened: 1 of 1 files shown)


imeckler (approved):
LGTM

PR #64, opened 2022-01-13
update docs link to official API reference

carterbrett:
Updated SnarkyJS docs link from github pages to docs on minaprotocol.com.


mitschabaude:
thanks brett! closing because the same change was just merged minutes before :)

PR #99, opened 2022-03-08
In-snark encryption

mitschabaude:
This adds an encrypt / decrypt API for snarkyjs, alsong with several conversion helpers to represent a string as an array of field elements and vice versa.

For reference: [Github gist with proposed algorithms](https://gist.github.com/imeckler/986d8299c67276ec497cc665ce1b6fce)

~~TODO: implement David's version of the encryption algorithm~~


mitschabaude:
Everything done!

mitschabaude:
This is the Mina commit that was necessary to add Poseidon.Sponge: https://github.com/MinaProtocol/mina/commit/2f263afe409c69b19a60b8f7235cf3f0892e33ec

src/lib/encryption.ts
```diff
@@ -0,0 +1,77 @@
...
+Poseidon.Sponge = Sponge;
+
+type CipherText = {
+  publicKey: Group;
+  cipherText: Field[];
+};
+
+function encrypt(message: Field[], otherPublicKey: PublicKey) {
+  // key exchange
+  let privateKey = Scalar.random();
```
mimoo, line 32: shouldn't you make sure that this is run as a witness thingy? I can't remember the exact API for that, but I'm wondering what `Scalar.random()` means in a circuit otherwise

src/lib/encryption.ts
```diff
@@ -0,0 +1,77 @@
...
+  sponge.absorb(sharedSecret.x); // don't think we need y, that's enough entropy
+
+  // encryption
+  let cipherText = [];
+  for (let messageChunk of message) {
+    let keyStream = sponge.squeeze();
+    let encryptedChunk = messageChunk.add(keyStream);
+    sponge.absorb(encryptedChunk); // absorb for the auth tag
+    cipherText.push(encryptedChunk);
+  }
```
mimoo, line 46: just a thought, but this might be inefficient as theoretically we can absorb two field elements without permuting (the rate of our sponge is 2), but this squeeze-then-absorb might trigger a permutation for every field element. I'm not sure what's the implementation anymore.

src/examples/encryption.ts
```diff
@@ -0,0 +1,41 @@
...
+
+if (decryptedMessage !== message) throw Error('decryption failed');
+console.log(`Original message: "${message}"`);
+console.log(`Recovered message: "${decryptedMessage}"`);
+
+// the same but in a checked computation
+
+await Circuit.runAndCheck(async () => {
+  // encrypt
+  let cipherText = Encryption.encrypt(messageFields, publicKey);
```
mimoo, line 28: probably want to do the encryption outside of the circuit

src/lib/encryption.ts
```diff
@@ -0,0 +1,77 @@
...
+  // decryption
+  let message = [];
+  for (let encryptedChunk of cipherText) {
+    let keyStream = sponge.squeeze();
+    let messageChunk = encryptedChunk.sub(keyStream);
+    sponge.absorb(encryptedChunk);
+    message.push(messageChunk);
+  }
+  // authentication tag
+  sponge.squeeze().assertEquals(authenticationTag!);
```
mimoo, line 74: what's the exclamation mark after `authenticationTag`? Is it an `unwrap` because `pop` returns an option?

mitschabaude: It's similar - an expression that tells typescript that authenticationTag is not undefined (which it could be after pop if the array is empty), to make typescript not complain. It has no runtime effect

src/examples/encryption.ts
mitschabaude: I think it can be useful though - what if we want to store a secret on-chain, and prove how it has been computed?

src/lib/encryption.ts
mitschabaude: Yep the underlying implementation in snarky_js_bindings creates variables [here](https://github.com/MinaProtocol/mina/blob/0587489752ea9bf4334455fa6301313a8a188c35/src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml#L801) (actually an array of boolean vars representing the bits of a scalar, which seems pretty awful to me)

src/lib/encryption.ts
mitschabaude: The relevant implementation is here: https://github.com/o1-labs/snarky/blob/master/sponge/sponge.ml#L287

As you say, the present version triggers one permutation per iteration - because we permute when absorbing and then squeezing. Without the in-between absorbing we'd only permute every second iteration.

src/lib/encoding.ts
```diff
@@ -0,0 +1,253 @@
...
+    toBytes: (fields: Field[]) => toBytesBijective(fields, q),
+    fromBytes: (bytes: Uint8Array) => toFieldsBijective(bytes, q),
+
+    toString(fields: Field[]) {
+      return new TextDecoder().decode(toBytesBijective(fields, q));
+    },
+    fromString(message: string) {
+      let bytes = new TextEncoder().encode(message);
+      return toFieldsBijective(bytes, q);
+    },
```
mimoo, line 106: would be nice to have some sanity check tests to make sure you can go back and forth

MINA_COMMIT
```diff
@@ -1,2 +1,2 @@
 The mina commit used to generate the backends for node and chrome is
-532588af567b7e9a6c157163a8bacc3b2491160f
+2f263afe409c69b19a60b8f7235cf3f0892e33ec
```
mimoo, line 2: I guess this is what generated `snarky_js_chrome.bc.js`? It'd be nice if CI verifies that the file hasn't been manually modified as well :P

src/lib/encoding.ts
```diff
@@ -0,0 +1,253 @@
...
+
+function bytesToFields(bytes: Uint8Array) {
+  // we encode 248 bits (31 bytes) at a time into one field element
+  let fields = [];
+  let currentBigInt = 0n;
+  let bitPosition = 0n;
+  for (let byte of bytes) {
+    currentBigInt += BigInt(byte) << bitPosition;
+    bitPosition += 8n;
+    if (bitPosition === 248n) {
```
mimoo, line 40: noob q: why `===` instead of `==` here?

mimoo (approved):
LGTM!

MINA_COMMIT
mitschabaude: Hm yeah, it would already be a win if my build script updated this file automatically, right now I do it manually

src/lib/encoding.ts
mitschabaude: Yeah I still need to find a testing flow that feels nice and works well with our JS setup.. Right now I'm adding "tests" as examples that I manually run to see if it works

mitschabaude: this is the file where it's "tested" - src/examples/encoding-bijective.ts

src/lib/encryption.ts
mimoo: we might want to create an issue to optimize this at some point :o (or you might want to process chunks two field elements by two field elements)

src/examples/encryption.ts
mimoo: yeah true, I guess only the scalar.random() sounds like it should be done outside the circuit. I remember asking that question a while ago, if we should forbid this function from being run within a circuit

src/lib/encoding.ts
mitschabaude: JS devs just always use `===` :smile: It's the real equality operator

src/lib/encryption.ts
```diff
@@ -29,7 +29,9 @@ type CipherText = {
 
 function encrypt(message: Field[], otherPublicKey: PublicKey) {
   // key exchange
-  let privateKey = Scalar.random();
+  let privateKey = Circuit.inCheckedComputation()
+    ? Circuit.witness(Scalar, () => Scalar.random())
+    : Scalar.random();
```
mimoo, line 34: Nice! IMO `Scalar.random` should do that out of the box, because someone else might make the same mistake

src/lib/encryption.ts
```diff
@@ -38,11 +40,13 @@ function encrypt(message: Field[], otherPublicKey: PublicKey) {
...
-  for (let messageChunk of message) {
+  for (let i = 0; i < message.length; i++) {
     let keyStream = sponge.squeeze();
-    let encryptedChunk = messageChunk.add(keyStream);
-    sponge.absorb(encryptedChunk); // absorb for the auth tag
+    let encryptedChunk = message[i].add(keyStream);
     cipherText.push(encryptedChunk);
+    // absorb for the auth tag (two at a time for saving permutations)
+    if (i % 2 === 1) sponge.absorb(cipherText[i - 1]);
+    if (i % 2 === 1 || i === message.length - 1) sponge.absorb(cipherText[i]);
```
mimoo, line 49: a bit hard to parse, but I'm not sure how you could make it easier to parse. LGTM though

PR #106, opened 2022-04-01
Feature: Add a fetch module to get zkapp account information from a specified endpoint

MartinMinkov:
**Description**
This PR exposes a new function called 'getAccount' which fetches zkapp related information from a specified GraphQL endpoint. The specified endpoint must support the following query which is native to our daemon GraphQL API:
```graphql
query GetAccountInfo($publicKey: PublicKey!) {
      account(publicKey: $publicKey) {
        publicKey
        nonce
        snappUri
        snappState
        receiptChainHash
        delegate
        balance {
          total
        }
    }
```

On a successful fetch, the function will return the account data. If there is an error with the endpoint, the HTTP status code and message are logged. There is an optional `timeout` parameter that can specify a timeout for the GraphQL query. The default is set to `5s`.

**Tested By**
Running a Berkeley node locally and running queries against localhost. Additionally deployed a Berkeley node to a GCP instance and specified the IP address of the instance and that also worked as expected.

### DIFF
--- a/src/fetch/index.mjs
+++ b/src/fetch/index.mjs
@@ -1,4 +1,4 @@
-import fetch, { AbortError } from 'node-fetch';
+import fetch, { AbortError } from 'isomorphic-fetch';
 
 // Specify 5s as the default timeout
 const defaultTimeout = 5000;
--- a/src/fetch/index.mjs
+++ b/src/fetch/index.mjs
@@ -1,0 +1,76 @@
+import fetch, { AbortError } from 'node-fetch';
+
+// Specify 5s as the default timeout
+const defaultTimeout = 5000;
+
+const query = `
+    query GetAccountInfo($publicKey: PublicKey!) {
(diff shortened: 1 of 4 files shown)


src/fetch/index.mjs
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Gets account information on the specified publicKey by performing a GraphQL query
+ * to the specified endpoint. This will call the 'GetAccountInfo' query which fetches
+ * zkapp related account information.
+ *
+ * If an error is returned by the specified endpoint, an error is thrown. Otherwise,
+ * the data is returned.
+ *
+ * @param {string} graphqlEndpoint The GraphQL endpoint to pull account information from
+ * @param {string} publicKey The specified account to get account information on
```
mitschabaude, line 46: Using JSDoc for types. I like it!

mitschabaude (approved):
Very nice! Thanks Martin

PR #116, opened 2022-04-15
Transaction Construction, Pt 2. Restructure snarky.js into folder

mitschabaude:
This is a minor PR (but fairly large diff)
* Refactor the previous `snarky.js` and `snarky.node.js` files, which contained lots of duplicated code, into their own folder `src/snarky`
* This also cleans up the `src` folder from top-level js files, simplifying the build scripts
* In the new folder, add a file `addons.js` which can be used to interface with Rust/Wasm directly, without going via Ocaml. In it, we add a method to serialize/recover the verification key for non-recursive, plain Circuits. This feature is also tested in one of the examples.


src/snarky/index.js
```diff
@@ -0,0 +1,30 @@
...
+  Scalar,
+  Ledger,
+  shutdown,
+  isReady,
+  picklesCompile,
+};
+
+let isReadyBoolean = false;
+let isReady = snarky_ready.then(() => (isReadyBoolean = true));
+let isItReady = () => isReadyBoolean;
```
MartinMinkov, line 20: Why do we declare a function that returns `isReadyBoolean` instead of using it directly?

MartinMinkov (approved):
LGTM! Just have some questions

src/snarky/index.js
mitschabaude: Because we need to access this at later points in time, when the value might have changed!

bkase (approved):

PR #117, opened 2022-04-15
Transaction Construction, Bonus: Int65 example

mitschabaude:
This adds a new example, `src/examples/int65.ts` containing an in-snark "int64" implementation where the signed integer is represented as a magnitude and sign. The magnitude is constrained to 64 bits like the existing UInt64 type, and the sign is constrained to be +-1.

This is intended to replace the existing, incomplete & unconstrained Int64 type in snarkyjs. For now, it serves as an example for implementing new data types on top of existing snarkyjs primitives.

### DIFF
--- a/src/examples/int65.ts
+++ b/src/examples/int65.ts
@@ -42,8 +42,9 @@ class Int65 extends CircuitValue {
   }
 
   toString() {
-    let sign = this.sign.equals(Field.one).toBoolean() ? '' : '-';
-    return sign + this.magnitude;
+    let abs = this.magnitude.toString();
+    let sgn = this.sign.equals(Field.one).toBoolean() || abs === '0' ? '' : '-';
+    return sgn + abs;
   }
 
   isConstant() {
@@ -123,5 +124,7 @@ async function test() {
   console.assert(x.sub(y).toString() === '-256');
   console.assert(y.add(x.neg()).toString() === '256');
   console.assert(x.mul(y).toString() == (-(128 ** 2)).toString());
+  console.assert(y.div(x).neg().toString() === '1');
+  console.assert(y.div(Int65.fromNumber(129)).toString() === '0');
   console.log('everything ok!');
 }
--- a/src/examples/int65.ts
+++ b/src/examples/int65.ts
@@ -51,7 +51,7 @@ class Int65 extends CircuitValue {
   }
 
   // --- circuit-compatible operations below ---
-  // the assumption here is that all DoubleInt64 values that appear in a circuit are already checked as valid
+  // the assumption here is that all Int65 values that appear in a circuit are already checked as valid
   // this is because Circuit.witness calls .check
   // so we only have to do additional checks if an operation on valid inputs can have an invalid outcome (example: overflow)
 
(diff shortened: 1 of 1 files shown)


src/examples/int65.ts
```diff
@@ -0,0 +1,160 @@
...
+import { Field, UInt64, prop, CircuitValue, Circuit, isReady } from 'snarkyjs';
+
+export { Int65 };
+
+class Int65 extends CircuitValue {
+  // * in the range [-2^64+1, 2^64-1], unlike a normal int64
+  // * under- and overflowing is disallowed, similar to UInt64, unlike a normal int64
+
+  @prop magnitude: Field; // absolute value, restricted like UInt64
+  @prop sign: Field; // +/- 1
```
bkase, line 12: Shouldn't we pack the sign bit into the same underlying `Field` corresponding to the magnitude to be more efficient?

Could we instead make this a class with a custom `toFields` rather than a circuit value?

src/examples/int65.ts
```diff
@@ -0,0 +1,160 @@
...
+    return sgn + abs;
+  }
+
+  isConstant() {
+    return this.magnitude.isConstant() && this.sign.isConstant();
+  }
+
+  // --- circuit-compatible operations below ---
+  // the assumption here is that all Int65 values that appear in a circuit are already checked as valid
+  // this is because Circuit.witness calls .check
```
bkase, line 63: oo I didn't know this, where is this defined?

bkase (changes requested):

src/examples/int65.ts
mitschabaude: It's here! https://github.com/MinaProtocol/mina/blob/5a1640263e9fcbec000d8e5709d5c3c5b0773ff5/src/lib/snarky_js_bindings/lib/snarky_js_bindings_lib.ml#L1323

src/examples/int65.ts
mitschabaude: I thought about this for a bit now..

`toFields` returns the in-circuit representation, so the main objective is to minimize the number of constraints that result from this representation. Therefore, I think the only candidate for an efficient 1-field representation is the one where the Int65 _is_ the field: `toFields = int65 => [int65.magnitude.mul(int65.sign)]`. Anything else involving bit packing would just lead to very inefficient circuit operations, IIUC.

So, is `magnitude * sign` ("1-field") a more efficient representation than `(magnitude, sign)` ("2-field")?
Several common operations like add, mul, etc, operate on 1-field so in 2-field they result in one additional multiplication constraint. However, the `check` operation (constraining to 64 bits + a sign) which is called at the introduction of _every_ witness, and also at the end of add, mul, etc, operates on 2-field. So here, the 1-field representation needs to add an additional `magnitude * sign = int65` multiplication constraint, which will typically cancel out most of the gains achieved by 1-field elsewhere.
There are some notable operations for which 2-field is definitely better:
* div and mod (which do integer division with rounding on the magnitude)
* converting the Int64 to a `Currency.Amount.Signed` (for the zkapp balance), which has the exact same `(magnitude, sign)` representation we use here. This is currently the main thing an Int64 is used for, and was my original motivation to use 2 fields.

Overall, I think the existing implementation is the optimal one, but happy to be corrected.

bkase: Sounds good! I love this -- please include this text in a doccomment for the class

PR #118, opened 2022-04-15
Transaction Construction, Pt 3. Prove smart contract execution

mitschabaude:
This adds `SmartContract.prove`, which executes methods inside a Pickles prover. The prover is an output of the already existing `compile()` method.

This PR also contains initial work on constructing the public input -- called the _statement_ in the code. The statement consists of certain hashes of the transaction and of the proving `Party` which is constructed during method execution.

In `SmartContract.prove`, a method is run _twice_: First outside the proof, to obtain the statement, and once in the prover, which takes the statement as input. The current transaction is hashed again _inside_ the prover, which asserts that the result equals the input statement, as part of the snark circuit. The block producer will also hash the transaction they receive and pass it as a public input to the verifier. Thus, the transaction is fully constrained by the proof - the proof couldn't be used to attest to a different transaction.

The PR also adds a pure JS example (for quicker iterating), which is maintained as a useful reference to use snarkyjs in environments without TS decorator support.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "prepublish:web": "NODE_ENV=production node src/build/buildWeb.mjs",
     "prepublish:both": "npm run prepublish:web && npm run prepublish:server",
     "prepublishOnly": "npm run prepublish:web && npm run prepublish:server",
+    "bootstrap-build": "npm run build && node src/build/extractJsooMethods.js && npm run build",
     "format": "prettier --write --ignore-unknown **/*",
     "test": "node --experimental-wasm-modules --experimental-modules --experimental-wasm-threads --experimental-vm-modules ./node_modules/jest/bin/jest.js --silent",
     "clean": "rm -rf dist"
--- a/src/examples/simple_snapp.mjs
+++ b/src/examples/simple_snapp.mjs
@@ -52,6 +52,15 @@ let account2 = Local.testAccounts[1].privateKey;
 let snappPrivKey = PrivateKey.random();
 let snappPubKey = snappPrivKey.toPublicKey();
 
+console.log('compile');
+let { provers, getVerificationKey } = SimpleSnapp.compile(snappPubKey);
+
+console.log('prove');
+let snapp = new SimpleSnapp(snappPubKey);
+let proof = snapp.prove(provers, 'update', [Field(3)]);
+console.log({ proof });
+
+console.log('deploy');
 await Mina.transaction(account1, async () => {
   let snapp = new SimpleSnapp(snappPubKey);
   const amount = UInt64.fromNumber(1e6);
--- a/src/examples/simple_snapp.mjs
+++ b/src/examples/simple_snapp.mjs
@@ -56,10 +56,10 @@ console.log('compile');
 let { provers, getVerificationKey } = SimpleSnapp.compile(snappPubKey);
 
 console.log('deploy');
-await Mina.transaction(account1, async () => {
+await Mina.transaction(account1, () => {
   let snapp = new SimpleSnapp(snappPubKey);
   const amount = UInt64.fromNumber(1e6);
-  const p = await Party.createSigned(account2);
+  const p = Party.createSigned(account2);
   p.balance.subInPlace(amount);
   snapp.deploy(amount, Field(1));
 })
@@ -73,9 +73,9 @@ let snapp = new SimpleSnapp(snappPubKey);
 let proof = await snapp.prove(provers, 'update', [Field(3)]);
 console.log({ proof });
 
-await Mina.transaction(account1, async () => {
+await Mina.transaction(account1, () => {
   let snapp = new SimpleSnapp(snappPubKey);
-  await snapp.update(Field(3));
+  snapp.update(Field(3));
 })
   .send()
   .wait();
--- a/src/examples/simple_snapp.mjs
+++ b/src/examples/simple_snapp.mjs
@@ -55,11 +55,6 @@ let snappPubKey = snappPrivKey.toPublicKey();
 console.log('compile');
 let { provers, getVerificationKey } = SimpleSnapp.compile(snappPubKey);
 
-console.log('prove');
-let snapp = new SimpleSnapp(snappPubKey);
-let proof = snapp.prove(provers, 'update', [Field(3)]);
-console.log({ proof });
-
 console.log('deploy');
 await Mina.transaction(account1, async () => {
   let snapp = new SimpleSnapp(snappPubKey);
@@ -72,6 +67,11 @@ await Mina.transaction(account1, async () => {
   .wait();
 var snappState = (await Mina.getAccount(snappPubKey)).snapp.appState[0];
 console.log('initial state: ' + snappState);
+
+console.log('prove');
+let snapp = new SimpleSnapp(snappPubKey);
+let proof = await snapp.prove(provers, 'update', [Field(3)]);
+console.log({ proof });
 
 await Mina.transaction(account1, async () => {
   let snapp = new SimpleSnapp(snappPubKey);
--- a/src/examples/simple_snapp.mjs
+++ b/src/examples/simple_snapp.mjs
@@ -1,0 +1,76 @@
+// this is a pure JS version of the simple_snapp.ts example, that could be the starting point
+// to develop a more explicit, less pretty API without decorators, which would have the huge benefit
+// of working without a custom TS config and even without TS at all
+
+import {
+  Field,
+  state,
+  State,
+  method,
+  UInt64,
+  PrivateKey,
+  SmartContract,
+  Mina,
+  Party,
+  isReady,
+  shutdown,
+} from 'snarkyjs';
+import 'reflect-metadata';
+
+await isReady;
+
+class SimpleSnapp extends SmartContract {
+  constructor(address) {
+    super(address);
+    this.x = State();
+  }
+  deploy(initialBalance, x) {
+    super.deploy();
+    this.balance.addInPlace(initialBalance);
+    this.x.set(x);
+  }
+  update(y) {
+    let x = this.x.get();
+    this.x.set(x.add(y));
+  }
+}
+
+// manually apply decorators:
+
+// @state(Field) x
+state(Field)(SimpleSnapp.prototype, 'x');
+
+// @method update(y: Field)
+Reflect.metadata('design:paramtypes', [Field])(SimpleSnapp.prototype, 'update');
+method(SimpleSnapp.prototype, 'update');
+
+// usual code
+let Local = Mina.LocalBlockchain();
+Mina.setActiveInstance(Local);
+let account1 = Local.testAccounts[0].privateKey;
+let account2 = Local.testAccounts[1].privateKey;
+let snappPrivKey = PrivateKey.random();
+let snappPubKey = snappPrivKey.toPublicKey();
+
+await Mina.transaction(account1, async () => {
+  let snapp = new SimpleSnapp(snappPubKey);
+  const amount = UInt64.fromNumber(1e6);
+  const p = await Party.createSigned(account2);
+  p.balance.subInPlace(amount);
+  snapp.deploy(amount, Field(1));
+})
+  .send()
+  .wait();
+var snappState = (await Mina.getAccount(snappPubKey)).snapp.appState[0];
+console.log('initial state: ' + snappState);
+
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -35,6 +35,7 @@
     "prepublish:web": "NODE_ENV=production node src/build/buildWeb.mjs",
     "prepublish:both": "npm run prepublish:web && npm run prepublish:server",
     "prepublishOnly": "npm run prepublish:web && npm run prepublish:server",
+    "bootstrap-build": "npm run build && node src/build/extractJsooMethods.js && npm run build",
```
MartinMinkov, line 38: Our `scripts` is starting to get pretty croweded. What are thoughts on using a `Makefile` or some sort of build tool for all these building scripts?

src/examples/simple_snapp.mjs
```diff
@@ -0,0 +1,85 @@
+// this is a pure JS version of the simple_snapp.ts example, that could be the starting point
```
MartinMinkov, line 1: Great idea to include this!

src/lib/party.ts
```diff
@@ -740,7 +751,15 @@ export class Party<P> {
     return this.body.publicKey;
   }
 
-  static createUnsigned(publicKey: PublicKey): Party<void> {
+  static defaultParty(address: PublicKey) {
+    const body = Body.keepAll(address);
+    const predicate = AccountPredicate.ignoreAll();
+    return new Party(body, predicate) as Party & {
```
MartinMinkov, line 757: We are using this union type `Party & { predicate }` a decent amount. Would it be helpful to write this as a type with a name that self documents it's purpose?

src/lib/party.ts
```diff
@@ -782,3 +801,78 @@ export class Party<P> {
...
+    predicate = { type: 'nonce', value: party.predicate };
+  } else {
+    predicate = { type: 'full', value: party.predicate };
+  }
+  return {
+    predicate,
+    body: {
+      ...party.body,
+      events: party.body.events.events,
+      depth: parseInt(party.body.depth.toString(), 10),
```
MartinMinkov, line 819: Would `events` and `body` ever be null? Wondering if it would be good to check for these values if it makes sense.

src/lib/snapp.ts
```diff
@@ -268,6 +272,83 @@ function isProof(typ: any) {
   return false; // TODO
 }
 
+type Statement = { transaction: Field; atParty: Field };
+
+function toStatement(
```
MartinMinkov, line 277: It would be nice if you included a short snippet from your PR description that describes what a `Statement` is and why it's used here.

Info I'm referencing:
... 
constructing the public input -- called the statement in the code. The statement consists of certain hashes of the transaction and of the proving Party which is constructed during method execution. 
...

Some of this information would be really helpful to have instead of hunting through PRs later on

src/lib/snapp.ts
```diff
@@ -268,6 +272,83 @@ function isProof(typ: any) {
...
+        ? (type, i) => Circuit.witness(type, () => witnesses![i])
+        : emptyWitness
+    );
+    func(...witnesses);
+    let tail = Field.zero;
+    checkStatement(statement, self, tail, protocolState);
+    onEnd?.();
+  }
+
+  return [0, name, main];
```
MartinMinkov, line 349: Why is a `0` returned here?

MartinMinkov (approved):

package.json
mitschabaude: TBH, I prefer the package.json as a place for these - that's where a JS developer would expect them. But maybe we don't really need `prepublish:web` and `prepublish:both` as standalone scripts

src/lib/party.ts
mitschabaude: No, they shouldn't be null! I'd say let's rely on the type system and check null/undefined only for values whose type includes null/undefined

src/lib/snapp.ts
mitschabaude: good point - added a comment: https://github.com/o1-labs/snarkyjs/pull/123/commits/65bbe45dfbe8599b6322278d804366f979536493

src/examples/simple_snapp.mjs
bkase: nit: should this be a doccomment? `///` , not sure exactly how that gets surfaced by tools when it's at the top of a file like this, but seems "doccommentish" to me

src/lib/party.ts
```diff
@@ -719,11 +729,12 @@ export class PartyBalance {
   }
 }
 
-export class Party<P> {
+type Predicate = undefined | UInt32 | AccountPredicate;
+export class Party {
   body: Body;
-  predicate: P;
+  predicate: Predicate;
```
bkase, line 735: Can you explain why you moved away from the generic type here?

src/lib/party.ts
```diff
@@ -782,3 +801,78 @@ export class Party<P> {
     return party;
   }
 }
+
+function toParty(party: Party): Party_ {
```
bkase, line 805: It's scary that we have this subtly different `Party` and `Party_` on the bridge. We've had similar weird glue in parts of the Mina code that led to pain which eventually we threw out. If it's easy enough, can we unify these now? If it's not easy enough or there are design questions to consider, can you make an issue and then we can triage it when we sort polish tasks? Or if you think we shouldn't unify these, feel free to explain in a reply here.

src/lib/snapp.ts
```diff
@@ -268,6 +272,83 @@ function isProof(typ: any) {
...
+
+let mainContext = undefined as
+  | {
+      witnesses?: unknown[];
+      self: Party & { predicate: AccountPredicate };
+      protocolState: ProtocolStatePredicate;
+    }
+  | undefined;
+
+function withContext(context: typeof mainContext, f: Function) {
```
bkase, line 320: Can we change this to not rely on a global variable? This could lead to scary bugs later.

We could pass it into `f` or flex our inner JavaScript and `bind` the context at the callsite to this: `f.apply(context)` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply

src/lib/snapp.ts
```diff
@@ -292,31 +373,59 @@ export class SmartContract {
...
+      },
+      () => picklesCompile(rules)
     );
-    let output = picklesCompile(inductiveRules);
-    console.log('output', output);
     return output;
   }
 
+  async prove(provers: any[], methodName: keyof this, args: unknown[]) {
+    let SnappClass = this.constructor as never as typeof SmartContract;
```
bkase, line 397: do you need the `as never` in there to get this to compile? is this the idiomatic way to force a cast in typescript?

src/lib/snapp.ts
```diff
@@ -292,31 +373,59 @@ export class SmartContract {
...
+    let ctx = {
+      self: Party.defaultParty(this.address),
+      protocolState: ProtocolStatePredicate.ignoreAll(),
+    };
+    // console.log(
+    //   'prove: running method directly in prover mode to get statement'
+    // );
+    let statement = Circuit.runAndCheckSync(() => {
+      mainContext = ctx;
+      (this[methodName] as any)(...args);
```
bkase, line 409: can we use `withContext` (and the version without the global variable)

src/snarky.d.ts
```diff
@@ -499,6 +496,8 @@ export class Circuit {
   // static runAndCheck<T>(f : () => Promise<(() => T)>): Promise<T>;
   static runAndCheck<T>(f: () => Promise<() => T>): Promise<T>;
 
+  static runAndCheckSync<T>(f: () => T): T;
```
bkase, line 499: when would you use `runAndCheckSync` vs `runAndCheck`? Is this for web-workers?

bkase (changes requested):

src/snarky.d.ts
mitschabaude: runAndCheck was an async version made by Izzy, but it caused some bugs, like this one: https://github.com/o1-labs/snarkyjs/issues/65
I don't think the async version is really fully compatible with how `snarky` works right now. At some point, I couldn't get something to work with the async version, so I added the sync version. The async version is not used anymore, should probably just replace it with the sync version.

src/lib/party.ts
mitschabaude: I agree! But first, I want to auto-generate some of the TS types in snarky.d.ts, i.e. `Party_` -- see https://github.com/MinaProtocol/mina/issues/10553
After that, I definitely want to unify these two type systems. It's crazy how the type conversions I have to do continue to expand. It's the first thing I want to tackle once all the Berkeley interactions are working end-to-end (which is almost the case)

mitschabaude: created issue https://github.com/o1-labs/snarkyjs/issues/132

src/lib/snapp.ts
mitschabaude: I've seen `as never` it a couple times. `as any` would also work, not sure which is cleaner. It seems to be needed here, because `this.constructor` is just `Function`. What I need here is access the child class -- the class that extends SmartContract. I didn't find a type-safe way to do that.

src/lib/snapp.ts
mitschabaude: I'm really glad you brought this up. When I added this originally, I guess it was fine because both compiling and proving were synchronous, so there was nothing that could conflict with the global variable while they were running (JS is single-threaded). However, since then, the Pickles provers became async. So there are probably real bugs in the code right now if you'd run `prove` multiple times concurrently. I'll have to fix this!
It's not as simple as passing it into `f` though. The reason I made it a global variable is that the pickles circuit needs to access private inputs, but it has no arguments to do so. The interface of Pickles sort of forced me to use a global variable, to store context that the circuit could access. The `f.apply(context)` thing is creative, hadn't thought of that, maybe I can even use that to solve the problem.

src/lib/party.ts
mitschabaude: I don't remember, I can only guess that keeping track of the type parameter in places that had nothing to do with the account predicate became too annoying. The union type we have now makes distinguishing between different variants more localized

src/lib/snapp.ts
mitschabaude: I couldn't come up with a solution with a behaviour that's safer than global variables.
So the way I fixed the potential bug for now, is observing how many times the mainContext is accessed, and throw an error if it's more than the expected number of times. This should reliably detect concurrent accesses. It's not ideal, but at least it protects correctness: https://github.com/o1-labs/snarkyjs/pull/123/commits/84a014996eaadf157dd7fc890b5e36fc51993f74

src/lib/snapp.ts
mitschabaude: done https://github.com/o1-labs/snarkyjs/commit/84a014996eaadf157dd7fc890b5e36fc51993f74

src/examples/simple_snapp.mjs
mitschabaude: I removed this comment at some point. To have some docs for pure JS usage, I instead chose to doc-comment the two functions that enable using decorator functionality without decorators: https://github.com/o1-labs/snarkyjs/pull/123/commits/fb01a0970d1fc102d9eea5ab8614b7fd7410255e

src/lib/party.ts
bkase: Seems fair enough 👍

src/lib/snapp.ts
bkase: Good workaround!

> that the pickles circuit needs to access private inputs, but it has no arguments to do so
Is this something we can fix by changing the pickles circuit interface? Can you make in issue tracking this? (in Mina repo?)

src/snarky.d.ts
bkase: Can you delete the async version from the code so people don't accidentally try to use it?

bkase (approved):
Looks good -- please address those last few things before closing this though

src/snarky.d.ts
mitschabaude: done https://github.com/o1-labs/snarkyjs/pull/123/commits/2a9c15e2568e114405e652ef3cb9e82048c17d45

src/lib/snapp.ts
mitschabaude: That change to pickles seems very far-reaching... not sure if it should be changed. But made an issue to surface the problem: https://github.com/MinaProtocol/mina/issues/10815

PR #119, opened 2022-04-15
Transaction Construction, Pt 4. Deploy

mitschabaude:
* Integrate derivers by exposing a `partiesToJson` function
* Add more party conversion functions to support `partiesToJson`
* Expose method `getVerificationKeyArtifact` to return the verification key from `Pickles.compile` in string format
* Add the `deploy()` method
* Refactor `Mina.transaction` to use the more flexible `createTransaction` under the hood - which is also used by `deploy`
* Add a full example using `compile` / `deploy` to create the deploy transaction for a SmartContract.
* Adapt some party types to changes on the protocol side (this PR is where OCaml-snarkyjs was first merged with develop after a long time)


src/lib/mina.ts
```diff
@@ -80,9 +139,9 @@ export const LocalBlockchain: () => MockMina = () => {
       Math.ceil((new Date().valueOf() - startTime) / msPerSlot)
     );
 
-  const addAccount = (pk: PublicKey, balance: number) => {
+  function addAccount(pk: PublicKey, balance: number) {
```
MartinMinkov, line 142: Why are we switching these to function declarations from function expressions?

MartinMinkov (approved):

src/lib/mina.ts
```diff
@@ -57,6 +54,68 @@ interface Mina {
...
+
+  try {
+    Circuit.runAndCheckSync(f);
+  } catch (err) {
+    currentTransaction = undefined;
+    // TODO would be nice if the error would be a bit more descriptive about what failed
+    throw err;
+  }
+
+  const otherParties: Array<Party_> = currentTransaction.parties.map((party) =>
```
bkase, line 80: nit: `currentTransaction.parties.map(toParty)`

src/lib/mina.ts
```diff
@@ -57,6 +54,68 @@ interface Mina {
...
+    feePayer = {
+      body: toPartyBody(Body.dummy()),
+      predicate: UInt32.zero,
+    };
+  }
+
+  const txn: Parties = { otherParties, feePayer };
+
+  nextTransactionId.value += 1;
+  currentTransaction = undefined;
```
bkase, line 104: we should do something similar w.r.t. getting rid of globals with this value -- same thing though as my note on the prior PR; we can make an issue if it's not an obvious fix

src/lib/snapp.ts
```diff
@@ -372,6 +373,7 @@ export class SmartContract {
   static compile(address?: PublicKey) {
     // TODO: think about how address should be passed in
     // if address is not provided, create a random one
+    // TODO: maybe PublicKey should just become a variable? Then compile doesn't need to know the address, which seems more natural
```
bkase, line 376: 👍 -- calling @mrmr1993 for a double-check

src/lib/snapp.ts
```diff
@@ -531,3 +528,65 @@ function emptyWitness<A>(typ: AsFieldElements<A>) {
...
+function deploy<S extends typeof SmartContract>(
+  SmartContract: S,
+  address: PublicKey,
+  verificationKey: string
+) {
+  let i = 0;
+  let tx = Mina.createUnsignedTransaction(() => {
+    let snapp = new SmartContract(address);
+    snapp.deploy();
+    i = Mina.currentTransaction!.nextPartyIndex - 1;
```
bkase, line 543: looks like `i` is unused now

bkase (changes requested):

src/lib/mina.ts
mitschabaude: it's already that way at the top of the PR stack 👍🏻

src/lib/snapp.ts
mitschabaude: already fixed

src/lib/mina.ts
mitschabaude: I made one: https://github.com/o1-labs/snarkyjs/issues/134

bkase (approved):
Looks good to me! But worth getting @mrmr1993 's 👀 s on that one bit

src/lib/snapp.ts
mrmr1993: Yea, make it a variable

mitschabaude: Will do that, thanks! Outsourced it to a separate issue, since it will be easier after other refactorings I want to do: https://github.com/o1-labs/snarkyjs/issues/136

PR #122, opened 2022-04-15
Transaction Construction, Pt 7. Fully working transactions

mitschabaude:
This includes the remaining fixes to make transactions work in an end-to-end test, and also to make the old mock API -- `Mina.transaction` -- work again.

This PR also makes improvements to how signing is handled in snarkyjs:
* Instead of passing JSON parties back and forth with OCaml to add signatures, add flexible methods to compute transaction commitments and to sign arbitrary field elements
* Implement "lazy signing", that is, decouple _declaring_ a party that should be signed with actually creating the signature. Important because signatures can only be created once the full transaction is known.

Finally, this also contains a large refactor of `party.ts`, to get rid of unnecessary classes (which are replaced by plain types and functions) and reduce code.

### DIFF
--- a/src/lib/party.ts
+++ b/src/lib/party.ts
@@ -334,7 +334,7 @@ export type Precondition = undefined | UInt32 | AccountPrecondition;
  *
  * TODO: We need to rename this still.
  */
-export class Body {
+export type Body = {
   /**
    * The address for this body.
    */
@@ -372,11 +372,13 @@ export class Body {
   accountPrecondition: Precondition;
   useFullCommitment: Bool;
   incrementNonce: Bool;
+};
 
+export let Body = {
   /**
    * A body that Don't change part of the underlying account record.
    */
-  static keepAll(publicKey: PublicKey): Body {
+  keepAll(publicKey: PublicKey): Body {
     function keep<A>(dummy: A): SetOrKeep<A> {
       return new SetOrKeep(new Bool(false), dummy);
     }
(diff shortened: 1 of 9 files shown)


src/lib/party.ts
```diff
@@ -772,13 +638,31 @@ export class Party {
     return this.body.publicKey;
   }
 
+  sign(privateKey?: PrivateKey) {
```
bkase, line 641: Can we keep these data types purely functional? This should lead to fewer gotchas for people.

ie. returning a copy of this Party but with the authorization field set

src/lib/party.ts
```diff
@@ -819,27 +703,100 @@ export class Party {
...
+
+type FeePayer = Party & {
+  authorization: Exclude<LazyControl, { kind: 'proof'; value: string }>;
+} & {
+  body: { accountPrecondition: UInt32 };
+};
+
+type Parties = { feePayer: FeePayer; otherParties: Party[] };
+
+function addMissingSignatures(
```
bkase, line 739: same here -- ideally we have a separate type for a "pending signatures Party" and a "finished Party"

bkase: See ExternalTransition.t and ExternalTransition.Validated.t in the OCaml codebase for an example of this

src/lib/zkapp.ts
```diff
@@ -643,38 +661,69 @@ async function callUnproved<S extends typeof SmartContract>(
...
   });
-  let txJson = tx.toJSON();
-  if (zkappKey !== undefined)
-    txJson = Ledger.signOtherParty(txJson, zkappKey, 0);
+  let txJson = tx.sign().toJSON();
   return txJson;
 }
 
-async function signJsonTransaction(
+async function addFeePayer(
```
bkase, line 673: We should purely functionalize here too -- this one is especially dangerous to use mutation because of the async/await

bkase (changes requested):

src/lib/party.ts
mitschabaude: Hmm.. So currently, our model for building up the transaction involves actions which mutate the SmartContract's `this.self` party, e.g.
```ts
this.balance.addInPlace(100); // mutates party.body.balance_change
this.x.set(1); // mutates party.body.update.app_state
this.self.update.permissions.setValue(...);
```
I'm not so excited about breaking this flow by introducing a command that replaces the entire party:
```ts
this.self = this.self.sign();
```

So how about this: I could make the `.sign` method on `Party` purely functional, but add another `.sign` method on SmarContract which performs the mutation `this.self = this.self.sign()`?

mitschabaude: I started doing this
https://github.com/o1-labs/snarkyjs/pull/123/commits/65937625263c6e9f464cd83f8ebd9aec9d335728
https://github.com/o1-labs/snarkyjs/pull/123/commits/7753a674552cbf9e1585a82eedce620ad3b6d336

I'm still having doubts that this is a good approach

src/lib/party.ts
mitschabaude: done, I made a type `PartiesValidated` where the signature can't be lazy https://github.com/o1-labs/snarkyjs/pull/123/commits/f10aca039cfda0630ffddbbc9ca3327f814a4d40

src/lib/zkapp.ts
mitschabaude: done

mitschabaude:
I found a nice way to copy party types: https://github.com/o1-labs/snarkyjs/pull/123/commits/f74daf81673ba83ec38b0c084f2bc44913be6991
With this, I made some steps towards functionalizing: https://github.com/o1-labs/snarkyjs/pull/123/commits/4cc467f388ec2a7687bb06053d76ed777bed2578

I think everything is addressed @bkase

bkase (approved):

PR #126, opened 2022-04-22
Enable interactions with Berkeley QANet

### LINKED ISSUE #51: Throw an error if a transaction fails
Currently transactions tend to fail silently, which is discovered later when an account is not available or a state update didn't happen

### LINKED ISSUE #104: Create good API for proving
EDIT (gregor):

* Just *call* methods inside transaction blocks to trigger (lazy?) proof creation
* Have an option like `proofLevel: "none"` to skip creating proofs (use dummy proofs) in the mock interface. Should also be recognize during deploy to add a dummy verification program.

### LINKED ISSUE #113: Integrate account fetching into SmartContracts
* Use the existing fetch methods magically within `createTransaction` and similar places to fetch account state from an API
* Add a cache to store existing fetch results in
* Refactor createTransaction, such that it either uses account state from the cache, or marks it as "to be fetched" (not fetch it directly!). For proof transactions, this enables us to first run it "test mode", then asynchronously fetch the missing data, then run it again (with all the data available) inside the synchronous prover. For non-proof transactions, there can be a simple async wrapper that does those three steps in sequence.
* Expose a method to explicitly populate the cache with arbitrary data, for testing scenarios / optional fine-grained control over where the data comes from (e.g. from disk instead of from an API).

Small TODO which should be enabled by this:
* Add nonce precondition to _all_ signed parties - currently only done in `createSigned`

### LINKED ISSUE #129: Implement the `Mina` interface for QAnet interactions
Analogous to the existing `LocalBlockchain` which implements `Mina`, we need to create an interface for interacting with a real, deployed network. For now, this will be the Berkeley QANet.

### LINKED ISSUE #145: Remove the `isSameAsFeePayer` option in `Party.createSigned`

mitschabaude:
* Closes #113
* Closes #129 
* Closes #104 
* Closes https://github.com/o1-labs/zkapp-cli/issues/131
* Closes #51
* Closes #145
* Add more examples
* Fix some unrelated things


mitschabaude:
At this point, everything except the integration with proving is done

src/lib/mina.ts
```diff
@@ -163,69 +205,151 @@ export function LocalBlockchain({
...
+      await Fetch.fetchMissingAccounts(graphqlEndpoint);
+      return createTransaction(sender, f, { fetchMode: 'cached' });
+    },
     testAccounts,
   };
 }
 
+function BerkeleyQANet(graphqlEndpoint: string): Testnet {
+  let whaleKey = PrivateKey.fromBase58(
+    'EKEtDLmauasBxaB3FsPVLHBdFHUoQMYa21y8phQwB9dV1fL4M2Cb'
```
bkase, line 309: this key is pwned now as well

src/lib/zkapp.ts
```diff
@@ -67,26 +78,26 @@ function createState<A>() {
     },
 
     getLayout() {
-      const layout: Map<string, { offset: number; length: number }> =
+      let layout: Map<string, { offset: number; length: number }> =
```
bkase, line 81: idk if I agree with the let -> const switching, but we can hash that out later

src/lib/fetch.ts
```diff
@@ -1,100 +1,385 @@
...
+ *
+ * If an error is returned by the specified endpoint, an error is thrown. Otherwise,
+ * the data is returned.
+ *
+ * @param publicKey The specified account to get account information on
+ * @param graphqlEndpoint The graphql endpoint to fetch from
+ * @param config An object that exposes an additional timeout option
+ * @returns zkapp information on the specified account or an error is thrown
+ */
+async function fetchAccount(
```
bkase, line 39: should we also be checking the account cache here in addition to writing to it?

src/lib/mina.ts
```diff
@@ -163,69 +205,151 @@ export function LocalBlockchain({
...
-    transaction,
-    applyJsonTransaction,
-    addAccount,
+    accountCreationFee: () => accountCreationFee,
+    currentSlot() {
+      throw Error(
+        'currentSlot() is not implemented yet for remote blockchains.'
+      );
+    },
+    getAccount(publicKey: PublicKey) {
```
bkase, line 265: or here is where we check the cache I guess?

bkase (changes requested):

src/examples/simple_zkapp.mjs
```diff
@@ -46,31 +38,26 @@ let account1 = Local.testAccounts[0].privateKey;
 let zkappKey = PrivateKey.random();
 let zkappAddress = zkappKey.toPublicKey();
 
-let initialBalance = 10_000_000_000;
 let initialState = Field(1);
+let zkapp = new SimpleZkapp(zkappAddress);
+
+console.log('compile');
+SimpleZkapp.compile(zkappAddress);
```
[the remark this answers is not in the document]
mitschabaude: If you didn't call this, down below the following error would be thrown:
`Error: Cannot prove execution of update(), no prover found. Try calling `await SimpleZkapp.compile()` first, this will cache provers in the background.`

Also, the deploy transaction would silently have now verification key in it (which wouldn't matter, because the mock interface doesn't require one)

One idea I had was to introduce a parameter on the `Mina` implementations, called `proofLevel` or similar, that would determine whether actual proofs and verification keys were required or not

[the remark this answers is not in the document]
mitschabaude: Re other questions: compile takes ages, and blocks the main thread if you do it like here

Compiling automatically whenever it's needed makes sense to me -- but I would possibly log some warning that it was triggered implicitly, so that devs don't think snarkyjs hangs for no reason

src/examples/simple_zkapp_berkeley.ts
```diff
@@ -0,0 +1,119 @@
...
+
+// hard-coded test account which has enough MINA to fund our zkapps
+let whaleKey = Berkeley.testAccounts[0].privateKey;
+let { account: whaleAccount } = await fetchAccount(whaleKey.toPublicKey());
+let { nonce, balance } = whaleAccount!;
+console.log(`using whale account with nonce ${nonce}, balance ${balance}`);
+
+// let zkappKey = PrivateKey.random();
+let zkappKey = PrivateKey.fromBase58(
+  'EKFQZG2RuLMYyDsC9RGE5Y8gQGefkbUUUyEhFbgRRMHGgoF9eKpY'
```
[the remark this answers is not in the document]
mitschabaude: as explained, this is a random key generated by me just for testing. should I still remove this line?

src/examples/simple_zkapp_berkeley.ts
```diff
@@ -0,0 +1,119 @@
...
+
+let transactionFee = 100_000_000;
+let initialState = Field(1);
+let doCompile = true;
+
+// compile the SmartContract to get the verification key (if deploying) or the cache the provers (if updating)
+// this can take a while...
+console.log('Compiling smart contract...');
+let { verificationKey } = doCompile
+  ? await compile(SimpleZkapp, zkappAddress)
```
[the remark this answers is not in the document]
mitschabaude: oh.. there are two versions, this mostly just calls `SimpleZkapp.compile`, do you think I should collapse them to one API? The zkapp cli uses this one right now

src/lib/fetch.ts
mitschabaude: hmm, I feel that it's nice to have one explicit method to definitely perform the fetch and bypass the cache

src/lib/mina.ts
mitschabaude: this only works when the cache is filled, to be synchronously callable

src/lib/mina.ts
mitschabaude: should I remove it? I'm worried about losing the smooth testing flow this enables

src/examples/sudoku/sudoku-zkapp.ts
```diff
@@ -0,0 +1,158 @@
...
+Mina.setActiveInstance(Local);
+const account1 = Local.testAccounts[0].privateKey;
+
+const zkappPrivkey = PrivateKey.random();
+let zkappAddress = zkappPrivkey.toPublicKey();
+
+async function deploy(sudoku: number[][]) {
+  let tx = await Mina.transaction(account1, () => {
+    const p = Party.createSigned(account1, { isSameAsFeePayer: true });
+    p.balance.subInPlace(Mina.accountCreationFee());
```
MartinMinkov, line 116: Here as well, could we default this behavior in `createdSigned`?

src/examples/sudoku/sudoku-zkapp.ts
```diff
@@ -0,0 +1,158 @@
...
+    let zkapp = new SudokuZkapp(zkappAddress);
+    let sudokuInstance = new Sudoku(sudoku);
+    zkapp.deploy({ zkappKey: zkappPrivkey });
```
MartinMinkov, line 119: Makes sense to rename `zkappKey` to `zkappPrivateKey` since that's how it's being used anyways :P

src/lib/party.ts
```diff
@@ -836,6 +860,57 @@ function addMissingSignatures(
   };
 }
 
+type PartiesProved = {
+  feePayer: FeePayer;
+  otherParties: (Party & { authorization: Control | LazySignature })[];
+};
+
+async function addMissingProofs(parties: Parties): Promise<PartiesProved> {
```
MartinMinkov, line 868: When does this get called or is intended to be called?

src/examples/simple_zkapp_berkeley.ts
```diff
@@ -0,0 +1,119 @@
...
+let zkappAddress = zkappKey.toPublicKey();
+
+let transactionFee = 100_000_000;
+let initialState = Field(1);
+let doCompile = true;
+
+// compile the SmartContract to get the verification key (if deploying) or the cache the provers (if updating)
+// this can take a while...
+console.log('Compiling smart contract...');
+let { verificationKey } = doCompile
```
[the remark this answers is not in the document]
mitschabaude: Nah, you're right, it's a testing artifact

src/examples/simple_zkapp_berkeley.ts
```diff
@@ -0,0 +1,119 @@
...
+
+// if the zkapp is not deployed yet, create a deploy transaction
+if (!isDeployed) {
+  console.log(`Deploying zkapp for public key ${zkappAddress.toBase58()}.`);
+  // the `transaction()` interface
+  let transaction = await Mina.transaction(
+    { privateKey: whaleKey, fee: transactionFee },
+    () => {
+      const p = Party.createSigned(whaleKey, { isSameAsFeePayer: true });
+      p.balance.subInPlace(Mina.accountCreationFee());
```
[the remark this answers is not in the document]
mitschabaude: Imo Party.createSigned is too important and general-purpose to make it default to that. But we could add another helper that achieves this use case in one line. Something like `fundNewAccount(feePayerKey)`?

src/lib/party.ts
mitschabaude: This is called under the hood by `transaction.prove()`

src/examples/simple_zkapp_berkeley.ts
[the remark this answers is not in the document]
MartinMinkov: I like the idea of a simple wrapper/helper for this!

src/examples/sudoku/sudoku-zkapp.ts
mitschabaude: I renamed `zkappPrivkey` -> `zkappKey` instead :D In typical crypto lingo, a "key" is a private key. A public key is an "address". So I don't think we have to be that verbose. (This also applies to `feePayerKey` vs `feePayerPrivateKey` @jasongitmail)

src/examples/simple_zkapp_berkeley.ts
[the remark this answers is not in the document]
mitschabaude: done

src/lib/mina.ts
mitschabaude: removed the testAccounts, including this key, now

src/examples/simple_zkapp_berkeley.ts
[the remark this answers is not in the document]
mitschabaude: I added this wrapper now!

mitschabaude:
@bkase I think I addressed everything. In particular, removed the whale key

src/lib/fetch.ts
bkase: Fair

bkase (approved):
Great!

PR #149, opened 2022-05-07
Fix bug in decrypt funciton

qcomps:
Fixes https://github.com/o1-labs/snarkyjs/issues/148

## Description of Bug

Decrypting a value longer than one field results in an error being thrown in the decrypt function `Error: assertEquals: not equal`.  The root cause of the error is that the sponge is meant to absorb the final chunk of ciphertext when it is on the final iteration of decrypting chunks, but the check for being on the final iteration is looking at the length of the plaintext that is being built, rather than the ciphertext.  This results in `ciphertext[i]` being absorbed by the sponge on every iteration.

## Fix

Simply replace the check from length of message to length of ciphertext

### DIFF
--- a/src/examples/encryption.ts
+++ b/src/examples/encryption.ts
@@ -35,4 +35,51 @@ Circuit.runAndCheckSync(() => {
   });
 });
 
+// With a longer message
+message = JSON.stringify({
+  'coinbase': {
+    'btc': 40000.00,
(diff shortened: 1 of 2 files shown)


mitschabaude:
Nice catch, thank you so much!

mitschabaude (approved):

PR #154, opened 2022-05-10
Integrate `deploy()` with qanet

mitschabaude:
This updates the top-level `deploy()` method to be actually usable with a remote blockchain, by using `Mina.transaction` which -- depending on the currently set Mina instance -- will fetch missing account info.

not finished, there are some further possibilities for cleanup in `deploy`.

TODO to make this actually usable by zkapp-cli:
* [ ] properly keep track of zkapp nonce

### DIFF
--- a/src/lib/mina.ts
+++ b/src/lib/mina.ts
@@ -20,7 +20,6 @@ export {
   BerkeleyQANet,
   LocalBlockchain,
   nextTransactionId,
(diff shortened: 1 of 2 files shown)


mitschabaude:
everything this PR had in it is merged. closing for house-keeping, the issue it wanted to solve is not urgent

PR #155, opened 2022-05-12
Adding String Type

qcomps:
## Problem
A ZKapp might want to manipulate "string"-like data or make assertions about a string other than its complete equality with another string.  Bijective encoding is an existing string to field representation, but it cannot fulfill every task that we might want from a string.  A canonical use case could be asserting that the string `Today's date: 05/12/2022` contains the string `05/12/2022`.

## Solution
Adding a string type which can be manipulated char by char, which is just a wrapper of an array of fields.

## Drawbacks
For now, each character takes up a whole field of storage.  This implementation includes a lot of looping and comparison at the per-character level which can't be super fast.

Fixes https://github.com/o1-labs/snarkyjs/issues/92


src/lib/zkstring.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  static fromString(str: string) {
+    const char = Field(str.charCodeAt(0));
+    return new Character(char);
+  }
+
+  static validateInput(value: Field): Boolean {
+    if (true) {
+      return true;
```
qcomps, line 34: Here could add things like is within certain range e.g. utf-8 1-byte characters, or any utf-8 character, or even this could be an abstract class and different character sets would need to implement their own validations?

src/lib/zkstring.ts
```diff
@@ -0,0 +1,85 @@
...
+  toFields(): Field[] {
+    return this.values.map(x => x.toField());
+  }
+
+  toString(): string {
+    return this.values.map(x => x.toString()).join('');
+  }
+
+  equals(str: ZKString): Bool {
+    try {
```
qcomps, line 67: Not sure this is valid in a circuit.  I had some index out of bounds issues with some examples, and thought it would be nice to ensure that lengths are equal before continuing.

qcomps:
@mitschabaude I dropped my original idea of packing bits into a Field because that never came up in any discussions about use-cases.  This implementation could probably be further optimized to not use a full Field per character.  I'm not really sure how much value there is in making that optimization though.  I wanted to get your thoughts on this so far.

mitschabaude:
Awesome! 👀

jasongitmail:
🙌 Awesome. Thanks for submitting this @qcomps

src/examples/encryption.ts
```diff
@@ -69,7 +69,7 @@ console.log(`Recovered message: "${decryptedMessage}"`);
 
 // the same but in a checked computation
 
-Circuit.runAndCheckSync(() => {
+Circuit.runAndCheck(() => {
```
qcomps, line 72: this must have been a typo?  Fixing.

qcomps:
@mitschabaude 

Ok, continued with this idea, and should be good for an actual review.  I'm not super confident with the checked computation stuff, and made the design decision that when things like `validateInput` and `Zkstring.contains(otherZkstring)` are run outside of checked computations, they just return true.  Not sure if that meets the spec of snarkyjs in general.

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+  static fromString(str: string) {
+    const char = Field(str.charCodeAt(0));
+    return new Character(char);
+  }
+
+  /* 
+  TODO: Add support for more character sets
+  ASCII reference: https://en.wikipedia.org/wiki/ASCII#Printable_characters
+  */
+  static validateInput(value: Field): Boolean {
```
mitschabaude, line 39: This function should be called `check`, then it will run automatically when a witness is created, so ZkStrings that are passed to a method (for example) will be constrained inside the circuit

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
+import { Bool, Circuit, Field, Poseidon } from '../snarky';
+import { CircuitValue, prop } from './circuit_value';
+
+export class Character extends CircuitValue {
+  @prop value: Field;
+
+  constructor(value: Field) {
+    super();
+
+    if (Character.validateInput(value)) {
```
mitschabaude, line 10: I don't think we should call validateInput in the constructor, the pattern we've been using is to let the constructor be dumb and use .check, see below

src/lib/zkstring.ts
mitschabaude: Also, let's remove the check for inProver / inCheckedComputation, if we don't call it implicitly in the constructor as suggested above

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  /* 
+  TODO: Add support for more character sets
+  ASCII reference: https://en.wikipedia.org/wiki/ASCII#Printable_characters
+  */
+  static validateInput(value: Field): Boolean {
+    if (Circuit.inProver() || Circuit.inCheckedComputation()) {
+      const ret = Bool.and(
+        value.gte(Field(32)),
```
mitschabaude, line 42: I would go for a full bit range, 0 to 255, because that can be implemented more efficiently.
Use rangeCheckHelper(8), like in UInt32

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+        value.lte(Field(126))
+      );
+
+      return ret.toBoolean();
```
mitschabaude, line 46: toBoolean breaks the model that this can be a variable. This should just add assertions and don't return anything

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return true;
+  }
+};
+
+export class ZKString extends CircuitValue {
+  @prop values: Character[];
```
mitschabaude, line 54: Character[] does not extend CircuitValue, so this won't work. We need arrayProp, and need to make the string fixed-width

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  constructor(values: Character[]) {
+    super();
+
+    this.values = values;
+  }
+
+  append(str: ZKString): ZKString {
+    return new ZKString([...this.values, ...str.values])
```
mitschabaude, line 63: This works re constraints (it's just reorganizing existing variables). But it should take fixed-length strings and return a string with a different fixed length

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  private _contains(str: ZKString): Bool {
+    let ret = new Bool(false);
+    const thisLength = Number(this.length().toString());
+    const strLength = Number(str.length().toString());
+    let i = 0;
+
+    while (i + strLength <= thisLength) {
+      ret = ret.or(this.substring(i, i + strLength).equals(str))
```
mitschabaude, line 132: Nice, this works 💯

mitschabaude: (for fixed length strings, that is)

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    return ret;
+  }
+
+  private _equals(str: ZKString): Bool {
+    let ret = new Bool(true);
+
+    this.values.forEach((value: Character, i: number) => {
+      ret = ret.and(value.equals(str.values[i]));
```
mitschabaude, line 143: 👍🏻

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+  hash(): Field {
+    return Poseidon.hash(this.values.map(x => x.value));
+  }
+
+  length(): Field {
+    return Field(this.values.length);
+  }
+
+  substring(start: number, end: number): ZKString {
+    return new ZKString(this.values.slice(start, end));
```
mitschabaude, line 93: 👍🏻

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+    this.values = values;
+  }
+
+  append(str: ZKString): ZKString {
+    return new ZKString([...this.values, ...str.values])
+  }
+
+  // returns TRUE if str is found in this zkstring
+  contains(str: ZKString): Bool {
+    if (Circuit.inProver() || Circuit.inCheckedComputation()) {
```
mitschabaude, line 68: Let's not discriminate between in-snark and out-of-snark like this. We want operations to work and be the same in both scenarios.

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+      let ret = new Bool(false);
+      const strLength = str.length();
+
+      Circuit.if(
+        this.length().gt(strLength),
+        ret = new Bool(false),
+        ret = this._contains(str)
+      )
```
mitschabaude, line 76: I think if lengths are static (as required above), we don't need to include length checks, just _contains. Could just throw an error if it doesn't match length constraints

src/lib/zkstring.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  toFields(): Field[] {
+    return this.values.map(x => x.toField());
+  }
+
+  toString(): string {
+    return this.values.map(x => x.toString()).join('');
+  }
+
+  equals(str: ZKString): Bool {
```
mitschabaude, line 104: Same comments here as for contains()

mitschabaude (changes requested):
Nice work. We'll need to find a way to make the strings fixed length though, which is likely a major refractor. Otherwise I just requested some quick to make changes

mitschabaude:
> For now, each character takes up a whole field of storage. This implementation includes a lot of looping and comparison at the per-character level which can't be super fast.

I don't think there's a more efficient way to do this. It's not really _storage_ that has to minimized in snarks, but _constraints_. Usually any bit packing would add a ton of boolean constraints which makes it much more expensive than an implementation like this. If operations are on the character level, 1 Field per character feels optimal.

mitschabaude:
> I'm not super confident with the checked computation stuff, and made the design decision that when things like validateInput and Zkstring.contains(otherZkstring) are run outside of checked computations, they just return true. Not sure if that meets the spec of snarkyjs in general.

General design goal of snarkyjs was to make adding constraints feel like the same as just doing operations. So, a thing like `x.add(y)` should just do an addition outside the circuit, but also add constraints inside the circuit, in a way that the user doesn't have to think that hard about circuits/constraints.

Therefore, not making it work as expected outside the circuit kind of breaks our model, and I'd recommend to change that

mitschabaude:
Some comments on naming:
* I think we should spell it `ZkString` rather than `ZKString`, this is proper camel case style IMHO (https://stackoverflow.com/questions/15526107/acronyms-in-camelcase)
* You might consider `CircuitString` instead, which would fit well with our existing `CircuitValue`

qcomps:
I like `CircuitString` and will change to that for the next commit

src/lib/zkstring.ts
qcomps: @mitschabaude do you mean I shouldn't do this if statement?  I was getting errors trying to run `gt` outside of a checked computation.  I'm not sure about this specific circumstance, but I was getting errors in general so added this line basically everywhere.

Or do you just mean I should replicate the same behavior in or out?

src/lib/zkstring.ts
qcomps: so `check` is a magic method name?  Are there others?

src/lib/zkstring.ts
qcomps: Hmm, the range here was intentionally chosen from: https://en.wikipedia.org/wiki/ASCII#Printable_characters

I didn't want to get too deep into the implementation of any one character set, but this range is specifically printable ascii characters.  In the future, people will probably want much richer characters which will require even more rules.  Is this less efficient version going to be like unacceptably slow?  If so, it seems to be a soft blocker on developing this string concept any further.

qcomps:
@mitschabaude 


TODO:

I want to comb through some nit details that I might have overlooked during my latest coding session.  I also want to move test cases from "examples" to an actual test and make the examples simpler and more about showing users the API than testing edge cases.

With that in mind, I'd appreciate another look over what I have now.  Do you think `usedLength` will work the way I have it, for instance in `append` or `contains`?  Or am I cheating somehow by referencing that property?  Does the null character make sense?

src/lib/circuit_string.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    // set the value of this.values to the input values, filled with null char
+    this.values = values.fill(new NullCharacter, usedLength);
+
+    this.usedLength = Field(usedLength);
+    this.maxLength = Field(maxLength);
+  }
+}
+
+export async function resolveCircuitStringArrayProps() {
```
qcomps, line 174: 💀

src/lib/circuit_string.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    return ret;
+  }
+};
+
+export class CircuitString8 extends CircuitString {
+  constructor(values: Character[]) {
+    super(values);
+
+    const usedLength = values.length;
```
qcomps, line 162: After calling super, this value is actually being _reset_ - is that allowed?  Is it better to break this logic into a separate static method to be implemented by each class?

src/lib/zkstring.ts
mitschabaude: I think `.gt` is not needed here, because you don't have to constrain the lengths in the circuit if they are static. So you can remove the if statement, and the length check (or check the length in pure JS if you want to handle the error case). In any case, the behavior should be the same in & out

src/lib/zkstring.ts
mitschabaude: no there aren't, at least for CircuitValue 😅

src/lib/zkstring.ts
mitschabaude: Gotcha. It's not _unacceptably_ slow. However, a full bit range can be made very fast with lookup tables (coming soon), and detailed ranges will probably never be as fast.

In the current implementation, with `String.fromCharCode` as the mapping between numbers and chars, I think the range 0 to 255 is just fine. It can be converted back and forth to a string unambiguously. So I don't see the argument for introducing a performance penalty, just so it belongs to a set of "printable" characters.

src/lib/circuit_string.ts
mitschabaude: Yeah I see no problem with that in particular

src/lib/circuit_string.ts
```diff
@@ -100,10 +130,10 @@ export class CircuitString extends CircuitValue {
...
-  private _contains(str: CircuitString): Bool {
+  private _contains(str: CircuitString, size: number): Bool {
     let ret = new Bool(false);
-    const thisLength = Number(this.length().toString());
-    const strLength = Number(str.length().toString());
+    const thisLength = Number(this.usedLength.toString());
+    const strLength = size;
     let i = 0;
 
     while (i + strLength <= thisLength) {
```
mitschabaude, line 138: like in `append`, here you are fixing the circuit to the length of the string. Again I think that could be avoided

src/lib/circuit_string.ts
```diff
@@ -32,38 +34,70 @@ export class Character extends CircuitValue {
...
 
-    this.values = values;
+    const usedLength = values.length;
+    const maxLength = DEFAULT_STRING_LENGTH;
+    values.length = maxLength;
+
+    // set the value of this.values to the input values, filled with null char
+    this.values = values.fill(new NullCharacter, usedLength);
+
+    this.usedLength = Field(usedLength);
```
mitschabaude, line 70: After seeing some of the problems in your circuits below, I think it's probably not helpful to store the `usedLength` on the class. Not storing it would force you to write your circuits in a way that they are agnostic to the actual length, instead using the static `maxLength`. This is the way to ensure that the usedLength actually stays dynamic, and not ends up being a constant that gets baked into the circuit

test/circuit_string.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      Circuit.runAndCheck(() => {
+        const str1 = CircuitString.fromString('abcd')
+        const str2 = CircuitString.fromString('efgh')
+        expect(str1.append(str2).toString()).toBe('abcdefgh');
+      });
+    });
+  });
+
+  describe('CircuitString8', () => {
+    // This test passes but it should not
```
qcomps, line 142: @mitschabaude any ideas here?  I would expect in a checked computation for a `String8` to either fail to init, or to only accept 8 characters.

test/circuit_string.test.ts
```diff
@@ -0,0 +1,164 @@
...
+      Circuit.runAndCheck(() => {
+        const str = CircuitString8.fromString('More than eight chars')
+        expect(str.toString()).toBe('More than eight chars')
+      });
+    })
+  });
+
+  describe('with invalid input', () => {
+    // This test passes but it should not
+    test('cannot use a character out of range', () => {
```
qcomps, line 153: @mitschabaude Similar issue here, I would expect `Field(10000)` to be an invalid Character and for `#check` to throw an error on init for a character.

mitschabaude:
@qcomps I fixed arrayProp so you don't have to `await isReady` before using it: https://github.com/o1-labs/snarkyjs/pull/173
I encourage you to merge that PR into yours so you can use it more easily

mitschabaude (approved):
I fixed up some things and commented the non-working parts for now (namely, `.contains` and the smaller `CircuitString8`).
I think this is good to merge, also has an important fix to CircuitValue now, plus the fix of arrayProp.

Nice work @qcomps, big thanks for helping with this!

PR #158, opened 2022-05-13
Auto-generate Parties types and JSON conversion

### LINKED ISSUE #132: Unify types in snarky.d.ts and party.ts
things like `Party` vs `Party_` etc

mitschabaude:
* Companion of https://github.com/MinaProtocol/mina/pull/10938
* add auto-generated types for json parties (i.e. the JSON.parse'd verison)
* add auto-generated types in a JS version, which is unified with `Parties` (https://github.com/o1-labs/snarkyjs/issues/132)
* add toJson conversion function between the two types above, relying on hand-written (but trivial) converters for the leafs and a recursive function for composed types like Parties
* add toFields conversion function from JS types to Field[], plus hand-written converters for leafs
* replace existing strategy for generating JSON transactions (which involved converting Parties to OCaml) with one that creates the JSON from JS directly
* replace existing strategy for hashing a party (which involved converting he party to OCaml) with one that converts the party to a list of Fields in JS, and reconstruct the Party.Checked.t from that in OCaml (the latter is snarky built-in functionality)
* remove differences of Party in TS with autogenerated type to a minimum; remove almost all conversion code
* Closes #132


mitschabaude:
Update: toJson works and is compatible with the original
TODO: actually use new toJson in place of old

src/lib/int.ts
```diff
@@ -381,6 +381,14 @@ export class Int64 {
     return new Int64(this.value.neg());
   }
 
+  static add(x: Int64, y: Int64 | UInt32 | UInt64) {
+    return new Int64(x.value.add(y.value));
```
bkase, line 385: can you delegate to the other add method here just to minimize code paths: ie.
`return new Int64(x.value).add(y.value)`

src/lib/int.ts
```diff
@@ -381,6 +381,14 @@ export class Int64 {
...
+  }
+
+  static sub(x: Int64, y: Int64 | UInt32 | UInt64) {
+    return new Int64(x.value.sub(y.value));
```
bkase, line 389: same here

src/snarky/js-layout.ts
```diff
@@ -0,0 +1,2344 @@
+export { jsLayout };
```
bkase, line 1: can you put this in a `gen/` folder (I'm assuming this is autogenerated). I think we should still check it in so we can codereview when it changes etc

bkase: what do we need to do in CI to make sure a test fails if the layout changes and it breaks something in snarkyjs?

src/snarky/parties-json.ts
```diff
@@ -0,0 +1,387 @@
+// this file is auto-generated - don't edit it directly
```
bkase, line 1: can this also go to `gen/`

src/snarky/parties-leaves-json.ts
```diff
@@ -0,0 +1,78 @@
...
+type GlobalSlot = UInt32;
+type CurrencyAmount = UInt64;
+type StateHash = Field;
+type Fee = UInt64;
+type BlockTime = UInt32;
+type TokenId = Field;
+
+// to what types in the js layout are mapped
+type TypeMap = {
+  PublicKey: PublicKey;
```
bkase, line 52: (I think) you can use field punning to reduce the noise here:

```typescript
type TypeMap = {
  PublicKey,
  Field,
  /* ... etc */
}
```

src/snarky/parties-leaves.ts
```diff
@@ -0,0 +1,249 @@
+import { Field, Bool, Group, Ledger } from '../snarky';
```
bkase, line 1: what is the difference between this file and `parties-leaves-json`? seems like they can share more content -- are these autogenerated?

bkase (changes requested):

src/build/jsLayoutToTypes.mjs
```diff
@@ -0,0 +1,223 @@
+import fs from 'node:fs/promises';
```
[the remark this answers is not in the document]
mitschabaude: so, currently, /build only contains utilities for building, not build artifacts (those live in /dist)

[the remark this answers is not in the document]
mitschabaude: I think it fits in that directory very well.. but I could rename the directory

src/snarky/parties-leaves-json.ts
mitschabaude: no, I checked.. for types it doesn't work

src/snarky/parties-leaves.ts
mitschabaude: No these are written manually!

There are two "Party" types - one which directly corresponds to the JSON schema, which is (autogenerated) in `parties-json.ts`, and one which corresponds to our existing snarkyjs types -- so it can be used in a circuit -- which is (autogenerated) in `parties.ts`.
For both of those types, we need to manually specify what the primitive types, i.e. the "scalars" in the graphql schema, should be. This is what's done in the `-leaves` files. So, for example,
* `parties-leaves-json.ts` declares that "Fee" should become `string`
* `parties-leaves.ts` declares that "Fee" should become `UInt64`

In addition, `parties-leaves.ts` has manually written methods to convert from the snarkyjs leaf types to the JSON leaf types, e.g. it says that the method to convert `UInt64` to `string` is `x => x.toString()`.

Similarly, `parties-leaves.ts` has manually written converters of leaf types to `Field[]`, e.g. `(x: UInt64) => x.toFields()`.

src/snarky/js-layout.ts
mitschabaude: I tried to make sure that, if the layout changes, either:
* 1. snarkyjs continues working, because everything that needs to change changes automatically, or
* 2. there's a type error in OCaml, or
* 3. there's a type error in TS

I think 2.) is the weakest point at the moment, because in some places we rely on strings in OCaml as hints to what JS types should be, and I want to reduce that further in the future.

I tried to ensure 3.) as much as possible; in particular, the manually written conversion functions for toJson / toFields are type-checked.

If there's a type error in TS, CI would fail pretty early, both in the snarkyjs unit test and in the "build JS integration test" job

mitschabaude:
@bkase addressed your feedback, added new folder `src/snarky/gen`

src/snarky/parties-leaves.ts
mitschabaude: and to answer your question: the two files have no potential for code sharing that is worth doing IMO

bkase: can you put this response into a comment on one of the files? (and on the other one say something like "see parties-leaves.ts for more details on what this is for")

bkase (commented):
What is the reasoning for 812b31dbda4effa7b7138331910545f938bce008 ?

I think it should be reverted -- the more specific types help document what to expect in those fields.

Also it opens the opportunity for us to add doccomments to the types (maybe in a follow up) to clearly explain to users what semantic meaning should be given to the number being used in this position (CurrencyAmount vs BlockTime for example)

mitschabaude:
> What is the reasoning for [812b31d](https://github.com/o1-labs/snarkyjs/commit/812b31dbda4effa7b7138331910545f938bce008) ?
> 
> I think it should be reverted -- the more specific types help document what to expect in those fields.
> 
> Also it opens the opportunity for us to add doccomments to the types (maybe in a follow up) to clearly explain to users what semantic meaning should be given to the number being used in this position (CurrencyAmount vs BlockTime for example)

The reasoning is to reduce what has to be added manually to the auto-generated code, or in other words, increase the power of code generation, thus making the process less brittle.

Previously, the JS deriver only knew that there were some types called "Fee", "Balance", "BlockTime", and it was left to manual JS code to specify that these all are implemented as UInt64. That these are all UInt64 is really part of the protocol, which is defined in OCaml, so it seems much more robust to automatically bring this information over from OCaml to JS. In fact, https://github.com/o1-labs/snarkyjs/commit/812b31dbda4effa7b7138331910545f938bce008 fixed a bug - `BlockTime` was a UInt32 in JS before this commit, but it should have been UInt64. Which perfectly illustrates how the change makes the system less brittle.

So, I very strongly feel the connection between specialized and generic types should be exposed to the JS deriver, and be part of the generated -- not manually written -- code. That said, we could combine this with what you want, and also auto-generate type aliases (from the existing names on scalar types in Ocaml), which are used on the generated party type fields. Then we could also bring doccomments on those type aliases over from Ocaml.

FWIW, I'm against that though, because I don't like type aliases 😅  I think they add opaqueness rather than documentation ("Huh? The function expects a `BlockTime`? What's a `BlockTime`?? *digging through the source code* ... Oh, it it's just a `UInt32`"). 

But yeah, it's an easy addition to make now, which should be made on top of the last commit rather than by reverting it.

bkase (approved):
Makes sense! We can chat about the merits of the aliases later

PR #164, opened 2022-05-16
Add testing spec for int.ts file

MartinMinkov:
**Description**
Adds initial test spec for all classes in the `int.ts` file. This includes `Int64`, `UInt64`, and `UInt32`.

There are a number of bugs uncovered while writing unit tests for these classes, some are documented as issues and others are not. I will be looking through our issues list and adding bugs that are not currently present.

The current tests that fail due to bugs are commented out with their error message as a comment. If one wishes to inspect further, uncomment the test and run `npm run test`


src/lib/int.test.ts
```diff
@@ -0,0 +1,876 @@
...
+
+describe('int', () => {
+  beforeAll(async () => {
+    await isReady;
+  });
+
+  afterAll(async () => {
+    setTimeout(async () => {
+      await shutdown();
+    }, 1500);
```
bkase, line 19: why is there a long timeout here before we shutdown?

bkase (changes requested):
Wow a lot of test cases, nice!

1. Can you make issues all the failing tests?
2. Is there some way we can test the circuits too? I think right now we're testing the "outside the circuit" version of the logic. We may need to do some crazy ocaml stuff in order to do constraint checking without proof generation (this will be fast, but still catch errors) -- maybe @mitschabaude already knows if this is exposed?

mitschabaude:
Oh wow, so many failure cases! This is great to have.

> Is there some way we can test the circuits too? I think right now we're testing the "outside the circuit" version of the logic. We may need to do some crazy ocaml stuff in order to do constraint checking without proof generation (this will be fast, but still catch errors) -- maybe @mitschabaude already knows if this is exposed?

I agree that we'd like to have some in-snark testing as well. There's `Circuit.runAndCheck`, I don't know in detail what it does but I think it's designed for exactly that

src/lib/int.test.ts
bkase: SetTimeout 0 as a defer seems okay to me 👍

bkase (approved):
Awesome!

Issue #179, opened 2022-05-23
Set preconditions based on current account / protocol state


mitschabaude:
# Precondition RFC

If users want to add preconditions on the account or protocol state, these will most often relate to the _current_ on-chain state of these fields. A very intuitive model is to let users just "use" the current state -- for example, let them access the "current account balance". The mental model is that this balance is a variable (not a constant!) that depends on the current chain. It can be implemented by fetching the current balance from the chain, using it in the prover as the value for the variable, and adding a precondition which fixes the balance to exactly this value. If we wouldn't add the precondition, then the balance would be unconstrained, which would be against the intuition that we prove a computation which uses the current balance as input.

Therefore, I think a good default is to just let users access the balance (and other fields) as the property of some object, and behind the scenes add the necessary precondition. We propose the following API to achieve that:

```ts
// inside method
let balance = this.currentAccount.balance; // automagically adds a precondition
balance.assertLte(100e9); // this method can only be called if current balance <= 100 MINA
```

Note that to use this API intuitively, one doesn't need to understand the concept of a precondition at all. It's enough to be aware of the `this.currentAccount` property. The returned `balance` is a plain `UInt64` variable; the added precondition stays completely hidden. Similarly, we propose to have a `this.currentProtocolState` property which exposes protocol state preconditions. On other parties, we would have corresponding `party.currentX` properties.

However, advanced users may want to declare arbitrary preconditions. For example, in the above example, it makes sense to actually _not_ use the current balance, but just to add a precondition directly. We propose the following API:

```ts
Party.assertBetween(this.accountPrecondition.balance, 0, 100e9);
```

This works for any party, not just the `this` party. It mutates `this.accountPrecondition.balance`, which _is_ the precondition (represented as an object `{lower: UInt64, upper: UInt64}`.

The last example achieves the same as the first one -- it constrains the balance to be less than 100 MINA -- but without the fragility that the current balance needs to stay unchanged until the transaction is accepted. However, to use an API like this, you do have to understand the concept of a precondition. That's why it's OK to use more jargon ("precondition"), and make this a bit less discoverable (have to use `Party`), and let it not do any magic. Similar to `this.accountPrecondition`, we propose to expose `this.protocolStatePrecondition`, which would just be shortcuts to access `party.body.accountPrecondition` and `party.body.protocolStatePrecondition`.

[...]

PR #206, opened 2022-05-25
Consolidate preconditions MinaProtocol/mina#11096

bkase:
SnarkyJS side of MinaProtocol/mina#11096 addressing the naming
consistency part of #179


mitschabaude:
if we want to merge this to main, it needs updated bindings - there's a script in Mina for that called `./scripts/update-snarkyjs-bindings.sh`. or we wait for #207 which has this one merged as well

PR #207, opened 2022-05-26
Preconditions

### LINKED ISSUE #179: Set preconditions based on current account / protocol state

mitschabaude:
- Closes #179
- Adds example of using the new API: https://github.com/o1-labs/snarkyjs/blob/dab7ed9aa93fe95fb88f1b184195de252046704c/src/examples/simple_zkapp.ts#L39
- Adds a method for checking deep equality of two party subtypes (where "party subtype" is defined very broadly to be future-proof): https://github.com/o1-labs/snarkyjs/pull/207/commits/5bf05cdc3341b5159498daabb11b1613f313d1a7. This is used to check if the user manually modified a precondition, which counts as explicitly constraining the value, to avoid the error that we throw when it's not constrained.

TODOs:
* [ ] change integration test so that it tests fetching from the graphql URI, and uses preconditions, and ideally is broken up into zkapps that test different features
* [x] investigate broken prover and decide if it should be fixed as part of this issue
* [ ] some precondition fields aren't implemented yet because the logic to parse them from an encoded string is missing. this could be done here or in a follow-up PR


src/lib/fetch.ts
```diff
@@ -218,34 +245,6 @@ function stringifyAccount(account: FlexibleAccount): FetchedAccount {
   };
 }
 
-async function checkResponseStatus(
```
mitschabaude, line 221: this code block was just moved to the bottom of the file. I often move unimportant / helper-type code to the bottom, and  the exported API / most important parts to the top, because I think this makes code more readable

src/lib/precondition.ts
```diff
@@ -0,0 +1,375 @@
...
+  return { account: Account(party), network: Network(party) };
+}
+
+// note: please keep the two precondition implementations separate
+// so we can add customized fields easily
+
+function Network(party: Party): Network {
+  // TODO there should be a less error-prone way of typing js layout
+  // e.g. separate keys list and value object, so that we can access by key
+  let layout = (jsLayout as any).Party.layout[0].value.layout[9].value.layout[0]
```
mitschabaude, line 36: the `jsLayout` provides the necessary info to produce `precondition` interfaces of the right structure at runtime

src/lib/precondition.ts
```diff
@@ -0,0 +1,375 @@
...
+function Account(party: Party): Account {
+  // TODO there should be a less error-prone way of typing js layout
+  // e.g. separate keys list and value object, so that we can access by key
+  let layout = (jsLayout as any).Party.layout[0].value.layout[9].value.layout[1]
+    .value as Layout;
+  let context = getPreconditionContextExn(party);
+  return preconditionClass(layout, 'account', party, context);
+}
+
+function preconditionClass(
```
mitschabaude, line 51: this function is recursive! it produces all the `preconditions.account` and `preconditions.network` fields

src/lib/precondition.ts
```diff
@@ -0,0 +1,375 @@
...
+  preconditionContexts.set(party, {
+    read: new Set(),
+    constrained: new Set(),
+    vars: {},
+    isSelf,
+  });
+}
+
+function assertPreconditionInvariants(party: Party) {
+  let context = getPreconditionContextExn(party);
```
mitschabaude, line 191: this is the function that throws an error if the user used `.get()` on a account / network field, without explicitly adding a precondition. calling `get` would have added an entry to `context.read`. adding a precondition would have added an entry to `context.constrained`.

src/snarky.d.ts
```diff
@@ -417,6 +428,15 @@ declare class Bool {
   toBoolean(): boolean;
 
   /* static members */
+  /**
+   * The constant [[`Bool`]] that is `true`.
+   */
+  static true: Bool;
```
mitschabaude, line 434: I added `Bool.true` and `Bool.false`, because in the end it won the community vote :D

src/lib/int.ts
```diff
@@ -1,15 +1,18 @@
...
+import { Circuit, Field, Types } from '../snarky';
 import { CircuitValue, prop } from './circuit_value';
 
 export { UInt32, UInt64, Int64 };
 
-class UInt64 extends CircuitValue {
+class UInt64 extends CircuitValue implements Types.UInt64 {
   @prop value: Field;
   static NUM_BITS = 64;
+  type = 'UInt64' as const;
```
mitschabaude, line 9: I added this field to `UInt64`, to make the type conform to the interface `{ value: Field; type: "UInt64" }`, which is used to signify `UInt64`s in the autogenerated party types.
the intention is to make the `UInt64` interface distinct from the `UInt32` one, to get advanced types in `precondition.ts`to work. this also fixes a couple of bug sources, where `UInt32` could be passed into functions that expected `UIn64` and vice versa

src/lib/int.ts
```diff
@@ -1,15 +1,18 @@
...
 
   constructor(value: Field) {
     super();
     this.value = value;
+    // TODO type breaks Circuit.if for some reason
+    delete (this as any).type;
```
mitschabaude, line 15: we actually can't keep the `.type` property on the class instance.. it's just for typescript

src/lib/party.ts
```diff
@@ -482,10 +483,26 @@ type LazyControl = Control | LazySignature | LazyProof;
...
+  constructor(body: Body, authorization = {} as LazyControl, isSelf = false) {
     this.body = body;
+    this.authorization = authorization;
+    let { account, network } = Precondition.preconditions(this, isSelf);
+    this.account = account;
+    this.network = network;
+    this.isSelf = isSelf;
+  }
+
+  static clone(party: Party) {
```
mitschabaude, line 502: Before adding this, `Party`s were just cloned with `cloneCircuitValue` directly. I changed it because the new `account` and `network` fields aren't safely clone-able, due to how they use closures / are attached to a party instance. This new cloning will lose precondition-related metadata attached to a party (i.e. which account / network fields were accessed / constrained). this seems appropriate, because cloning a party also doesn't magically add it to the current transaction

mitschabaude (commented):
Some comments to help review:

src/lib/int.ts
bkase: Can we use a phantom type here instead of actually polluting the object at runtime? https://dev.to/busypeoples/notes-on-typescript-phantom-types-kg9 You can put the type parameter on the interface to avoid people having to see it when they use Uint64 .

src/lib/party.ts
bkase: 👍

src/snarky.d.ts
bkase: Nice

src/lib/int.ts
bkase: See below, I think I solved it nicely

bkase (changes requested):
I don’t know if I believe in the WeakMap over storing secret things on the object.
The benefit is you can easily see everything associated with that entity by looking at the class. When there are WeakMap’s floating around, you have to know where to look to find the extra associated data. 

Also, popular and well-thought-out libraries do this (like React), so I don’t think it’s controversial.

What is your reason for reaching for weakmaps here?

If we do end up sticking with WeakMaps, having just one for all the secret things at least mitigates the “hard to find” associated data problem a bit. Rather than one for each kind of associated datum.

mitschabaude:
> I don’t know if I believe in the WeakMap over storing secret things on the object. (...) Also, popular and well-thought-out libraries do this (like React), so I don’t think it’s controversial.

Ok, this is good feedback. I might have overreached with the intention to move all secret props away. Specifically, I'm not sure if I can defend the move of `SmartContract._executionState` to a weakmap, because that prop mainly gets used *internally* by the `SmartContract` class. In that case, I guess it's more natural to make the prop `private` so it's removed from the TS type / doesn't pollute intellisense, but still present at runtime.

> What is your reason for reaching for weakmaps here?

> If we do end up sticking with WeakMaps, having just one for all the secret things at least mitigates the “hard to find” associated data problem a bit. Rather than one for each kind of associated datum.

I do think there are cases where weakmaps are better than extra class props, and it's also better to not have them in one place. For example, the `precondition.ts` module stores metadata associated with each `Party` instance, in a weakmap. I like the weakmap approach in this case because it makes the metadata book-keeping be localized to the module that is concered with it - `precondition.ts`. I think keeping modules encapsulated helps with understanding and maintaining the code. In this case, seeing a map of precondition-related metadata being initialized on the `Party` class in `party.ts` would be more confusing, because there's no other code in `party.ts` concerned with that metadata. It would mix concerns of the `Party` class with concerns of the `precondition` module.

An alternative would be to secretly set a property `party._preconditionData = ...` inside `precondition.ts`, but I think this is just not as clean, not as type-able etc as maintaining a weakmap for that data.

The same reasoning applies for the `@state` logic that's now encapsulated in `state.ts`. We used to store `_states` and `_layout()` fields on `SmartContract` subclasses from inside the `@state` logic. I think it's strictly less confusing and more maintainable now that we have a state-related weakmap inside `state.ts`, and not mess with the `SmartContract`s from that module.

Since only `precondition.ts` / `state.ts` ever need to be concerned with their respective metadata, I don't see the value of having a central weakmap which combines their stuff. Also, the keys and values of that weakmap would be a mix of completely different things if we do that, making accurate typing harder.

TL;DR, I think both of thos weakmaps help with encapsulation, code clarity and strict typing!

@bkase, what do you think of a middle-ground approach where we don't try to avoid private props in principle, but we use weakmaps if a module wants to associate data with objects that originate in a different module? I would revert the `_executionState` weakmap in that case.

src/snarky/parties-leaves.ts
```diff
@@ -12,8 +12,8 @@ export {
 
 export { toJson, toJsonLeafTypes, toFields, toFieldsLeafTypes };
 
-type UInt64 = { value: Field };
-type UInt32 = { value: Field };
+type UInt64 = { value: Field; type: 'UInt64' };
```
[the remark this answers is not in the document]
mitschabaude: Unfortunately doesn't seem to solve my use case.. I need `UInt64` to not extend `UInt32` and vice versa.

I tried the following snippet:
```ts
type FieldLike<N> = { value: Field };
type UInt64 = FieldLike<{ type: 'UInt64' }>;
type UInt32 = FieldLike<{ type: 'UInt32' }>;
type X = UInt32 extends UInt64 ? 'not working' : 'it works!';
```
For me, `X` is of type `"not working"`..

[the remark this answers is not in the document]
mitschabaude: I think my original solution can be made slightly better by making the type prop optional:

```ts
type UInt64 = { value: Field; type?: 'UInt64' };
type UInt32 = { value: Field; type?: 'UInt32' };
```
that way, they don't extend each other and we can just instantiate them without the `type` field at runtime.

it's still not perfect because the `type` field shows up in intellisense

src/lib/fetch.ts
```diff
@@ -317,6 +347,163 @@ function addCachedAccountInternal(
   };
 }
 
+async function fetchLastBlock(graphqlEndpoint = defaultGraphqlEndpoint) {
```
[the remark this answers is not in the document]
mitschabaude: very nice, I did that and it became much faster!

src/lib/fetch.ts
```diff
@@ -317,6 +347,163 @@ function addCachedAccountInternal(
...
+      totalCurrency,
+      slot,
+      slotSinceGenesis,
+      nextEpochData,
+      stakingEpochData,
+    },
+  },
+}: FetchedBlock): NetworkValue {
+  return {
+    snarkedLedgerHash: Field.zero, // TODO
```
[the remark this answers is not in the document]
mitschabaude: done 👍🏻

src/lib/precondition.ts
```diff
@@ -0,0 +1,375 @@
...
+  for (let preconditionPath of context.read) {
+    // check if every precondition that was read was also contrained
+    if (context.constrained.has(preconditionPath)) continue;
+
+    // check if the precondition was modified manually, which is also a valid way of avoiding an error
+    let precondition = getPath(party.body.preconditions, preconditionPath);
+    let dummy = getPath(dummyPreconditions, preconditionPath);
+    if (!circuitValueEquals(precondition, dummy)) continue;
+
+    // we accessed a precondition field but not constrained it explicitly - throw an error
```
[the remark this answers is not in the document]
mitschabaude: I added a file with unit tests. Please check whether you think it covers everything: https://github.com/o1-labs/snarkyjs/pull/207/commits/899d14ce8d349c7d5d7876104f28e26f6c57cb10

src/snarky/parties-leaves.ts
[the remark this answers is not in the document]
bkase: That's weird that they extend each other, shows that I do not understand typescript's type system enough.

I think this is good with them optional -- since it appears in intellisense can you use some kind of prefix so people know to avoid it: like `_type` or `$type` (I'm not sure if we have a standard for this yet)

bkase:
Renaming the `type?` to `_type?` (or something) + refactoring that WeakMap for executionState and then I'll approve 💯

bkase (approved):

PR #223, opened 2022-05-31
Restrict Group.fromJSON type to an object literal

MartinMinkov:
**Description**
Fixes the weird behavior outlined in https://github.com/o1-labs/snarkyjs/issues/210

The `Group.fromJSON` function was defined in the `snarky.d.ts` file to accept parameters that the OCaml side was not handling. This PR changes the types that `Group.fromJSON` expects to be only valid ones, being an object literal with `x` and `y` values defined as either numbers or strings.


mitschabaude:
thanks!

PR #229, opened 2022-06-06
Remove spin lock

mitschabaude:
- update bindings
- 0.4.0
- 0.4.1
- fix serve:web script
- don't modify function.name, not supported well
- some code cleanup, logging
- add missing methods to worker_spec; fixes proving
- add logging to web worker intf


mitschabaude:
stale, and doesn't contain any valuable work on the issue yet

PR #235, opened 2022-06-06
`@state` refactor

mitschabaude:
* moves the logic required for the `@state` decorator into its own file, `state.ts`. the original file `zkapp.ts` was already too big to comfortably navigate. https://github.com/o1-labs/snarkyjs/pull/235/commits/484d4a263108265b4180634f7f7a0cfc79ae4ffe
* moves secret / internal methods from the object returned by `State()`, to standalone functions that are internal to our module. this avoids mixing internal interfaces with public APIs. https://github.com/o1-labs/snarkyjs/pull/235/commits/40d9ce845883ff9c277f1e8f886781eecd1010f0
* move secret properties that were stored on subclasses of `SmartContract` to an internal `WeakMap`. this also avoids mixing internal interfaces with public APIs, and makes `@state` code more self-contained / easier to reason about. https://github.com/o1-labs/snarkyjs/pull/235/commits/e881a921a7cbc7fa427615ccea26251ba9f2da5a

moving away from properties secretly stored on public-facing objects, or objects from other modules, is a general direction I want to take, so this would be a good opportunity to speak up against it (if you're opposed)

in general I think using `WeakMap` is a good way of doing book-keeping à la "this `State` belongs to this `SmartContract`"

### DIFF
--- a/src/lib/state.ts
+++ b/src/lib/state.ts
@@ -17,12 +17,115 @@ type State<A> = {
   fetch(): Promise<A | undefined>;
   assertEquals(a: A): void;
 };
-
-/**
- * Gettable and settable state that can be checked for equality.
- */
 function State<A>(): State<A> {
   return createState<A>();
+}
+
+/**
+ * A decorator to use within a zkapp to indicate what will be stored on-chain.
+ * For example, if you want to store a field element `some_state` in a zkapp,
+ * you can use the following in the declaration of your zkapp:
+ *
+ * ```
+ * @state(Field) some_state = State<Field>();
+ * ```
+ *
+ */
+function state<A>(ty: AsFieldElements<A>) {
+  return function (
+    target: SmartContract & { constructor: any },
+    key: string,
+    _descriptor?: PropertyDescriptor
+  ) {
+    const ZkappClass = target.constructor;
+    if (reservedPropNames.has(key)) {
+      throw Error(`Property name ${key} is reserved.`);
+    }
+
+    if (ZkappClass._states == undefined) {
+      ZkappClass._states = [];
+      let layout: Map<string, { offset: number; length: number }>;
+      ZkappClass._layout = () => {
+        if (layout === undefined) {
+          layout = new Map();
+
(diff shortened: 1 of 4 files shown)


src/lib/state.ts
```diff
@@ -0,0 +1,277 @@
...
+          stateAsFields = Array(layout.length).fill(Field.zero);
+        } else {
+          stateAsFields = [];
+          for (let i = 0; i < layout.length; ++i) {
+            stateAsFields.push(account.zkapp.appState[layout.offset + i]);
+          }
+        }
+        // in prover, create a new witness with the state values
+        // outside, just return the state values
+        stateAsFields = inProver
```
bkase, line 194: Wouldn’t this make the circuit have different constraints if you are inProver vs outside? The witness statement makes variables, right? But the other side of the ternary doesn’t make any variables.

src/lib/state.ts
```diff
@@ -0,0 +1,277 @@
...
+          'fetch can only be called when the State is assigned to a SmartContract @state.'
+        );
+      if (Mina.currentTransaction !== undefined)
+        throw Error(
+          'fetch is not intended to be called inside a transaction block.'
+        );
+      let layout = getLayoutPosition(this._contract);
+      let address: PublicKey = this._contract.instance.address;
+      let { account } = await fetchAccount(address);
+      if (account === undefined) return undefined;
```
bkase, line 226: Should we throw here? If you try to call set and it returned undefined here we’ll throw and it will just be more confusing

bkase (changes requested):

src/lib/state.ts
mitschabaude: yes exactly. there are three cases which are all handled here:
* in compile, make a variable without a value
* in the prover, get the value and make a variable with the value inside
* outside any circuits, get the value and just return it

src/lib/state.ts
bkase: 👍

bkase (approved):

PR #241, opened 2022-06-09
Preconditions 2

mitschabaude:
this is the same as https://github.com/o1-labs/snarkyjs/pull/207, which I accidentally merged into the wrong branch 😵‍💫 
so it needs another approval stamp


georgeee (approved):
Approving it without a review as @bkase approved a similar PR before.

PR #244, opened 2022-06-13
Add memo to transaction creation

MartinMinkov:
**Description**
Adds the memo property to various methods in SnarkyJS. The following methods now have a memo property:
1. signFeePayer
2. addFeePayer
3. Mina.transaction
4. Deploy

These changes allow attaching a memo to a `Parties` structure. If there is no memo passed into these methods, they will simply reuse the memo property in the `Parties` structure. 

All the code to create a signature with the memo field is already included in the backend OCaml code. It just needed to be exposed in SnarkyJS.

**Tested By**
Manual testing

### DIFF
--- a/src/lib/mina.ts
+++ b/src/lib/mina.ts
@@ -65,7 +65,7 @@ function setCurrentTransaction(transaction: CurrentTransaction) {
 
 type SenderSpec =
   | PrivateKey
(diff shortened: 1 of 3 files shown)


mitschabaude (approved):
LGTM!

PR #245, opened 2022-06-13
Recursion Pt 1

### LINKED ISSUE #89: Recursion / proof composition RFC

### LINKED ISSUE #247: Recursion / Proof Composition

mitschabaude:
* Closes #247 
* RFC: https://github.com/o1-labs/snarkyjs/issues/89
* This has reduced scope compared to the original goal, to make review easier: It only adds the necessary machinery for merging proofs, and makes it fully work for `SmartContract`s. That is, smart contracts can now verify their own proofs as well as proofs from other smart contracts
* What this PR *doesn't* do is add an interface to create non-smart-contract proofs. Implementation-wise, this will be only a small addition, which was left for later because it turned out to be an orthogonal change.


src/examples/simple_zkapp_with_proof.ts
```diff
@@ -0,0 +1,129 @@
+import {
```
mitschabaude, line 1: This file tests the additions of this PR!

src/lib/circuit_value.ts
```diff
@@ -278,6 +288,64 @@ function circuitMain(
 }
 
 let primitives = new Set(['Field', 'Bool', 'Scalar', 'Group']);
+let complexTypes = new Set(['object', 'function']);
+
+// TODO properly type this at the interface
+// create recursive type that describes JSON-like structures of circuit types
+function circuitValue<T>(typeObj: any): AsFieldElements<T> {
```
mitschabaude, line 295: helper function which provides functionality equivalent to `CircuitValue`, but is less brittle, doesn't require decorators and can be applied in situations where a circuit value should be created from a plain JS object
i.e. 
```ts
let MyCircuitValue = circuitValue({ a: Group, b: Bool, c: [Field, Field] });
```
is equivalent to (but is more precisely typed than)
```ts
class MyCircuitValue extends CircuitValue {
  @prop a: Group;
  @prop b: Bool;
  @arrayProp(Field, 2) c: Field[];
  
  constructor(a: Group, b: Bool, c: Field[]) {
    this.a = a;
    this.b = b;
    this.c = c;
  }
}
```
`circuitValue` is only used in one place right now but I want to explore further usage, for example to simplify non-decorator ways of specifying circuit inputs

src/lib/proof_system.ts
```diff
@@ -1,22 +1,40 @@
...
+export { Proof, CompiledTag };
 
-export function branch(
-  target: any,
-  propertyName: string,
-  _descriptor?: PropertyDescriptor
-): any {}
+class Proof<T> {
+  static publicInputType: AsFieldElements<any> = undefined as any;
+  static tag: () => { name: string } = () => {
```
mitschabaude, line 7: the new `Proof` class needs a `tag` to identify the compiled `SmartContract` that this proof belongs to

src/snarky.d.ts
```diff
@@ -460,18 +460,19 @@ declare class Bool {
...
   static fromJSON(x: JSONValue): Bool | null;
+  static check(x: Bool): void;
 }
 
 declare interface AsFieldElements<T> {
   toFields(x: T): Field[];
   ofFields(x: Field[]): T;
   sizeInFields(): number;
-  check?: (x: T) => void;
+  check(x: T): void;
```
mitschabaude, line 475: I made this change to facilitate inferring `<T>` from and `AsFieldElements<T>`.
That every circuit type has a `check` method is now also enforced by `Circuit.witness`.
@imeckler had mentioned that he should've made `check` required instead of optional, so I went with that; it makes for simpler types and less special-casing.
This change has minor repercussions throughout the code base

src/snarky.d.ts
```diff
@@ -595,13 +591,15 @@ declare class Scalar {
...
+  static check(x: Scalar): void;
 }
 
-declare class EndoScalar {
-  static toFields(x: Scalar): Field[];
-  static ofFields(fields: Field[]): Scalar;
-  static sizeInFields(): number;
-}
+// TODO: Add this when OCaml bindings are implemented:
+// declare class EndoScalar {
```
mitschabaude, line 598: removed this because it didn't actually exist!

src/snarky/wrapper.node.js
```diff
@@ -9,12 +9,14 @@ function getWasm() {
...
+// TODO get rid of shutdown
 let didShutdown = false;
 async function shutdown() {
-  if (global.wasm_rayon_poolbuilder && !didShutdown) {
-    didShutdown = true;
-    global.wasm_rayon_poolbuilder.free();
-    await Promise.all(global.wasm_workers.map((worker) => worker.terminate()));
-    process.exit(0);
-  }
+  process.exit(0);
```
mitschabaude, line 15: `shutdown` as previously implemented failed to always shut down.. so I just made it exit the node process directly

mitschabaude (commented):
Added some comments to aid review:

src/snarky.d.ts
mitschabaude: Yeah you're right. It was made to enable some typing that I intended to add in this PR, but ended up saving for the next PR.

src/snarky.d.ts
```diff
@@ -775,9 +790,13 @@ declare const Pickles: {
    * * `verify` - a function which can verify proofs from any of the provers
    * * `getVerificationKeyArtifact` - a function which returns the verification key used in `verify`, in base58 format, usable to deploy a zkapp
    */
-  compile: (rules: [0, string, (statement: Statement) => void][]) => {
-    provers: ((statement: Statement) => Promise<unknown>)[];
-    verify: (statement: Statement, proof: unknown) => Promise<boolean>;
+  compile: (
+    rules: Pickles.Rule[],
+    statementSize: number
```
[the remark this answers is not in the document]
mitschabaude: I understand the objection. This is not intended as a user-facing API though, it's internal functionality.
That `Pickles` ends up being exported from snarkyjs is an unintentional side effect of exporting `*` from this file.
I propose to fix this by removing it from the exports. OK?
(It's a good idea anyway to reduce unintentional exporting from snarkyjs)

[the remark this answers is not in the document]
mrmr1993: > I propose to fix this by removing it from the exports. OK?

Sgtm

src/lib/zkapp.ts
```diff
@@ -207,32 +259,36 @@ export class SmartContract {
...
-    address ??= PrivateKey.random().toPublicKey();
     let instance = new this(address);
 
-    let rules = (this._methods ?? []).map(({ methodName, witnessArgs }) =>
+    let rules = (this._methods ?? []).map((methodEntry) =>
       picklesRuleFromFunction(
-        methodName,
-        (...args: unknown[]) => (instance[methodName] as any)(...args),
-        witnessArgs
+        // TODO: it's problematic that this refers to the instance that was passed in during compile
```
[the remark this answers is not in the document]
mitschabaude: fixed https://github.com/o1-labs/snarkyjs/pull/245/commits/9f60ff4cd8af70f5340d62fa8bf36e58a2092a45

src/snarky.d.ts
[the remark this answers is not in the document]
mitschabaude: fixed! https://github.com/o1-labs/snarkyjs/pull/245/commits/e7d69ba338f0285545e8dec79d9ee593216cc3bb

src/lib/circuit_value.ts
bkase: Yeah I think we should move towards the more precise typed versions of things with generics in general.

src/lib/circuit_value.ts
```diff
@@ -278,6 +288,64 @@ function circuitMain(
 }
 
 let primitives = new Set(['Field', 'Bool', 'Scalar', 'Group']);
+let complexTypes = new Set(['object', 'function']);
+
+// TODO properly type this at the interface
```
bkase, line 293: Can you add a todo for unit tests here too?

bkase (approved):

PR #250, opened 2022-06-20
Recursion, Pt 2

mitschabaude:
* Follow-up to https://github.com/o1-labs/snarkyjs/pull/245
* RFC: https://github.com/o1-labs/snarkyjs/issues/89
* Adds new `Program` function to create a custom "proof system"
* Most of the logic for `Program` shares its implementation with `SmartContract`, even though `SmartContract` does not (yet?) use `Program` directly; common functionality was moved from `zkapp.ts` to `proof_system.ts`, and smart-contract-specific stuff was split out
* adds new example `src/examples/program.ts` which serves as rudimentary documentation for `Program`

PR #260, opened 2022-06-23
Recursion pt 3

mitschabaude:
Follow-up to #250, fills in some missing pieces
* [x] serialize proofs
* [x] simplify accessing `Proof` classes
* [x] serialize verification key
* [x] (final) verification of ZkProgram proofs


src/lib/proof_system.ts
```diff
@@ -32,16 +33,53 @@ class Proof<T> {
   verifyIf(condition: Bool) {
     this.shouldVerify = condition;
   }
-  toString() {
+  toTransactionString() {
```
bkase: Why is this called `toTransactionString` and not toString?

bkase (approved):

PR #263, opened 2022-06-27
Fix `lte` for UInt64 and UInt32 and tests

MartinMinkov:
**Description**
Fixes https://github.com/o1-labs/snarkyjs/issues/174 and related tests

**Tested**
Updated unit tests and ran successfully


MartinMinkov:
Fixed in https://github.com/o1-labs/snarkyjs/pull/307

PR #264, opened 2022-06-27
Update bindings for simplified fee-payer

mrmr1993:
This PR is the counterpart of Minaprotocol/Mina#11332.


mitschabaude (approved):
lgtm!

mitschabaude:
I would wait with merging this though until the mina PR lands, so that we can add up to date bindings on this one (which are compatible with Mina/develop), by running `./scripts/update-snarkyjs-bindings.sh`

PR #268, opened 2022-06-28
Adds circuitDigest; see MinaProtocol/mina#11261

bkase:
I'm fighting the OCaml build right now, so I can't update the ocaml artifacts here sadly


mitschabaude:
note: shouldn't be merged yet because it's not working yet, see https://github.com/MinaProtocol/mina/pull/11261#issuecomment-1170156329

PR #269, opened 2022-06-30
Improve CircuitValue

mitschabaude:
this makes significant changes to the `CircuitValue` class, some non-controversial and some potentially controversial:
* change the types of static methods on CircuitValue -- they weren't inferred correctly before, now they are, so type-safe extending of these methods works now
* might be controversial: make the constructor on CircuitValue generic.
  - the **benefit** is that now users of CircuitValue don't have to write a (very boiler-plate-y) constructor.
  - a **downside** is that, with default TS configs, people not using a constructor on classes that extend CircuitValue will get a TS error on their props:  
  `"Property has no initializer and is not definitely assigned in the constructor"`  
  this can be fixed locally by adding a `!` to the prop, e.g. `@prop value!: Field`. it can also be fixed globally in the project by adding [a TS config option](https://github.com/o1-labs/snarkyjs/pull/269/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0R17), which we should probably do in the CLI template project
  - another small **downside** is that overriding the constructor becomes a bit more error-prone - you have to pass on your constructor arguments to `super()`. as such, this change probably breaks a couple of existing code, in a way that fixing it might not be very obvious. if we go with this, I think going forward we should not encourage the pattern of overriding the constructor in examples, but instead add static `from` methods which add any additional logic and invoke the original constructor. e.g., see [how I changed the sudoku example](https://github.com/o1-labs/snarkyjs/pull/269/files#diff-4f28cac68e5aaf84bb21ddd42725b110f5ba1665c49ffe63164e6a9ad0c97165R23)


src/lib/circuit_value.ts
```diff
@@ -15,22 +15,41 @@ export {
...
 
 function asFieldElementsToConstant<T>(typ: AsFieldElements<T>, t: T): T {
   const xs: Field[] = typ.toFields(t);
   return typ.ofFields(xs);
 }
 
-// TODO: Synthesize the constructor if possible (bkase)
-//
 abstract class CircuitValue {
+  constructor(...props: any[]) {
```
mitschabaude, line 26: this function was written verbatim, completely correctly, by github copilot on the first try!!!

src/lib/circuit_value.ts
```diff
@@ -122,21 +166,6 @@ abstract class CircuitValue {
   }
 }
 
-(CircuitValue as any).check = function (v: any) {
```
mitschabaude, line 125: was able to remove this hack because I got the types working on `CircuitValue`

src/lib/int.ts
```diff
@@ -364,9 +354,7 @@ class Int64 extends CircuitValue implements BalanceChange {
   // Overall, I think the existing implementation is the optimal one.
 
   constructor(magnitude: UInt64, sgn = Field.one) {
-    super();
-    this.magnitude = magnitude;
-    this.sgn = sgn;
+    super(magnitude, sgn);
```
mitschabaude, line 357: I left this constructor because it adds something nice -- a default value for one of the arguments

tsconfig.json
```diff
@@ -14,6 +14,7 @@
     "useDefineForClassFields": false, // ensure correct behaviour of class fields with decorators
 
     "strict": true, // for the full TypeScript experience
+    "strictPropertyInitialization": false, // to enable generic constructors, e.g. on CircuitValue
```
mitschabaude, line 17: if we go with this PR, this should be added in the zkapp CLI template projects as well

mitschabaude (commented):
some comments:

src/lib/circuit_value.ts
```diff
@@ -75,7 +75,23 @@ abstract class CircuitValue {
...
+    if (fields === undefined || fields === null) {
+      return;
+    }
+
+    for (let i = 0; i < fields.length; ++i) {
+      const [key, propType] = fields[i];
+      const value = (v as any)[key];
+      if (propType.check === undefined)
+        throw Error('bug: circuit value without .check()');
+      propType.check(value);
```
MartinMinkov, line 111: Does this error get thrown if a CircuitValue class does not define a `check` function for itself?

MartinMinkov (approved):
The changes look good to me as long as everyone agrees on this controversial change :D 

I added an issue in the zkapp-cli repo to update the Sudoku example if this lands in it's current state.

src/lib/circuit_value.ts
ymekuria: That is really cool! I'm going to try out copilot.

tsconfig.json
ymekuria: I agree we should fix this globally in the CLI and push users to do the same instead fixing locally by adding a `!` to the prop, e.g. `@prop value!: Field`.

mitschabaude:
> Introducing a new pattern of adding a static from method will be more involved. It would involve changing existing code, updating docs, and user eduction. I think the upside is probably worth all this.
Hm we also don't have to do this. I just thought it's cleaner than the ovrridden constructor, but definitely not urgent to change this everywhere. But if you don't prefer it than we can also stick with constructors in our examples, or just don't single out one way of doing it

ymekuria:
> > Introducing a new pattern of adding a static from method will be more involved. It would involve changing existing code, updating docs, and user eduction. I think the upside is probably worth all this.
> 
> Hmm ok, we also don't have to do this. If you don't prefer it then we can also stick with constructors in our examples, or just don't single out one way of doing it

I'm ok to do this if we all agree that this pattern is better. We should be clear and consistent on our recommendations of best practices.

bkase (approved):
This looks great, I think the tradeoffs are all worth it 💯

PR #270, opened 2022-06-30
Hashing primitives for events

mitschabaude:
This adds core functionality for implementing events / sequence events, namely, the building blocks for hashing them together.

As a first milestone, we successfully recreate the "empty" events / sequence events hashes, which were previously hard-coded.

This PR also does a fairly large refactor of the Poseidon module -- wrapping the primitives on the JS side instead of exporting them directly from OCaml, for added flexibility.


src/lib/global-context.ts
```diff
@@ -96,5 +98,9 @@ function inCompile() {
   return !!mainContext?.inCompile;
 }
 function inCheckedComputation() {
-  return !!mainContext?.inCompile || !!mainContext?.inProver;
+  return (
+    !!mainContext?.inCompile ||
+    !!mainContext?.inProver ||
+    !!mainContext?.inCheckedComputaton
+  );
```
MartinMinkov, line 105: What's the difference between these cases now?

Given this function is named `inCheckedComputation` and we check for `inCheckedComputaton`, could we refactor `inCompile` and `inProver` to use `inCheckedComputation` instead?

src/lib/circuit_value.ts
```diff
@@ -271,12 +272,15 @@ function circuitMain(
   }
 
   target.snarkyMain = (w: Array<any>, pub: Array<any>) => {
-    let args = [];
-    for (let i = 0; i < numArgs; ++i) {
-      args.push((publicIndexSet.has(i) ? pub : w).shift());
-    }
+    let [, result] = withContext({ inCheckedComputaton: true }, () => {
```
MartinMinkov, line 275: Nit: Spelling of `inCheckedComputaton` (and other references too it)

src/lib/hash.ts
```diff
@@ -39,3 +43,25 @@ const Poseidon = {
...
   Sponge,
 };
+
+const prefixes = Poseidon_.prefixes;
+
+function salt(prefix: string) {
+  return Poseidon.update(Poseidon.initialState, [prefixToField(prefix)]);
+}
+
+// same as Random_oracle.prefix_to_field in OCaml
```
MartinMinkov, line 53: Why don't we use the OCaml binding of this function instead of defining our own version here?

MartinMinkov (approved):
Small nit and some questions, otherwise LGTM!

src/lib/circuit_value.ts
mitschabaude: good catch!

src/lib/global-context.ts
mitschabaude: So, I added the additional inCheckedComputation because I don't yet have a good way of knowing whether we're in prove or conpile, in the case of plain Circuits. And so far, we also don't need that distinction there, we only need to be able to know if it's one of them.
In SmartContracts however, this distinction is important in some places, so we need to keep the more detailed information provided by inProver / inCompile

mitschabaude: For example, when doing this.state.get(), behaviour is different for inProver vs inCompile

inCheckedComputation just means "one of the two". This is the info we need for Poseidon. We also have an OCaml way of getting this info, but that was buggy in the past, so I'm trying to avoid it now

For Poseidon, we just need to know whether we

mitschabaude: See https://github.com/o1-labs/snarkyjs/issues/65
I'm really scared of this breaking things and causing some of our known bugs, we should dig into it soon

src/lib/hash.ts
mitschabaude: I could've done that, yeah. I guess I just try to slowly move things away from OCaml and to JS when I can. Makes better error messages and easier to look at the code for us and outside contributors. And maybe, eventually, smaller size if we can get rid of some big chunk of OCaml dependencies, like mina_base for example

MartinMinkov: Sounds like a good idea to me :D

Issue #233, opened 2022-06-06
[Tokens RFC] - Research desired functionality to expose


MartinMinkov:
**Description**

Before implementing token functionality in SnarkyJS, it is important that we align on what we intended to expose and how it works. Some pre-homework to this task is to do a quick survey of existing token implementations in other chains. Following that, this issue has been broken into 3 steps.

1. Examine protocol code to see what functionalities are exposed with tokens
2. Interview a select number of protocol and product engineers to agree on interfaces that should be exposed in SnarkyJS

- [x] - Brandon
- [x] - Gregor
- [x] - Matthew
- [x] - Izaak

3. Finalize RFC

MartinMinkov:
# Tokens RFC
## Overview
Custom tokens are supported at a protocol level, and we wish to bring that functionality to SnarkyJS. Custom tokens are denoted by a token identifier which is used to distinguish between different tokens and a token account. A token account is similar to a regular account, but its balance is stored in the chosen custom token. One may use the same public key for multiple token accounts for different custom tokens.

In addition to token accounts, there is a notion of a token owner account. The token owner account is the owner of a custom token and supervises the behavior of accounts whose balance is stored in that custom token. They are the only account that can provide certain invariants on a custom token before the supervised accounts are updated. The token owner must be included in the Party's transaction layout of a custom token transaction for the update to be authorized.

Token owner accounts are created by specifying a public key and a token identifier. For the purposes of most token owner accounts, using the default, MINA token identifier can be used to derive a new custom token identifier.  

In this RFC, we propose a bare-bones API for SnarkyJS to support creating token accounts and transacting in custom tokens. 

**Note**: Tokens cannot be used to pay for fees or be used in stake delegations.

## Creating Token Accounts
Given an existing account, a new token account is created by simply receiving a payment denoted in a custom token. When such a transaction occurs, the protocol will create a token account for the public key receiving the new custom token.

When one wants to create a new custom token, it must be derived from an existing account and an existing token identifier. If an existing public key is specified in deriving a new custom token, that public key will be the token owner. The native MINA token identifier can be used when setting an existing token identifier. In this model, developers will want their zkApp to be the token owner of their custom token to automate any authorization for other users to interact.

To derive a new custom token identifier from a public key and an existing token identifier, we can expose a function that will accept a public key and token identifier and spit out a new custom token identifier. This new custom token identifier can be used by the specified public key to mint new tokens.

```ts
import { createNewToken } from "snarkyjs";

[...]

PR #273, opened 2022-07-05
Tokens

MartinMinkov:
# Description
This PR adds support for custom tokens and token symbols for SnarkyJS. It implements the requirements mentioned in the following RFC: #233 

Mina PR: https://github.com/MinaProtocol/mina/pull/11449

To enable minting, burning, and sending, changes to the Party structure was made to support `this.token()...` as well as `this.tokenSymbol.set(...)`. Additionally, changes to account fetching have been made to support querying for a specific token id in addition to a public key.

# Tested
Implemented unit tests in `token.test.ts` to check different behaviors that tokens offer. These tests are only related to the local ledger and thus are not tested with a live network yet. To test with a live network, a new QA network should be deployed and then a zkApp deployed that acts similar to the zkApp in the test file.


src/lib/party.ts
```diff
@@ -231,7 +242,7 @@ let Permissions = {
   }),
 };
 
-const getDefaultTokenId = () => Field.one;
+const getDefaultTokenId = () => Ledger.fieldToBase58(Field.one);
```
mitschabaude, line 245: For dealing with variables, it's probably best / most convenient to have tokens represented as `Field` most of the time in snarkyjs, and only using the string versions when doing IO. At least that's what we generally do with other values, like public/private keys, hashes, balances etc

mitschabaude: Which is to say I'd probably revert this change (but don't do it if it's less convenient that way)

src/lib/party.ts
```diff
@@ -492,6 +503,52 @@ type UnfinishedSignature = undefined | LazySignature | string;
...
+    let { tokenOwner, parentTokenId } = options ?? {};
+
+    // Reassign to default tokenId if undefined
+    parentTokenId ??= getDefaultTokenId();
+
+    // Check if we can deserialize the parentTokenId
+    try {
+      Ledger.fieldOfBase58(parentTokenId);
+    } catch (e) {
+      throw new Error(`Invalid parentTokenId\n(error: ${(<Error>e).message})`);
```
ymekuria, line 521: Did you mean to have this error message? 

`Invalid parentTokenId\n(error: ${e.message})`;

If not, I'm just curious what this does `<Error>` tag is.

src/lib/precondition.ts
```diff
@@ -178,7 +178,7 @@ To write a correct circuit, you must avoid any dependency on the concrete value
   let value: U;
   if (accountOrNetwork === 'account') {
     let address = party.body.publicKey;
-    let account: any = Mina.getAccount(address);
+    let account: any = Mina.getAccount(address, getDefaultTokenId());
```
mitschabaude, line 181: we probably want to handle other token ids on the zkapp account, in places like this? if yes, it would be good to track it somehow

src/lib/mina.ts
```diff
@@ -225,18 +226,26 @@ function LocalBlockchain({
         Math.ceil((new Date().valueOf() - startTime) / msPerSlot)
       );
     },
-    getAccount(publicKey: PublicKey): Account {
-      let ledgerAccount = ledger.getAccount(publicKey);
+    getAccount(
+      publicKey: PublicKey,
+      tokenId: string = getDefaultTokenId()
```
mitschabaude, line 231: maybe this is the place where we should do the conversion from `Field` to string? because here we know we need the actual value

mitschabaude (commented):
started review, will continue tomorrow!

src/lib/party.ts
```diff
@@ -492,6 +503,52 @@ type UnfinishedSignature = undefined | LazySignature | string;
...
+      Ledger.customTokenID(tokenOwner, Ledger.fieldOfBase58(parentTokenId));
+    } catch (e) {
+      throw new Error(
+        `Could not create a custom token id:\n(error: ${(<Error>e).message})`
+      );
+    }
+    this.parentTokenId = parentTokenId;
+    this.tokenOwner = tokenOwner;
+    this.id = Ledger.customTokenID(
+      tokenOwner,
```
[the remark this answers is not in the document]
mitschabaude: Actually it's more precise to switch on `tokenOwner.isConstant() && parentTokenId.isConstant()` and not `inCheckedComputation()`

src/lib/fetch.ts
```diff
@@ -160,13 +176,15 @@ type Account = {
 type FlexibleAccount = {
   publicKey: PublicKey | string;
   nonce: UInt32 | string | number;
+  tokenId: string;
```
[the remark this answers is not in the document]
MartinMinkov: Yup, fixed!

src/lib/party.ts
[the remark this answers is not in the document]
MartinMinkov: I couldn't find a good way to check for `isConstant()` on a `PublicKey` type so I defaulted to using `inCheckedComputation`. Could you advise how to check a PublicKey if it's a var or constant?

src/lib/party.ts
MartinMinkov: Good call, fixed!

src/snarky.d.ts
```diff
@@ -1,3 +1,5 @@
+import { TokenId } from 'snarky/parties-leaves-json';
```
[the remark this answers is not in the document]
MartinMinkov: Good catch, removed.

src/lib/mina.ts
MartinMinkov: Good call! Made the `getAccount` accept a `Field` for the tokenId. Should the `Account` tokenId also be a Field or should that stay as a string? I changed the type to be a Field instead, wondering what's best here :/

src/lib/party.ts
```diff
@@ -492,6 +503,52 @@ type UnfinishedSignature = undefined | LazySignature | string;
...
+    } catch (e) {
+      throw new Error(`Invalid parentTokenId\n(error: ${(<Error>e).message})`);
+    }
+
+    // Check if we can create a custom tokenId
+    try {
+      Ledger.customTokenID(tokenOwner, Ledger.fieldOfBase58(parentTokenId));
+    } catch (e) {
+      throw new Error(
+        `Could not create a custom token id:\n(error: ${(<Error>e).message})`
```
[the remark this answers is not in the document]
MartinMinkov: Same as previous

src/lib/party.ts
ymekuria: Oh nice!

src/lib/party.ts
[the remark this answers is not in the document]
mitschabaude:
```ts
tokenOwner.toFields().every((x) => x.isConstant())
```
-- it's constant if all its parts are constant

I actually added this as a method on `CircuitValue` in a recently merged PR, but it has a bug :D Fix is in an unmerged PR. You could either add that fix yourself or inline the code above

src/lib/mina.ts
mitschabaude: Yes `Field` on `Account ` feels right to me, because the other things on there are snarky types as well

src/lib/state.ts
```diff
@@ -203,7 +203,7 @@ function createState<T>(): InternalStateType<T> {
       if (!GlobalContext.inCompile() && !GlobalContext.inAnalyze()) {
         let account: Account;
         try {
-          account = Mina.getAccount(address);
+          account = Mina.getAccount(address, getDefaultTokenId());
```
mitschabaude, line 206: could we use `party.body.tokenId` for the token id here as well?

src/lib/party.ts
```diff
@@ -516,8 +577,105 @@ class Party {
...
+
+        // Require signature from the receiver account being deducted
+        receiverParty.authorization = {
+          kind: 'lazy-signature',
+        };
+      },
+
+      send({ from, to, amount }: SendParams) {
+        // Check if the sender party is the current zkApp address
+        if (from === thisParty.publicKey) {
```
mitschabaude, line 622: hm, just using `from === thisParty.publicKey` to compare public keys seems wrong and clever at the same time 😆

src/lib/state.ts
MartinMinkov: Unless I'm mistaken, there is no `party.body.tokenId` in scope here, the only access to a Party here is through `this._contract.instance.self` which doesn't seem right to be referencing the tokenId property.

src/lib/party.ts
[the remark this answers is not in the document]
MartinMinkov: Sounds good! Stole your fix from the unmerged PR :P

src/lib/party.ts
```diff
@@ -516,8 +575,105 @@ class Party {
...
+          // If so, create a new party for the zkApp to send the amount to the receiver
+          let senderParty = Party.createUnsigned(thisParty.publicKey, {
```
mitschabaude, line 624: Actually, I think in the `if` branch, we also have to sign the senderParty. It's a separate party, not connected to whatever authorization the `thisParty` already has (it's even for a different tokenId, so a different account). It also has to be a child of `thisParty` according to the token rules (AFAIU). So, it seems the logic of the `if` branch is actually equivalent to the `else` branch and we can just remove the `if` branch

src/lib/party.ts
```diff
@@ -516,8 +575,105 @@ class Party {
...
+          senderParty.authorization = {
+            kind: 'lazy-signature',
+          };
+        }
+
+        let receiverParty = Party.createUnsigned(to, {
+          caller: this.id,
+          tokenId: this.id,
+          callDepth: 1,
+          useFullCommitment: Bool(true),
```
mitschabaude, line 651: this shouldn't set `useFullCommitment`, the receiver doesn't need replay protection

src/lib/party.ts
```diff
@@ -516,8 +575,105 @@ class Party {
...
+      id: customToken.id,
+      parentTokenId: customToken.parentTokenId,
+      tokenOwner: customToken.tokenOwner,
+
+      mint({ address, amount }: MintOrBurnParams) {
+        let receiverParty = Party.createUnsigned(address, {
+          caller: this.id,
+          tokenId: this.id,
+          callDepth: 1,
+          useFullCommitment: Bool(true),
```
mitschabaude, line 594: this shouldn't set `useFullCommitment`, a receiver doesn't need replay protection

src/lib/state.ts
mitschabaude: `this._contract.instance.self.body.tokenId` is fine! not pretty but it will reliably hold the party connected to the smart contract for which this is the state. we need to get the state from the account with the right token id

src/lib/party.ts
```diff
@@ -516,8 +575,105 @@ class Party {
...
+      },
+    };
+  }
+
+  get tokenSymbol() {
+    let party = this;
+
+    return {
+      set(tokenSymbol: string) {
+        Party.setValue(party.update.tokenSymbol, {
```
mitschabaude, line 666: yes 💯 this is the API we should use for all updates. maybe we'll put them all on `this.account` eventually to have fewer props on `this`, but this is good for now

src/lib/party.ts
```diff
@@ -516,8 +575,105 @@ class Party {
...
+          data: tokenSymbol,
+          hash: salt(tokenSymbol)[0],
```
mitschabaude, line 668: based on a quick check, this seems very much like how the hash is computed in OCaml :O

src/lib/token.test.ts
```diff
@@ -0,0 +1,349 @@
...
+        zkapp.sign(zkappKey);
+      }).catch((e) => {
+        expect(e).toBeDefined();
+      });
+    });
+
+    it('should error if token owner mints negative amount tokens', async () => {
+      await Mina.transaction(feePayer, () => {
+        Party.fundNewAccount(feePayer);
+        zkapp.mint(tokenAccount1, UInt64.from(-100_000));
```
mitschabaude, line 194: this throws an error because `UInt64.from(<negative number>)` isn't possible, so it's not exercising the test

src/lib/token.test.ts
```diff
@@ -0,0 +1,349 @@
...
+
+      expect(() => {
+        tx.send();
+      }).toThrow();
+    });
+
+    it('should error if token owner burns negative amount tokens', async () => {
+      await Mina.transaction(feePayer, () => {
+        Party.fundNewAccount(feePayer);
+        zkapp.burn(tokenAccount1, UInt64.from(-100_000));
```
mitschabaude, line 253: `UInt64.from(-100_000)` throws

src/lib/token.test.ts
```diff
@@ -0,0 +1,349 @@
...
+          zkapp.send(tokenAccount1, tokenAccount2, UInt64.from(10_000));
+          zkapp.sign(zkappKey);
+        })
+      )
+        .sign([tokenAccount1Key])
+        .send();
+
+      expect(
+        Mina.getBalance(tokenAccount1, zkapp.token().id).value.toBigInt()
+      ).toEqual(BigInt(90_000));
```
mitschabaude, line 285: tip: you can write bigints like this: `90_000n`

mitschabaude (commented):
Finished reviewing! Just a couple of smaller suggestions from me -- great work!

The most important change I'm suggesting is to remove the tokenId argument in `token(tokenId?)`, because it has to be the same as `this.tokenId` anyway. (Correct me if I'm wrong.)

src/lib/circuit_value.ts
```diff
@@ -77,8 +77,8 @@ abstract class CircuitValue {
     Circuit.assertEqual(this, x);
   }
 
-  isConstant(x: this) {
-    return x.toFields().every((x) => x.isConstant());
+  toConstant(): this {
```
mitschabaude, line 80: wait, it confuses me that you changed `isConstant` to `toConstant`. don't we want both?

src/lib/party.ts
```diff
@@ -492,6 +503,56 @@ type UnfinishedSignature = undefined | LazySignature | string;
 
 type LazyControl = Control | LazySignature | LazyProof;
 
+class Token {
```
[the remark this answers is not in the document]
mitschabaude: (no need to change it though if you feel the class has a use case)

src/lib/circuit_value.ts
MartinMinkov: My bad, I got confused here. I reverted the change. :sweat_smile:

src/lib/party.ts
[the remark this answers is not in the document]
MartinMinkov: I'll leave it as a class for now since that was what the RFC suggested when we agreed upon everything. It's nice to have all these different properties lumped together in one object as well :D

src/lib/party.ts
```diff
@@ -492,6 +503,56 @@ type UnfinishedSignature = undefined | LazySignature | string;
...
+
+  constructor(options: { tokenOwner: PublicKey; parentTokenId?: Field }) {
+    let { tokenOwner, parentTokenId } = options ?? {};
+
+    // Reassign to default tokenId if undefined
+    parentTokenId ??= getDefaultTokenId();
+
+    // Check if we can deserialize the parentTokenId
+    try {
+      Ledger.fieldToBase58(parentTokenId);
```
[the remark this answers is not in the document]
MartinMinkov: You're right, removed!

src/lib/party.ts
MartinMinkov: Nice catch, fixed!

src/lib/party.ts
MartinMinkov: That was a bad call on my part :sweat_smile: Ended up removing this check due to the comment you mentioned about the conditions being the same thing.

src/lib/party.ts
MartinMinkov: You're right, fixed!

src/lib/token.test.ts
MartinMinkov: The accepted types for `mint`, `burn` and `send` are all UInts so I guess this test doesn't do a great job at its goal. I removed these tests since it's really a check on how UInt behaves

src/lib/party.ts
```diff
@@ -516,8 +577,105 @@ class Party {
     return new (Party as any)(body, authorization, party.isSelf);
   }
 
+  token(tokenId?: Field) {
```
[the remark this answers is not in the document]
MartinMinkov: This comment is a little confusing to me, I'll have to come back to it :P

src/lib/circuit_value.ts
```diff
@@ -77,8 +77,8 @@ abstract class CircuitValue {
     Circuit.assertEqual(this, x);
   }
 
-  toConstant(): this {
-    return (this.constructor as any).toConstant(this);
+  isConstant(x: this) {
+    return x.toFields().every((x) => x.isConstant());
```
mitschabaude, line 81: should be
```ts
isConstant() {
    return this.toFields().every((x) => x.isConstant());
}
```

mitschabaude (approved):
lgtm now!

src/lib/token.test.ts
```diff
@@ -0,0 +1,313 @@
...
+    let maxAmountInCirculation = this.maxAmountInCirculation.get();
+    this.maxAmountInCirculation.assertEquals(maxAmountInCirculation);
+
+    let newTotalAmountInCirculation = totalAmountInCirculation.add(amount);
+
+    newTotalAmountInCirculation.value
+      .lte(maxAmountInCirculation.value)
+      .assertTrue();
+
+    this.token().mint({
```
jasongitmail, line 53: Would developers need to use this assertion every time they use the mint method? 

Asking because it seems like potentially a common error case to forget this, if so. Could it be handled automatically inside of `mint()`? I wouldn't imagine anyone ever needing to deviate from this condition.

src/lib/token.test.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  describe('Create existing token', () => {
+    beforeEach(async () => {
+      await setupLocal();
+    });
+
+    it('should have a valid custom token id', async () => {
+      const tokenId = zkapp.token().id;
+      const expectedTokenId = new Token({ tokenOwner: zkappAddress }).id;
+      expect(tokenId).toBeDefined();
```
jasongitmail, line 137: could delete this line `expect(tokenId).toBeDefined();` given the following assertion, no? Or is there a chance it returns undefined?

src/lib/token.test.ts
```diff
@@ -0,0 +1,313 @@
...
+    super.deploy(args);
+    this.setPermissions({
+      ...Permissions.default(),
+      editState: Permissions.proofOrSignature(),
+    });
+  }
+
+  @method init() {
+    this.tokenSymbol.set(tokenSymbol);
+    this.totalAmountInCirculation.set(UInt64.zero);
```
jasongitmail, line 36: I think we should consider using the term Quantity, instead of Amount, everywhere. On trading dashboards, qty refers to the number of tokens and amount prefers to `price * amount`, which would be USD etc. when trading on fiat markets. So qty is more consistent with how it's used on most financial platforms.

src/lib/mina.ts
```diff
@@ -232,18 +233,25 @@ function LocalBlockchain({
...
-      let ledgerAccount = ledger.getAccount(publicKey);
+    getAccount(
+      publicKey: PublicKey,
+      tokenId: Field = getDefaultTokenId()
+    ): Account {
+      let ledgerAccount = ledger.getAccount(publicKey, tokenId);
       if (ledgerAccount == undefined) {
         throw new Error(
-          `getAccount: Could not find account for public key ${publicKey.toBase58()}`
+          `getAccount: Could not find account for public key ${publicKey.toBase58()} with the tokenId ${Ledger.fieldToBase58(
```
bkase, line 243: can you omit the `with the tokenId ...` in the case that the tokenId is the default one?

src/lib/mina.ts
```diff
@@ -297,21 +305,31 @@ function RemoteBlockchain(graphqlEndpoint: string): Mina {
...
+        let account = Fetch.getCachedAccount(
+          publicKey,
+          tokenId,
+          graphqlEndpoint
+        );
         if (account !== undefined) return account;
       }
       throw Error(
-        `getAccount: Could not find account for public key ${publicKey.toBase58()}.\nGraphql endpoint: ${graphqlEndpoint}`
+        `getAccount: Could not find account for public key ${publicKey.toBase58()} with the tokenId ${Ledger.fieldToBase58(
```
bkase, line 330: same here

src/lib/mina.ts
```diff
@@ -394,11 +416,14 @@ let activeInstance: Mina = {
     ) {
       let account = Fetch.getCachedAccount(
         publicKey,
+        tokenId,
         Fetch.defaultGraphqlEndpoint
       );
       if (account === undefined)
         throw Error(
-          `getAccount: Could not find account for public key ${publicKey.toBase58()}. Either call Mina.setActiveInstance first or explicitly add the account with addCachedAccount`
+          `getAccount: Could not find account for public key ${publicKey.toBase58()} with the tokenId ${Ledger.fieldToBase58(
```
bkase, line 424: same here (should this string be a constant stored somewhere instead of inlined all the time?)

src/lib/party.ts
```diff
@@ -494,6 +498,50 @@ type UnfinishedSignature = undefined | LazySignature | string;
...
+      );
+    }
+
+    this.parentTokenId = parentTokenId;
+    this.tokenOwner = tokenOwner;
+    if (
+      tokenOwner.toFields().every((x) => x.isConstant()) &&
+      parentTokenId.isConstant()
+    ) {
+      this.id = Ledger.customTokenIdChecked(tokenOwner, this.parentTokenId);
```
bkase, line 527: What's the difference between these branches?

src/lib/party.ts
```diff
@@ -523,8 +571,94 @@ class Party {
...
+          caller: this.id,
+          tokenId: this.id,
+          useFullCommitment: Bool(true),
+        });
+
+        // Sub the amount to burn from the receiver's account
+        receiverParty.body.balanceChange =
+          receiverParty.body.balanceChange.sub(amount);
+
+        // Require signature from the receiver account being deducted
```
bkase, line 608: Is this always true? Can't certain tokens be burned via proof?

src/lib/party.ts
```diff
@@ -523,8 +571,94 @@ class Party {
...
+        let senderParty = createChildParty(thisParty, from, {
+          caller: this.id,
+          tokenId: this.id,
+          useFullCommitment: Bool(true),
+        });
+
+        senderParty.body.balanceChange =
+          senderParty.body.balanceChange.sub(amount);
+
+        // Require signature from the sender party
```
bkase, line 625: same here -- aren't there certain senders that need a proof instead?

src/lib/token.test.ts
bkase: I also typically see (and we use consistently in the backend code) the term "totalSupply" for this -- probably for consistency we should go with that.

bkase (approved):
This looks great! I just have a few questions that either you can just explain away in comments or perhaps I found a bug

src/lib/party.ts
mitschabaude: hey @MartinMinkov, I think these branches are wrong.. the else branch should use the checked version, not the if branch

@bkase, if the token owner or parent token id is a variable then we need to derive the token id in-snark. (Not sure if we could always use the checked version, depends on how flexible it is written, but IIRC checked hashing returns a new variable even when the inputs are constant, which wouldn't work outside the circuit)

MartinMinkov: Yikes! Good catch, switch these. :sweat_smile:

src/lib/fetch.ts
```diff
@@ -160,13 +169,15 @@ type Account = {
...
+  tokenSymbol?: string;
   balance?: UInt64 | string | number;
   zkapp?: { appState: (Field | string | number)[] };
 };
 
 // TODO provedState
-const accountQuery = (publicKey: string) => `{
-  account(publicKey: "${publicKey}") {
+const accountQuery = (publicKey: string, tokenId: string) => `{
+  account(publicKey: "${publicKey}" tokenId: "${tokenId}") {
```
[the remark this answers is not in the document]
MartinMinkov: Good catch!

src/lib/party.ts
MartinMinkov: Good question, I'm not sure how this would work with the current SnarkyJS API. @mitschabaude, is there a way to require either a proof or signature? Also, how would one construct a proof to do this?

src/lib/token.test.ts
MartinMinkov: Gregor beat me to this! All the points above are solid reasons why we shouldn't push down these custom behaviors of zkApps into the library unless there's a very good reason too. In the endgame of zkApps, I imagine there would be libraries you could import that are similar to contracts you extend in Solidity to get these checks for the developer.

src/lib/mina.ts
MartinMinkov: Good call, fixed!

src/lib/party.ts
mitschabaude: Aaah yes. I forgot about that. It depends on the senderParty's account, and the permissions on there, whether it needs a proof or a signature.

mitschabaude: This is a puzzle 🤔 because different zkApps have completely different ways of authorizing transactions. So my fear is token contracts will typically impose a certain interface on zkApps for that token.

I'm pretty sure @mrmr1993 has thought deeper about this

In any case, I think there's no easy solution here. The `token()` API we designed seems fundamentally not flexible enough to handle parties with proofs, because they are all different. The API must be one where users provide their own parties, not where we create parties for them. But even then it seems very tricky to make a circuit for the token owner contract that's flexible enough to handle arbitrary zkApp children. So I'm not even sure we can _ever_ provide a ready-made, general implementation for parties with proofs.

That makes it all the more important to at least expose the low-level pieces to create token contracts -- I think that's the `Token` class (which I still think should be just a function)

mitschabaude: Maybe a fairly general solution is not that hard to build after all.
The token() functions must witness their child parties instead of hardcoding them into the circuit, and only assert the properties of them that they care about -  balanceChanges and tokenId.

The challenge is how to create a good API for this. I imagine its functions must take a party as input, which might mean we have to make Party one of the possible arguments of a smart contract method. A smart contract could sort of "call" a token contract with 'this' as an argument, and that token contract would make the smart contract its child. A short of reversed zkapp composability

mitschabaude: I guess the natural API from the protocol perspective would be if the token contract methods would accept a callback, which is like a transaction block in that you can run a (or possibly multiple) smart contract methods in it, which would create parties that are authorized with a proof. And the token contract would then assert certain conditions on the resulting parties, for example
* They can't have children
* There must be exactly / at most X of them
* The balances must cancel in send
* In mint / burn, conditions X, Y, Z must hold

mitschabaude: I created a separate issue so that we can discuss the priority & path forward on this. It's out of scope for this PR, it's already nice to have token transfers authorized by a proof
https://github.com/o1-labs/snarkyjs/issues/300

src/lib/token.test.ts
jasongitmail: cool

PR #274, opened 2022-07-06
Sequencing events

### LINKED ISSUE #88: Sequencing support (a.k.a. sequence events)

mitschabaude:
- Closes #88 
- RFC: https://github.com/o1-labs/snarkyjs/issues/265


src/lib/zkapp.ts
```diff
@@ -370,6 +450,134 @@ export class SmartContract {
...
+      { maxTransactionsWithUpdates = 32 } = {}
+    ): { state: S; stateHash: Field } {
+      if (updateLists.length > maxTransactionsWithUpdates) {
+        throw Error(
+          `stateUpdate.applyUpdates: Exceeded the maximum number of lists of updates, ${maxTransactionsWithUpdates}.
+Use the optional \`maxTransactionsWithUpdates\` argument to increase this number.`
+        );
+      }
+      let methodData = contract.analyzeMethods();
+      let possibleUpdatesPerTransaction = [
```
mitschabaude, line 524: one of the magical things this function does is it knows how many events each of the smart contract methods emits, so it can perform computations for each of those numbers-of-events and then pick the actual one.
to enable that, I added `SmartContract.analyzeMethods` in this file and `synthesizeMethodArguments` in proof_systems.ts

src/examples/state_update_rollup.ts
```diff
@@ -0,0 +1,140 @@
...
+      state.counter = state.counter.add(update);
+      return state;
+    },
+  });
+
+  // on-chain version of our state. it will typically lag behind the
+  // version that's implicitly represented by the list of updates
+  @state(Field) counter = State<Field>();
+  // helper field to store the point in the update history that our on-chain state is at
+  @state(Field) stateHash = State<Field>();
```
bkase, line 37: let's change this to whatever you pick on the RFC-side like `actionsHash` or `actionsCommitment`

src/examples/state_update_rollup.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  // emits an update
+  @method incrementCounter(increment: Field) {
+    this.stateUpdate.emit(increment);
+  }
+
+  @method rollupStateUpdate() {
```
bkase, line 44: -> reduce

src/lib/circuit_value.test.ts
```diff
@@ -0,0 +1,15 @@
+import { Bool, Circuit, isReady, shutdown, Int64 } from '../../dist/server';
+
+describe('circuit', () => {
```
bkase, line 3: can you add tests expecting failures if there are more than one true and zero true?

bkase (commented):
This still needs updating to the latest version of the RFC before we should land it even under the "experimental" flag -- and don't forget to change the names of things everywhere! I stopped mentioning it after I realized you haven't updated this yet.

src/lib/circuit_value.test.ts
mitschabaude: done. for some reason returning 0 for zero true seemed natural to me. but I can also throw in that case if you think that's better

src/lib/circuit_value.ts
```diff
@@ -467,3 +469,39 @@ function circuitValueEquals<T>(a: T, b: T): boolean {
     ([key, value]) => key in b && circuitValueEquals((b as any)[key], value)
   );
 }
+
+function pickOne<T, A extends AsFieldElements<T>>(
```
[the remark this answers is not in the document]
mitschabaude: switch is great

src/lib/circuit_value.test.ts
bkase: good point yeah 0 is fine, but can you add a test for the zero case succeeding too?

src/lib/zkapp.ts
```diff
@@ -370,6 +441,119 @@ export class SmartContract {
...
+type Reducer<Action> = { actionType: AsFieldElements<Action> };
+
+type ReducerReturn<Action> = {
+  dispatch(action: Action): void;
+  reduce<State>(
+    actions: Action[][],
+    stateType: AsFieldElements<State>,
+    reduce: (state: State, action: Action) => State,
+    initial: { state: State; actionsHash: Field },
+    advancedOptions?: { maxTransactionsWithActions?: number }
```
bkase, line 453: this is nice

bkase: probably just `options?` is fine

bkase (approved):

src/examples/state_update_rollup.ts
```diff
@@ -0,0 +1,135 @@
...
+
+console.log('action 1');
+tx = await Mina.transaction(feePayer, () => {
+  zkapp.incrementCounter();
+  if (!doProofs) zkapp.sign(zkappKey);
+});
+if (doProofs) await tx.prove();
+tx.send();
+// update internal state
+pendingActions.push([INCREMENT]);
```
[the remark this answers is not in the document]
mitschabaude: yeah I agree. I think this would be achieved if `reducer.getActions` -- which will fetch actions in the real network interaction -- gets them from memory in the LocalBlockchain case

PR #278, opened 2022-07-14
Properly constrain zkApp party

### LINKED ISSUE #98: Ensure that the zk app circuits generated by @method have all the proper constraints needed
Not needed for end-to-end testing, but important for release

mitschabaude:
- Stacked on top of https://github.com/o1-labs/snarkyjs/pull/274
- Closes #98 - zkApps are now properly constrained 🎉 
- This means that zkApps take the party they produced, hash it and assert that it equals the public input. So the public input can't be anything else than the hash of the exact same party. Essentially, the party _is_ the public input.
- Without this assertion, a transaction containing an entirely different party, which might not comply with the smart contract logic at all, could also have had a valid proof!
- This PR was easier than originally expected, thanks to the fact that the protocol definition of the zkApp public input changed, to a hash of the zkApp party we had already implemented. (It used to be something that didn't really make sense,)

### DIFF
--- a/src/lib/circuit_value.ts
+++ b/src/lib/circuit_value.ts
@@ -326,7 +326,15 @@ let complexTypes = new Set(['object', 'function']);
 // TODO properly type this at the interface
 // create recursive type that describes JSON-like structures of circuit types
 // TODO unit-test this
-function circuitValue<T>(typeObj: any): AsFieldElements<T> {
+function circuitValue<T>(
+  typeObj: any,
+  advanced?: { customObjectKeys: string[] }
+): AsFieldElements<T> {
+  let objectKeys =
+    typeof typeObj === 'object'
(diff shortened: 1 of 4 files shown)


src/lib/circuit_value.ts
```diff
@@ -326,7 +326,15 @@ let complexTypes = new Set(['object', 'function']);
 // TODO properly type this at the interface
 // create recursive type that describes JSON-like structures of circuit types
 // TODO unit-test this
-function circuitValue<T>(typeObj: any): AsFieldElements<T> {
+function circuitValue<T>(
+  typeObj: any,
+  advanced?: { customObjectKeys: string[] }
```
bkase, line 331: how about `options?` ? Same suggestion on the other PR too

bkase (approved):

PR #285, opened 2022-07-15
zkApps composability, pt 1: Child parties

mitschabaude:
- Stacked on top of https://github.com/o1-labs/snarkyjs/pull/280
- Adds the ability to make some parties children of others in the "call forest" structure
- Adds the hashing logic to compute the public input for a party with children, plus a test which compares this against the Ocaml implementation
- This doesn't yet deal with calling other zkApps, but is an important step. It also has other use cases -- for example, for creating another party inside a smart contract method which should be constrained by the proof


MartinMinkov (approved):
LGTM! Can't wait to have this, at the moment I am hardcoding the `callDepth `for token transfers so this API will be great :D

bkase (approved):

PR #287, opened 2022-07-15
Use `.from` instead of constructors

MartinMinkov:
**Description**
Switching all uses of UInt32/UInt64/Int64 constructors with `.from` instead.


mitschabaude:
I agree with the usage of `from` outside the UInt64 class, but not in internal functions like `add`. `UInt64.from` does additional checks which duplicate stuff that operations like `add` do anyway, and I don't see the value of using the type-safer version internal to the class itself

PR #294, opened 2022-07-20
zkApps composability, pt 3: witness Party

mitschabaude:
* Stacked on top of #296 

This adds the capability to "witness a Party" -- i.e., doing something like `Circuit.witness(() => party)`. (Which isn't possible with `Circuit.witness` directly).

Motivation: This was the main hurdle to take for implementing zkApp composability. When calling another zkApp, we actually don't want to run its computation inside the caller's circuit. Instead, we want the callee to be its own party, with its own proof. In the caller's circuit we just want to pull the result of the call out of our hat (= witness it). That result is the callee's party, plus the input and output arguments of the call. We connect it to the actual computation & its proof by having the callee's party's hash as part of the public input in the caller circuit.

To create a party witness we want to leverage the auto-generated party structure from OCaml -- to avoid hard-coding the party structure again. So, this PR
* expands the capabilities of the code using that autogenerated structure, to be able to output auxiliary parts of the party in addition to its field elements, and create a party from fields + aux data.
* makes a couple of changes to achieve that `Party implements Types.Party`. I.e., the Party class exactly conforms to the TS type that is generated. The changes needed were
  * have an extra class for the `Sign` that's part of `Int64`, so that in the party we can have the `Balance` be a `Uint64` + `Sign` instead of an `Int64` directly
  * move "lazy" forms of authorization to a separate property on Party, away from the `authorization` field which has to be compatible with the OCaml type

The PR also significantly changes the generated `js-layout`, by changing the js deriver (see Mina companion), but this is actually an orthogonal change, only a small part of it was really needed here.


MartinMinkov (approved):

ymekuria (approved):

src/snarky/parties-leaves.ts
```diff
@@ -156,23 +161,207 @@ function toFields<K extends keyof TypeMap>(typeName: K, value: TypeMap[K]) {
...
+      .toField()
+      .mul(someY)
+      .add(isTheRightY.not().toField().mul(someY.neg()));
+    return new PublicKey(new Group(x, y));
+  },
+  Field(fields: Field[]) {
+    return fields.pop()!;
+  },
+  Bool(fields: Field[]) {
+    return Bool.Unsafe.ofField(fields.pop()!);
```
bkase, line 251: this is only invoked on generated code we control from the json layout right?

src/lib/int.ts
```diff
@@ -404,14 +430,9 @@ class Int64 extends CircuitValue implements BalanceChange {
 
   // --- circuit-compatible operations below ---
   // the assumption here is that all Int64 values that appear in a circuit are already checked as valid
-  // this is because Circuit.witness calls .check
+  // this is because Circuit.witness calls .check, which calls .check on each prop, i.e. UInt64 and Sign
   // so we only have to do additional checks if an operation on valid inputs can have an invalid outcome (example: overflow)
```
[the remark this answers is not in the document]
mitschabaude: The comment? No. I did mean to remove the custom check method, because now the check method on the base class does what we need

src/snarky/parties-leaves.ts
mitschabaude: Yes! So it's safe because we know the field we're reading out was a Bool originally

Issue #282, opened 2022-07-14
Scaffold for zkApp VM application test


bkase:
Build the scaffold for "Application A" of the zkapp VM application tests and sort out decisions related to where these should live in the repository and how the tests should behave and execute.

Specifically:

- [ ] Add the application A example to the repo (just stub everything for now)
- [ ] Add one example integration test (just stub it for now to always pass)
- [ ] Add the integration test to the GitHub Actions so these tests run on every PR

These tasks will imply decisions on several things which should be rationalized via an RFC-issue if it is large enough or a sentence or two in the PR.

These decisions include (but may not be limited to):

- [ ] Where should these zkApp application examples live?
- [ ] Where should tests for these applications live?
- [ ] How should these tests be written? How are they executed?
- [ ] What's the pattern for adding tests to GitHub Actions

PR #295, opened 2022-07-20
Scaffold for zkApp VM application test

ymekuria:
**Description**

Addresses #282 

This PR proposes a scaffold structure to add zkApp test applications and tests to Snarkyjs.

**File Structure and Location**

The smart contract used for this scaffold is located in the `src/examples/hellow_world` directory of the SnarkyJS repo. Creating future test applications, and adding them to the examples directory can serve the dual purpose of testing and providing more examples for the community. The tests are colocated with smart contract in a tests directory with the naming convention `<SmartContractName>.integration.test.ts` . Here is a basic folder structure for the contract and tests.

```
- hello_world
  hellow_world.ts
  - tests
    hellow_world.integration.ts 
```

**Testing and CI**

These test are created using  the Jest testing framework that is already included in the SnarkyJS repo. The Mina.LocalBlochain API is used to setup a local ledger to deploy the contract to and test it's logic. Theses tests and all future test application tests will be included in the SnarkyJS CI tests that are run here:

https://github.com/o1-labs/snarkyjs/blob/b70511aa2f4dd9472f8b2e2b7edfa0486a699863/.github/workflows/build-action.yml#L24-L26


src/examples/hello_world/hello_world.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const adminPrivateKey = PrivateKey.random();
+export const adminPublicKey = adminPrivateKey.toPublicKey();
+
+export class HelloWorld extends SmartContract {
+  @state(Field) x = State<Field>();
+
+  deploy(input: DeployArgs) {
+    super.deploy(input);
+    this.x.set(Field(3));
```
MartinMinkov, line 20: Since we intend these tests/examples to be user-facing for others to see, we should switch the state initialization to its own function as previously discussed :D

src/examples/hello_world/hello_world.ts
```diff
@@ -0,0 +1,33 @@
...
+    Party.setValue(this.self.update.delegate, adminPublicKey);
```
MartinMinkov, line 21: What does `this.self.update.delegate` refer too and why are we setting it to `adminPublicKey`?

mitschabaude (commented):
I would prefer to write large zkApp tests as linear scripts rather than jest tests, like the existing examples. For me that scripting workflow is a kind of local optimum, because it makes debugging quite easy and you get good error traces (especially if you use `node --stack-trace-limit=1000`).

The error traces jest shows are so short that if a test in mina-signer broke (which uses jest) I often ended up copying the code out of the test, into a little script that I could run with `node --stack-trace-limit=1000`.

I also just think programming is easier when you don't have to conform to the structure dictated by `describe` and `it`. That structure adds value for unit tests where you have a large number of independent test cases, but -- as I think this example shows -- it is needlessly convoluted when all you're writing is really one big, interconnected test, compared to a plain script.

In normal libraries jest brings the extra benefit that it runs everything in parallel, but snarkyjs is so heavy that this is nearly impossible, I have to run them one at a time locally, so there is nothing over scripts that jest adds here.

Other comments:
* Re file structure, for a good balance of number of folders vs number of files per folder, I think we should put all of those big zkapp examples in a common folder, e.g. `examples/zkapps`, and in that folder IMO the files and tests can just be all colocated
* I have mixed feelings about separating smart contracts and tests / script. The benefit is it gives us "clean" example smart contracts, which could for instance be deployed as-is. The downside is more jumping around when reading and writing the scripts.

src/examples/hello_world/hello_world.ts
mitschabaude: @MartinMinkov, I think the conclusion in our stand-up was that we would find a way to "automate" state initialization, e.g. based off of a default value the user passes to `State`. Using `deploy` until that happens seems fine to me

ymekuria: +1 to @mitschabaude comment. The contract is based off of the example that @bkase and others have been presenting at meetups. We decided to leave the deploy method with the state initialization in the example until we can automate it.

src/examples/hello_world/hello_world.ts
ymekuria: This line and the deploy method is from the Hello World example presentation. @bkase is there a reason you added this to the presentation example?

ymekuria:
> I would prefer to write large zkApp tests as linear scripts rather than jest tests, like the existing examples. For me that scripting workflow is a kind of local optimum, because it makes debugging quite easy and you get good error traces (especially if you use `node --stack-trace-limit=1000` -- the `./run` script already has this built in).
> 
> The error traces jest shows are so short that if a test in mina-signer broke (which uses jest) I often ended up copying the code out of the test, into a little script that I could run with `node --stack-trace-limit=1000`, just to see where the error comes from. I'm not excited about having to do that in snarkyjs as well.

That's a good point. Are you referring to the run.ts scripts we created in the CLI examples like this https://github.com/o1-labs/zkapp-cli/blob/main/examples/sudoku/ts/src/run.ts ?

ymekuria:
> * Re file structure, for a good balance of number of folders vs number of files per folder, I think we should put all of those big zkapp examples in a common folder, e.g. `examples/zkapps`, and in that folder IMO the files and tests can just be all colocated

In this structure you mentioned, would every zkApp have its own folder with the contracts and tests within it? Something like this ?

```
- examples
  - zkapps
    - example_a
      example_a.ts
      run_example_a.ts 
    - example_b
      example_a.ts
      run_example_a.ts
```

ymekuria:
> * I have mixed feelings about separating smart contracts and tests / script. The benefit is it gives us "clean" example smart contracts, which could for instance be deployed as-is. The downside is more jumping around when reading and writing the scripts.

I struggle with this also. If the benefits of building these contracts is only to test SnarkyJS and the protocol, I would include the tests in the contract file. It feels like it would bee a missed opportunity if we don't build "clean" smart contracts that can be deployed and used as examples for the community.

mitschabaude:
> > I would prefer to write large zkApp tests as linear scripts rather than jest tests, like the existing examples. For me that scripting workflow is a kind of local optimum, because it makes debugging quite easy and you get good error traces (especially if you use `node --stack-trace-limit=1000` -- the `./run` script already has this built in).
> > The error traces jest shows are so short that if a test in mina-signer broke (which uses jest) I often ended up copying the code out of the test, into a little script that I could run with `node --stack-trace-limit=1000`, just to see where the error comes from. I'm not excited about having to do that in snarkyjs as well.
> 
> That's a good point @mitschabaude . Are you referring to the run.ts scripts we created in the CLI examples like this https://github.com/o1-labs/zkapp-cli/blob/main/examples/sudoku/ts/src/run.ts ?

I was mainly referring to existing examples in snarkyjs, like these:
* https://github.com/o1-labs/snarkyjs/blob/main/src/examples/simple_zkapp.ts
* https://github.com/o1-labs/snarkyjs/blob/main/src/examples/state_update_rollup.ts
* https://github.com/o1-labs/snarkyjs/blob/main/src/examples/simple_zkapp_with_proof.ts
These are similar to what we are building now -- examples that exercise several zkApp features in a single, interconnected test. They are just not run in CI right now (but they should!)

mitschabaude:
> In this structure you mentioned, would every zkApp have its own folder with the contracts and tests within it? Something like this ?

I was thinking all files in one folder. One point in favor of that is that it makes them easier to look at in the github UI, where jumping between folders is a bit inconvenient. Other than that, it's down to aesthetic preference I guess -- what's your opinion?

ymekuria:
> > In this structure you mentioned, would every zkApp have its own folder with the contracts and tests within it? Something like this ?
> 
> I was thinking all files in one folder. One point in favor of that is that it makes them easier to look at in the github UI, where jumping between folders is a bit inconvenient. Other than that, it's down to aesthetic preference I guess -- what's your opinion?

I see your point about the github UI. I thought folders would be a good choice because some of the larger zkapps may have multiple files associated with them. If there will only be a contract file and run script file for every zkApp, I'm fine with putting them all in one folder.

mitschabaude:
> I see your point about the github UI. I thought folders would be a good choice because some of the larger zkapps may have multiple files associated with them. If there will only be a contract file and run script file for every zkApp, I'm fine with putting them all in one folder.

Oh you're right, I hadn't thought of that. Some of them will be pretty big! We should definitely have separate folders for those, so having a separate folder for this one is consistent 👍🏻

src/examples/hello_world/hello_world.ts
bkase: This was added to the presentation to show off the account preconditions features. It’s probably nice to mirror the presentation so I think we should leave it in

bkase (approved):
This looks good to me! Excited to see what the run script looks like!

PR #296, opened 2022-07-20
zkApp composability, pt 2: multiple proofs

mitschabaude:
Small PR which fixes the case of multiple independent proofs in one transaction.
This isn't composability, but a case that came up on initial testing that turned out to be broken. The PR fixes it and adds an example which combines the proofs of two smart contracts together in one transaction block

The PR also has a couple of other minor changes, to collect unrelated stuff which would deter from reviewing the bigger chunks (see next PR)

### DIFF
--- a/src/examples/zkapps/simple_and_counter_zkapp.ts
+++ b/src/examples/zkapps/simple_and_counter_zkapp.ts
@@ -1,0 +1,249 @@
+/**
+ * This is just two zkapps mixed together in one file, with their respective interactions bundled
+ * in the same transaction, to check that this actually works.
+ * -) "simple zkapp", testing state updates + events + account preconditions + child parties
+ * -) "counter rollup", testing state updates + sequence events / reducer
+ */
+
+import {
+  Field,
+  state,
+  State,
+  method,
+  UInt64,
+  PrivateKey,
+  SmartContract,
+  Mina,
+  Party,
+  isReady,
+  Permissions,
+  DeployArgs,
+  UInt32,
+  Bool,
+  PublicKey,
+  Circuit,
+  Experimental,
+} from 'snarkyjs';
+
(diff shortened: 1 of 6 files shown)


src/examples/zkapps/simple_and_counter_zkapp.ts
```diff
@@ -0,0 +1,249 @@
+/**
```
mitschabaude, line 1: the new example!

src/lib/party.ts
```diff
@@ -657,8 +657,14 @@ class Party {
   }
 
   hash() {
-    let fields = Types.Party.toFields(toPartyUnsafe(this));
-    return Ledger.hashPartyFromFields(fields);
+    // these two ways of hashing are (and have to be) consistent / produce the same hash
```
mitschabaude, line 660: this is what fixes the new example. the previous version was always returning variables, even outside the circuit

MartinMinkov (approved):
LGTM!

src/examples/zkapps/simple_and_counter_zkapp.ts
```diff
@@ -0,0 +1,249 @@
...
+
+    // compute the new counter and hash from pending actions
+    // remark: it's not feasible to pass in the pending actions as method arguments, because they have dynamic size
+    let { state: newCounter, actionsHash: newActionsHash } =
+      this.reducer.reduce(
+        offchainStorage.pendingActions,
+        // state type
+        Field,
+        // function that says how to apply an action
+        (state: Field, _action: Field) => {
```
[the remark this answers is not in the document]
MartinMinkov: Nvm, just realized that `actionsHash` is being passed in. I guess the use would be extra information we want to pass into the reducer function to do something specific?

PR #297, opened 2022-07-22
zkApp composability

### LINKED ISSUE #185: Support interactions between 2 smart contracts (i.e. composability, call stack)

mitschabaude:
This is the final PR for zkApp composability. 

* Implements the RFC here: https://github.com/o1-labs/snarkyjs/issues/303 (please read before reviewing this)
* Stacked on top of https://github.com/o1-labs/snarkyjs/pull/294
* Closes #185


src/examples/zkapps/composability.ts
```diff
@@ -0,0 +1,118 @@
+/**
```
mitschabaude, line 1: this is the example that demonstrates composability

src/lib/circuit_value.ts
```diff
@@ -533,3 +546,38 @@ Circuit.constraintSystem = function <T>(f: () => T) {
   );
   return result;
 };
+
+let memoizationContext = Context.create<{
```
mitschabaude, line 550: new global context that is used to remember prover values which are created during `createTransaction`, to be reused in the prover for the same method. In snarkyjs, only used for the `blindingValue` right now, but I figured it would be nice to also expose that capability to the outside world -- that's what the `memoizeWitness` function below is for.

src/lib/proof_system.ts
```diff
@@ -467,6 +472,43 @@ function synthesizeMethodArguments(
   return args;
 }
 
+function methodArgumentsToConstant(
+  { allArgs, proofArgs, witnessArgs }: MethodInterface,
```
mitschabaude, line 476: two new helper functions we need to convert the input arguments of a zkapp call to inputs to our hash function

src/lib/zkapp.ts
```diff
@@ -93,84 +110,277 @@ function method<T extends SmartContract>(
...
-      // important to run this with a fresh party everytime, otherwise compile messes up our circuits
-      // because it runs this multiple times
-      let [, result] = Mina.currentTransaction.runWith(
-        {
-          sender: undefined,
-          parties: [],
-          fetchMode: inProver() ? 'cached' : 'test',
-          isFinalRunOutsideCircuit: false,
-        },
+    if (!smartContractContext.has()) {
```
mitschabaude, line 138: This function -- `wrappedMethod` -- is what always gets called when you call any smart contract method. It is where most of the new logic lives. The function is now quite complex!

This first, large if branch is handling the case that was supported already -- calling a zkApp method when you're _not_ inside another zkApp method.

src/lib/zkapp.ts
```diff
@@ -93,84 +110,277 @@ function method<T extends SmartContract>(
...
-          checkPublicInput(publicInput, this.self);
-
-          // check the self party right after calling the method
-          // TODO: this needs to be done in a unified way for all parties that are created
-          assertPreconditionInvariants(this.self);
-          cleanPreconditionsCache(this.self);
-          assertStatePrecondition(this);
-          return result;
+          if (inCheckedComputation()) {
+            // important to run this with a fresh party everytime, otherwise compile messes up our circuits
```
mitschabaude, line 143: the code below didn't really change in its entirety. It's just marked green because there is more nesting, causing two more spaces on the left

src/lib/zkapp.ts
```diff
@@ -93,84 +110,277 @@ function method<T extends SmartContract>(
...
+                  actualArgs
+                );
+                if (methodIntf.returnType) {
+                  callDataFields.push(
+                    ...methodIntf.returnType.toFields(result)
+                  );
+                }
+                callDataFields.push(getMethodId(methodIntf, methodIndex));
+                callDataFields.push(blindingValue);
+                party.body.callData = Poseidon.hash(callDataFields);
```
mitschabaude, line 184: This whole code block is the case that we're in a circuit. So, these commands are what is added to the circuit on top of the user code in a `@method`. It is also how a called method is executed inside its own prover (but not in the prover of the caller -- that case is further below!). On this line, we're populating our own `callData` field -- as explained [in the RFC](https://github.com/o1-labs/snarkyjs/issues/303), this is done every time, to make the method callable by default

src/lib/zkapp.ts
```diff
@@ -93,84 +110,277 @@ function method<T extends SmartContract>(
...
-          args: clonedArgs,
-          // proofs actually don't have to be cloned
-          previousProofs: getPreviousProofsForProver(actualArgs, methodIntf),
-          ZkappClass,
-        });
-      }
-      return result;
+      )[1];
     }
+    // if we're here, this method was called inside _another_ smart contract method
```
mitschabaude, line 261: this is what happens inside the caller, _inside_ calling another method. the code below has some comments to explain what's happening

mitschabaude (commented):
some comments to help reviewing!

src/lib/zkapp.ts
```diff
@@ -93,84 +110,277 @@ function method<T extends SmartContract>(
...
+        // so we have to explain to the user how to do that
+        let { returnType } = methodIntf;
+        let noReturnTypeError =
+          `To return a result from ${methodIntf.methodName}() inside another zkApp, you need to declare the return type.\n` +
+          `This can be done by annotating the type at the end of the function signature. For example:\n\n` +
+          `@method ${methodIntf.methodName}(): Field {\n` +
+          `  // ...\n` +
+          `}\n\n` +
+          `Note: Only types built out of \`Field\` are valid return types. This includes snarkyjs primitive types and custom CircuitValues.`;
+        // if we're lucky, analyzeMethods was already run on the callee smart contract, and we can catch this error early
```
ymekuria, line 277: Clarifying question. When does analyzeMethods run for the first time on a contract?

mitschabaude: It runs
* in `SmartContract.compile()` (before actually compiling)
* when computing the `SmartContract.digest()`
* in a `Mina.transaction` block, it runs if one of the methods needs it under the hood; this is only the case when reducing sequence events right now.
So, it definitely runs when we're proving things (because we `compile` first), but likely not if we don't.

The code above will throw the helpful error message if we're not proving, so that case is covered.

The case that's problematic is when we compile the caller first, and the callee second. Then, the "if we're lucky" case above is not true in the caller circuit. Then, we won't be able to nicely catch the missing return type annotation, and the caller's `compile` will throw an ugly, unhelpful error instead.

Thanks for poking into this :D Maybe I should make it so that analyzeMethods figures out all the SmartContracts called inside a method, and also runs analyzeMethods on them. This way, the callee would definitely be analyzed already in the code above, so we'd avoid the "ugly error" case.

src/examples/zkapps/composability.ts
```diff
@@ -0,0 +1,118 @@
...
+tx.send();
+
+console.log('call interaction');
+tx = await Mina.transaction(feePayer, () => {
+  // we just call one contract here, nothing special to do
+  zkapp.callAddAndEmit(Field(5), Field(6));
+  if (!doProofs) zkapp.sign(zkappKey);
+});
+if (doProofs) {
+  console.log('proving (3 proofs.. can take a bit!)');
```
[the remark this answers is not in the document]
mitschabaude: takes 42 sec for me

src/lib/zkapp.ts
mitschabaude: Update: I tried this out, but wasn't able to make it work. The ugly error is already thrown during analyzeMethods, so I have no way of finding the list of callees before. I think we'll have to live with a suboptimal error message in this edge case

ymekuria: > Update: I tried this out, but wasn't able to make it work. The ugly error is already thrown during analyzeMethods, so I have no way of finding the list of callees before. I think we'll have to live with a suboptimal error message in this edge case

At least we are aware of the edge case error behavior.

MartinMinkov (approved):

ymekuria (approved):

mitschabaude:
Update: Implemented callData hashing as proposed here: https://github.com/o1-labs/snarkyjs/issues/303#issuecomment-1196441140
I think this is a solution that already makes sense, so leaving further tweaks to a separate PR

Issue #283, opened 2022-07-14
Flesh out Application A and P0 test cases


bkase:
Flesh out Application A and the P0 test cases that are identified for it. Specifics still being iterated on.

PR #302, opened 2022-07-25
Application A and P0 test cases

ymekuria:
**Description**
Addresses 
#283 

This PR adds an update smart contract to the examples folder along with a run.ts script used to test the contract. 

**File Structure and Location**

The smart contract and the run script are located in the  `src/examples/zkapps` directory of the SnarkyJS repo. Future test applications and run scripts can be added to the zkapps examples directory with a folder structure similar to this:

```
- examples
  - zkapps
    - hello_world
      hello_world.ts
      run.ts 
```

**Testing and CI**

The decision was made to keep the tests in a separate file so the examples contracts can be easily deployed as is in the future. The tests are written as a linear script in the `run.ts` file. An npm `test:integration` script command was added to the package.json and also included in the SnarkyJS CI. Future test application run scripts can be included in the CI by appending `./run` and the path to the `run.ts` file in the package.json test:integration script.

`"test:integration": "./run src/examples/zkapps/hello_world/run.ts  &&  ./run src/examples/zkapps/example_b/run.ts"`

### DIFF
--- a/src/examples/zkapps/hello_world/hello_world.ts
+++ b/src/examples/zkapps/hello_world/hello_world.ts
@@ -22,7 +22,7 @@ export class HelloWorld extends SmartContract {
       ...Permissions.default(),
       editState: Permissions.proofOrSignature(),
     });
-    this.x.set(Field(3));
+    this.x.set(Field(2));
 
     Party.setValue(this.self.update.delegate, adminPublicKey);
   }
--- a/src/examples/zkapps/hello_world/hello_world.ts
+++ b/src/examples/zkapps/hello_world/hello_world.ts
@@ -29,7 +29,7 @@ export class HelloWorld extends SmartContract {
 
   @method update(squared: Field, admin: PrivateKey) {
     const x = this.x.get();
-    this.x.assertEquals(this.x.get());
+    this.x.assertNothing();
     x.square().assertEquals(squared);
     this.x.set(squared);
 
--- a/src/examples/zkapps/hello_world/hello_world.ts
+++ b/src/examples/zkapps/hello_world/hello_world.ts
@@ -34,6 +34,7 @@ export class HelloWorld extends SmartContract {
     this.x.set(squared);
 
     const adminPk = admin.toPublicKey();
+
     this.account.delegate.assertEquals(adminPk);
   }
 }
--- a/src/examples/zkapps/hello_world/hello_world.ts
+++ b/src/examples/zkapps/hello_world/hello_world.ts
@@ -7,7 +7,7 @@ import {
   method,
   DeployArgs,
   PrivateKey,
-} from '../../../dist/server';
+} from '../../../../dist/server';
 
 export const adminPrivateKey = PrivateKey.random();
 export const adminPublicKey = adminPrivateKey.toPublicKey();
--- a/src/examples/zkapps/hello_world/hello_world.ts
+++ b/src/examples/zkapps/hello_world/hello_world.ts
@@ -7,7 +7,8 @@ import {
   method,
   DeployArgs,
   PrivateKey,
-} from '../../../../dist/server';
+  Permissions,
+} from 'snarkyjs';
 
 export const adminPrivateKey = PrivateKey.random();
 export const adminPublicKey = adminPrivateKey.toPublicKey();
@@ -17,13 +18,18 @@ export class HelloWorld extends SmartContract {
 
   deploy(input: DeployArgs) {
     super.deploy(input);
+    this.setPermissions({
+      ...Permissions.default(),
+      editState: Permissions.proofOrSignature(),
+    });
     this.x.set(Field(3));
+
     Party.setValue(this.self.update.delegate, adminPublicKey);
   }
 
   @method update(squared: Field, admin: PrivateKey) {
     const x = this.x.get();
-    this.x.assertEquals(x);
+    this.x.assertEquals(this.x.get());
     x.square().assertEquals(squared);
     this.x.set(squared);
 
--- a/src/examples/zkapps/hello_world/run.ts
+++ b/src/examples/zkapps/hello_world/run.ts
@@ -27,15 +27,49 @@ try {
 }
 
 const initialState = zkAppInstance.x.get().toString();
+let currentState;
 
-console.log(`Updating from ${initialState} to 9 failed with error`);
+console.log(`Updating state from ${initialState} to 9 with Admin Private Key`);
 
 txn = await Mina.transaction(deployerAccount, () => {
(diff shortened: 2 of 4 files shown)


src/examples/zkapps/hello_world/hello_world.ts
```diff
@@ -17,17 +18,23 @@ export class HelloWorld extends SmartContract {
...
+    });
+    this.x.set(Field(2));
+
     Party.setValue(this.self.update.delegate, adminPublicKey);
   }
 
   @method update(squared: Field, admin: PrivateKey) {
     const x = this.x.get();
-    this.x.assertEquals(x);
+    this.x.assertEquals(this.x.get());
```
shimkiv, line 32: We [decided](https://docs.google.com/document/d/1ha2pyPn7ksIcKRd2w2Oj9J-xFvAxoCw0DNsApVpic9s/edit?disco=AAAAcV5-Vd4) to remove this assertion.

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+  if (currentState !== '4') {
+    throw new Error(
+      `Current state of ${currentState} does not match 4 after calling update with 4`
+    );
+  } else {
+    console.log(`Current state succesfully updated to ${currentState}`);
+  }
+} catch (err: any) {
+  throw new Error(`
+    Updating from ${initialState} to 4 failed with error ${err.message}`);
```
mitschabaude, line 65: same comment as above re: preserving original errors

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+);
+txn = await Mina.transaction(feePayer1, () => {
+  zkAppInstance.update(Field(16), wrongAdminPrivateKey);
+  zkAppInstance.sign(zkAppPrivateKey);
+});
+
+try {
+  txn.send().wait();
+
+  currentState = await Mina.getAccount(
```
mitschabaude, line 81: the logic here might be removed if we expect `txn.send()` to fail anyway

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+console.log(
+  `Attempting to update state from ${currentState} to 16 with incorrect Admin Private Key ...`
+);
+txn = await Mina.transaction(feePayer1, () => {
+  zkAppInstance.update(Field(16), wrongAdminPrivateKey);
+  zkAppInstance.sign(zkAppPrivateKey);
+});
+
+try {
+  txn.send().wait();
```
mitschabaude, line 79: this test case should _fail_ if `txn.send().wait()` incorrectly succeeds. currently, if it succeeds nothing bad happens.

you need something like
```ts
let correctlyFails = false;
try {
  txn.send().wait();
} catch (err) {
  correctlyFails = true;
}
if (!correctlyFails) throw Error('We could update the state with the wrong admin key');
```

BONUS: Ideally, in the catch block you should check if the error is exactly the error we expect. In this case, we expect the following line to fail:
```ts
this.account.delegate.assertEquals(adminPk);
```
I think this would throw an `assert_equal` error if the admin key is wrong, so we could check if `err.message.includes("assert_equal")`, and re-throw the original error otherwise

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+
+  // expected to fail and current state stays at 4
+  txn = await Mina.transaction({ feePayerKey: feePayer1, fee: '10' }, () => {
+    zkAppInstance.update(Field(256), adminPrivateKey);
+    zkAppInstance.sign(zkAppPrivateKey);
+  });
+
+  txn.send();
+
+  currentState = await Mina.getAccount(
```
mitschabaude, line 138: same comments as before on properly checking the failure case. what you do -- throwing a different error after something that should've thrown -- would also properly test for the expected failure, but _only if_ you distinguish between those two errors in the catch block

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+  txn3.send();
+
+  currentState = await Mina.getAccount(
+    zkAppAddress
+  ).zkapp?.appState[0].toString();
+
+  console.log(`Update successful. Current state is ${currentState}.`);
+} catch (err: any) {
+  throw new Error(
+    `State was updated to ${currentState} with error ${err.message}`
```
mitschabaude, line 193: same comment as above re: preserving original errors

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+
+  txn2.send();
+
+  currentState = await Mina.getAccount(
+    zkAppAddress
+  ).zkapp?.appState[0].toString();
+
+  console.log(`Update successful. Current state is ${currentState}.`);
+} catch (err: any) {
+  throw new Error(
```
mitschabaude, line 172: same comment as above re: preserving original errors

src/examples/zkapps/hello_world/run.ts
```diff
@@ -1,26 +1,214 @@
...
+  // expected to fail and current state remains 256
+  txn4 = await Mina.transaction({ feePayerKey: feePayer1, fee: '1' }, () => {
+    zkAppInstance.update(Field(16), adminPrivateKey);
+    zkAppInstance.sign(zkAppPrivateKey);
+  });
+
+  txn4.send().wait();
+
+  throw new Error(`State was incorrectly updated to ${currentState}`);
+} catch (err: any) {
```
mitschabaude, line 207: same comment as above re: properly testing for an expected failure

mitschabaude (changes requested):
I like this! Thanks for putting it in CI too 💯 

Left some comments, I think some of the testing logic needs a bit of refinement, and we should strive to have useful error traces in case something breaks

src/examples/zkapps/hello_world/run.ts
mitschabaude: Or, you can save the `correctlyFails` variable if you throw two different errors in the try block, one on the failure case and one in the (incorrect) success case, and distinguish between those errors in the catch block (same comment below)

src/examples/zkapps/hello_world/hello_world.ts
ymekuria: An error will be produced without this precondition with the current version of SnarkyJS.

mitschabaude (approved):
LGTM!

PR #306, opened 2022-07-26
Circuit value improvements

mitschabaude:
This fixes some issues with CircuitValue
* It no longer breaks when you declare a constructor with different argument structure than the base constructor. That's because `ofFields` and `fromJSON` no longer use `new this(...)` to instantiate the class
* There's now a `CircuitValue.fromObject` method, which is nicely typed (in contrast to the generic constructor). It expects an objects that has the same props as the CircuitValue, but without the methods.


Trivo25 (approved):
a much needed improvement, makes working with CircuitValues a lot easier!

PR #310, opened 2022-07-30
Hash inputs in JS

### LINKED ISSUE #180: Fix setting timestamp precondition

mitschabaude:
- snarkyjs side of https://github.com/MinaProtocol/mina/pull/11590
- implements computing _hash inputs_ on the JS side:
  - Before, we had to convert snarkyjs parties to `Party.Checked.t` on the OCaml side, where it would use `Party.Checked.digest` to compute its hash. Under the hood, the most of the logic in `Party.Checked.digest` is about converting the party to a list of field elements, the "hash input", which is then hashed.
  - In this PR, we implement `Types.Party.toInput()` completely on the JS side, removing a crucial dependency of snarkyjs on `mina_base`. Party hashing in JS now only depends on OCaml via `Poseidon.update` which provides the "raw", non-application-specific hashing logic
  - `toInput` is implemented as usual, with a recursive function leveraging the auto-generated party structure
  - the PR contains a refactor of `parties-helpers.ts` which achieves sharing code between most of the functions which walk the auto-generated structure
- consistency between the OCaml and JS sides of computing party hashes is enforced in various ways:
  1. a new unit test that runs in Mina CI: https://github.com/MinaProtocol/mina/pull/11590/files#diff-9bd2909dd52a761e3aab0e4ad4ed357803f352286e1bdadcab95732b774b97a6R1
  2. We still use the old OCaml logic for computing party hashes _outside_ the circuit. This can be removed at some point, but it provides a very tight consistency check for now: https://github.com/o1-labs/snarkyjs/pull/310/files#diff-7df11bef616dc6cacab4cb8ee003eeab5fb37eaa6c39678415965ed52c8a57d3R815
- in support for consistent hash inputs, fixes how token symbols are handled in snarkyjs, adds a new `TokenSymbol` circuit type
- unrelatedly, this has some minor changes related to fixing timestamp preconditions (i.e., this is also the snarkyjs companion for https://github.com/MinaProtocol/mina/pull/11577)
- closes #180


src/examples/to-hash-input.ts
```diff
@@ -0,0 +1,179 @@
+import {
```
mitschabaude, line 1: this is a sort of unit test, which is also run in Mina CI. I left it here just because for the Mina version, I had to remove the TS types, and it's easier to read & modify with the types in place

src/lib/fetch.ts
```diff
@@ -139,7 +139,7 @@ function toPermission(p: AuthRequired): Permission {
 type FetchedAccount = {
   publicKey: string;
   nonce: string;
-  tokenId: string;
+  token: string;
```
mitschabaude, line 142: this file has some unrelated fixes which turned up when debugging snarkyjs transactions against a local Mina node, while debugging timestamp preconditions

src/lib/hash.ts
```diff
@@ -78,3 +86,70 @@ function prefixToField(prefix: string) {
...
+      currentPackedField = field;
+    }
+  }
+  packedBits.push(currentPackedField);
+  return fields.concat(packedBits);
+}
+
+type Input = { fields?: Field[]; packed?: [Field, number][] };
+
+type TokenSymbol = { symbol: string; field: Field };
```
mitschabaude, line 117: new `TokenSymbol` type!

src/lib/hash.ts
```diff
@@ -78,3 +86,70 @@ function prefixToField(prefix: string) {
     .flat();
   return Field.ofBits(bits);
 }
+
+/**
+ * Convert the {fields, packed} hash input representation to a list of field elements
+ * Random_oracle_input.Chunked.pack_to_fields
+ */
+function packToFields({ fields = [], packed = [] }: Input) {
```
mitschabaude, line 94: this is the last step that's necessary to convert the result of `toInput` to something which can be hashed

src/lib/party.ts
```diff
@@ -810,9 +812,14 @@ class Party implements Types.Party {
...
+    // TODO: there's no reason anymore to use two different hashing methods here!
+    // -- the "inCheckedComputation" branch works in all circumstances now
+    // we just leave this here for a couple more weeks, because it checks consistency between
+    // JS & OCaml hashing on *every single party proof* we create. It will give us 100%
+    // confidence that the two implementations are equivalent, and catch regressions quickly
     if (inCheckedComputation()) {
-      let fields = Types.Party.toFields(this);
-      return Ledger.hashPartyFromFields(fields);
+      let input = Types.Party.toInput(this);
+      return hashWithPrefix(prefixes.body, packToFields(input));
```
mitschabaude, line 822: this is the main achievement of this PR!

src/lib/precondition.ts
```diff
@@ -57,8 +57,6 @@ let unimplementedPreconditions: LongKey[] = [
   'account.isNew',
   // this is partially unimplemented because the field is not returned by the local blockchain
   'account.delegate',
-  // this is unimplemented because setting this precondition made the integration test fail
-  'network.timestamp',
```
mitschabaude, line 61: fixed the timestamp precondition, so removed it from the list of unimplemented preconditions!

src/lib/state.ts
```diff
@@ -257,7 +257,7 @@ To write a correct circuit, you must avoid any dependency on the concrete value
         throw Error(
           'fetch can only be called when the State is assigned to a SmartContract @state.'
         );
-      if (Mina.currentTransaction !== undefined)
+      if (Mina.currentTransaction.has())
```
mitschabaude, line 260: unrelated fix

src/lib/zkapp.ts
```diff
@@ -794,8 +794,11 @@ async function deploy<S extends typeof SmartContract>(
       let amount = UInt64.fromString(String(initialBalance)).add(
         Mina.accountCreationFee()
       );
-      let party = Party.createSigned(feePayerKey);
+      let feePayerAddress = feePayerKey.toPublicKey();
+      let party = Party.defaultParty(feePayerAddress);
+      party.body.useFullCommitment = Bool(true);
```
mitschabaude, line 799: unrelated fix, necessary for intg test

src/snarky/parties-helpers.ts
```diff
@@ -52,183 +56,115 @@ function asFieldsAndAux<T, TJson>(typeData: Layout, customTypes: CustomTypes) {
   };
 }
 
-function toJson(typeData: Layout, value: any, customTypes: CustomTypes): any {
```
mitschabaude, line 55: lots of refactoring here -- see the new `mapReduce` function at the bottom of the file, which abstracts what most of the functions here were doing!

mitschabaude (commented):
some comments to guide review:

src/snarky/parties-helpers.ts
```diff
@@ -297,46 +229,122 @@ function fromFieldsReversed(
...
+        return {};
+      },
+      customTypes,
+    },
+    typeData,
+    value
+  );
+}
+
+type MapReduceSpec<T, R> = {
```
bkase, line 294: I think this is semantically more of a `fold` ala `Fields.fold` in the ocaml code that’s generated from `deriving fields`. I think something like FoldSpec and genericFold or layoutFold?

bkase (approved):
Looks good! Just one naming nit

src/lib/fetch.ts
MartinMinkov: Good catch on figuring out these fixes!

src/lib/party.ts
MartinMinkov: Awesome work!

ymekuria (approved):

src/examples/to-hash-input.ts
```diff
@@ -0,0 +1,179 @@
...
+  Ledger.hashInputFromJson.accountPrecondition,
+  account
+);
+
+// network precondition
+let NetworkPrecondition = asFieldsAndAux<NetworkPrecondition, any>(
+  jsLayout.Party.entries.body.entries.preconditions.entries.network
+);
+let network = party.body.preconditions.network;
+party.network.stakingEpochData.ledger.hash.assertEquals(Field.random());
```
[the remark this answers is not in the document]
mitschabaude: It changes a precondition on the `Party`. I'm trying to add as much changes to the party here as possible, to not miss a part it where the hashing logic doesn't agree between JS and OCaml (if I were just using the default party that I start out with, the hash input would be mostly zeroes -- so two of those zeros could be reversed and we wouldn't notice)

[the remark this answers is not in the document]
ymekuria: Ok that makes sense.

PR #322, opened 2022-08-03
Use 'rimraf' instead of OS specific 'rm'.

shimkiv:


mitschabaude (approved):
LGTM. I wondered at first whether we couldn't use a trivial nodejs script instead of an external dependency (trying to avoid dependencies in general..). however the `rimraf` package is widely used -- 55M downloads per week -- and its code looks fine to me. also, it handles a fair amount of edge cases so it provides value, and doesn't introduce additional nested dependencies.

PR #323, opened 2022-08-03
`fetchEvents` for LocalBlockchain

Trivo25:
### Description

This is the first part of extending `LocalBlockchain` to support `fetchEvents` and later in a second PR `fetchSequenceEvents`. 

### Under the hood

A smart contract can emit events 

```ts
@method emitMe(value: Field) {
	this.emit("someEvent", value);
}
````

but the `LocalBlockchain` did not store emitted events until now. After the transaction has successfully been applied to the `Ledger` via `applyJsonTransaction`, all events are stored in a variable `events`. Events are identified by `publicKey` and `tokenId`.

`LocalBlockchain` now exposes a closure `fetchEvents()`, which returns all events that have been emitted.

The `SmartContract` class has been extended by adding a new method `fetchEvents(start: UInt32, end?: UInt32): { type: string; event: AsFieldElements<any> }[]`, which returns a list of all events that have been emitted by a `SmartContract` instance, using the `tokenId` and `publicKey` as identifier and `start` and `end` as range options.

Note: There has been a [discussion](https://github.com/o1-labs/snarkyjs/issues/248#issuecomment-1172714578) regarding what type to use in order to filter events. I went with using `slots` as an indicator for now.

### Example

Developers can now call the `fetchEvents(start: UInt32, end?: UInt32)` method on an instance of a smart contract that uses the LocalBlockchain as active instance.

```ts

class MyApp extends SmartContract {
	
	events = {
		someEvent: Field,
	};

	@method emitMe() {
		this.emit("someEvent", Field.zero);
	}
}

// ...

let events = myAppInstance.fetchEvents(UInt32.from(0))
console.log(events)
/*
[
	{ type: "someEvent", event: Field(0)  }
]

*/


````

### CODE REVIEW

File: src/lib/zkapp.ts [hunk 1]
```diff
@@ -611,6 +611,50 @@ class SmartContract {
...
+          ? start.lte(slot).toBoolean()
+          : start.lte(slot).toBoolean() && slot.lte(end).toBoolean();
+      })
+      .map((el: any) => el.events)
+      .flat();
+
+    // used to match field values back to their original type
+    let sortedEventTypes = Object.keys(this.events).sort();
+
+    return events.map((event: any) => {
```
Trivo25, line 633: It would probably be smarter to align the events in a stack-like structure, where new events are at the beginning of the array instead of the end. How is it enforced on the protocol level, or rather archive api?

mitschabaude: hmm, not sure what's better to use 🤔 yeah we should probably mirror whatever the graphql API ends up doing

src/lib/mina.ts
```diff
@@ -266,10 +271,33 @@ function LocalBlockchain({
...
+
       ledger.applyJsonTransaction(
-        JSON.stringify(partiesToJson(txn.transaction)),
+        JSON.stringify(partiesJson),
         String(accountCreationFee)
       );
+
+      // fetches all events from the transaction and stores them
+      // events are identified and associated with a publicKey and tokenId
+      partiesJson.otherParties.forEach((p: any) => {
```
mitschabaude, line 284: nit: I believe the `any` annotation in the callback unnecessarily removes the type information on `p`. typescript should now that `p: Party` without a type annotation

src/lib/mina.ts
```diff
@@ -290,6 +318,9 @@ function LocalBlockchain({
     applyJsonTransaction(json: string) {
       return ledger.applyJsonTransaction(json, String(accountCreationFee));
     },
+    fetchEvents(publicKey: PublicKey, tokenId: Field): any {
+      return events[publicKey.toBase58()][Ledger.fieldToBase58(tokenId)];
```
mitschabaude, line 322: could we return an empty array here instead of failing / returning `undefined`, if there are no events stored for that account / tokenId yet?
Also, I know we are bad at typing events right now, but I'd prefer `any[]` as the return type over `any`

mitschabaude: Another thing: `tokenId` should be optional and use the default id by default, like `getAccount`

src/lib/zkapp.ts
```diff
@@ -611,6 +611,50 @@ class SmartContract {
     party.body.events = Events.pushEvent(party.body.events, eventFields);
   }
 
+  fetchEvents(
+    start: UInt32,
```
mitschabaude, line 615: could we also make `start` optional, and default to `0`? David suggested that on the events RFC as a common use case

src/lib/zkapp.ts
```diff
@@ -611,6 +611,50 @@ class SmartContract {
     party.body.events = Events.pushEvent(party.body.events, eventFields);
   }
 
+  fetchEvents(
```
mitschabaude, line 614: I think we should give this an `async` signature. `Mina.fetchEvents` as well. In the local blockchain case, its not really async, but in the remote blockchain case it is, so we need `async` to cover both cases

mitschabaude (changes requested):
nice work on your first PR!  left a couple of suggestions

src/examples/local_events_zkapp.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const doProofs = false;
+
+await isReady;
+
+class Event extends CircuitValue {
+  @prop pub: PublicKey;
+  @prop value: Field;
+
+  constructor(pub: PublicKey, value: Field) {
```
ymekuria, line 31: We had a discussion awhile back about adding a static `from` method that invokes the original constructor instead of overriding the constructor in examples to encourage this pattern. Here is the [discussion](https://github.com/o1-labs/snarkyjs/pull/269) for background and here is an [example](https://github.com/o1-labs/snarkyjs/pull/269/files#diff-4f28cac68e5aaf84bb21ddd42725b110f5ba1665c49ffe63164e6a9ad0c97165R23).

What you are doing works (passing constructor arguments to super) but we want to encourage the `.from `pattern to the community. @mitschabaude do you still think we should encourage this `.from` pattern in our examples?

ymekuria (commented):
Nice work Florian! I left some comments.

src/examples/local_events_zkapp.ts
mitschabaude: Uhm, to be honest, I'm not sure. I think it's fine either way. My original idea was that you'd have to do nothing and use the generic constructor. `.from` would've only be for the few cases where you'd want to transform the input arguments. However, by now I think a constructor like this, which is a no-op, is still nice to add, just so it becomes properly typed

src/lib/mina.ts
```diff
@@ -318,8 +318,13 @@ function LocalBlockchain({
...
     },
-    fetchEvents(publicKey: PublicKey, tokenId: Field): any {
-      return events[publicKey.toBase58()][Ledger.fieldToBase58(tokenId)];
+    async fetchEvents(
+      publicKey: PublicKey,
+      tokenId: Field = getDefaultTokenId()
+    ): Promise<any[]> {
+      let eventsForAccount =
+        events?.[publicKey.toBase58()]?.[Ledger.fieldToBase58(tokenId)];
+      return eventsForAccount === undefined ? [] : eventsForAccount;
```
mitschabaude, line 327:
```ts
return events?.[publicKey.toBase58()]?.[Ledger.fieldToBase58(tokenId)] ?? []
```
😉

src/examples/local_events_zkapp.ts
ymekuria: That works for me. I'm fine leaving it as is if we are in agreement.

MartinMinkov (approved):
Nothing to add that others haven't said for feedback, approving preemptively! 
Great job! :D

mitschabaude (approved):

ymekuria (approved):

src/lib/mina.ts
Trivo25: ahh, perfect

PR #324, opened 2022-08-03
Finish unimplemented preconditions

### LINKED ISSUE #240: Finish unimplemented preconditions
This is a left-over from https://github.com/o1-labs/snarkyjs/pull/207, which was already big enough in scope

A couple of preconditions are not implemented yet for various reasons. Tasks:
* [ ] Parse the fetch result of various "hashes", which have a base58-like enconding, into `Field`
* [ ] Add some missing fields to the `LocalBlockchain`
* [ ] Nice-to-have: enable to set network state for local blockchain, so that network preconditions can be tested locally

mitschabaude:
- Closes #240 
- Implements all remaining preconditions, except two that are blocked by protocol work.
  * Ones that are still missing: `network.epochData.seed` is not checked on the protocol side yet; `account.provedState` is not available via graphql yet
- Expose most of the account fields for accounts in the local ledger to JS, so we can use them for account preconditions
- Add JS equivalents of encoding functions for various hashes: receipt chain hash, ledger hash etc. These are implemented as follows:
  * The `Field` that represents the hash is serialized to bytes in a way that mirrors the bin_prot serialization on the OCaml side (which is pretty straightforward -- just the bytes of the `Field` plus an optional version number at the beginning)
  * The conversion from bytes to a base58 string is done by using the `base58_check` lib from OCaml. "Version bytes" for different types which are expected by this lib are also imported from OCaml
- Move `getDefaultTokenId` and base58 encoders for token ids to a common namespace, `TokenId`, which is used internally. Export that namespace to the outside by putting it on `Token.Id`, i.e. as sub-namespace on the existing `Token` class


Trivo25 (approved):
A general namespace for base58 encoders is a great improvement!

ymekuria (approved):
LGTM!

Issue #281, opened 2022-07-14
Implement API for sending MINA tokens via Party transactions


MartinMinkov:
# Description
During an internal discussion of the [Tokens RFC](https://github.com/o1-labs/snarkyjs/issues/233), the product team decided to not include an API for transfers of MINA for zkApps. We instead decided to punt on this for a later PR as we wanted more time to decide on an easy to use API.

Once we have a good design in place for tokens, this should be implemented as part of this issue.

MartinMinkov:
# RFC

We wish to expose an intuitive and easy-to-use API for supporting sending MINA in party transfers. Adding transactions for MINA payments was discussed in #233 but was cut out of scope and should be implemented in another PR. Here, we propose an API that zkApp developers can use to support transactions with default MINA that is separate from the custom tokens API.

## Sending inside a zkApp method
An easy-to-use API should be exposed on the zkApp instance inside a method's execution. For example, a method name such as `send` can be attached to the `this` property inside a zkApp method. The send method will take as input an address to send MINA to as the receiver and the amount to send. The following code would be an example of such a method:

```ts
@method sendMINATokens(receiverAddress: PublicKey) {
    this.send({
        to: receiverAddress,
        amount: 10e9
    })
}
```
When the `sendMINATokens` method is invoked, the method will execute `this.send` with the specified properties. `this.send` will construct the parties necessary to represent a valid balance change in both the zkApp balance and the receiverAddress. 


## Sending using a constructed Party

In addition to exposing a way to send MINA tokens on a zkApp method execution, we would also like to support generating the parties structure by the zkApp developer outside of a zkApp. This can be done by adding a `send` method to the `Party` class inside SnarkyJS. This `send` method will behave similarly to the zkApp version by taking in an address to send to as the receiver and the amount. The following code would be an example of such method:

```ts
let party = Party.createSigned(privateKey); // create party from a private key
let party2 = Party.createUnsigned(publicKey); // Can also create a party from a public key but require authorization later before txn is submitted
party.send({ to: someAddress, amount: 10e9 }); // send MINA from the party's account
```
This way to construct a signed party and specify a send operation can be done inside a transaction block. This means that sending MINA in SnarkyJS isn't dependent on interacting with a zkApp and can be done anywhere a zkApp uses it.

The combination of these two approaches (inside a zkApp and with a signed party) should cover the primary use cases that a zkApp developer needs to send MINA in an easy-to-use way. 

## Testing

[...]

PR #325, opened 2022-08-03
Add `send` API to zkApp and Party classes

MartinMinkov:
Implements #281 

Adds the `.send()` method on both the Party and zkApp classes. Additionally added a test script to verify correct behavior.


src/lib/party.ts
```diff
@@ -670,6 +671,21 @@ class Party implements Types.Party {
     };
   }
 
+  send({ to, amount }: Omit<SendParams, 'from'>) {
+    let party = this;
+    let receiverParty = Party.createUnsigned(to);
```
mitschabaude, line 676: If the `this` party has a non-default token id, then this method could be used to send tokens (e.g., inside the smart contract callback we give to a token contract). This would be cool functionality! However, for this to work, the receiver party must get the same token id. (The current implementation would add Mina on the receiver party but subtract tokens from the `this` party, which would just make the txn fail)

So, I suggest to add the following line here:
```ts
receiverParty.body.tokenId = party.body.tokenId
```

mitschabaude: Another thing: We should think about whether we want to make the receiverParty the child of the sender here, by default. Otherwise, the receiver is not linked to the circuit, so I'm scared that a user would add checks on the receiver which would do nothing, and they'd have a security hole that could be exploited

mitschabaude (changes requested):

src/lib/party.ts
MartinMinkov: Ahh, that's a good point I didn't think about. When you say by default, is there a case that the receiverParty wouldn't want to be a child in this case? Seems like we would always want to link the receiver to the circuit to verify its correctness.

mitschabaude: Yeah let's do it always for now! I think it might make our lives a little harder when doing the token proofs authorization thing, but we'll figure it out then

MartinMinkov: Gotcha, done! :D

mitschabaude (approved):

ymekuria (approved):

src/examples/simple-zkapp-payment.ts
```diff
@@ -0,0 +1,131 @@
+import {
```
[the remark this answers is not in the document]
mitschabaude: good point, I think we'll accumulate more & more of those, so I'd add most of them to `/src/examples/zkapps` and reserve `/src/examples` for the most fundamental examples, to not overwhelm a user who explores these

Trivo25 (approved):

src/examples/simple-zkapp-payment.ts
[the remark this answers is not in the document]
MartinMinkov: Sure thing, I'll move it over

PR #327, opened 2022-08-04
add `getActions` to `LocalBlockchain`

Trivo25:
# Description

This is a follow up PR to https://github.com/o1-labs/snarkyjs/pull/323 that introduces `getActions` (`fetchSequenceEvents`) to the`LocalBlockchain` and the reducer on smart contracts.

In order to get it working https://github.com/o1-labs/snarkyjs/pull/324 is required, since it adds the needed `sequenceState` field to zkapp accounts. (DONE)

# Under the hood

Similar to `fetchEvents`, emitted actions are now stored in a variable called `actions` and the `LocalBlockchain` exposes a closure 'getActions` that returns all emitted actions.

zkApp instances can now fetch sequence events (actions) as follows:

```ts
this.reducer.getActions({
    fromActionHash?: Field;
    endActionHash?: Field;
}): Field[][]

```

Question: `getActions` currently is a part of the `reducer`. Should it stay there or become its own method on the `SmartContract` class, similar to `fetchEvents`?  

The example `src/examples/state_update_rollup.ts` has been extended to support multiple action rollups 

closes https://github.com/o1-labs/snarkyjs/issues/284


src/lib/zkapp.ts
```diff
@@ -826,6 +833,58 @@ Use the optional \`maxTransactionsWithActions\` argument to increase this number
...
+    getActions({
+      fromActionHash,
+      endActionHash,
+    }: {
+      fromActionHash?: Field;
+      endActionHash?: Field;
+    }): A[][] {
+      let actionsForAccount: A[][] = [];
+
+      Circuit.asProver(() => {
```
Trivo25, line 845: I wrapped the logic into `Circuit.asProver` because I am doing string comparisons, that didnt seem to work. Is this a bad thing to do? The logic itself doesnt need to be a part of the circuit, since its just a bit of ordering and sorting

src/lib/zkapp.ts
```diff
@@ -826,6 +833,58 @@ Use the optional \`maxTransactionsWithActions\` argument to increase this number
...
+          ? Ledger.fieldToBase58(endActionHash)
+          : undefined;
+
+        let actions;
+
+        actions = Mina.getActions(contract.address, contract.self.tokenId);
+
+        // gets the start/end indices of our array slice
+        let startIndex = start
+          ? actions.findIndex((e) => e.hash === start) + 2
```
Trivo25, line 865: the `+2` was a weird workaround.. I did that because the on-chain `actionHash` was always an action "behind", so I had to skip one

src/lib/zkapp.ts
mitschabaude: it isn't run during compile, so in compile `actionsForAccount` stays undefined

what you want to do is return `actionsForAccount` from a `Circuit.witness` block

src/lib/mina.ts
```diff
@@ -318,8 +324,37 @@ function LocalBlockchain({
...
           });
         }
-      });
 
+        // actions/sequencing events
+        let latestActionsHash =
+          ledger.getAccount(
+            PublicKey.fromBase58(addr),
+            Ledger.fieldOfBase58(tokenId)
+          )?.zkapp?.sequenceState[0] ?? Events.emptySequenceState();
```
mitschabaude, line 333: I don't think getting the latestActionsHash from the ledger here is correct in the most general case, where there are multiple parties from the same account that have sequence events, in one transaction

src/lib/mina.ts
```diff
@@ -318,8 +324,37 @@ function LocalBlockchain({
...
+
+        let actionList = p.body.sequenceEvents;
+        let eventsHash = Events.hash(
+          actionList.map((e) => e.map((f) => Field(f)))
+        );
+
+        latestActionsHash = Events.updateSequenceState(
+          latestActionsHash,
```
mitschabaude, line 341: latestActionHash is supposed to not change if the actions list is empty. here we always modify it, so this needs to be changed

src/lib/zkapp.ts
mitschabaude: let's fix computing action hashes (if actions are empty) and see if the workaround is still needed then
i would expect `+1`, because the hash stored on `actions` is the hash _after_ applying those actions, but with fromActionsHash we mean the hash _before_ the whole list of actions

src/lib/zkapp.ts
mitschabaude: hm, to return it from a witness block, we have to make it fixed size. however, `reducer.reduce` already handles dynamic size. so.. seems complicated

mitschabaude:
> Question: getActions currently is a part of the reducer. Should it stay there or become its own method on the SmartContract class, similar to fetchEvents?

should stay part of reducer, that gives us much better typing than we have with events. I think we should change events as well at some point

src/lib/zkapp.ts
mitschabaude: I thought about this again. It's simple actually! Since reduce handles creating the witnesses with the right structure, this doesn't need to deal with that. So, the asProver approach is close to what we want. In compile, the array should just be made empty. If asProver doesn't work, we can differentiate explicitly between the different snark states managed in JS:
inCompile, inAnalayze - in those the array can be empty
inProver, else case - in those the array should be filled

src/lib/mina.ts
Trivo25: good call, this also fixed the `+2` thing!

src/lib/zkapp.ts
Trivo25: I see; its now clearer how exactly asProver and witness behaves. 
Currently, it should return an empty array if asProver wont run - I tested it with the state rollup example and seems to work

mitschabaude (approved):
lgtm!

PR #329, opened 2022-08-05
expose network precondition setters to `LocalBlockchain` 

Trivo25:
## Description

Exposes various setters to individually set network preconditions. Network state is now entirely being managed on the side of SnarkyJS and can easily be customized.

## Example

```ts

  @method blockheightEquals(y: UInt32) {
    let length = this.network.blockchainLength.get();
    this.network.blockchainLength.assertEquals(length);

    length.assertEquals(y);
  }

}

// ... create a local blockchain instance

let Local = Mina.LocalBlockchain();

// ... set the network state
Local.setBlockchainLength(500);


let tx = await Mina.transaction(feePayer, () => {
  zkapp.blockheightEquals(UInt32.from(500));
});
tx.send();
```

## Under the hood

`LocalBlockchain` has been extended to manage the network state within a new variable `networkState`, instead of relying on a hard-coded value within OCaml. `LocalBlockchain` now exposes the following closures:

```ts
setTimestamp: (ms: UInt64) => void;
setGlobalSlot: (slot: UInt32) => void;
setGlobalSlotSinceHardfork: (slot: UInt32) => void;
setBlockchainLength: (height: UInt32) => void;
setTotalCurrency: (currency: UInt32) => void;
```

When a transaction is being processed, the network state is now being passed in as an additional parameter.

```ts
applyJsonTransaction(
	txJson: string,
    accountCreationFee: string,
    networkState: string
): Account[];
```

bindings not included, requires https://github.com/MinaProtocol/mina/pull/11639


src/lib/mina.ts
```diff
@@ -217,11 +222,13 @@ function LocalBlockchain({
...
   const ledger = Ledger.create([]);
 
+  let networkState = defaultNetworkState();
+
   function addAccount(pk: PublicKey, balance: string) {
     ledger.addAccount(pk, balance);
   }
 
-  let testAccounts = [];
+  let testAccounts: any[] = [];
```
MartinMinkov, line 231: Is there a way to type this? Not sure what the accounts data looks like but it would be a nice touch! Could we define a type in TS and just cast it over?

src/lib/mina.ts
```diff
@@ -328,6 +336,21 @@ function LocalBlockchain({
...
+    setGlobalSlot(slot: UInt32) {
+      networkState.globalSlotSinceGenesis = slot;
+    },
+    setGlobalSlotSinceHardfork(slot: UInt32) {
+      networkState.globalSlotSinceHardFork = slot;
+    },
+    setBlockchainLength(height: UInt32) {
+      networkState.blockchainLength = height;
+    },
+    setTotalCurrency(currency: UInt32) {
```
MartinMinkov, line 351: Just a small question, will we ever need a `UInt64` for any of these as well as `UInt32`?

MartinMinkov (approved):
LGTM! Just some small questions

src/lib/mina.ts
mitschabaude: argh.. turns out that `UInt32` extends `UInt64` (but not the other way). I wonder if this is a bug which we should fix, or if it makes sense in a weird way, because a `UInt32` also fits in 64 bits

Trivo25: Ah thats right, currency is also supposed to be UInt64. I should probably set it to UInt64 as well, since it is also defined like this in all the Precondition interfaces

mitschabaude (approved):
this is great! I appreciate the example

PR #330, opened 2022-08-08
Compressed public key

mitschabaude:
- this changes the representation that snarkyjs uses for public keys from uncompressed (x, y) to compressed (x, isOdd), to comply with the representation in the protocol and remove crazy inefficiencies in smart contracts
- closes https://github.com/o1-labs/snarkyjs/issues/298


src/lib/signature.ts
```diff
@@ -51,25 +54,56 @@ export class PrivateKey extends CircuitValue {
...
+    // compute y from elliptic curve equation y^2 = x^3 + 5
+    // TODO: we have to improve constraint efficiency by using range checks
+    let { x, isOdd } = this;
+    let ySquared = x.mul(x).mul(x).add(5);
+    let someY: Field;
+    if (ySquared.isConstant()) {
+      someY = ySquared.sqrt();
+    } else {
+      someY = Circuit.witness(Field, () => ySquared.toConstant().sqrt());
+      someY.square().equals(ySquared).or(x.equals(Field.zero)).assertTrue();
```
mrmr1993, line 72: Why do we need this `or`? Presumably this is to let us decompress the 'empty' public key, but I can't think why that would be a useful thing to do..

mitschabaude: Good catch! I needed this for some reason (can't remember which) before this PR, when decompression was done on every `party.body.publicKey`. Now I don't have any test that fails if I remove this, and can't think of a reason to allow it. So I'll remove it for now!

mrmr1993 (approved):

PR #331, opened 2022-08-08
Generic bytes & base58 encoding

mitschabaude:


mitschabaude:
this is superseeded by https://github.com/o1-labs/snarkyjs/pull/551 which also includes these changes

PR #339, opened 2022-08-11
Move tokens API to `this.experimental` for release

MartinMinkov:
In preparation for a new SnarkyJS release, we comment on the current tokens API as there is not adequate developer documentation for the feature. Make sure to uncomment out the API in a later commit that is not being used for the current release.

### DIFF
--- a/src/lib/party.ts
+++ b/src/lib/party.ts
@@ -596,75 +596,76 @@ class Party implements Types.Party {
     return clonedParty;
   }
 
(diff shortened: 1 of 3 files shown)


mitschabaude:
Update: instead of commenting it out, we decided to introduce a new `this.experimental` property where `token()` now lives

mitschabaude (approved):

PR #343, opened 2022-08-11
Add Experimental Merkle Tree Implementation

### LINKED ISSUE #342: Add Merkle Tree Implementation to SnarkyJS

ymekuria:
**Description**
resolves #342 
resolves o1-labs/proj-management#15

This PR adds a Merkle Tree implementation to SnarkyJS. This initial version was added here to support developers building more zkApp use cases in upcoming hackathons. We expect to add more features, and the API surface to change as we do so in the future. Because of this the `MerkleTree` and `MerkleWitness` classes are placed under the Experimental namespace.  

**Testing**

The MerkleTree class was tested with unit tests.


src/lib/merkle_tree.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    return hash;
+  }
+
+  calculateIndex(): Field {
+    let powerOfTwo = Field(1);
+    let index = Field(0);
+
+    for (let i = 1; i < 256; ++i) {
+      index = Circuit.if(this.isLeft[i - 1], index, index.add(powerOfTwo));
```
mitschabaude, line 109: this uses 2x the multiplication gates it has to, will fix

src/lib/merkle_tree.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  fill(leaves: Field[]) {
+    leaves.forEach((value, index) => {
+      this.setLeaf(BigInt(index), value);
+    });
+  }
+}
+
+class MerkleWitness extends CircuitValue {
+  @arrayProp(Field, 255) path: Field[];
```
Trivo25, line 82: we should probably make this a dependency of the actual merkle tree height, otherwise we will have to do a lot of hashing

mitschabaude (approved):
I'm happy now!

Trivo25 (approved):

PR #356, opened 2022-08-19
Voting Application B

Trivo25:
depends on
- [x] https://github.com/o1-labs/snarkyjs/pull/327

and fixes for 
- [x] https://github.com/o1-labs/snarkyjs/issues/355
- [ ] https://github.com/o1-labs/snarkyjs/issues/359


src/examples/zkapps/voting/demo.ts
```diff
@@ -0,0 +1,326 @@
...
+
+let contracts = await VotingApp(params);
+
+let voterStore = new OffchainStorage<Member>(8);
+let candidateStore = new OffchainStorage<Member>(8);
+let votesStore = new OffchainStorage<Member>(8);
+
+let initialRoot = voterStore.getRoot();
+try {
+  tx = await Mina.transaction(feePayer, () => {
```
bkase, line 60: isn't this too many actions to put into one txn (if this is executed on-chain)? I know it will work for the local blockchain case, but I wonder if we should keep these as consistent as we can with what we could do on-chain too

src/examples/zkapps/voting/factory.ts
```diff
@@ -0,0 +1,77 @@
+/**
+ * Produces a set of three contracts, Voting, Voter Membership and Candidate Membership SCs.
```
bkase, line 2: this is a nice abstraction 👍

src/examples/zkapps/voting/off_chain_storage.ts
```diff
@@ -0,0 +1,36 @@
+// Merkle Tree and off chain storage
```
bkase, line 1: sorry to be annoying, but can we add this in a separate PR (and include at least a few basic tests) -- maybe a test contract?

src/examples/zkapps/voting/demo.ts
Trivo25: yep, we can! This is also only the "demo" script, to demonstrate a basic run through - the tests will look better

src/examples/zkapps/voting/off_chain_storage.ts
Trivo25: It was only meant to be a small wrapper for the off-chain storage so we dont have to work with two separate APIs (map and merkle tree) - but if you want I can put that into its own PR + tests and we might even add it to the Experimental namespace?

src/lib/zkapp.ts
```diff
@@ -766,14 +773,16 @@ class ${contract.constructor.name} extends SmartContract {
...
+      /*    let methodData = (
         contract.constructor as typeof SmartContract
       ).analyzeMethods(contract.address);
       let possibleActionsPerTransaction = [
         ...new Set(Object.values(methodData).map((o) => o.sequenceEvents)).add(
           0
         ),
-      ].sort((x, y) => x - y);
+      ].sort((x, y) => x - y); */
+      let possibleActionsPerTransaction = [0, 1];
```
mitschabaude, line 785: instead of hacking it here, you could also fix the error in userland by calling `MyContract.analyzeMethods` or `MyContract.compile` before doing transactions

src/examples/zkapps/voting/off_chain_storage.ts
bkase: ah okay, yeah this is fine for now

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,166 @@ import {
...
+    // but for a voter we simply use UInt64.MAXINT() as the maximum
+    member.balance
+      .gte(participantPreconditions.minMina)
+      .and(member.balance.lte(participantPreconditions.maxMina)).assertTrue;
+
+    let accumulatedMembers = this.accumulatedMembers.get();
+    this.accumulatedMembers.assertEquals(accumulatedMembers);
+
+    // checking if the member already exists within the accumulator
+    let { state: exists } = this.reducer.reduce(
```
bkase, line 88: given that we're using a reducer to handle the case where concurrent membership registrations are occurring shouldn't we using some kind of random identifier (like a public key) rather than an incrementing index (I see this in demo.ts) for registrations? That could be a source of races.

Even if we don't execute on a change like this, we should call out in a comment why you may not want to use an incrementing integer in a case like this to help inform those trying to learn.

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,166 @@ import {
...
+      this.reducer.getActions({
+        fromActionHash: accumulatedMembers,
+      }),
+      Bool,
+      (state: Bool, _action: Member) => {
```
bkase, line 93: nit: I think we should avoid the leading underscore on parameters that are used in functions

bkase: ie `_action` -> `action`

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,166 @@ import {
...
+    let committedMembers = this.committedMembers.get();
+    this.committedMembers.assertEquals(committedMembers);
+
+    let { state: newCommittedMembers, actionsHash: newAccumulatedMembers } =
+      this.reducer.reduce(
+        this.reducer.getActions({
+          fromActionHash: accumulatedMembers,
+        }),
+        Field,
+        (state: Field, _action: Member) => {
```
bkase, line 148: same here

src/examples/zkapps/voting/voting.ts
```diff
@@ -6,41 +6,232 @@ import {
...
+    this.accumulatedVotes.assertEquals(accumulatedVotes);
+
+    let committedVotes = this.committedVotes.get();
+    this.committedVotes.assertEquals(committedVotes);
 
+    let { state: newCommittedVotes, actionsHash: newAccumulatedVotes } =
+      this.reducer.reduce(
+        this.reducer.getActions({ fromActionHash: accumulatedVotes }),
+        Field,
+        (state: Field, _action: Member) => {
```
bkase, line 213: likewise

src/examples/zkapps/voting/voting.ts
```diff
@@ -6,41 +6,232 @@ import {
...
+          // checking that the member is part of the merkle tree
+          // TODO: make work
```
bkase, line 215: reminder to sort this out

bkase (changes requested):

src/examples/zkapps/voting/membership.ts
Trivo25: do you mean this part in the demo.ts?

```ts
// creating and registering a new voter
    let m = registerMember(
      0n, // incremented int
...
```

if so - i totally agree that a public key would be a better identifier, but it would unfortunately result in a much larger merkle tree since we would have to, in theory, cover the entire key range

Trivo25: we could make the off-chain store take a public key instead of an int and transform it automatically, but that would result in the merkle witness having a length of 256 - if i am not mistaken. thoughts?

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,171 @@ import {
...
+        return action.equals(member).or(state);
+      },
+      // initial state
+      { state: Bool(false), actionsHash: accumulatedMembers }
+    );
+
+    /*
+    we cant really branch the control flow - we will always have to emit an event no matter what, 
+    so we emit an empty event if the member already exists
+    it the member doesnt exist, emit the "real" member
```
shimkiv, line 108: `doesnt` -> `doesn't`

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,171 @@ import {
...
   }
 
-  @method isMember(accountId) {
+  /**
+   * Method used to check whether a member exists within the committed storage.
+   * @param accountId
+   * @returns true if member exists
+   */
+  @method isMember(member: Member): Bool {
     // Verify membership (voter or candidate) with the accountId via merkletree committed to by the sequence events and returns a boolean
```
shimkiv, line 124: `merkletree` -> `merkle tree`

src/examples/zkapps/voting/voting.ts
```diff
@@ -6,41 +6,232 @@ import {
...
+    CandidateContract.publish();
   }
 
-  vote(candidate) {
-    // Check if Voter and Candidate exist by calling the isMember method of corresponding Smart-Contracts 
-    // Emit corresponding Sequence Event with the Vote for Candidate information
+  /**
+   * Method used to cast a vote to a specific candidate.
+   * Dispatches a new vote sequence event.
+   * @param member
```
shimkiv, line 182: `candidate` and `voter`

src/examples/zkapps/voting/demo.ts
```diff
@@ -0,0 +1,379 @@
...
+
+    contracts.voting.voterRegistration(m);
+
+    if (!params.doProofs) contracts.voting.sign(votingKey);
+  });
+  if (params.doProofs) await tx.prove();
+  tx.send();
+
+  /*
+  since the voting contact calls the voter membership contract via invoking voterRegister,
```
ymekuria, line 161: typo change to -> `since the voting contract`

src/examples/zkapps/voting/preconditions.ts
```diff
@@ -0,0 +1,29 @@
+import { UInt32, UInt64 } from 'snarkyjs';
+
+export class ElectionPreconditions {
+  startElection: UInt32;
+  endElection: UInt32;
+
+  static get default(): ElectionPreconditions {
```
ymekuria, line 7: Did we remove the default preconditions method? Is that version in a different PR?

Trivo25: Not sure, I think can't remember removing it

mitschabaude (changes requested):
Hey, `examples/zkapps/voting` need to be updated to replace `Party` with `AccountUpdate`

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,175 @@ import {
...
+   * @param member
+   */
+  @method addEntry(member: Member): Bool {
     // Emit event that indicates adding this item
     // Preconditions: Restrict who can vote or who can be a candidate
+
+    // since we need to keep this contract "generic", we always assert within a range
+    // even tho voters cant have a maximum balance, only candidates
+    // but for a voter we simply use UInt64.MAXINT() as the maximum
+    let party = Experimental.createChildAccountUpdate(
```
mitschabaude, line 80: party

src/examples/zkapps/voting/membership.ts
```diff
@@ -6,31 +6,175 @@ import {
...
+      this.self,
+      member.publicKey
+    );
+    party.account.balance.assertEquals(party.account.balance.get());
+
+    let balance = party.account.balance.get();
```
mitschabaude, line 86: can you grep for "party" in these new file an rename to accountUpdate?

mitschabaude (approved):

src/examples/zkapps/voting/membership.ts
Trivo25: ahhh yeep, good catch!

bkase (approved):

PR #360, opened 2022-08-26
Update SequenceEvents to match OCaml

psteckler:
Update `SequenceEvents` to match changes in OCaml. Those changes will be pushed, along with the bump from this PR, to the branch in https://github.com/MinaProtocol/mina/pull/11659.

Main change: there are now separate salts for the sequence events, and the sequence state.

Tested by running `test-snarkyjs-bindings.sh` locally.


mitschabaude (approved):
very nice!

MartinMinkov (approved):

PR #361, opened 2022-08-26
Voting Application B VM logic tests

ymekuria:
This PR includes a wide range of different tests for the voting application. See the testing app b doc for reference 

- minor fix for disabling proving with the `proofsEnabled: boolean` flag when creating child account updates


bkase:
I changed the base of this PR to be on top of the other one -- also it's a bit hard to review this (I know it's still in progress anyway), but maybe we can have some way to link the test cases here in the code to the ones in the doc?

bkase (commented):
Only had time for a skim so far, but looking good!

There a few kinds of tests that I think slipped through the cracks but are very important to exercise:

```
* Extend with verification key manipulation
    “Verification key”
    Transactions are still valid depending on the order of verification key updates when working with multiple parties
    Ie: Upgrade the source code of one of the “underlying voting app” and see how the failures are handled in the “wrapper voting app”
* Extend with permissions changes 
    (ie Constrain candidate registration and voter registration with a permission on “sending transactions”=Signature). If someone changes these permissions later votes can’t go through.
```

We can rephrase the first one as "what happens if we change the source code within a compound transaction"? and "what happens if we change the permissions within a compound transaction"

src/examples/zkapps/voting/test.ts
```diff
@@ -1,14 +1,25 @@
...
+import { assertValidTx, printResult, registerMember, vote } from './voting_lib';
 
 type Votes = OffchainStorage<Member>;
 type Candidates = OffchainStorage<Member>;
 type Voters = OffchainStorage<Member>;
 
+let feePayer: PrivateKey;
 /**
  * Function used to test a set of contracts and precondition
  * @param set A set of contracts
```
shimkiv, line 25: Param `set` was renamed to `contracts`.

src/examples/zkapps/voting/voting_lib.ts
```diff
@@ -0,0 +1,106 @@
...
+    console.log(cb.toString());
+    console.error('with error message: ');
+    throw Error(err);
+  } else if (failed && !expectToBeValid) {
+    if (err.message.includes(msg ?? 'NO__EXPECTED_ERROR_MESSAGE_SET')) {
+      console.log('> transaction failed, as expected!');
+    } else {
+      throw Error('transaction failed, but got a different error message!');
+    }
+  } else {
```
shimkiv, line 103: Before checking for all other cases I'd add:

```
else if (!failed && !expectToBeValid) {
    throw Error('transaction passed but should have failed');
}
```

src/examples/zkapps/voting/test.ts
```diff
@@ -31,43 +42,640 @@ export async function testSet(
...
+      let candidate = Member.from(
+        PrivateKey.random().toPublicKey(),
+        Field.zero,
+        params.candidatePreconditions.minMina.add(1)
+      );
+      voting.candidateRegistration(candidate);
+      voting.sign(votingKey);
+    },
+    feePayer,
+    'Expected'
```
shimkiv, line 677: For this and similar cases above. 
In real world I'd not only check the presence of single word in error message.
But probably `Ok` for this case.

src/examples/zkapps/voting/voting_lib.ts
```diff
@@ -0,0 +1,106 @@
+import { Member, MerkleWitness } from './member';
+import { OffchainStorage } from './off_chain_storage';
+import { Voting_ } from './voting';
+import { Mina, PrivateKey } from 'snarkyjs';
+import { Printer } from 'prettier';
+/**
+ * Updates off-chain storage when regestering a member or candidate
```
shimkiv, line 7: `regestering` -> `registering`

src/examples/zkapps/voting/voting_lib.ts
```diff
@@ -0,0 +1,106 @@
...
+ * @param {bigint} i                            index of memberStore or candidatesStore
+ * @param {OffchainStorage<Member>} votesStore  votes off-chain storage
```
shimkiv, line 9: - Not all params are documented;
- There is no such param as `votesStore`.

src/examples/zkapps/voting/voting_lib.ts
```diff
@@ -0,0 +1,106 @@
...
+  m.witness = new MerkleWitness(store.getWitness(i));
+
+  return m;
+}
+
+/**
+ * Updates off-chain storage after voting
+ * @param {bigint} i                            index of candidateStore and votesStore
+ * @param {OffchainStorage<Member>} votesStore  votes off-chain storage
+ * @param {OffchainStorage<Member>} votesStore  candidates off-chain storage
```
shimkiv, line 32: Duplicated `votesStore`

src/examples/zkapps/voting/deployContracts.ts
```diff
@@ -0,0 +1,63 @@
+import { Experimental, Field, Mina, Party, PrivateKey } from 'snarkyjs';
+import { VotingAppParams } from './factory';
+
+import { Membership_ } from './membership';
+
+import { Voting_ } from './voting';
+
+/**
+ * Function used to deploy a set of contracts for a given set of preconditions
```
shimkiv, line 9: Not all params are documented.

src/examples/zkapps/voting/test.ts
```diff
@@ -31,43 +42,640 @@ export async function testSet(
...
+      - no state change
+  */
+  console.log('attempting to count votes but no votes were casted...');
+
+  let beforeAccumulator = voting.accumulatedVotes.get();
+  let beforeCommitted = voting.committedVotes.get();
+  await assertValidTx(
+    true,
+    () => {
+      voting.countVotes();
```
shimkiv, line 461: It will be good to print results as part of the `countVotes` method invocation rather then as part of another helper method `printResult`. Wdyt?

src/examples/zkapps/voting/test.ts
```diff
@@ -31,43 +42,640 @@ export async function testSet(
...
+    throw Error(
+      'votesStore merkle root does not match on-chain committed votes'
     );
-    voting.accumulatedVotes.set(Experimental.Reducer.initialActionsHash);
-  });
-  tx.send();
+  }
+
+  console.log('election is over, printing results');
+  printResult(voting, votesStore);
```
shimkiv, line 626: We also should check the correctness of counting itself I believe.

src/examples/zkapps/voting/test.ts
Trivo25: unfortunately not, that would be a pretty hacky way since we dont have access to the actual vote-data inside the contract, only a reference to it (the committed storage root) :/

src/examples/zkapps/voting/test.ts
```diff
@@ -28,46 +41,769 @@ export async function testSet(
...
     });
 
-    voting.deploy({ zkappKey: params.votingKey });
-    candidateContract.deploy({ zkappKey: params.candidateKey });
-    voterContract.deploy({ zkappKey: params.voterKey });
+    tx.send();
+  } catch (err: any) {
+    if (!err.toString().includes('precondition_unsatisfied')) {
+      throw Error('Transaction should have failed but went through!');
+    }
```
ymekuria, line 95: We should also handle the cases when the transaction fails for any other reason than the expected failure case of 'precondition_unsatisfied'.

    if (!err.toString().includes('precondition_unsatisfied')) {
      throw Error('Transaction should have failed but went through!');
    } else {
      // Throw an error on all unexpected failure cases
      throw Error(error)
    }

src/examples/zkapps/voting/test.ts
```diff
@@ -28,46 +41,769 @@ export async function testSet(
...
+  try {
+    let tx = await Mina.transaction(sequenceOverflowSet.feePayer, () => {
+      sequenceOverflowSet.voting.authorizeRegistrations();
+      sequenceOverflowSet.voting.sign(votingKey);
+    });
+    tx.send();
+  } catch (err: any) {
+    if (!err.toString().includes('the maximum number of lists of actions')) {
+      throw Error('Transaction should have failed but went through!');
+    }
```
ymekuria, line 167: Same comment as above about handling the cases when the transaction fails for any other reason than the expected failure case. 

    ```
if (!err.toString().includes('the maximum number of lists of actions')) {
      throw Error('Transaction should have failed but went through!');
    } else {
      // Throw an error on all unexpected failure cases
      throw Error(error)
  }
```

File: src/examples/zkapps/voting/test.ts (hunk 10 above)
Comment (Trivo25) [reply src/examples/zkapps/voting/test.ts:95]: good point!

File: src/examples/zkapps/voting/test.ts [hunk 12]
```diff
@@ -1,17 +1,36 @@
-import { Experimental, Mina, AccountUpdate } from 'snarkyjs';
+import {
```
mitschabaude, line 1: Should this test script be run as part of `./run-integration-tests.sh`, so that voting tested in CI?
(And does this mean the `voting/run.ts` script doesn't need to run in CI anymore? Sorry I haven't really looked at either of these yet)

Trivo25: `run.ts` sets the preconditions and parameters for the tests we want to run and then calls into `test.ts` to execute all the tests, so it's being run although maybe a bit confusing :D

bkase (changes requested):

bkase (approved):
Woo 🥳

PR #366, opened 2022-08-29
Make snarkyjs an ES module

### LINKED ISSUE #364: Fix Jest tests in CI

mitschabaude:
Switches snarkyjs to a `"type": "module"` package.json.
The main resulting change is that TS compiles to ESM format.

Impact: 
* Simpler node build, basically just `npx tsc` without additional esbuild compilation
* Everything is more similar to zkapp-cli projects. For example, internal imports in TS now use the `"<relatve-path>.js"` syntax
* Better integration with node which uses ESM (import / export) syntax. For example, we no longer have to use `.mjs` files. Also, nodejs unit tests (and other scripts) written in TS can now be compiled with the normal build pipeline to JS and just run with node.
* Prepares for using the native node test runner in place of jest

Apart from the above, this
* makes Jest test run in sequence (to make sure they are working)
* Fixes #364 
* changes the dist folder `dist/server` to `dist/node`
* fixes an unrelated issue which broke the web version

TODOs:
* [x] fix up web build
* [x] fix up CJS export (or decide to stop supporting CJS)
* [x] fix up Jest test running or move to `node:test` + `expect` directly


MartinMinkov (approved):
LGTM!

bkase (approved):
Ship it 🚢

Trivo25 (approved):

PR #367, opened 2022-08-29
Unit tests using node:test

mitschabaude:
ports unit tests to native node tests

TODOs:
* [ ] with `node --test` not working, the script's not failing on failing tests. but we need this for CI
* [ ] fix token tests


mitschabaude:
Closing this, as we decided against a full switch of the unit test framework for now. But the techniques here might be useful for selected unit tests, e.g. to remove snarkyjs exports only needed (these don't even have to use the `node:test` library -- the main point is about running tests via node instead of Jest)

PR #374, opened 2022-08-30
Fix simple zkapp example

### LINKED ISSUE #373: Fix simple zkapp example

mitschabaude:
- Fixes #373 
- Fixes events hashing, which caused the issue above
- Fixes prettier config
- Makes simple zkApp example nicer, both by using newer APIs and by embellishing the `this.send` API
- `this.send` optionally can take a `Party` instead of a `PublicKey` for the receiver account
- Inline the input types for send (mint, burn) APIs, to give a better UX when using these functions
- Add a way to run plain node scripts as unit tests with access to internal APIs
- Add a unit test which asserts that we are calculating hash inputs consistently
- Run unit tests in CI


MartinMinkov (approved):
Great changes! Thanks! :D

bkase (approved):

PR #376, opened 2022-09-01
DEX v0 implementation

### LINKED ISSUE #375: DEX zkApp, initial implementation
doesn't include adding test cases beyond "it runs the happy path"

does include:
* [x] witnessing arbitrary parties layouts

mitschabaude:
- Closes #375 
- ~~Note: tests don't currently pass but will with updated bindings (https://github.com/MinaProtocol/mina/pull/11791)~~
- Adds initial DEX contract
- Adds `Party.witnessTree` which allows you to witness a layout of child parties
- Other changes are called out in my review comments below

Design / parties layout of the swap method:

swap-parties


src/lib/party.ts
```diff
@@ -819,57 +823,48 @@ class Party implements Types.Party {
...
-    fallbackToZeroNonce = false
+    privateKey?: PrivateKey
   ) {
-    feePayer.body.nonce = this.getNonce(feePayer, fallbackToZeroNonce);
+    feePayer.body.nonce = this.getNonce(feePayer);
     feePayer.authorization = Ledger.dummySignature();
     feePayer.lazyAuthorization = { kind: 'lazy-signature', privateKey };
   }
 
-  // TODO this needs to be more intelligent about previous nonces in the transaction, similar to Party.createSigned
```
mitschabaude, line 844: the changes here and in `Party.createSigned` unify the logic of computing the right nonce, and make it correct in more cases (e.g., when child parties exist)

src/lib/party.ts
```diff
@@ -927,42 +932,29 @@ class Party implements Types.Party {
     return { body, authorization: Ledger.dummySignature() };
   }
 
-  static createUnsigned(publicKey: PublicKey) {
-    let party = Party.defaultParty(publicKey);
-    Mina.currentTransaction()?.parties.push(party);
+  static create(publicKey: PublicKey, tokenId?: Field) {
+    let party = Party.defaultParty(publicKey, tokenId);
+    if (smartContractContext.has()) {
+      makeChildParty(smartContractContext.get().this.self, party);
```
mitschabaude, line 938: apart from changing the (IMO, confusing) `createUnsigned` name to `create`, I also now automatically create a child party (instead of a top level party) when inside a zkapp. This is because typically when a zkapp creates a party, it wants to make statements about it. If it isn't a child party, those statements would not go in the proof, which I think is a huge security footgun

src/lib/party.ts
```diff
@@ -1025,21 +1017,85 @@ class Party implements Types.Party {
...
+    party.children = proverParty?.children ?? { parties: [] };
     party.parent = proverParty?.parent;
     return { party, result: fieldsAndResult.result };
   }
+
+  /**
+   * Like Party.witness, but lets you specify a layout for the party's children,
+   * which also get witnessed
+   */
+  static witnessTree<T>(
```
mitschabaude, line 1029: This function is one of the main additions in this PR! It allows you to witness not just one party, but an entire tree with a specific layout

src/lib/party.ts
```diff
@@ -1052,42 +1108,44 @@ const CallForest = {
...
 
 function createChildParty(
   parent: Party,
   childAddress: PublicKey,
-  options?: {
-    caller?: Field;
-    tokenId?: Field;
-    useFullCommitment?: Bool;
-  }
+  tokenId?: Field
```
mitschabaude, line 1138: simplified the options here to make it have the same `publicKey` + `tokenId` interface as almost all other methods that create a party

src/lib/party.ts
```diff
@@ -1177,7 +1235,7 @@ function addMissingSignatures(
     let { privateKey } = party.lazyAuthorization;
     if (privateKey === undefined) {
       let i = additionalPublicKeys.findIndex((pk) =>
-        pk.equals(party.body.publicKey)
+        pk.equals(party.body.publicKey).toBoolean()
```
mitschabaude, line 1238: this was wrong all the time 🙄 🤦🏻

src/lib/zkapp.ts
```diff
@@ -206,13 +208,16 @@ function wrapMethod(
             assertStatePrecondition(this);
             return result;
           } else {
-            // in a transaction, also add a lazy proof to the self party
-            // (if there's no other authorization set)
+            // called smart contract at the top level, in a transaction!
+            // => attach ours to the current list of parties
+            let party = context.selfUpdate;
+            if (!context.isCallback) {
+              Mina.currentTransaction()?.parties.push(party);
```
mitschabaude, line 215: important change! the `this.self` party isn't attached to the list of parties automatically anymore. so we have to do it explicitly here. (it was just annoying that parties that were intended to be child parties popped up on the top level as well)

src/lib/zkapp.ts
```diff
@@ -583,48 +595,59 @@ class SmartContract {
...
 
-  sign(zkappKey?: PrivateKey, fallbackToZeroNonce?: boolean) {
-    this.self.signInPlace(zkappKey, fallbackToZeroNonce);
+  sign(zkappKey?: PrivateKey) {
+    this.self.sign(zkappKey);
   }
 
-  private executionState(): ExecutionState {
-    if (!Mina.currentTransaction.has()) {
+  private executionState(): Party {
```
mitschabaude, line 613: this was changed to use the party from `smartContractContext` if we're inside a smart contract, so it's less confusing where the `this.self` party originates (previously, it was the place where `this.self` was called first)

we also don't auto-attach the party to the transaction anymore in that case

src/lib/zkapp.ts
```diff
@@ -1031,6 +1052,14 @@ async function deploy<S extends typeof SmartContract>(
   return tx.sign().toJSON();
 }
 
+function Account(address: PublicKey, tokenId?: Field) {
```
mitschabaude, line 1055: convenient helper (suggested by @Trivo25) to inspect another account in a zkApp. somewhat dangerous because it adds an extra party which you can't easily reuse for something else

run-integration-tests.sh
```diff
@@ -1,6 +1,6 @@
-./run src/examples/zkapps/hello_world/run.ts
-./run src/examples/zkapps/voting/run.ts
-./run src/examples/simple_zkapp.ts
-./run src/examples/zkapps/reducer/reducer_composite.ts
-./run src/examples/zkapps/composability.ts
-./run src/examples/zkapps/token_with_proofs.ts
+./run src/examples/zkapps/hello_world/run.ts || exit 1
```
mitschabaude, line 1: change to make sure that the intg tests really made CI fail if one of them failed

src/examples/zkapps/dex/dex.ts
```diff
@@ -0,0 +1,382 @@
+import {
```
mitschabaude, line 1: the DEX contracts!

src/examples/zkapps/dex/erc20.ts
```diff
@@ -0,0 +1,211 @@
+import {
```
mitschabaude, line 1: draft of an ERC20-like token interface, not used yet but maybe will use later.
not everything ERC20 has is easily possible on Mina though (or possible at all, not sure)

mitschabaude (commented):
some comments to help you review this!

src/lib/mina.ts
```diff
@@ -62,7 +62,7 @@ type Account = Fetch.Account;
 
 type FetchMode = 'fetch' | 'cached' | 'test';
 type CurrentTransaction = {
-  sender?: PrivateKey;
+  sender?: PublicKey;
```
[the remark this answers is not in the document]
MartinMinkov: Sounds like a good change to me!

src/lib/mina.ts
```diff
@@ -301,12 +301,21 @@ function LocalBlockchain({
...
-      );
+      try {
+        ledger.applyJsonTransaction(
+          JSON.stringify(partiesJson),
+          String(accountCreationFee),
+          JSON.stringify(networkState)
+        );
+      } catch (err: any) {
+        try {
+          // reverse errors so they match order of account updates
```
MartinMinkov, line 312: I've been thinking that instead of just throwing error messages, we could catch them and match them up with the string descriptions that live here: https://github.com/MinaProtocol/mina/blob/91acb8f454111bc65ec85af9888c73217951efd2/src/lib/mina_base/transaction_status.ml#L324

Would this try block be the place we would do it? If so, what's the best way of generating these strings without manually copying them so we can instead throw the messages instead? Would we just expose these in the SnarkyJS bindings and have a dictionary we could work with?

src/lib/zkapp.ts
MartinMinkov: Why is `this.self` not being added anymore?

src/lib/zkapp.ts
MartinMinkov: > Why is this.self not being added anymore?

I guess cause it's happening here?

MartinMinkov (approved):
LGTM, just some questions! :thinking:

src/examples/zkapps/token_with_proofs.ts
```diff
@@ -55,9 +58,13 @@ class TokenContract extends SmartContract {
   @method sendTokens(
     senderAddress: PublicKey,
     receiverAddress: PublicKey,
-    callback: Experimental.Callback
+    callback: Experimental.Callback<any>
   ) {
-    let senderParty = Experimental.partyFromCallback(this, callback, true);
+    let senderParty = Experimental.partyFromCallback(
+      this,
+      [undefined],
```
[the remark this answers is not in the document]
mitschabaude: not a dummy party, but one that can have any children. so we're not constraining what children the party has (we're just witnessing the children's hash as a Field, instead of computing it in the circuit)

src/lib/mina.ts
mitschabaude: yeah it seems we could just import them to `snarky_js_bindings_lib.ml` and replace the statuses with strings there -- that should be fairly easy!

run-integration-tests.sh
ymekuria: This is great!

PR #377, opened 2022-09-01
Make SnarkyJS run inside a web worker

mitschabaude:


maht0rz:
I have tested this fix manually in Chrome v104 with `new Worker(...)`, it no longer crashes when calling `await isReady` inside a web worker. I have not been able to test any further, since i am running into an infinite loop while compiling in a browser (while it works in the CLI).

bkase (approved):

PR #394, opened 2022-09-08
Add example doc comments for Merkle Tree

### LINKED ISSUE #388: template doc comments for merkle tree

Trivo25:
adds example doc comments for the Merkle Tree implementation

closes #388

### DIFF
--- a/src/lib/merkle_tree.ts
+++ b/src/lib/merkle_tree.ts
@@ -28,8 +28,8 @@ class MerkleTree {
 
   /**
    * Creates a new, empty [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree).
-   * @param {number} height The height of Merkle Tree.
-   * @returns {MerkleTree} A new MerkleTree
+   * @param height The height of Merkle Tree.
+   * @returns A new MerkleTree
    */
   constructor(public readonly height: number) {
     this.zeroes = [Field(0)];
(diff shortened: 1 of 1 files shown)


src/lib/merkle_tree.ts
```diff
@@ -1,3 +1,8 @@
+/**
+ * This file contains all code related to the [Merkle Tree](https://en.wikipedia.org/wiki/Merkle_tree) implementation available in SnarkyJS.
+ * @author TODO: should we add authors? Some Name <name@o1labs.org>
```
Trivo25, line 3: we can probably ignore a dedicated list of author, or wdyt?

MartinMinkov: For most core functionality, I would agree that we could probably ignore it but I like having authors for data structures that are added on top. Things like this and `CircuitString` can add authors and give credit on the people implementing these! Just my thoughts though :D

ymekuria: I think we can leave this out.

ymekuria (approved):

ymekuria (approved):

src/lib/merkle_tree.ts
```diff
@@ -8,23 +13,45 @@ export { Witness, MerkleTree, MerkleWitness, BaseMerkleWitness };
...
     for (let i = 1; i < height; i++) {
       this.zeroes.push(Poseidon.hash([this.zeroes[i - 1], this.zeroes[i - 1]]));
     }
   }
 
+  /**
+   * Returns a node which lives at a given index and level.
+   * @param {number} level Level of the node.
+   * @param {bigint} index Index of the node.
+   * @returns {Field} The data of the node.
```
MartinMinkov, line 45: I personally don't like adding types in documentation since TS can infer these types. The comments could drift and disagree with what TS is saying and can be confusing.

MartinMinkov (approved):
Looks great to me! I would prefer to remove types but interested to here what others think.

src/lib/merkle_tree.ts
Trivo25: Good point! didn't think of it that way - its probably better to just leave them out then

PR #406, opened 2022-09-09
Make address a variable

mitschabaude:
- made address & tokenId witnesses
- fix .get() functions to extract prover value of address / tokenId in a witness block
- fix party.witness -- make return values constant

### DIFF
--- a/src/lib/zkapp.ts
+++ b/src/lib/zkapp.ts
@@ -13,6 +13,7 @@ import {
   circuitValue,
   cloneCircuitValue,
   getBlindingValue,
+  witness,
   memoizationContext,
   toConstant,
 } from './circuit_value.js';
@@ -136,6 +137,9 @@ function wrapMethod(
 ) {
   return function wrappedMethod(this: SmartContract, ...actualArgs: any[]) {
     cleanStatePrecondition(this);
+    // witness address and tokenId so the circuit doesn't depend on them
+    let address = witness<PublicKey>(PublicKey, () => (this as any)._address);
+    let tokenId = witness<Field>(Field, () => (this as any)._tokenId);
     // TODO: the callback case is actually more similar to the composability case below,
     // should reconcile with that to get the same callData hashing
     if (!smartContractContext.has() || smartContractContext()?.isCallback) {
@@ -144,7 +148,7 @@ function wrapMethod(
           this: this,
           methodCallDepth: 0,
           isCallback: false,
-          selfUpdate: selfParty(this.address, this.nativeToken),
+          selfUpdate: selfParty(address, tokenId),
         },
         (context) => {
           if (inCheckedComputation() && !context.isCallback) {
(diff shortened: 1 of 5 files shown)


src/examples/circuit_string.ts
```diff
@@ -23,7 +23,7 @@ await isReady;
 let address = PrivateKey.random().toPublicKey();
 
 console.log('compile...');
-await MyContract.compile(address);
```
MartinMinkov, line 26: Great change :D

src/lib/errors.ts
```diff
@@ -0,0 +1,51 @@
...
+export { invalidTransactionError };
+
+function invalidTransactionError(
+  transaction: Types.ZkappCommand,
+  errors: string[][][],
+  { accountCreationFee }: { accountCreationFee: string | number }
+) {
+  let lastUpdate = errors[errors.length - 1];
+
+  if (lastUpdate.some(([msg]) => msg === 'Invalid_fee_excess')) {
```
MartinMinkov, line 14: Minor nit but can we refactor `invalidTransactionError` a bit so that `Invalid_fee_excess` is less coupled from the function, so it's easier to add more errors later down the road?

src/lib/zkapp.ts
```diff
@@ -620,16 +640,16 @@ class SmartContract {
     let inTransaction = Mina.currentTransaction.has();
     let inSmartContract = smartContractContext.has();
     if (!inTransaction && !inSmartContract) {
-      // TODO: it's inefficient to return a fresh account update everytime, would be better to return a constant "non-writable" account update,
-      // or even expose the .get() methods independently of any account update (they don't need one)
-      return selfAccountUpdate(this.address, this.nativeToken);
+      // TODO: it's inefficient to return a fresh party everytime, would be better to return a constant "non-writable" party,
+      // or even expose the .get() methods independently of any party (they don't need one)
```
MartinMinkov, line 644: I think we accidentally renamed `account update` to `party` here

src/lib/zkapp.ts
```diff
@@ -555,24 +571,28 @@ class SmartContract {
     });
   }
 
-  static async compile(address: PublicKey, tokenId: Field = TokenId.default) {
+  static async compile() {
```
MartinMinkov, line 574: While we're changing `compile` and `analyzeMethods` in this PR, could you add doc comments as well if you don't mind? :D

MartinMinkov (approved):
Awesome changes! :1st_place_medal:

src/lib/zkapp.ts
mitschabaude: nice catch!

src/lib/zkapp.ts
mitschabaude: sloppy merge conflict handling, ty!

Issue #415, opened 2022-09-14
Replace Safari-breaking RegExp with function


mitschabaude:
It's in `removeJsonQuotes`: https://github.com/o1-labs/snarkyjs/blob/235dc35450132090414592225c0522874fe266d0/src/lib/fetch.ts#L577-L583

The task is to get rid of the ReqExp in there, or replace it with one that works in Safari.

Requirements:
Should remove the quotes from the keys of an arbitrary JSON object (but not quotes in string values)
Should handle any escaped quotes in string values. Can assume keys don't have escaped quotes.

mitschabaude:
it's in `removeJsonQuotes`

shimkiv:
<img width="2048" alt="image" src="https://user-images.githubusercontent.com/4096154/190239590-f51f1387-2c7b-4ba1-91c4-a9ead12a71b0.png">

PR #419, opened 2022-09-17
Simplified regex in removeJsonQuotes

jdsteinhauser:
Resolves Issue #415. I'm not satisfied with how I had to test it, since Jest wasn't happy with any way I tried to import `src/lib/fetch.[jt]s` . If an RFC is needed, I'll gladly write one but I was assuming this was small and straightforward enough to not require one.

- cleaned up the tests
- added some tests
- made sure it works with safari


shimkiv:
Seems to work because now it reveals [another issue](https://stackoverflow.com/a/65099653) (no stacktrace available):

image

src/lib/fetch.test.ts
```diff
@@ -0,0 +1,88 @@
...
+  let cleaned = JSON.stringify(JSON.parse(json), null, 2);
+  return cleaned.replace(/\"(\S+)\"\s*:/gm, "$1 :");
+}
+
+// function removeJsonQuotes(json: string) {
+//   let cleaned = JSON.stringify(JSON.parse(json), null, 2);
+//   return cleaned.replace(/\"([\w-]+)\"\s*:/gm, "$1 :");
+// }
+
+describe('fetch helper functions', () => {
```
mitschabaude, line 16: can we also test nested objects + arrays, e.g.
```json
{
  "array": [
    {
      "data": "string with \"escaped quotes\": 1"
    }
  ]
}
```

mitschabaude (commented):
Very nice! One more test and I'm happy :D

The web workers thing is a bummer, but we'll manage to work around that as well

mitschabaude:
@jdsteinhauser re: testing. Our Jest setup currently doesn't manage to build the whole project, so right now with Jest we can only test stuff that is exported from snarkyjs (all the other Jest test import `"snarkyjs"` which resolves to the build output in `/dist/node`).

We recently introduced a way to test internal stuff as well -- you can see it demonstrated in `src/lib/hash-inputs.unit-test.ts`.
-> Just make a new file with the ending `unit-test.ts`, and it will automatically be built and run as part of `npm run test:unit`.
You can't use `describe` and `it` in these files, but you _can_ use `expect` (which is just a utility function that throws an error if a test fails), so you get the same testing capabilities, just with less fancy console output

bkase:
@jsteinhauser could you rebase the PR off of `main` instead?

Trivo25:
- cleaned up the tests
- added some tests
- made sure it works with safari

mitschabaude (approved):
awesome!

jdsteinhauser:
Oh my goodness, thanks for finishing this one out @Trivo25 !

Trivo25:
> Oh my goodness, thanks for finishing this one out @Trivo25 !

Sorry it took so long :D

PR #423, opened 2022-09-18
verify transactions locally

Trivo25:
Recovers a verify function from the verificationKey that lives in the zkApp account and implements functionality to check AccountUpdates based on a set of Permissions that are associated with that account. Before, proofs were not verified at all.

Because checking now happens in JS, we now get (imo) nicer error messages as well

so instead of `Error: [[],[["Update_not_permitted_app_state"]]]` we get `Error: Transaction verification failed: Cannot update field 
'appState' because permission for this field is 'Proof', but the required authorization was not provided or is invalid.`

Also makes `tx.send()` async, which shouldn't cause too many issues and is probably a logical step if we consider that 'sending' something is supposed to be async anyway. 

For the remote blockchain, sending the request to the graphql endpoint now happens in `send` directly, instead of `wait` which is more logical IMO. `wait` still exists and could be used to 'wait for inclusion in a block' if we decide so - I wouldn't add the 'waiting for block' directly to `send` because it could take a couple of minutes which would simply be unfeasible

`hash()` is not async anymore, because the request is now being resolved directly in `send`- `hash` just returns the tx hash string

closes https://github.com/o1-labs/snarkyjs/issues/150


src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+      case 'permissions':
+        return perm.setPermissions;
+      case 'zkappUri':
+        return perm.setZkappUri;
+      case 'tokenSymbol':
+        return perm.setTokenSymbol;
+      case 'timing':
+        return 'None'; // TODO: do timing?
+      case 'votingFor':
+        return perm.setVotingFor; // TODO: do timing?
```
bkase, line 848: comment?

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+  // we are essentially only checking if the update is empty or an actual update
+  function includesChange(val: any): boolean {
+    if (val?.constructor === Array) {
+      return !(val as Array<any>).every((v) => v == null);
+    } else {
+      return val != null;
+    }
+  }
+
+  function permissionForUpdate(key: string): string {
```
bkase, line 831: can we type this more precisely than `string` so we can catch if we ever add any new keys here? @mitschabaude did something fancy with the account-update type to do this in a different part of snarkyjs

If that's too complex or intractable, can we at least `throw` in the default branch so we can notice (at least when we run it once) that we need to revisit this function?

src/lib/mina.ts
```diff
@@ -307,6 +316,29 @@ function LocalBlockchain({
...
+            party.body.tokenId
+          );
+          if (account) {
+            await verifyAccountUpdate(
+              account!,
+              party,
+              commitments,
+              proofsEnabled
+            );
+          }
```
bkase, line 336: Should we somehow merge these errors with the ones in this next `try/catch/finally` block rather than skipping them (if `account` is falsey) or console.log-ing (if the error is caught)

bkase (changes requested):

src/lib/mina.ts
```diff
@@ -307,6 +316,29 @@ function LocalBlockchain({
     sendTransaction(txn: Transaction) {
       txn.sign();
 
+      let commitments = Ledger.transactionCommitments(
+        JSON.stringify(zkappCommandToJson(txn.transaction))
+      );
+
+      txn.transaction.accountUpdates.forEach(async (party) => {
```
mitschabaude, line 323: We should await these promises at the level of `sendTransaction`. Right now, they're escaping the synchronous context this runs in. This requires refactoring of `sendTransaction` to be async -- which is a breaking change in the API, but fine IMO, "sending" is something inherently async, this won't complicate any existing flows of using snarkyjs

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+    const proof = SmartContract.Proof().fromJSON({
+      maxProofsVerified: 2, // TODO: this can probably vary?
+      proof: accountUpdate.authorization.proof!,
+      publicInput: publicInputFields.map((f) => f.toString()),
+    });
+
+    try {
+      let verificationKey = account.zkapp?.verificationKey!;
+      isValidProof = await verify(proof, verificationKey);
+    } catch (error) {
```
mitschabaude, line 883: I'm generally against silencing errors like this. If at some point, a proof breaks `verify` for some reason, I want it to throw the actual error with the actual stack trace. You could accomplish that and still display the nice error message below by doing something like
```ts
err.message = additionalMessage + (err.message ? '\n\n' + err.message : '')
```

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+      : commitment;
+
+    // checking permissions and authorization for each party individually
+    try {
+      isValidSignature = Ledger.checkAccountUpdateSignature(
+        JSON.stringify(accountUpdate.toJSON()),
+        txC
+      );
+    } catch (error) {
+      isValidSignature = false;
```
mitschabaude, line 900: same here about swallowing the error

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+    }
+  }
+
+  function checkPermission(p: string, field: string) {
```
mitschabaude, line 904: `p: Types.Json.AuthRequired`

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+
+  // we don't check if proofs aren't enabled
+  if (!proofsEnabled) isValidProof = true;
+
+  if (accountUpdate.authorization.proof && proofsEnabled) {
+    let publicInput = accountUpdateToPublicInput(accountUpdate);
+    let publicInputFields = ZkappPublicInput.toFields(publicInput);
+
+    const proof = SmartContract.Proof().fromJSON({
+      maxProofsVerified: 2, // TODO: this can probably vary?
```
mitschabaude, line 875: comment: the `maxProofsVerified` given here currently aren't used by `verify`, that's why this doesn't matter. `verify` normalizes whatever input proof it gets to have `maxProofsVerified=2` -- the Mina node does the same

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+
+    if (!verified) {
+      throw Error(
+        `Transaction verification failed: Cannot update field '${field}' because permission for this field is '${p}', but the required authorization was not provided or is invalid.`
+      );
+    }
+  }
+
+  // goes through the update field on a transaction
+  Object.entries(update).forEach(async ([key, value]) => {
```
mitschabaude, line 930: this shouldn't be async right?

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+  },
+  proofsEnabled: boolean
+): Promise<void> {
+  let perm = account.permissions;
+
+  let { commitment, fullCommitment } = transactionCommitments;
+
+  // we are essentially only checking if the update is empty or an actual update
+  function includesChange(val: any): boolean {
+    if (val?.constructor === Array) {
```
mitschabaude, line 824: I think if you use `Array.isArray(val)` as a test, TS knows that it's an array afterwards

src/lib/mina.ts
```diff
@@ -773,3 +805,143 @@ function defaultNetworkState(): NetworkValue {
...
+      return !(val as Array<any>).every((v) => v == null);
+    } else {
+      return val != null;
```
mitschabaude, line 827: it's a tiny nit, but can we in general be explicit about the distinction between `null` and `undefined`? The JSON types all have `null` for indicating a missing field:
https://github.com/o1-labs/snarkyjs/blob/f1a6b131d6bcc66003fa640e10e9e9ddfa3145dd/src/snarky/gen/transaction-json.ts#L32
(maybe more general nit: can we use `===` instead of `==` always if we don't _specifically_ want to handle the cases where `==` does casting?)

mitschabaude (changes requested):
Left some comments!

src/lib/mina.ts
mitschabaude: Is there a reason to not just throw the error here? Do we want to make sure other problems with the transaction are logged as well? I'm OK if we don't

src/lib/mina.ts
Trivo25: ah ye lol
VSCode auto complete does some weird things at times

src/lib/mina.ts
Trivo25: after thinking about it, I would agree its probably better to just throw right away

mitschabaude (approved):
Awesome work!!

bkase (approved):

PR #428, opened 2022-09-20
Callbacks via composability

### LINKED ISSUE #402: Dex: Internals continued
Dependencies tracked in other issues
* [x] #136 for reusing a smart contract for two different accounts
* [x] #363 for passing in a verification key to a method; and maybe for passing an account update without introducing another special case for method arguments
* [ ] #431 -- account update structure not working as expected
* [ ] #429 -- return values to transaction block

This issue
* [x] properly prove that we're executing a callback with certain arguments, in the zkApp A (see definitions below). this should be doable by directly using composability for A calling C, getting the party of C and wiring it up to be a child of B. 
```
A - creates callback, passes it to B
B - gets callback, witnesses a superset of parties, does generic check on them
C - is called by the callback
```

mitschabaude:
* Closes #402
* Enables to pass AccountUpdates to methods (by making the class implement `Provable<AccountUpdate>`)
* Add an experimental  `authorize` method, which can (should) be used on AccountUpdates that are passed to methods:
```ts
@method authorizingMethod(accountUpdate: AccountUpdate) {
  this.experimental.authorize(accountUpdate);
}
```
In this case, `authorize` simply takes the AccountUpdate and makes it a child of the zkApp ("adopts it").

There's also an advanced second parameter to witness a certain layout of child updates:
```ts
this.experimental.authorize(accountUpdate, [0, 1, [null]]);
```
The weird layout syntax is documented here:
https://github.com/o1-labs/snarkyjs/blob/2bf421ba36780c4dc644552e02a6d2dcfe51036c/src/lib/account_update.ts#L1211-L1230

`authorize` can be used to achieve what the callback was all about: instead of taking a "zkApp callback" and extracting the account update in the authorizing method, we let it be up to the method's caller to get the account update, and pass it in directly.
The caller could, for example, do this
* call a third zkApp
* take the account update off of `thirdZkapp.self`
* pass that account update to a method like `authorizingMethod` above, which uses `authorize` (and then does checks on the witnessed account update tree, like summing up the balances)

If this is done, the functionality is equivalent to a callback, with the difference that the caller can also see the result of the callback! (which, it turns out, is what we often want)

Let's call this flow "manual callback".

The "manual callback" flow is safer than the old callback, because our zkApps composability machinery ensures that we properly prove in the caller's circuit that `thirdZkapp` was really called, with certain arguments and outputs.

This PR also:
* Refactors the DEX to use "manual callback" instead of callbacks
* Refactor callbacks to use the "manual callback" flow under the hood, to be secure
* Adds helpers for pretty-printing a transaction or account update, with `.toPretty()`
* Unrelatedly, makes some tweaks to the `./run` script
* Unrelatedly, tweaks how `Circuit.witness` works


src/examples/zkapps/dex/dex.ts
```diff
@@ -338,14 +267,12 @@ class TokenContract extends SmartContract {
...
-  @method authorize(callback: Experimental.Callback<any>) {
-    let layout = [[3, 0, 0], 0, 0]; // these are 10 child account updates we allow, in a left-biased tree of width 3
-    // TODO: this should also return what the callback returns, and authorize should pass it on!
-    let zkappUpdate = Experimental.accountUpdateFromCallback(
-      this,
-      layout,
-      callback
-    );
+  @method authorize(zkappUpdate: AccountUpdate) {
+    let layout = [[[3, 0, 0], 0, 0]]; // these are 10 child account updates we allow, in a left-biased tree of width 3
```
MartinMinkov, line 271: This part of the API feels really low level to me, what are your thoughts on creating an abstracting around `layout` in a later PR?

I'm not sure how exactly this would look like but some of the cases documented are duplicates of each other in terms of functionality, it would be great to have type/types to capture the intent.

For example:
Perhaps we can have a type called `EmptyAccountUpdate`.

```
* []         // an empty accountUpdates list -- EmptyAccountUpdate
* 0          // same as [] -- EmptyAccountUpdate
```
```
 * [0]        // a list of one accountUpdate, which doesn't have children 
 * 1          // same as [0] 
 * 2          // same as [0, 0] 
```
In these cases, we can have `ChildlessAccountUpdate` and encapsulate all the cases by wrapping them with indexes in a list, e.g. 

 * [0]        `[ChildlessAccountUpdate]`
 * 1          // same as [0] - `[ChildlessAccountUpdate]`
 * 2          // same as [0, 0] - `[ChildlessAccountUpdate, ChildlessAccountUpdate]`
 
 Just to clarify, I'm not suggesting we do this in this PR, just wondering your thoughts on a follow-up later on when we have a better idea of what types/API we want for this?

src/build/buildExample.js
```diff
@@ -37,6 +37,29 @@ async function build(srcPath) {
   return absPath;
 }
 
+async function buildOne(srcPath) {
```
MartinMinkov, line 40: Awesome build config :D What's the intention of adding this in? Just one-off builds?

src/examples/zkapps/dex/dex.ts
```diff
@@ -105,19 +103,13 @@ class Dex extends SmartContract {
...
-    // construct any account update, e.g. draining all our lq tokens -.-
-    let callback = Experimental.Callback.create(dexX, 'redeemLiquidity', [
-      user,
-      dl,
-      this.tokenY,
-    ]);
-    tokenX.authorize(callback);
-    // TODO: get return value from callback and return here
+    let dxdy = dexX.redeemLiquidity(user, dl, this.tokenY);
+    tokenX.authorize(dexX.self);
```
MartinMinkov, line 111: Bless :pray:

src/lib/zkapp.ts
```diff
@@ -489,70 +488,61 @@ class Callback<Result> extends GenericArgument {
...
+  }
+
+  private constructor(self: Callback<any>) {
+    super();
+    Object.assign(this, self);
   }
 }
 
-// TODO: prove call signature in the outer circuit, just like for composability!
 function accountUpdateFromCallback(
```
MartinMinkov, line 537: Sorry to be that guy in the PR but could we add some doc comments here since we are modifying the function? :D

MartinMinkov (approved):
Pre approving! Looks great, awesome work :rocket:

src/examples/zkapps/dex/dex.ts
mitschabaude: That's a great suggestion, having names instead of numbers will go a long way towards making it clearer :D

Maybe two functions are enough:
* `StaticChildren(n: number)` / `StaticChildren(...children)` to replace 0, 1, .., [0, ...], ...
* `AnyChildren` to replace `null`

Any maybe `NoChildren` as alias for `StaticChildren(0)`

So the layout in the DEX example `[[3, 0, 0], 0, 0]` would become
```ts
// in a namespace to not bloat exports
let { StaticChildren, NoChildren } = AccountUpdateLayout;

let layout = StaticChildren(
  StaticChildren(StaticChildren(3), NoChildren, NoChildren),
  NoChildren,
  NoChildren
);
```

mitschabaude: or should we be even more verbose (and potentially, clearer) with something like `AccountUpdateWithChildren(3)`, `ChildlessAccountUpdate`, `AccountUpdateWithAnyChildren`?

mitschabaude: ended up with the shorter names above, but put them under a nested namespace: `AccountUpdate.Layout`

src/lib/zkapp.ts
mitschabaude: Thanks for being that guy :D Since `accountUpdateFromCallback` was the almost the same as `authorize` anyway, instead of doc-commenting it I ended up removing it and making `authorize` also take a callback. API feels cleaner now. I also added doc-comments to `authorize`!

PR #435, opened 2022-09-22
Add caller to token.send call

MartinMinkov:
# Description
This commit adds an optional `caller` input argument to the token.send call. By setting `caller` to the accountUpdate that's being used to authorize a token transfer, we let the token contract be callable by other zkApps.

If one zkApp calls on a token contract to authorize a transfer, the `caller` field of the token contract AccountUpdate must be set to the token id of the calling zkApp.

This is an initial stab at fixing https://github.com/o1-labs/snarkyjs/issues/431

# Code Example Changes

The code for a transfer method on a token contract would now look like:

```ts
@method transfer(
    from: PublicKey,
    to: PublicKey,
    value: UInt64,
    caller: Field
  ) {
    this.experimental.token.send({ from, to, amount: value, caller });
  }
```

The code for a zkApp to call into the token contract would look like:

```ts
@method supplyLiquidity(user: PublicKey, dx: UInt64) {
  let tokenX = new TokenContract(this.tokenX);
  tokenX.transfer(user, this.address, dx, this.experimental.token.id);
}
```

### DIFF
--- a/src/lib/account_update.ts
+++ b/src/lib/account_update.ts
@@ -681,12 +681,15 @@ class AccountUpdate implements Types.AccountUpdate {
         from,
         to,
         amount,
+        caller,
       }: {
         from: PublicKey;
         to: PublicKey;
         amount: number | bigint | UInt64;
+        caller?: Field;
       }) {
         // Create a new accountUpdate for the sender to send the amount to the receiver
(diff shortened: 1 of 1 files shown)


MartinMinkov:
This was a draft PR to unblock some features. This is no longer needed.

PR #452, opened 2022-09-28
Fetch Account's verification key.

shimkiv:
This should allow users determine if SC was already deployed using particular account.


mitschabaude (approved):
LGTM!

ymekuria (approved):
LGTM. Thanks for doing this @shimkiv.

PR #459, opened 2022-09-29
Dex Application - supplyLiquidity Happy Path

MartinMinkov:
# Description
This updates the `run.ts` and `dex.ts` files to properly run the `supplyLiquidity` methods. The script will call `Dex.supplyLiquidityBase` and `Dex.supplyLiquidity`. There is an [existing issue with caller fields](https://github.com/o1-labs/snarkyjs/issues/431) not being set correctly so the `run.ts` script contains a hack workaround of just mutating the `caller` field in the returned transaction.

There is also another [existing bug that prevents @methods with return values](https://github.com/o1-labs/snarkyjs/issues/429) to be called directly in a transaction block. In this PR, I have just commented out the return types in the related Dex methods, for now, to get around this bug.

Additionally, there is an error in the `Dex.supplyLiquidityBase` method on one of the constraints, where a failure is thrown. This check was passed before but after merging in local proof validation, it seems to have broken. Perhaps mutating the `caller` field is having weird side effects here. 

In a follow-up PR, I will add some tests around the failure methods of supplying liquidity.

Additionally hard coded some private keys to make debugging the returned transactions easier. 

# Tested By
Ran the script and inspected the results manually.

### DIFF
--- a/src/examples/zkapps/dex/dex.ts
+++ b/src/examples/zkapps/dex/dex.ts
@@ -378,13 +378,20 @@ function balanceSum(accountUpdate: AccountUpdate, tokenId: Field) {
 }
 
 /**
- * Random accounts keys, labeled by the input strings
+ * Predefined accounts keys, labeled by the input strings. Useful for testing/debugging with consistent keys.
  */
 function randomAccounts<K extends string>(
   ...names: [K, ...K[]]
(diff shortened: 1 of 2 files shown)


mitschabaude (approved):
LGTM!

PR #462, opened 2022-09-29
add enableProofs

Trivo25:
Missing piece to https://github.com/o1-labs/snarkyjs/issues/150

- if `enableProofs == false`,
	- `tx.prove()` will be "skipped" and a dummy proof will be inserted instead 
	- proofs wont be verified


mitschabaude (approved):
nice!

PR #467, opened 2022-09-30
Error message if token symbol length too long

### LINKED ISSUE #465: Add a check to `TokenSymbol` to match the same assertion as OCaml
# Description
In OCaml, we have an assertion to [check the length of a token symbol](https://github.com/MinaProtocol/mina/blob/04b85f369ee4013ca6278d41842381cafbeb26c1/src/lib/mina_base/account.ml#L99) to be a max length of `6`. We should additionally do this check-in SnarkyJS as well, to save the user time before sending it off to the network.

Presumably, SnarkyJS is already doing this check but the length seems to be off. If we pass in a very long token symbol, we get the following stack trace:

```
Error: prefix too long
    at prefixToField (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/hash.js:56:15)
    at Object.from (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/hash.js:141:21)
    at Object.set (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/account_update.js:379:86)
    at TokenContract.init (file:///workspace/deploy2berkeley/build/src/token-minaexplorer-repro.js:27:26)
    at file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/zkapp.js:89:113
    at runWith (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/global-context.js:55:18)
    at Function.runWith (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/global-context.js:14:37)
    at file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/zkapp.js:89:61
    at runWith (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/global-context.js:55:18)
    at Function.runWith (file:///workspace/deploy2berkeley/node_modules/snarkyjs/dist/node/lib/global-context.js:14:37)
```

We should just do the initial check of length `6` when setting the token symbol.

garethtdavies:
Fixes #465

### DIFF
--- a/src/lib/hash.ts
+++ b/src/lib/hash.ts
@@ -170,6 +170,7 @@ const TokenSymbol = {
   },
 
   from(symbol: string): TokenSymbol {
+    if (symbol.length > 6) throw Error('Token symbol length should be a maximum of 6');
     let field = prefixToField(symbol);
     return { symbol, field };
   },


mitschabaude (approved):
Thank you!

ymekuria (approved):

mitschabaude:
Merged as part of another PR!

PR #468, opened 2022-09-30
Updating burn method to update circulating supply

garethtdavies:
This replicates the code for minting tokens but subtracts the burnt tokens from the circulating supply


garethtdavies:
@MartinMinkov Updated this to update the circulating currency when using the burn in line with the mint method. As it's used as an example in the docs, probably worth including? This breaks the test though 😕 Any pointers to fix those up?

PR #470, opened 2022-10-05
Add optional assertion failure messages to SnarkyJS methods

MartinMinkov:
# Description
Added an optional string parameter to be used in assertion methods in the SnarkyJS OCaml bindings.

# Tested By
By building SnarkyJS locally and using it in test scripts. Ensures that both passing in an optional string and omitting it works as expected.

Companion too https://github.com/MinaProtocol/mina/pull/11964

Closes https://github.com/o1-labs/snarkyjs/issues/286


Trivo25:
very awesome addition! do you also want to add it to the changelog right in this PR as well?

mitschabaude (approved):
very cool! changelog would be nice

mitschabaude (approved):

PR #476, opened 2022-10-10
Fix verifcation on the web

### LINKED ISSUE #474: Failure to verify proofs in the browser
it seems the wasm js bindings need to be updated, and this should be added as a test to the browser e2e test

mitschabaude:
fixes #474


MartinMinkov (approved):
:rocket:

PR #477, opened 2022-10-11
Add proof verification into the tests.

shimkiv:
#474

### DIFF
--- a/tests/artifacts/javascript/on-chain-state-mgmt-zkapp-ui.js
+++ b/tests/artifacts/javascript/on-chain-state-mgmt-zkapp-ui.js
@@ -1,20 +1,16 @@
+import { logEvents } from './e2eTestsHelpers.js';
 import {
-  Mina,
-  PrivateKey,
+  adminPrivateKey,
+  HelloWorld,
+} from './examples/zkapps/hello_world/hello_world.js';
+import {
   AccountUpdate,
   Field,
   isReady,
-  SmartContract,
-  state,
-  State,
-  method,
-  Permissions,
+  Mina,
+  PrivateKey,
+  verify,
 } from './index.js';
-import {
-  HelloWorld,
-  adminPrivateKey,
-} from './examples/zkapps/hello_world/hello_world.js';
-import { logEvents } from './e2eTestsHelpers.js';
 
 await isReady;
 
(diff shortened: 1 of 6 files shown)


tests/artifacts/javascript/on-chain-state-mgmt-zkapp-ui.js
```diff
@@ -96,7 +93,12 @@ updateButton.addEventListener('click', async (event) => {
       zkAppInstance.sign(zkAppPrivateKey);
     });
 
-    await transaction.prove();
+    const proof = await transaction.prove();
+
+    if (verificationKey) {
+      verify(proof[0], verificationKey.data);
```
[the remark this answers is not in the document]
shimkiv: Forgotten about the `Promise` in return, thank you!

MartinMinkov (approved):
Looks great, thanks! :rocket:

shimkiv:
@mitschabaude @MartinMinkov 
Weird that tests still fail by:

```
Cannot read properties of undefined (reading 'proof')
```

Am I missing something?

MartinMinkov:
You are storing the result of `Mina.transaction` into `deploymentTransaction` but then calling `transaction.prove` and `transaction.send()`. Maybe it should be `deploymentTransaction` instead?

mitschabaude:
Not sure what causes the error you mention @shimkiv! Another thing: To make verification work in the browser, you probably have to merge `releases` since that is where I pushed the fix

mitschabaude:
@shimkiv I think you're signing the tx, that's why no proof is created, so `proofs[0]` is undefined`

tests/artifacts/javascript/on-chain-state-mgmt-zkapp-ui.js
```diff
@@ -96,7 +93,13 @@ updateButton.addEventListener('click', async (event) => {
       zkAppInstance.sign(zkAppPrivateKey);
```
mitschabaude, line 93: this causes it to create no proof

shimkiv: Thank you!
What was the driver behind this decision (if sign present no proofs will be generated)?
Is it documented somewhere?

mitschabaude: not documented, but planned to be!
sign means a signature is to be used as authorization. 'signature' and 'proof'  are two mutually exclusive forms of authorization. you can't have both.

shimkiv: Got it, thank you!

PR #484, opened 2022-10-13
Dex happy path

### LINKED ISSUE #403: Dex: Happy path flow

### LINKED ISSUE #445: DEX Example - redeemLiquidity method should run correctly
# Description

The current [redeemLiquidity](https://github.com/o1-labs/snarkyjs/blob/200be0ca0f575e08af2a62e1a3a001e7f53fd787/src/examples/zkapps/dex/dex.ts#L107) method on the DEX example throws an error from OCaml. 

Further, inspection is needed to understand why the error is being thrown and how to fix it.

### LINKED ISSUE #482: StackOverflow error caused by cloning `lazyAuthorization` inside `toAuxiliary`
# Description
When calling `this.experimental.authorize(accountUpdate)` inside a smart contract method, a stack overflow error occurs. This happens because inside `toAuxiliary`, the function will call `cloneCircuitValue` on `a.lazyAuthorization` and recursively call `cloneCircuitValue` until it blows up the stack.

See this line here: https://github.com/o1-labs/snarkyjs/blob/ecffd00ecb39f6aa24f2ead47e51eda570ab8071/src/lib/account_update.ts#L1095

This was confirmed by commenting out the `cloneCircuitValue` call and ensuring that no stack overflow error occurs.

MartinMinkov:
This PR finishes the happy path of the first DEX version, without proofs. Proofs currently run into an exotic Pickles error when executing "redeemLiquidity". We checked that the account update structures created by the DEX methods make sense and are as expected.

This also
* makes some changes to how Circuit.witness works,
* adds a `Circuit.log` function (use like console.log) to remove `Circuit.asProver` boilerplate when debugging,
* makes adding account updates to the tx more explicit in some cases where it happened in a very hidden way, causing bugs
* Closes #482
* Closes #403 
* Closes #445

### DIFF
--- a/src/examples/zkapps/dex/arbitrary_token_interaction.ts
+++ b/src/examples/zkapps/dex/arbitrary_token_interaction.ts
@@ -1,0 +1,65 @@
+import {
+  isReady,
+  Mina,
+  AccountUpdate,
+  UInt64,
+  shutdown,
+  Token,
+} from 'snarkyjs';
+import { TokenContract, addresses, keys, tokenIds } from './dex.js';
+
+await isReady;
+let doProofs = true;
+
+let Local = Mina.LocalBlockchain({ proofsEnabled: doProofs });
+Mina.setActiveInstance(Local);
+let accountFee = Mina.accountCreationFee();
+
+let [{ privateKey: userKey, publicKey: userAddress }] = Local.testAccounts;
+let tx;
+
+console.log('-------------------------------------------------');
+console.log('TOKEN X ADDRESS\t', addresses.tokenX.toBase58());
+console.log('USER ADDRESS\t', userAddress.toBase58());
+console.log('-------------------------------------------------');
+console.log('TOKEN X ID\t', Token.Id.toBase58(tokenIds.X));
+console.log('-------------------------------------------------');
+
+// compile & deploy all 5 zkApps
+console.log('compile (token)...');
+await TokenContract.compile();
+
+let tokenX = new TokenContract(addresses.tokenX);
+
+console.log('deploy & init token contracts...');
+tx = await Mina.transaction(userKey, () => {
+  // pay fees for creating 2 token contract accounts, and fund them so each can create 1 account themselves
+  let feePayerUpdate = AccountUpdate.createSigned(userKey);
+  feePayerUpdate.balance.subInPlace(accountFee.mul(1));
+  tokenX.deploy();
+});
+await tx.prove();
+tx.sign([keys.tokenX]);
+await tx.send();
+
+console.log('arbitrary token minting...');
+tx = await Mina.transaction(userKey, () => {
+  // pay fees for creating user's token X account
+  AccountUpdate.createSigned(userKey).balance.subInPlace(accountFee.mul(1));
+  // 😈😈😈 mint any number of tokens to our account 😈😈😈
+  let tokenContract = new TokenContract(addresses.tokenX);
+  tokenContract.experimental.token.mint({
+    address: userAddress,
+    amount: UInt64.from(1e18),
+  });
+});
+await tx.prove();
+console.log(tx.toPretty());
+await tx.send();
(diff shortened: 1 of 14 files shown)


src/examples/zkapps/dex/arbitrary_token_interaction.ts
```diff
@@ -0,0 +1,65 @@
...
+console.log('compile (token)...');
+await TokenContract.compile();
+
+let tokenX = new TokenContract(addresses.tokenX);
+
+console.log('deploy & init token contracts...');
+tx = await Mina.transaction(userKey, () => {
+  // pay fees for creating 2 token contract accounts, and fund them so each can create 1 account themselves
+  let feePayerUpdate = AccountUpdate.createSigned(userKey);
+  feePayerUpdate.balance.subInPlace(accountFee.mul(1));
```
Trivo25, line 38: Is there a reason for multiplying it times one? Probably not, just curious tho

src/examples/zkapps/dex/dex.ts
```diff
@@ -58,6 +64,7 @@ class Dex extends SmartContract {
     let isXZero = dexXBalance.equals(UInt64.zero);
     let xSafe = Circuit.if(isXZero, UInt64.one, dexXBalance);
 
+    // FIXME
```
Trivo25, line 67: is this important?

src/lib/circuit_value.ts
```diff
@@ -1003,18 +1009,30 @@ Circuit.constraintSystem = function <T>(f: () => T) {
   return result;
 };
 
+Circuit.log = function (...args: any) {
```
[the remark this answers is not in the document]
Trivo25: great addition

Trivo25 (approved):
lgtm, just left a few questions :)

src/examples/zkapps/dex/arbitrary_token_interaction.ts
```diff
@@ -0,0 +1,65 @@
...
+});
+await tx.prove();
+tx.sign([keys.tokenX]);
+await tx.send();
+
+console.log('arbitrary token minting...');
+tx = await Mina.transaction(userKey, () => {
+  // pay fees for creating user's token X account
+  AccountUpdate.createSigned(userKey).balance.subInPlace(accountFee.mul(1));
+  // 😈😈😈 mint any number of tokens to our account 😈😈😈
```
ymekuria, line 49: I love the comment emojis ha!

src/examples/zkapps/dex/run.ts
```diff
@@ -9,9 +16,9 @@ import {
 } from './dex.js';
```
ymekuria, line 17: Was there a reason you created a separate `arbitrary_token_interaction.ts` file and this `run.ts` file? Is the goal to only execute the `run.ts` file in CI and not the `arbitrary_token_interaction.ts` file?

src/lib/circuit_value.ts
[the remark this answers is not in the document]
ymekuria: This is awesome!

ymekuria (approved):
LGTM! I just had some minor comments/questions.

src/examples/zkapps/dex/arbitrary_token_interaction.ts
mitschabaude: no there isn't :D it was copied from another place where it was multiplied by 2 I guess

src/examples/zkapps/dex/dex.ts
mitschabaude: yes! reminds us to fix it

src/examples/zkapps/dex/run.ts
ymekuria: Ok that makes sense. Maybe we should document the file with this info so we don't forget later on.

PR #496, opened 2022-10-19
Bigger hard-coded circuits

mitschabaude:


zhfnjust:
still can not compile the recursion sha256.

error:
image




https://gist.github.com/zhfnjust/4996fc397b38562eb230c73a059e4a3f

mitschabaude:
Yes as I said, the bigger circuit size will only work with the `Circuit` API. Increasing sizes of `ZkProgram` would need a bit more digging. But I don't see the point of doing that because it just will run into the same OOM error.

Maybe the issue with the recursion SHA256 is that, since it includes a verifier circuit in addition to the SHA256 circuit, it is probably too big already with 1 SHA256 invocation, so needs to be split up further

zhfnjust:
how to  split up SHA256 circuit?

xhliu:
We still need this.

mitschabaude:
> We still need this

Btw @xhliu we have a sha256 in the works that can hash ~10 blocks per circuit
https://github.com/o1-labs/o1js/pull/1285

xhliu:
When can it be used? @mitschabaude

PR #501, opened 2022-10-20
Dex, vesting + finish supply/redeem/swap

### LINKED ISSUE #455: Dex: Application C - Liquidity Redemption - Happy Path (No vesting)
# Description
Test the happy path methods during attempts to redeem liquidity with no vesting applied.

## Action
User calls the "Liquidity Redemption" SC method providing the amount of IqXY tokens one is willing to redeem

## Expected Results
Asked amount of IqXY tokens is burned off the user's account
- Check the balance before and after

Calculated amounts of X and Y tokens are transferred to the user's token accounts
- Check the balances on sender and receiver sides

### LINKED ISSUE #456: Dex: Application C - Liquidity Redemption - Happy Path (Vesting period)
# Description
Test the happy path methods during attempts to redeem liquidity with vesting applied.

## Actions
Same case as https://github.com/o1-labs/snarkyjs/issues/455, but instead this happy case is processed with a vesting period applied to IqXY tokens.

## Expected Results
Check that it's impossible to redeem IqXY tokens without respecting the timing 
Check that it's possible to redeem IqXY tokens when timing conditions are met

### LINKED ISSUE #457: Dex: Application C - Liquidity Redemption - Test Failures
# Description
Test the failure methods during attempts to redeem liquidity when:
- [x] There are not enough IqXY tokens available for a users account
- [x] Resulting operation will overflow users receiving tokens by type or any other applicable means
- [ ] There are not enough X and/or Y tokens available on the SC side in order to complete the requested redemption.

mitschabaude:
This PR finishes all remaining tests in the categories "Supply liquidity", "Redeem liquidity", and "Swap".

See comments in `dex/run.ts` for details on how the tests were done, and the testing Google doc for comments about why certain tests were left out.

- closes #455
- closes #456 
- closes #457
- exposes the `timing` field from accounts on local blockchain
- Mina companion: https://github.com/MinaProtocol/mina/pull/12020


src/lib/zkapp.ts
```diff
@@ -593,9 +593,9 @@ class SmartContract {
   address: PublicKey;
   tokenId: Field;
 
-  private _executionState: ExecutionState | undefined;
+  #executionState: ExecutionState | undefined;
```
mitschabaude, line 596: changes in this file are to work around some TS glitches when returning a `SmartContract` from a function (i.e., using a class factory)

src/lib/mina.ts
```diff
@@ -459,11 +471,11 @@ function LocalBlockchain({
     setTimestamp(ms: UInt64) {
       networkState.timestamp = ms;
     },
-    setGlobalSlot(slot: UInt32) {
-      networkState.globalSlotSinceGenesis = slot;
-    },
-    setGlobalSlotSinceHardfork(slot: UInt32) {
-      networkState.globalSlotSinceHardFork = slot;
+    incrementGlobalSlot(increment: UInt32 | number) {
```
mitschabaude, line 474: improvement to the API of setting the global slot: `globalSlotSinceGenesis` and `globalSlotSinceHardfork` should always be updated together, not independent of each other. Also, the only update that makes sense is increasing them.

src/lib/account_update.ts
```diff
@@ -940,7 +940,8 @@ class AccountUpdate implements Types.AccountUpdate {
     // if the fee payer is the same account update as this one, we have to start the nonce predicate at one higher,
     // bc the fee payer already increases its nonce
     let isFeePayer = Mina.currentTransaction()?.sender?.equals(publicKey);
-    if (isFeePayer?.toBoolean()) nonce++;
+    let shouldIncreaseNonce = isFeePayer?.and(tokenId.equals(TokenId.default));
```
mitschabaude, line 943: small bug fix here when computing nonces in an edge case

mitschabaude:
Note: tests are failing because the bindings are not updated, to make review easier

src/examples/zkapps/dex/dex.ts
```diff
@@ -19,250 +19,339 @@ import {
...
+      // but for that, we need the @method argument generalization
+      let dexXBalance = tokenX.getBalance(this.address);
+      let dexYBalance = tokenY.getBalance(this.address);
+
+      // // assert dy == [dx * y/x], or x == 0
+      let isXZero = dexXBalance.equals(UInt64.zero);
+      let xSafe = Circuit.if(isXZero, UInt64.one, dexXBalance);
+
+      // FIXME
+      // Error: Constraint unsatisfied (unreduced): Equal 0 1
```
MartinMinkov, line 76: Should we remove these comments now that the line below has been enabled? I'm also curious to know what was causing this error if you know the cause?

src/examples/zkapps/dex/dex.ts
```diff
@@ -19,250 +19,339 @@ import {
...
+         * exercise the "timing" (vesting) feature to lock the received liquidity tokens.
+         *
+         * THIS IS HERE FOR TESTING!
+         *
+         * In reality, the timing feature is a bit awkward to use for time-locking liquidity tokens.
+         * That's because, if there is currently a vesting schedule on an account, we can't modify it.
+         * Thus, a liquidity provider would need to wait for their current tokens to unlock before being able to
+         * supply liquidity again (or, create another account to supply liquidity from).
+         */
+        let amountLocked = dl;
```
MartinMinkov, line 100: Should we spin up a ticket to add a nicer API for updating timing functionality?

src/lib/mina.ts
MartinMinkov: Does it matter to add `UInt64` to this type as well?

MartinMinkov (approved):
Incredible work as always :D

src/examples/zkapps/dex/dex.ts
mitschabaude: It's covered by this one: https://github.com/o1-labs/snarkyjs/issues/186

src/examples/zkapps/dex/dex.ts
mitschabaude: I didn't test the line again with proving enabled - just wanted it enabled for fuller testing in the non proving case. All of these tests only work with not proving by now. I have an issue open just for investigating this line: https://github.com/o1-labs/snarkyjs/issues/499

src/lib/mina.ts
mitschabaude: I don't think so, slots are always just UInt32 in Mina

src/examples/zkapps/dex/run.ts
```diff
@@ -36,160 +31,487 @@ console.log('TOKEN X ID\t', Token.Id.toBase58(tokenIds.X));
...
+    dex.supplyLiquidityBase(addresses.user, UInt64.from(1e9), UInt64.from(1e9));
+  });
+  await tx.prove();
+  tx.sign([keys.user]);
+  await expect(tx.send()).rejects.toThrow(/Overflow/);
+
+  /**
+   * - Resulting operation will overflow the SC’s receiving token by type or by any other applicable limits;
+   *
+   * note: this throws not at the protocol level, but because the smart contract multiplies two UInt64s which overflow.
```
bkase, line 252: what do you mean here? There is a constraint failure due to the overflow, right? It's just due to the smart contract.

bkase (approved):
Looks great to me! @shimkiv should take a look too!

I think we need to at least sort out #499 before calling this project done though

src/examples/zkapps/dex/run.ts
mitschabaude: Yes there is a constraint failure. So we don't get to the point where we could trigger an overflow failure in the protocol logic. In that sense, this is not a proper test as it was intended

PR #514, opened 2022-10-26
Dex upgradability test

Trivo25:
depends on https://github.com/o1-labs/snarkyjs/pull/501 and https://github.com/o1-labs/snarkyjs/pull/503

- closes https://github.com/o1-labs/snarkyjs/issues/508
- closes https://github.com/o1-labs/snarkyjs/issues/507
- closes https://github.com/o1-labs/snarkyjs/issues/505
- closes https://github.com/o1-labs/snarkyjs/issues/504
- also: exposes the `verificationKey` of an account when calling `Mina.getAccount(): Account`

~~While the tests are all written, we have to wait for a fix for https://github.com/o1-labs/snarkyjs/issues/498 to make sure the tests actually do what we expect them to do. Without proving, we also can't verify the proofs against the deployed and upgraded verification keys~~


.github/workflows/build-action.yml
```diff
@@ -12,7 +12,7 @@ on:
 
 jobs:
   Build-And-Test-Server:
-    timeout-minutes: 180
+    timeout-minutes: 210
```
Trivo25, line 15: CI now takes up to 3.5hrs with the remaining DEX tests :( ill try to fix that while I am working on the other CI issue

run-integration-tests.sh
```diff
@@ -5,3 +5,4 @@
 ./run src/examples/zkapps/composability.ts || exit 1
 ./run src/examples/zkapps/dex/run.ts --bundle || exit 1
 ./run src/examples/zkapps/dex/happy-path-with-proofs.ts --bundle || exit 1
+./run src/examples/zkapps/dex/upgradability.ts --bundle || exit 1
```
Trivo25, line 8: I had to separately run the initial DEX tests and the new tests because I was running into a weird wasm bug, will open an issue soon

src/examples/zkapps/dex/dex.ts
```diff
@@ -293,6 +303,31 @@ function createDex({
     }
   }
 
+  class ModifiedDexTokenHolder extends DexTokenHolder {
+    /**
+     * This swap method has a slightly changed formula
+     */
+    @method swap(
```
Trivo25, line 310: the only change here is a slightly adjusted formula

Trivo25 (commented):
see the initial google doc for reference

src/examples/zkapps/dex/upgradability.ts
```diff
@@ -0,0 +1,456 @@
...
+  let newDelegate = PrivateKey.random().toPublicKey();
+  tx = await Mina.transaction(feePayerKey, () => {
+    let dexAccount = AccountUpdate.create(addresses.dex);
+    AccountUpdate.setValue(dexAccount.update.delegate, newDelegate);
+    dexAccount.sign();
+  });
+  await tx.prove();
+  await tx.sign([keys.dex]).send();
+
+  Mina.getAccount(addresses.dex).delegate?.assertEquals(newDelegate);
```
nicc, line 159: Would calls to smart contract methods fail at this point since the delegate is set to a random address? It's not a problem (delegates are exercised in the upgradability section), just checking that I understand correctly.

nicc (approved):
Lgtm. It’s nice that the test structure matches the initial set of tests identified in the google doc (atomic actions / upgradability).

In the interest of my onboarding, I’ve written out my understanding and asked some questions below and in comments. I’d appreciate corrections and input 🙏. This does not need to block the merge.

———
Okay so we’re testing the ability to upgrade a smart contract in the DEX example. We’ve extended the `DEX` smart contract class to make `ModifiedDEX`, which implements a change to the `swapX` method via a modified token holder dependency.

This simply supports a new integration test that exercises upgrades. The example itself is effectively unaltered since the `main` function in `src/examples/zkapps/dex/run.ts` does not run through an upgrade.

The test itself is linear, incrementally building on itself to cover all required cases. This invovles some coupling between tests and can get brittle as we add new test cases but it reads nicely and involves less setup. Seems liek a reasonable choice. We start by testing the actual transactions that set delegates (i.e. deploys changes to the smart contract), then move on to an integration flow that actually deploys some updates and verifies that behaviour has changed.

src/examples/zkapps/dex/upgradability.ts
```diff
@@ -0,0 +1,456 @@
...
+  tx = await Mina.transaction(feePayerKey, () => {
+    // pay fees for creating 3 dex accounts
+    AccountUpdate.createSigned(feePayerKey).balance.subInPlace(
+      accountFee.mul(3)
+    );
+    dex.deploy();
+    tokenX.deployZkapp(addresses.dex, DexTokenHolder._verificationKey!);
+    tokenY.deployZkapp(addresses.dex, DexTokenHolder._verificationKey!);
+
+    // setting the setDelegate permission field to impossible
```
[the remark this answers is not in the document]
Trivo25: good point, that should help better understand the flow of the test when you follow it in the console :)

src/examples/zkapps/dex/upgradability.ts
Trivo25: I think in that case it should be fine and not fail

PR #516, opened 2022-10-26
Add args to the callback method invocation

MartinMinkov:
# Description
Fixes https://github.com/o1-labs/snarkyjs/issues/515 
This PR adds a spread operation of the arguments into the called method on a callback. This properly passes all argument values into the method invocation.

# Tested:
Manual testing


Trivo25 (approved):
👍

mitschabaude (approved):
Good catch!

PR #518, opened 2022-10-27
Fix zkapp URI hash

mitschabaude:


mitschabaude:
will not be needed for Mina PR after all, closing

PR #522, opened 2022-10-31
Add additional token tests

### LINKED ISSUE #513: Token Tests Requirements
Gathering all requirements in a single issue for easier tracking.

According to an internal testing document, we would like the following aspects of tokens to be tested:
- [x] Token Contract Creation/Deployment.
    - [x] - Check that the proper token owner exists
    - [x] - Check that an account creation fee is taken when applied to the ledger
    - [x] - Check that the token owner contract exists in the ledger
    - [x] - Check that submitting a successful token symbol is successful and applied to the ledger
    - [x] - Check that submitting an unsuccessful token symbol fails
    - [x] Authorization
        - [x] - Check that a child token account can be deployed under a token owner with proper authorization

- [x] Minting
    - [x] - Check that a token owner can mint (signature and proof) and balances are applied to the ledger
    - [x] - Check that minting in an overflow will fail
    - [x] - Check that minting a negative amount of tokens fails

- [x] Burning
    - [x] - Check that a token owner can burn (signature and proof) and balances are applied to the ledger
    - [x] - Check that burning in an overflow will fail
    - [x] - Check that burning a negative amount of tokens fails

- [x] Transfer
    - [x] - Check that a token owner can transfer it's own tokens to a token account that doesn't exist
    - [x] - Check that a token owner can transfer it's own tokens to a token account that does exist
    - [x] - Check that sending a number of tokens greater than the balance fails
    - [x] - Check that sending a negative amount of tokens fails
    - [x] Authorization
        - [x] - Check that a child token account can get authorization from a token owner account to transfer tokens
        - [x] - Check that a child token account cannot transfer without proper authorization from the token owner
        - [x] - Check that a token owner can inspect different types of layouts for authorization of child token accounts

Related issue: https://github.com/o1-labs/snarkyjs/issues/404

MartinMinkov:
Closes #513


src/lib/token.test.ts
```diff
@@ -13,314 +14,523 @@ import {
...
+      test('correct token id can be derived with an existing token owner', () => {
+        expect(tokenId).toEqual(Token.getId(tokenZkappAddress));
+      });
 
-    it('should have a valid token symbol', async () => {
-      const symbol = Mina.getAccount(zkappAddress).tokenSymbol;
-      expect(tokenSymbol).toBeDefined();
-      expect(symbol).toEqual(tokenSymbol);
-    });
+      test('deployed token contract exists in the ledger', () => {
```
bkase, line 232: can you organize the tests and comment how they map to the token tests doc? Similar to https://github.com/o1-labs/snarkyjs/blob/main/src/examples/zkapps/voting/test.ts#L54 . This will help us double-check that we covered everything in the doc -- and I think you mentioned you added a few more cases that weren't covered in the doc, can you add it there too?

bkase (changes requested):

bkase (approved):

Issue #511, opened 2022-10-24
Remove static members that are replaceable by constructor invocations


bkase:
This means we should remove Field.zero and Field.one and Bool.true and Bool.false

mitschabaude:
There's also `Field.minusOne`, and `UInt.{one,zero}`, and `Int.{one,zero,minusOne}`

bkase:
I don't see UInt or Int constants in here and it looks like the bool ones were already removed

mitschabaude:
here!

https://github.com/o1-labs/snarkyjs/blob/928bc47ec2e568bc159ab0d468cb282747d9a788/src/lib/int.ts#L13-L19

https://github.com/o1-labs/snarkyjs/blob/928bc47ec2e568bc159ab0d468cb282747d9a788/src/lib/int.ts#L205-L211

https://github.com/o1-labs/snarkyjs/blob/928bc47ec2e568bc159ab0d468cb282747d9a788/src/lib/int.ts#L13-L19

https://github.com/o1-labs/snarkyjs/blob/928bc47ec2e568bc159ab0d468cb282747d9a788/src/lib/int.ts#L494-L502

mitschabaude:
@bkase actually we should probably wait on the simplified `UInt32(number)` constructors before removing these constants

bkase:
ah okay, they're just not used by us yet I guess

mitschabaude:
We use them quite a bit!

mitschabaude:
![image](https://user-images.githubusercontent.com/20989968/199316998-cd9de96f-7a55-4396-a8ec-519ac5bbf550.png)

PR #524, opened 2022-11-01
Removes static one/zero/-1 from Field

bkase:
Part of #511


bkase:
How do I rebuild the reference docs for the docs2 site? @MartinMinkov

MartinMinkov:
> How do I rebuild the reference docs for the docs2 site? @MartinMinkov

I don't think we need to rebuild the reference docs quite yet since we are not publishing a new version on NPM. If we rebuild the docs for these changes, it won't reflect what users are actually downloading off NPM, so we should wait until there's a new SnarkyJS version :D

mitschabaude (approved):
unfortunately looks good to me!
but better to deprecate instead of removing the props

Trivo25 (approved):

PR #525, opened 2022-11-01
Add shutdown to end of primitive unit test

MartinMinkov:
This issue currently is breaking CI by letting jobs run for their extended execution time (6 hours). Fixes it by calling `shutdown` to end the primitive test process.

### DIFF
--- a/src/lib/primitives.unit-test.ts
+++ b/src/lib/primitives.unit-test.ts
@@ -1,5 +1,5 @@
 import { Circuit, circuitMain } from './circuit_value.js';
-import { isReady } from '../snarky.js';
+import { isReady, shutdown } from '../snarky.js';
 import { UInt64, UInt32 } from './int.js';
 import { expect } from 'expect';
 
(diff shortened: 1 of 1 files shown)


src/lib/primitives.unit-test.ts
```diff
@@ -19,5 +19,6 @@ let keypair = Primitives.generateKeypair();
 let proof = Primitives.prove([], [], keypair);
 let ok = Primitives.verify([], keypair.verificationKey(), proof);
 expect(ok).toEqual(true);
+shutdown();
```
mitschabaude, line 22: can we move this to the very bottom, so that the success line is displayed? :D

MartinMinkov: My bad, wasn't thinking there :sweat_smile:

Trivo25 (approved):

ymekuria (approved):
LGTM! Nice catch!

PR #528, opened 2022-11-04
Mina-signer JS MVP

mitschabaude:
* closes https://github.com/MinaProtocol/mina/issues/11855
* see [src/mina-signer/consistency.unit-test.ts](https://github.com/o1-labs/snarkyjs/pull/528/files/60daef0da62ab1a36094f6378bf5d07c5d77b6a0#diff-0e85394c6f171f45e2cf2c31db89e1d20c11d01668fbbaa867b9c2500c90f71a) for a good overview of the achievements of this PR

TODOs:
* [x] share js-deriver code between snarkyjs and mina-signer by making it generic over the leaf types
* [x] implement mina-signer version of js derivers, with leaf types based on bigint
* [x] add `fromJSON`
* [x] make `AccountUpdate.fromJSON` work with existing hash / base58 dependencies
* [x] create memo from string
* [x] compute memo hash with existing hash / base58 dependencies
* [x] compute transaction commitment with existing hash / base58 dependencies
* [x] get signZkappCommand compiling in pure JS, with stubbed out hash/base58/signature dependencies
* [ ] make JS crypto backend usable in a normal JS environment (vs js_of_ocaml)
* [ ] implement Poseidon
* [ ] implement signatures
* [ ] implement base58check or pull from a library
* [x] implement PublicKey <-> base58
* [ ] implement PrivateKey <-> base58
* [ ] implement Signature <-> base58
* [ ] generate code with version numbers / version bytes for base58 conversions from OCaml (now: runtime dependency)
* [ ] generate code with hash prefixed from OCaml (now: runtime dependency)
* [ ] get `signZkappCommand` fully working & matching what OCaml produces

EDIT: I think it's time to review this line of work and roll over the TODOs left to a different PR, otherwise this will become unreviewable


src/lib/events.ts
```diff
@@ -0,0 +1,162 @@
+import { prefixToField } from '../provable/binable.js';
```
mitschabaude, line 1: Events logic was moved here to make it generic over `Field`

src/lib/circuit_value.ts
```diff
@@ -569,6 +568,22 @@ function provable<A>(
     );
     return Object.fromEntries(keys.map((k, i) => [k, values[i]]));
   }
+  function fromJSON(typeObj: any, json: any, isToplevel = false): any {
```
mitschabaude, line 571: a major theme of this PR is that `fromJSON` functions were added to all provables

src/lib/encoding.ts
```diff
@@ -256,71 +260,3 @@ function bigIntArrayToBytes(bigints: bigint[], bytesPerBigInt: number) {
   }
   return bytes.slice(0, i + 1);
 }
-
-// encoding of fields as base58, compatible with ocaml encodings (provided the versionByte and versionNumber are the same)
-
-function fieldToBase58(x: Field, versionByte: number, versionNumber?: number) {
```
mitschabaude, line 262: this code was moved to `provable/binable.ts`

src/lib/hash.ts
```diff
@@ -35,14 +35,16 @@ class Sponge {
 
 const Poseidon = {
   hash(input: Field[]) {
-    let isChecked = inCheckedComputation();
+    let isChecked = !input.every((x) => x.isConstant());
```
mitschabaude, line 38: this is a better (more general) way of determining whether to use a "checked" implementation, enables us to clean up code below

src/mina-signer/TSTypes.ts
```diff
@@ -0,0 +1,66 @@
+export type UInt32 = number | bigint | string;
```
mitschabaude, line 1: this file was just dumped here from Mina/mina-signer and doesn't do anything currently

src/provable/derived-leaves.ts
```diff
@@ -0,0 +1,135 @@
+import { GenericBool, GenericField, GenericHashInput } from './generic.js';
+import { createProvable } from './provable-generic.js';
+import * as Json from './gen/transaction-json.js';
+import { fieldEncodings, prefixToField } from './binable.js';
+
+export { derivedLeafTypes };
+
+function derivedLeafTypes<Field, Bool>({
```
mitschabaude, line 8: one of the many cases where we made an implementation generic over `Field`, so that we can reuse it between snarkyjs & mina-signer

src/provable/from-layout.ts
```diff
@@ -0,0 +1,563 @@
+import {
```
mitschabaude, line 1: this file used to be called `transaction-helpers.ts`. it is the core of our "derived JS from OCaml" infrastructure. it was extended by adding the `fromJSON` and `emptyValue` functions, and was made generic over `Field`

src/provable/transaction-leaves-bigint.ts
```diff
@@ -0,0 +1,64 @@
+import { Field, Bool, UInt32, UInt64, Sign } from './field-bigint.js';
```
mitschabaude, line 1: there are now three "transaction-leaves" files. there used to be two
* one for snarkyjs types
* one for JSON types

now we have this third one for mina-signer (bigint-based) types

these files were made much less complex, by auto-generating more code and by moving some common logic to `derived-leaves.ts`

mitschabaude (commented):
some comments to aid review!

mitschabaude (commented):
some comments to aid review!

mitschabaude:
Note for reviewers: In the last commit, I copied over the js crypto backend from the Mina repo, without changing anything about it. The intention of doing this here is so that in the next PR, you can better see & review the changes I'm making to it

bkase (changes requested):
Thanks for the comments to help with review, but I think it's not enough. Sorry for being annoying -- but this is too important to be this difficult to review.

In the future, definitely try to stack PRs for big changes like this instead of having one big one.

But for now, can you rewrite your commit history so a reviewer can read the commits one at a time to understand the PR in smaller chunks? Hopefully, it's not too annoying to set that up.

src/provable/from-layout.ts
```diff
@@ -285,6 +294,67 @@ function ProvableFromLayout<
     return (TypeMap as any)[typeData.type].fromFields(fields, aux);
   }
 
+  function emptyValue(typeData: Layout): any {
```
mitschabaude, line 297: this function creates a 'dummy' for an arbitrary type, derived from the layout. this turned out to be necessary, to replace the simpler `emptyValue` function above which just created the value from all-zero field elements. some leaf types have custom `emptyValues`, which aren't all zeroes

src/provable/from-layout.ts
```diff
@@ -123,19 +120,30 @@ function ProvableFromLayout<
...
+        case 'closedInterval':
         case 'flaggedOption': {
-          if (json === null) {
-            return { isSome: false, value: emptyValue(inner) };
+          let isSome = TypeMap.Bool.fromJSON(json !== null);
+          let value;
+          if (json !== null) {
+            value = fromJSON(optionTypeData.inner, json);
+          } else {
+            value = emptyValue(optionTypeData.inner);
```
mitschabaude, line 132: these changes fix `fromJSON` in the case the JSON has a `null` value in place of a precondition that is specified as a closed interval. the new implementation recreates the proper default interval which is bounded by the min and max values. these values are annotated in OCaml and read off the layout here

src/provable/gen/js-layout.ts
```diff
@@ -338,7 +338,9 @@ let jsLayout = {
                         },
                         timestamp: {
                           type: 'option',
-                          optionType: 'flaggedOption',
+                          optionType: 'closedInterval',
```
mitschabaude, line 341: changes here are caused by the Mina companion: https://github.com/MinaProtocol/mina/pull/12162

src/js_crypto/bigint256.ts
```diff
@@ -0,0 +1,85 @@
+/* global joo_global_object, caml_js_to_bool, caml_jsstring_of_string, caml_string_of_jsstring,
```
mitschabaude, line 1: the files in `js_crypto` are just dumped here to modify in the next PR -- please ignore

bkase (approved):

PR #529, opened 2022-11-04
Expose constraint system as JSON

mitschabaude:
- introduce printing constraint system fn
- print script
- read out gate coeffs
- print test
- simplify


Trivo25 (approved):
lgtm

PR #536, opened 2022-11-08
Make `JSONValue` internally used only

MartinMinkov:
## Description
Move the definition of JSONValue into circuit_value.ts since there are many types that depend on JSONValue inside that file. We also removed any usages of JSONValue from public fromJSON methods since that type is intended for internal use.

Addresses https://github.com/o1-labs/snarkyjs/issues/353 as `JSONValue` was originally exposed externally but was then removed.

### DIFF
--- a/src/lib/circuit_value.ts
+++ b/src/lib/circuit_value.ts
@@ -32,7 +32,6 @@ export {
   InferCircuitValue,
   Provables,
   HashInput,
-  JSONValue,
 };
 
 type Constructor<T> = new (...args: any) => T;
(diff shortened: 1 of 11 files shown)


mitschabaude (approved):
Nice!

ymekuria (approved):

src/lib/circuit_value.ts
```diff
@@ -434,6 +434,7 @@ function circuitMain(
 let primitives = new Set(['Field', 'Bool', 'Scalar', 'Group']);
 let complexTypes = new Set(['object', 'function']);
 
+// TODO: Should `ProvableExtension` use `any` as well?
```
MartinMinkov, line 437: @mitschabaude Should we update `ProvableExtension` and all related types to use `any` instead of `JSONValue` as well?

mitschabaude: I agree! Let's merge it

PR #537, opened 2022-11-09
JS crypto for mina-signer

mitschabaude:
Follow-up to https://github.com/o1-labs/snarkyjs/pull/528

This PR makes the js crypto backend usable for snarkyjs and mina-signer:
* Port code from js_of_ocaml syntax/compiler directives to modern TS
* without changing any of the logic (please check this during review)
* Makes sure the old unit tests still run


Trivo25 (approved):
lgtm, made sure the logic didn't change. This is way cleaner

bkase:
Looks right to me, but can you get @mrmr1993 to double-check too?

PR #546, opened 2022-11-11
implement merkle map

es92:
Can be used for nullifiers, membership, and key -> value maps


src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+  Bool,
+  SelfProof,
+  CircuitValue,
+  arrayProp,
+  Poseidon,
+  prop,
+  Circuit,
+} from 'snarkyjs'
+
+import assert from 'assert';
```
mitschabaude, line 15: assert won't work on the browser!

src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const printDebugs = false;
+
+export class MerkleMap {
+  tree: InstanceType<typeof Experimental.MerkleTree>;
+
+  // ------------------------------------------------
+
+  constructor() {
+    assert(bits <= 255, 'bits must be <= 255')
```
mitschabaude, line 27: something like
```ts
if (bits > 255) throw Error('bits must be <= 255');
```

src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+    this.tree = new Experimental.MerkleTree(bits+1);
+  }
+
+  // ------------------------------------------------
+
+  _keyToIndex(key: Field) {
+    // the bit map is reversed to make reconstructing the key during proving more convenient
+    let keyBits = key.toBits().slice(0,bits).reverse().map((b) => b.toBoolean());
+
+    var n = BigInt(0);
```
mitschabaude, line 40: nit: can we use `let` instead of `var`? And bigint literals?

src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+    for (var i = 0; i < keyBits.length; i++) {
+      var b = keyBits[i] ? 1 : 0;
+      n += BigInt(2)**BigInt(i) * BigInt(b);
+    }
+
+    return BigInt(n)
```
mitschabaude, line 46: nit: `n` is already a bigint => `return n;`
another nit: would be nice to run this code through `prettier` (so we don't get another large formatting commit when someone with auto-formatting enabled makes a change to this file later)

src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    class MerkleWitness extends Experimental.MerkleWitness(bits + 1) {}
+
+    const witness = new MerkleWitness(this.tree.getWitness(index));
+
+    if (printDebugs) {
+      // witness bits and key bits should be the reverse of each other, so
+      // we can calculate the key during recursively traversing the path
+      console.log('witness bits', witness.isLeft.map((l) => l.toBoolean() ? '0' : '1').join(', '));
+      console.log('key bits', key.toBits().slice(0,bits).map((l) => l.toBoolean() ? '1' : '0').join(', '));
```
mitschabaude, line 82: this confuses me. in `_keyToIndex`, I thought the bit sequence was reversed, i.e. `11001` -> `10011`
Here, instead it seems like a bitwise NOT is applied: `11001` -> `00110`, giving a different result, or am I missing something?

src/lib/merkle_map.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  computeRootAndKey(value: Field) {
+    let hash = value;
+
+    const isLeft = this.isLefts;
+    const siblings = this.siblings;
+
+    let key = Field(0);
+
+    for (let i = 1; i <= bits; ++i) {
```
mitschabaude, line 111: I think the loop would be easier to read with `i` starting from 0 and going while `i < bits`, then all the `i-1` below could be replaced with `i`

mitschabaude (commented):
The implementation looks good to me with minor comments!

I have a more high-level question though: To provide a usable tool for nullifiers, don't we need something which computes the key from the leaf, and uses it for getting / setting, _inside the circuit_? Because we need to connect the leaf to the key inside the circuit?

Currently it seems like only `MerkleMapWitness` is suitable for in-circuit use, but not `MerkleMap`, similar to `MerkleWitness` vs `MerkleTree`.

Comdex:
Occasionally found this pr, I also share my previous research on the use of nullifier(That needs to be handled in the circuit), it seems that we can only use the sparse merkle tree (SMT) with a tree height of 256(If I'm not mistaken, since the result of the poseidon hash in snarkyjs actually only affects 254 bits, we only need trees of height 254?) to deal with the nullifier in snarkyjs now, because SMT can generate non-membership proof and we can verify the proof in the circuit, but the disadvantage is that the verification of this proof requires more hash times, the efficiency is relatively low and it is easy to reach the upper limit of the snarkyjs circuit constraint size.

src/lib/merkle_map.ts
es92: I think both are true actually - since both `isLeft` is reading out `true -> 0` and `false -> 1` - and computing the path from `isLeft` is computing the key in the "opposite" order than the key itself was entered (so the reversing step in `_keyToIndex`)

mitschabaude: Alright, thanks!

mitschabaude:
> which nullifier should follow from, just with **keys derived from preimages of commitments**

This is the thing that I thought was missing. It's possible though: The `MerkleMapWitness` has the `key` baked into it implicitly, and re-computes it in the circuit when doing `witness.computeRootAndKey`. At that point the user can take the key and check that it matches whatever is the scheme for deriving the key from the leaf / value.

I think this is fine, but IMO it would add value to have such a key-deriving-and-checking scheme baked in already, so there's less load on users to come up with their own scheme and implement the checks correctly.

In general, the whole MerkleTree implementation imposes a 'reverse thinking' to the user, where you compute something directly _outside_ the circuit and then verify that it matches inside the circuit. IMO, where it makes sense this should be abstracted from the user such that there are simple direct functions, like "check non-inclusion of this value" or "add this value to the tree" available for use in the circuit. (We can do the witness-computation-outside-the-circuit within snarkyjs, in a `Circuit.witness` block, hidden from the user) Do you agree @es92?

However, let's address this later and move forward with this PR as-is

mitschabaude (approved):

es92:
> > which nullifier should follow from, just with **keys derived from preimages of commitments**
> 
> This is the thing that I thought was missing. It's possible though: The `MerkleMapWitness` has the `key` baked into it implicitly, and re-computes it in the circuit when doing `witness.computeRootAndKey`. At that point the user can take the key and check that it matches whatever is the scheme for deriving the key from the leaf / value.
> 
> I think this is fine, but IMO it would add value to have such a key-deriving-and-checking scheme baked in already, so there's less load on users to come up with their own scheme and implement the checks correctly.
> 
> In general, the whole MerkleTree implementation imposes a 'reverse thinking' to the user, where you compute something directly _outside_ the circuit and then verify that it matches inside the circuit. IMO, where it makes sense this should be abstracted from the user such that there are simple direct functions, like "check non-inclusion of this value" or "add this value to the tree" available for use in the circuit. (We can do the witness-computation-outside-the-circuit within snarkyjs, in a `Circuit.witness` block, hidden from the user) Do you agree @es92?
> 
> However, let's address this later and move forward with this PR as-is

This makes sense to me - I think I wasn't aware of some of those features too (Circuit.witness sounds very cool).

And yep the idea was to just copy the structure of Merkle Tree Witness. Would be supportive of other implementations down the road (but also like moving forward with this as is to get the functionality out at the least)

PR #548, opened 2022-11-14
Mina Signer Pt 1 - generic transaction helpers

mitschabaude:
First part of https://github.com/o1-labs/snarkyjs/pull/528, splitting up the initial work on mina-signer. This contains many of the changes of https://github.com/o1-labs/snarkyjs/pull/528 in a few commits and so will make reviewing the rest easier.

Main things that happen here:
* Some initial experiments on rebuilding mina-signer in JS are dumped here (from the Mina repo, in a branch that didn't get reviewed). I think these can be ignored for the purpose of this PR review, because:
  * the stuff inside `mina-signer/copied` will gradually be integrated into / reconciled with snarkyjs in later commits, and moved to different folders, and it makes more sense to review it at this later stage when it's actually integrated 
  * the files at the top level of `mina-signer` is just the original mina-signer code, copied verbatim and not yet functional
* The important file `transaction-helpers.ts` is refactored, such that it becomes generic over a `TypeMap` which contains leaf types, i.e. `TypeMap = { Field, Bool, .. }`. This is needed for now because mina-signer must have its own implementations of those leaf types, different from the snarkyjs ones which depend on OCaml


src/mina-signer/copied/curve.ts
```diff
@@ -0,0 +1,14 @@
+import { Bool, Field } from './field.js';
```
mitschabaude, line 1: note: don't review anything in `src/mina-signer` for this PR. this stuff will be integrated in a proper form in snarkyjs piece by piece, but the versions here are in large parts redundant with snarkyjs code or copied from the old mina-signer

src/snarky/transaction-helpers.ts
```diff
@@ -1,385 +1,414 @@
...
+  AuthRequired: AuthRequired;
+  AuthorizationKind: AuthorizationKind;
+  TokenId: TokenId;
+  // builtin
+  number: number;
+  null: null;
+  undefined: Undefined;
+  string: string;
+};
+type AnyTypeMap = GenericTypeMap<
```
bkase, line 32: lol

bkase (approved):
Looks good 👍

PR #551, opened 2022-11-14
Mina Signer Pt 4 - fromJSON

mitschabaude:
- stacked on https://github.com/o1-labs/snarkyjs/pull/550
- this adds `fromJSON` methods on all provables in the code base. `fromJSON` is needed by mina-signer as a first step when translating a JSON transaction to one built from bigint-based leaf types.
- this PR also starts a new file for unit testing the mina-signer-related work


src/provable/binable.ts
```diff
@@ -30,6 +30,44 @@ function withVersionNumber<T>(
   };
 }
 
+type Tuple<T> = [T, ...T[]] | [];
+
+function compose<Types extends Tuple<any>>(
```
mitschabaude, line 35: compose two binables to create a binable for the composed type. this is used in this PR to re-create Mina's encoding of public keys in JS

src/provable/curve-bigint.ts
```diff
@@ -1,14 +1,34 @@
...
+// TODO generate
+let FIELD_VERSION = 1;
+let PUBLIC_KEY_VERSION = 1;
+
+let FieldWithVersion = withVersionNumber(Field, FIELD_VERSION);
+let BinablePublicKey = withVersionNumber(
+  compose([FieldWithVersion, Bool]),
+  PUBLIC_KEY_VERSION
+);
+let Base58PublicKey = base58(
```
mitschabaude, line 19: base58 encoding of public keys, which is used to implement toJSON / fromJSON for public keys

src/lib/circuit_value.ts
```diff
@@ -302,6 +302,14 @@ function circuitArray<T>(
       }
       return array.map((v) => elementType.toJSON(v));
     },
+    fromJSON(json) {
+      if (!('fromJSON' in elementType)) {
```
bkase, line 306: why is this checked at runtime? To avoid introducing more layers of "ProvableExtendedExtended" or something? What are the cases in which the fromJSON would be missing?

src/provable/binable.ts
bkase: nit: Can we call this `tuple` or `pair`? There are a few ways to compose values and the tupling action could be more explicit from the name

src/lib/circuit_value.ts
mitschabaude: In `circuitArray`, I also accept a `Provable<T>` as input, without the 'extended'. The general idea was to accept every `Provable<T>` in most APIs that a user would interact with, and don't require methods like `fromJSON` on every type, to have an interface that is as light-weight as possible. Most use cases for provables don't involve the 'extended' stuff but just the basic to/fromFields method for using these types in a circuit

bkase (approved):

src/lib/circuit_value.ts
bkase: 👍

PR #560, opened 2022-11-15
Remove timestamp field in zkapp_precondition

ghost-not-in-the-shell:
This PR removes `timestamp` and also `globalSlotSinceHardFork` field in `zkapp_precondition`


mitschabaude (approved):
LGTM!

mitschabaude:
but we shouldn't merge yet - should set up and run a CI workflow for this branch first

src/lib/mina.ts
```diff
@@ -395,7 +395,7 @@ function LocalBlockchain({
           }
           events[addr][tokenId].push({
             events: p.body.events,
-            slot: networkState.globalSlotSinceHardFork.toString(),
+            slot: networkState.globalSlotSinceGenesis.toString(),
```
ghost-not-in-the-shell, line 398: @mitschabaude Since I removed `globalSlotSinceHardFork`, I replaced this with `globalSlotSinceGenesis`, would that be okay?

PR #565, opened 2022-11-16
Update snarkyjs bindings for hashprefix

MartinMinkov:
## Description

Generate new SnarkyJS OCaml bindings to include a new change to the protocol made regarding a seed for signatures.


jasongitmail (approved):
Awesome!

PR #567, opened 2022-11-16
Fix Struct issues

mitschabaude:
This fixes type inference in the following code example:
```ts
class A extends Struct({a: Field}) {
  print() {
    console.log(this.a.toString());
  }
}

class B extends Struct({a: A, b: Bool}) {
  print() {
    this.a.print(); // FIXED!
  }
}
```
The type inference is changed in a way that it tries to not affect cases where inferring a class type already worked (e.g., CircuitValues like `UInt32`) 

EDIT: this PR now is about fixing more issues with Struct that came up

* closes https://github.com/o1-labs/snarkyjs/issues/576
* makes all (I hope) our APIs work with customized Structs
* fixes `Struct.fromJSON`


src/lib/circuit_value.test.ts
```diff
@@ -31,4 +40,28 @@ describe('circuit', () => {
...
+
+    const original = new MyStruct({ values: [Field(0), Field(1)] });
+
+    const serialized = MyStruct.toJSON(original);
+    const reconstructed = MyStruct.fromJSON(serialized);
+
+    Circuit.assertEqual<MyStruct>(MyStruct, original, reconstructed);
+  });
+
+  it('can serialize nested Struct', async () => {
```
Trivo25, line 57: should we maybe also test something like just to be sure? 

```ts
class OtherStruct extends Struct({
      x: Field,
}) {}

class MyStruct extends Struct({
      y: OtherStruct,
}) {}
```

Trivo25 (approved):
lgtm, just left a small comment

src/lib/circuit_value.test.ts
mirceanis: When I added this test, I got an identical stack trace using either a CircuitValue (in my case PrivateKey) or explicitly defined Struct, so I only posted the PrivateKey variant for brevity.

It makes sense to add an explicit nested Struct test too but if CircuitValue types will be migrated to Struct anyway it will likely become redundant.

mitschabaude: good catch!

Comdex:
@mitschabaude Hey, can you export these two types for external use? This allows developers to use these types to support various custom Structs as parameters of generic methods.
```typescript
type FlexibleProvable<T> = Provable<T> | Struct<T>;
type FlexibleProvablePure<T> = ProvablePure<T> | StructPure<T>;
```

PR #572, opened 2022-11-17
Poseidon JS

mitschabaude:
* closes https://github.com/o1-labs/snarkyjs/issues/573
* implements Poseidon in JS
* establishes a new auto-generated JS file `constants.ts` where we dump constants from OCaml, at compile time
  * see the Mina companion https://github.com/MinaProtocol/mina/pull/12229
  * side effect of this that's not related to Poseidon is that we use base58 version bytes from the generated file instead of from the runtime JSOO

Note: The Poseidon implementation is tested by `mina-signer/consistency.unit-test.ts`, which wasn't changed here but which uses the Poseidon from `poseidon-bigint.ts` -- which now contains the pure JS implementation -- and checks it against various hashes produced by snarkyjs with the OCaml Poseidon


mrmr1993 (approved):

mrmr1993 (approved):

src/js_crypto/constants.ts
```diff
@@ -0,0 +1,328 @@
+// @gen this file is generated - don't edit it directly
+export { prefixes, versionBytes, poseidonParams }
+
+let prefixes = {
```
jspada, line 4: To avoid human error, I'd suggest automatically padding these and asserting the correct length.

src/js_crypto/constants.ts
```diff
@@ -0,0 +1,328 @@
...
+  "receiptChainHash": 12,
+  "ledgerHash": 5,
+  "epochSeed": 13,
+  "stateHash": 16,
+  "publicKey": 203,
+  "userCommandMemo": 20,
+  "privateKey": 90,
+  "signature": 154
+};
+let poseidonParams = {
```
jspada, line 26: For clarity, and to avoid confusion, it would be good to name these according to our naming conventions here
https://github.com/o1-labs/proof-systems/tree/master/poseidon/src/pasta
For example, Kimchi or Legacy + Fp or Fq

src/js_crypto/poseidon.ts
```diff
@@ -0,0 +1,101 @@
...
+    rate,
+    power: power_,
+    roundConstants: roundConstants_,
+    mds: mds_,
+  }: PoseidonParameters
+) {
+  if (hasInitialRoundConstant) {
+    throw Error("we don't support an initial round constant");
+  }
+  if (partialRounds !== 0) {
```
jspada, line 35: Nice

jspada (commented):
I didn't notice any tests, but it would be very good to import and run our standard set of poseidon test vectors.  We have them here: https://github.com/o1-labs/proof-systems/tree/master/poseidon/src/tests/test_vectors

If they are not in a desirable format for you, then you can update https://github.com/o1-labs/proof-systems/tree/master/poseidon/export_test_vectors to generate them in a javascript friendly format :-)

src/js_crypto/constants.ts
mitschabaude: these are auto-generated from OCaml, and I think we avoid human error there :)

jspada: Cool 👍🏼

src/js_crypto/test_vectors/dump-test-vectors.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -e
+
+pushd ../../../../../crypto/proof-systems/oracle/export_test_vectors
+  cargo run -p export_test_vectors -- b10 kimchi ../../../../snarky_js_bindings/snarkyjs/src/js_crypto/test_vectors/testVectors.json
```
jspada, line 6: You could also just generate them offline and save them to the repo somewhere, so that you don't need to have this dependency.

jspada (approved):
Looks good to me!

src/js_crypto/test_vectors/dump-test-vectors.sh
mitschabaude: Hm I didn't think of that :D Sounds like an idea

PR #585, opened 2022-11-21
Small fix to fromFields

mitschabaude:


nicc (approved):
A nullable argument fix makes sense given the error I encountered.

Issue #559, opened 2022-11-15
Discussion: Refactor internal CircuitValues to use Struct instead


MartinMinkov:
## Description
We have several useful primitives exposed in SnarkyJS that are implemented with the deprecated `CircuitValue` class. An example of such a primitive is the Merkle Tree implementation.

Should we refactor such primitives to use `Struct` instead?

bkase:
Will start with a subset -- and will identify the list of tasks

MartinMinkov:
Identified the following classes that extend `CircuitValue` that is used internally:

**In src/lib**
```sh
$ find src/lib -type f -name '*.ts' | xargs grep -n '.* extends CircuitValue'
src/lib/merkle_map.ts:116:export class MerkleMapWitness extends CircuitValue {
src/lib/merkle_tree.ts:149:class BaseMerkleWitness extends CircuitValue {
src/lib/signature.ts:11:class PrivateKey extends CircuitValue {
src/lib/signature.ts:77:class PublicKey extends CircuitValue {
src/lib/signature.ts:184:class Signature extends CircuitValue {
src/lib/string.ts:9:class Character extends CircuitValue {
src/lib/string.ts:37:class CircuitString extends CircuitValue {
src/lib/int.ts:12:class UInt64 extends CircuitValue {
src/lib/int.ts:248:class UInt32 extends CircuitValue {
src/lib/int.ts:480:class Sign extends CircuitValue {
src/lib/int.ts:525:class Int64 extends CircuitValue implements BalanceChange {
```
**In src/examples**
```sh
$ find src/examples/ -type f -name '*.ts' | xargs grep -n '.* extends CircuitValue'
src/examples/sudoku/sudoku-zkapp.ts:15:class Sudoku extends CircuitValue {
src/examples/schnorr_sign.ts:12:class Witness extends CircuitValue {
src/examples/zkapps/merkle_tree/merkle_zkapp.ts:41:class Account extends CircuitValue {
src/examples/zkapps/local_events_zkapp.ts:27:class Event extends CircuitValue {
src/examples/zkapps/voting/member.ts:21:export class Member extends CircuitValue {
src/examples/zkapps/voting/election_preconditions.ts:3:export default class ElectionPreconditions extends CircuitValue {
src/examples/zkapps/voting/participant_preconditions.ts:3:export default class ParticipantPreconditions extends CircuitValue {
src/examples/rollup/wip.ts:43:class RollupAccount extends CircuitValue {
[...]

PR #594, opened 2022-11-23
Refactor Signature classes to use `Struct` instead of `CircuitValue`

MartinMinkov:
## Description
This PR refactors the classes in `signature.ts` to use `Struct` instead of `CircuitValue`

Addresses some of #559


MartinMinkov:
This is getting stale, so I am closing it for now.

PR #612, opened 2022-11-29
Schnorr signatures in JS

### LINKED ISSUE #577: Implement signatures for mina-signer

mitschabaude:
- closes #577 
- this adds `signature.ts`, which is a JS implementation of [Schnorr.Chunked.sign](https://github.com/MinaProtocol/mina/blob/aebd4e552b8b4bcd78d1e24523169e8778794857/src/lib/signature_lib/schnorr.ml#L223-L255)
- this concludes the implementation of `signZkappCommand` for mina-signer JS
- adds a test comparing the JS signature on a non-trivial zkApp transaction with the OCaml implementation
- some supporting work, like base58 serialization of private keys and signatures
- **new npm dependency:** [blakejs](https://www.npmjs.com/package/blakejs), a package that implements the blake2b hash which is used to derive the nonce in the signature algorithm. the package has 750,000 weekly downloads and no dependencies itself.


bkase (approved):
Looks fine from a product perspective -- definitely I'd want to rely on a thorough crypto review and I think we'll need to include this in the next audit

src/lib/core.ts
```diff
@@ -38,3 +39,17 @@ Bool.fromBytes = function ([b]) {
   return Bool(!!b);
 };
 Bool.sizeInBytes = () => 1;
+
+Scalar.toFieldsCompressed = function (s: Scalar) {
+  let isConstant = s.toFields().every((x) => x.isConstant());
+  let constantValue: Uint8Array | undefined = (s as any).constantValue;
+  if (!isConstant || constantValue === undefined)
+    throw Error('Scalar.toFieldsCompressed is not available in provable code.');
```
MartinMinkov, line 47: Nit: Could we communicate what "provable" code is better in this error message? For users that are brand new to the Mina dev experience, this can cause some confusion on what this really means. Could we simplify this as to provide a hint on the correct usage?

src/provable/base58.ts
```diff
@@ -0,0 +1,66 @@
...
+  toBase58(t: T): string;
+  fromBase58(base58: string): T;
+};
+
+function base58<T>(binable: Binable<T>, versionByte: number): Base58<T> {
+  return {
+    toBase58(t) {
+      let bytes = binable.toBytes(t);
+      let binaryString = String.fromCharCode(...bytes);
+      let ocamlBytes = { t: 9, c: binaryString, l: bytes.length };
```
MartinMinkov, line 17: What's the significance of the `9` value here? Could we add a comment to explain why this number is specifically this value? If this was changed, what's the downstream effects? :O

package.json
```diff
@@ -76,6 +77,7 @@
     "typescript": "^4.8.2"
   },
   "dependencies": {
+    "blakejs": "^1.2.1",
```
MartinMinkov, line 80: Should we be fixing this version for a crypto dependency like this? Curious to your thoughts

MartinMinkov (approved):
Just left some comments and questions! I leave it up to the crypto team to review the more low level code here :D

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+) {
+  return sign({ fields: [message] }, privateKey, networkId);
+}
+
+function sign(
+  message: HashInput,
+  privateKey: PrivateKey,
+  networkId: NetworkId
+): Signature {
+  let publicKey = Group.scale(Group.one, privateKey);
```
jspada, line 60: For reasons not clear yet, it's probably best to call this `Group.generator_mina`

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+  let kPrime = deriveNonce(message, publicKey, privateKey, networkId);
+  if (Scalar.equal(kPrime, 0n)) throw Error('sign: derived nonce is 0');
+  let { x: r, y: ry } = Group.scale(Group.one, kPrime);
```
jspada, line 63: suggest to call this `rx` instead

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+  let isEven = !(ry & 1n);
+  let k = isEven ? kPrime : Scalar.negate(kPrime);
+  let e = hashMessage(message, publicKey, r, networkId);
+  let s = Scalar.add(k, Scalar.mul(e, privateKey));
+  return { r, s };
```
jspada, line 68: looks good.

src/provable/binable.ts
```diff
@@ -175,3 +116,22 @@ function bytesToBits(bytes: number[]) {
     })
     .flat();
 }
+
```
jspada, line 119: This could use some comments describing what it does, preconditions and postconditions

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+
+function signFieldElement(
+  message: Field,
+  privateKey: PrivateKey,
+  networkId: NetworkId
+) {
+  return sign({ fields: [message] }, privateKey, networkId);
+}
+
+function sign(
```
jspada, line 55: This interface should be commented and documented.

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (Scalar.equal(kPrime, 0n)) throw Error('sign: derived nonce is 0');
+  let { x: r, y: ry } = Group.scale(Group.one, kPrime);
+  let isEven = !(ry & 1n);
+  let k = isEven ? kPrime : Scalar.negate(kPrime);
+  let e = hashMessage(message, publicKey, r, networkId);
+  let s = Scalar.add(k, Scalar.mul(e, privateKey));
+  return { r, s };
+}
+
+function deriveNonce(
```
jspada, line 71: Also, this should be commented and described

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,109 @@
...
+  let inputBits = packedInput.map(Field.toBits).flat();
+  let inputBytes = bitsToBytes(inputBits);
+  let bytes = blake2b(Uint8Array.from(inputBytes), undefined, 32);
+  // drop the top two bits to convert into a scalar field element
+  // (creates negligible bias because q = 2^254 + eps, eps << q)
+  bytes[bytes.length - 1] &= 0x3f;
+  return Scalar.fromBytes([...bytes]);
+}
+
+function hashMessage(
```
jspada, line 95: Comment this function and the format of what it's hashing.

src/provable/curve-bigint.ts
```diff
@@ -9,20 +12,42 @@ import {
...
+const Group = {
+  toProjective({ x, y }: Group): GroupProjective {
+    return Pallas.ofAffine({ x, y, infinity: false });
+  },
+  fromProjective(point: GroupProjective): Group {
+    let { x, y, infinity } = Pallas.toAffine(point);
+    if (infinity) throw Error('Group.fromProjective: point is infinity');
+    return { x, y };
+  },
+  get one(): Group {
```
jspada, line 39: Suggest calling this `generator_mina` because when we update to the official Pasta curves it will mean there's another generator called `generator_pasta`

jspada (commented):
Nice job with this!

This is just my first pass.  This is security sensitive, so I'll do another later.

Meanwhile, I've left a few comments.  I recommend in particular to implement signature verification and then add a bunch of tests.

We have a standard set test vectors for signatures that you should use: https://github.com/o1-labs/proof-systems/blob/d7aa2d68d1b421a39c9ce1be39a5d936026090b2/signer/tests/signer.rs#L125

These tests cover payments and delegations on both the mainnet and testnet, i.e. Legacy ones.  For the Chunked I guess you have the freedom to create your own! :-)

package.json
mitschabaude: Good question! To me it makes sense to do this.

src/provable/base58.ts
mitschabaude: I added a comment!

mitschabaude: It's a JSOO thing

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,207 @@
...
+  signature: Signature,
+  message: HashInput,
+  publicKey: PublicKey,
+  networkId: NetworkId
+) {
+  let { r, s } = signature;
+  let pk = PublicKey.toGroup(publicKey);
+  let e = hashMessage(message, pk, r, networkId);
+  let { scale, one, sub } = Pallas;
+  let R = sub(scale(one, s), scale(Group.toProjective(pk), e));
```
jspada, line 199: Check $R \ne \infty$

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,207 @@
...
+ * Verifies a signature created by {@link sign}, returns `true` if (and only if) the signature is valid.
+ *
+ * @see {@link https://github.com/MinaProtocol/mina/blob/develop/docs/specs/signatures/description.md detailed spec of the algorithm}
+ *
+ * In contrast to the spec above, this uses the "chunked" style of hash input packing, implemented in {@link packToFields}.
+ *
+ * @param publicKey the public key has to be passed in as a compressed {@link PublicKey}.
+ * It can be created from a base58 string with {@link PublicKey.fromBase58}.
+ */
+function verify(
```
jspada, line 189: Baring question about infinity check below, this implementation looks correct to me.

src/mina-signer/signature.ts
```diff
@@ -0,0 +1,207 @@
...
+  let pk = PublicKey.toGroup(publicKey);
+  let e = hashMessage(message, pk, r, networkId);
+  let { scale, one, sub } = Pallas;
+  let R = sub(scale(one, s), scale(Group.toProjective(pk), e));
+  try {
+    let { x: rx, y: ry } = Group.fromProjective(R);
+    let isEven = !(ry & 1n);
+    return isEven && Field.equal(rx, r);
+  } catch {
+    return false;
```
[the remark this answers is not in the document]
jspada: If it's doing that check, then I think it's good to comment it explicitly.

src/mina-signer/signature.unit-test.ts
```diff
@@ -0,0 +1,103 @@
...
+  let okMainnetMainnet = verify(sigMain, msg, pk, 'mainnet');
+  expect(okTestnetTestnet).toEqual(true);
+  expect(okMainnetTestnet).toEqual(false);
+  expect(okTestnetMainnet).toEqual(false);
+  expect(okMainnetMainnet).toEqual(true);
+}
+
+// check with different random private keys
+for (let i = 0; i < 10; i++) {
+  let key = PrivateKey.random();
```
jspada, line 59: Is this using a seeded RNG so that the tests are reproducible?

src/mina-signer/consistency.unit-test.ts
```diff
@@ -178,6 +193,49 @@ let fullCommitment = hashWithPrefix(prefixes.accountUpdateCons, [
 ]);
 expect(fullCommitment).toEqual(ocamlCommitments.fullCommitment.toBigInt());
 
+// private key to/from base58
```
jspada, line 196: It would be nice to serialize these tests to JSON or some format and have this code deserialise them and perform the tests.  This way other signer implementations can use them for tests also.  But that seems like a lot of extra work and it's a good first step to have tests in the first place.  So I suggest this just to note the idea for future.

src/mina-signer/consistency.unit-test.ts
```diff
@@ -178,6 +193,49 @@ let fullCommitment = hashWithPrefix(prefixes.accountUpdateCons, [
...
+// signature
+let signature = signFieldElement(fullCommitment, feePayerKey, 'testnet');
+let signatureBase58 = Signature.toBase58(signature);
+let signatureOcaml = Ledger.signFieldElement(
+  ocamlCommitments.fullCommitment,
+  feePayerKeySnarky
+);
+expect(signatureBase58).toEqual(signatureOcaml);
+
+let isValid = verifyFieldElement(
```
jspada, line 212: Would be good to have at least one valid `mainnet` testcase

jspada (approved):
Approved with some minor suggestions and questions.

src/mina-signer/signature.unit-test.ts
mitschabaude: no, uses different random numbers every time

src/mina-signer/signature.ts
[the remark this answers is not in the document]
mitschabaude: yes, it catches it -- I added comments

src/mina-signer/consistency.unit-test.ts
mitschabaude: that's a good point! I made an issue: https://github.com/o1-labs/snarkyjs/issues/634
It shouldn't be that hard, let's revisit when (if) we implement chunked signatures in Rust?

src/mina-signer/signature.unit-test.ts
mitschabaude: How about this: since we don't have a seeded RNG readily available, I'm just going to hard-code 10 random private keys, to make the tests reproducible

src/mina-signer/consistency.unit-test.ts
mitschabaude: Good point! I exposed them from OCaml now: https://github.com/MinaProtocol/mina/pull/12337

src/mina-signer/signature.unit-test.ts
jspada: sgtm!

src/mina-signer/consistency.unit-test.ts
jspada: deal

PR #618, opened 2022-12-01
check for byte size instead of char length

Trivo25:
closes https://github.com/o1-labs/snarkyjs/issues/616


mitschabaude (approved):
Thanks!

PR #624, opened 2022-12-02
Prover bug

### LINKED ISSUE #498: Proving bug in Dex
This happens when creating a proof for `Dex.redeemLiquidity()`

In the Pickles wrap prover, the `compute_witness` step (in `Plonk_constraint_system`) fails when it tries to access `external_vars` at out-of-bounds indices.

The fact that this is the wrap prover makes the issue more mysterious because we wouldn't expect the wrap circuit to be influenced by a badly written snarkyjs circuit (which, for example, created less witnesses than expected or similar)

mitschabaude:
- closes #498 
- Minimal fix for prover bug
- Makes `dex/run.ts` fully work with proofs enabled
  - adds happy path script for DEX with proofs
  - removes the simpler, now redundant `token_with_proofs.ts` from integration tests

### DIFF
--- a/src/examples/zkapps/dex/happy-path-with-proofs.ts
+++ b/src/examples/zkapps/dex/happy-path-with-proofs.ts
@@ -9,7 +9,10 @@ let doProofs = true;
 tic('Happy path with proofs');
 console.log();
 
-let Local = Mina.LocalBlockchain({ proofsEnabled: doProofs });
+let Local = Mina.LocalBlockchain({
+  proofsEnabled: doProofs,
+  enforceTransactionLimits: false,
+});
 Mina.setActiveInstance(Local);
 let accountFee = Mina.accountCreationFee();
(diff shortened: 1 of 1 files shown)


src/examples/zkapps/dex/happy-path-with-proofs.ts
```diff
@@ -0,0 +1,118 @@
...
+import { expect } from 'expect';
+import { tic, toc } from '../tictoc.js';
+
+await isReady;
+let doProofs = true;
+
+tic('Happy path with proofs');
+console.log();
+
+let Local = Mina.LocalBlockchain({ proofsEnabled: doProofs });
```
Trivo25, line 12: I noticed most token examples exceed the new account update limits, so just FYI if this also is the case here we might have to add `enforceTransactionLimits: false`

Trivo25 (approved):
awesome work!!!

PR #628, opened 2022-12-04
fix `fetchEvents`

Trivo25:
closes https://github.com/o1-labs/snarkyjs/issues/627

### DIFF
--- a/src/lib/zkapp.ts
+++ b/src/lib/zkapp.ts
@@ -1001,11 +1001,11 @@ super.init();
         // if there are multiple events we have to use the index event[0] to find the exact event type
         let type = sortedEventTypes[event[0]];
         // all other elements of the array are values used to construct the original object, we can drop the first value since its just an index
-        event.shift();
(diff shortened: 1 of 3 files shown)


src/lib/zkapp.ts
```diff
@@ -1001,11 +1001,11 @@ super.init();
         // if there are multiple events we have to use the index event[0] to find the exact event type
         let type = sortedEventTypes[event[0]];
         // all other elements of the array are values used to construct the original object, we can drop the first value since its just an index
-        event.shift();
+        let eventProps = event.slice(1);
```
mitschabaude, line 1004: isn't this equivalent? 😮 
I'd love to understand what the issue here was and how this fixes it

Trivo25: The bug appeared when we called `fetchEvents` more than once, because `shift` manipulates the original array, `slice` doesn't. 
So every time we called `fetchEvents`, we would drop an element of the original array until the array would be empty

PR #637, opened 2022-12-07
Don't increment nonces on accounts that are the fee payer

### LINKED ISSUE #271: Don't increment nonces on accounts that are the fee payer
This is an important fix because with some upcoming changes on the protocol side, transactions will likely fail (or partially fail, which is even worse) if a fee payer is present in the other parties and also increments its nonce in that other party.

Here's how we could handle creation of a signed Party:

* see if there's a current transaction that has a fee payer
* if yes, look if our new party is the same account
* if yes, useFullCommitment. If no, incrementNonce

And here's how we should handle adding a fee payer to an existing transaction:
(this needs to happen in mina-signer as well! cc @MartinMinkov)

* see if the fee payer account is already present in other parties
* if yes, see if that other party increments its nonce
* if yes, throw an error.
* add the fee payer as usual

We also have to adapt the zkapp CLI, because in there we manually increment the nonce on the zkapp account which is also the fee payer. We should useFullCommitment instead.

mitschabaude:
- closes #271
- tweak the implementation of `AccountUpdate.createSigned`, to use `create` and `requireSignature()`. Deprecate the API where a private key is passed to `createSigned()` -- we want to move away from private keys in transactions. Instead, add another function signature that takes a public key.
- breaking change to the circuit used in `requireSignature()` -- this will break many deployed contracts and thus require a minor version bump

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,15 +17,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/o1-labs/snarkyjs/compare/d880bd6e...HEAD)
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


MartinMinkov (approved):
Awesome change! Everything checks out on my end :D

PR #644, opened 2022-12-12
profile times for integration tests

Trivo25:
- profiling for integration test times in CI 
	- adds the result to the summary page of each CI run, see https://github.com/o1-labs/snarkyjs/actions/runs/3677985581/attempts/1#summary-10042318738
	- right now, only the integration tests as a whole are being profiled, but if we want we can also profile each test case/transaction
	- can be extended if desired

closes https://github.com/o1-labs/snarkyjs/issues/604

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           npm ci
           npm run build:node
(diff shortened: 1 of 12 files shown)


MartinMinkov (approved):
looks great to me! :D

PR #653, opened 2022-12-14
remote integration test

### LINKED ISSUE #535: Connect to testnet in CI to guard against incompatibility

Trivo25:
Connects a basic zkApp contract to the berkeley testnet 

TODO: same thing in the mina repo

SnarkyJS part for https://github.com/MinaProtocol/mina/issues/11248

closes #535

### DIFF
--- a/src/examples/zkapps/hello_world/run_berkeley.ts
+++ b/src/examples/zkapps/hello_world/run_berkeley.ts
@@ -1,0 +1,79 @@
+// live Berkeley integration test
+
+import { chownSync } from 'fs';
+import {
+  Field,
+  state,
+  State,
+  method,
+  PrivateKey,
+  SmartContract,
+  Mina,
+  AccountUpdate,
+  isReady,
+  shutdown,
+  DeployArgs,
+  fetchAccount,
+} from 'snarkyjs';
+import { adminPrivateKey, HelloWorld } from './hello_world.js';
+
+await isReady;
+
+let Berkeley = Mina.Network('https://proxy.berkeley.minaexplorer.com/graphql');
+Mina.setActiveInstance(Berkeley);
+
+let feePayerKey = PrivateKey.fromBase58(
+  'EKEr6V6V5VuyxUvkS8Ugye58ePQq6e8mKKXRvAMSrZKAY7vch8Kd'
+);
+let response = await fetchAccount({ publicKey: feePayerKey.toPublicKey() });
+if (response.error) throw Error(response.error.statusText);
+let { nonce, balance } = response.account;
+console.log(`Using fee payer account with nonce ${nonce}, balance ${balance}`);
+
+let zkappKey = PrivateKey.random();
+let zkappAddress = zkappKey.toPublicKey();
+
+let transactionFee = 100_000_000;
+
+console.log('Compiling smart contract...');
+let { verificationKey } = await HelloWorld.compile();
+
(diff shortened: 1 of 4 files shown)


mitschabaude:
cool! arguably it's a separate issue, which we had already tracked: #535

src/lib/mina.ts
```diff
@@ -675,7 +675,7 @@ function Network(graphqlEndpoint: string): Mina {
             } else if (maxAttempts && attempts === maxAttempts) {
               return reject(
                 new Error(
-                  `Exceeded max attempts. TransactionId: ${txId}, attempts: ${attempts}`
+                  `Exceeded max attempts. TransactionId: ${txId}, attempts: ${attempts}, last received status: ${res}`
```
nicc, line 678: Nice

nicc (approved):
TL;DR: looks good to me. 

Just writing out what I understand to be happening here...
- init the HelloWorld app
- check if it’s deployed by fetching some state
- deploy it if not
- check that we can fetch accounts
- check that we can update some state
- check that the state has been updated by fetching it again

We’re not making test assertions here, just running a JS script that will error if there’s a problem. So we’re not running any test frameworks and will rely on the exit code of the script. Seems like a very reasonable choice in this context.

The only feedback that comes to mind is that `faucet()` could be a useful utility for other tests and I wonder if its existence will be known to future contributors. Two questions: do you think it's reuasable? And is there a place for test utils?

src/examples/zkapps/hello_world/run_berkeley.ts
```diff
@@ -0,0 +1,129 @@
...
+  let transaction = await Mina.transaction(
+    { feePayerKey, fee: transactionFee },
+    () => {
+      AccountUpdate.fundNewAccount(feePayerKey);
+      zkapp.deploy({ zkappKey, verificationKey });
+    }
+  );
+
+  console.log('Sending the transaction..');
+  await (await transaction.send()).wait();
```
mitschabaude, line 59: wow, really powerful to have wait() here

Trivo25:
> The only feedback that comes to mind is that faucet() could be a useful utility for other tests and I wonder if its existence will be known to future contributors. Two questions: do you think it's reuasable? And is there a place for test utils?

Very good point. I feel like we can now create a dedicated test lib file, as there are a handful of helper functions we use (tictoc, profiler, and now the faucet/waiting for tx). Will add that!

mitschabaude (approved):
awesome! I hope it passes :D

src/examples/zkapps/hello_world/run_berkeley.ts
```diff
@@ -0,0 +1,129 @@
...
+let transactionFee = 100_000_000;
+
+console.log('Compiling smart contract..');
+let { verificationKey } = await HelloWorld.compile();
+
+let zkapp = new HelloWorld(zkappAddress);
+let x = await zkapp.x.fetch();
+let isDeployed = x?.equals(0).not().toBoolean() ?? false;
+
+if (!isDeployed) {
```
mitschabaude, line 47: nit: we don't need this check here. there's no way the random public key can already have a deployed zkapp on it

mitschabaude:
> QQ: Do we want to introduce a Testing namespace that supplies those helper functions to the outside world?

I'd say no for now, because the testing helpers will use node JS APIs, while snarkyjs has to be web compatible as well.

PR #655, opened 2022-12-15
Renames sequence events to actions

mitschabaude:
companion of https://github.com/MinaProtocol/mina/pull/12367


src/lib/account_update.ts
```diff
@@ -360,7 +360,7 @@ interface Body extends AccountUpdateBody {
    *
    * [Check out our documentation about Actions!](https://docs.minaprotocol.com/zkapps/advanced-snarkyjs/actions-and-reducer)
    */
-  sequenceEvents: Events;
+  actions: Events;
```
mitschabaude, line 363: would be nice to change the comment above this!

src/lib/account_update.ts
```diff
@@ -585,14 +585,14 @@ class AccountUpdate implements Types.AccountUpdate {
   network: Precondition.Network;
   children: {
     callsType:
-      | { type: 'None' }
-      | { type: 'Witness' }
-      | { type: 'Equals'; value: Field };
+    | { type: 'None' }
```
mitschabaude, line 588: we should revert all this autoformatting

mitschabaude (commented):
LGTM with some comments (someone else needs to approve since I opened this PR)

EDIT: However, I think there is a `SequenceEvents` module / type on the snarkyjs side that we should still rename

Trivo25 (approved):

bkase:
I figured the other SequenceEvents name internally could be handled in another PR

src/lib/zkapp.ts
```diff
@@ -1306,7 +1306,16 @@ Use the optional \`maxTransactionsWithActions\` argument to increase this number
               )
             )
           );
-      });
+      }
+
+      if (Circuit.inCheckedComputation()) {
```
Trivo25, line 1311: A recent update in the Mina repo changed how snarky behaves. It used to set `inCheckedComputation = true` after *any* contract compilation - that now changed for some reason

src/lib/int.ts
```diff
@@ -108,7 +108,7 @@ class UInt64 extends CircuitValue {
       };
     }
 
-    y_ = y_.seal();
+    y_ = y_.seal(); //Circuit.inCheckedComputation() ? y_.seal() : y_;
```
bkase, line 111: @Trivo25 is it important we keep this comment?

bkase (approved):
Looks good!

Issue #381, opened 2022-09-06
Document tx construction, account updates, compiling, proving and signing


jasongitmail:

mitschabaude:
In my judgement, this is the most important open doc issue we have currently, and I want to pursue it before testnet launch.

Here's a proposal:
* We add a new explainer section, maybe titled "Transactions, account updates and proofs", at the bottom of "How to write a zkApp"
* We give succinct descriptions of what transactions and account updates are in Mina/snarkyjs
* We introduce the `Mina.transaction` API, in generality -- i.e. without reference to any specific flow like "interacting with Auro wallet via Mina provider" or "running a unit test with Jest". (These specialized use cases should live in later sections IMO.)
* We explain how account updates are typically created under the hood by more high-level methods, and we give a concrete example: creating a payment with the `send` API
* We explain how calling a zkApp creates an account update under the hood (which needs a proof for authorization)
* We show how to use the `Mina.transaction` output to create a proof, and how you have to compile the SmartContract first
* We hint at how the finished transaction would be used in a web app -- passed to the wallet via Mina provider and signed there. (When it exists, we would link to the Mina provider docs here.) We also explain how a transaction can be signed and sent with snarkyjs alone (hint at use cases: testing, server-side/automated interaction with Mina)

mitschabaude:
Further thing that came up:
we should document how to fetch accounts and explain the caching that we do in some places. `Mina.fetchAccount` vs `Mina.getAccount`, `zkApp.x.fetch()` maybe

mitschabaude:
More stuff that came up: it should be documented that the transaction code runs multiple times, so using mutated global state as method inputs is not supported

PR #656, opened 2022-12-16
Tweak send API with example & fix

mitschabaude:
This came out of writing docs for payments (#381)
* adapts the `AccountUpdate.send()` API so that the `to` field can also be a smart contract. (this saves us from creating an additional account update, if we pass in `this` for example)
* clean up the `simple_zkapp_payment.ts` example and ensure it covers all cases of payments and matches the docs, and works
* making the example work uncovered a subtle bug, which is fixed here: the nonce precondition _value_ has to be the default if we don't use a nonce precondition, because otherwise the hash created from the transaction (which only sees `null` for the precondition) won't match the hash computed in the prover

### DIFF
--- a/src/examples/zkapps/simple_zkapp_payment.ts
+++ b/src/examples/zkapps/simple_zkapp_payment.ts
@@ -5,7 +5,6 @@ import {
   AccountUpdate,
   PrivateKey,
   SmartContract,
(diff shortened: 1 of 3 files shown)


Trivo25 (approved):
nice :)

PR #658, opened 2022-12-19
Verification key hashes in account updates

mitschabaude:
companion of https://github.com/MinaProtocol/mina/pull/12380


MartinMinkov (approved):
Killer work as always :D

PR #660, opened 2022-12-20
Mina-signer: Legacy payment signatures

### LINKED ISSUE #632: Mina-signer: sign legacy/mainnet tx

mitschabaude:
closes #632

- re-implements signing of payments, stake delegations and strings in JS
- tested against reference signatures created with the OCaml mina-signer (`test_signatures.js`)
- to support legacy signatures, we also implement Poseidon with legacy parameters
- Legacy Poseidon is tested against test vectors created with the Rust tool


Trivo25 (approved):
lgtm

src/mina-signer/sign-legacy.ts
```diff
@@ -0,0 +1,248 @@
...
+
+function signUserCommand(
+  command: UserCommand,
+  privateKeyBase58: string,
+  networkId: NetworkId
+) {
+  let input = toInputLegacy(command);
+  let privateKey = PrivateKey.fromBase58(privateKeyBase58);
+  let signature = signLegacy(input, privateKey, networkId);
+  return Signature.toBase58(signature);
```
MartinMinkov, line 45: Does `Signaure.toBase58` have a chance to panic/crash? Do we handle that further down somewhere?

MartinMinkov (approved):
This looks good to me! Only thing I would caution on is if any of the cryptography code has a point of crashing that isn't handled. I leave it up to you to confirm if everything is handled on that end :D

src/mina-signer/signature.ts
```diff
@@ -204,3 +215,90 @@ function verify(
     return false;
   }
 }
+
+// legacy signatures
+
+/**
+ * Same as {@link sign}, but using the "legacy" style of hash input packing.
+ */
+function signLegacy(
```
jspada, line 224: This feels like a lot of code duplication, since the signing algorithm between legacy and kimchi versions are equivalent.  They only call different versions of serialization, which could be encapsulated at a serialization layer.   This is just a note-- not saying it needs to be changed for this PR, but you might want to think about it.

src/mina-signer/signature.ts
```diff
@@ -204,3 +215,90 @@ function verify(
...
+  let k = Field.isEven(ry) ? kPrime : Scalar.negate(kPrime);
+  let e = hashMessageLegacy(message, publicKey, rx, networkId);
+  let s = Scalar.add(k, Scalar.mul(e, privateKey));
+  return { r: rx, s };
+}
+
+/**
+ * Same as {@link verify}, but using the "legacy" style of hash input packing.
+ */
+function verifyLegacy(
```
jspada, line 242: Same comment above. You might want to refactor to keep a single implementation of the signing and verifying algorithms.

src/mina-signer/signature.ts
```diff
@@ -204,3 +215,90 @@ function verify(
...
+
+/**
+ * Same as {@link sign}, but using the "legacy" style of hash input packing.
+ */
+function signLegacy(
+  message: HashInputLegacy,
+  privateKey: PrivateKey,
+  networkId: NetworkId
+): Signature {
+  let publicKey = Group.scale(Group.generatorMina, privateKey);
```
jspada, line 229: I'm guessing TS doesn't have some sort of "safe/secure string" for storing secrets so that they can be wiped from memory after use, but it would be nice.

src/mina-signer/signature.ts
```diff
@@ -204,3 +215,90 @@ function verify(
...
+  let kPrime = deriveNonceLegacy(message, publicKey, privateKey, networkId);
+  if (Scalar.equal(kPrime, 0n)) throw Error('sign: derived nonce is 0');
+  let { x: rx, y: ry } = Group.scale(Group.generatorMina, kPrime);
+  let k = Field.isEven(ry) ? kPrime : Scalar.negate(kPrime);
+  let e = hashMessageLegacy(message, publicKey, rx, networkId);
+  let s = Scalar.add(k, Scalar.mul(e, privateKey));
+  return { r: rx, s };
```
jspada, line 236: Looks good to me.

src/mina-signer/signature.ts
```diff
@@ -204,3 +215,90 @@ function verify(
...
+  let { scale, one, sub } = Pallas;
+  let R = sub(scale(one, s), scale(Group.toProjective(pk), e));
+  try {
+    // if `R` is infinity, Group.fromProjective throws an error, so `verify` returns false
+    let { x: rx, y: ry } = Group.fromProjective(R);
+    return Field.isEven(ry) && Field.equal(rx, r);
+  } catch {
+    return false;
+  }
+}
```
jspada, line 260: Looks good to me.

src/mina-signer/test-vectors/legacySignatures.ts
```diff
@@ -0,0 +1,206 @@
...
+    },
+  },
+  {
+    body: { delegator: keypair.publicKey, newDelegate },
+    common: {
+      fee: '10',
+      feePayer: keypair.publicKey,
+      nonce: '1000',
+      validUntil: '8192',
+      memo: 'enough stake to kill a vampire',
```
jspada, line 66: hahha. nice one.

src/mina-signer/test-vectors/legacySignatures.ts
```diff
@@ -0,0 +1,206 @@
+import { DelegationJson, PaymentJson } from '../sign-legacy.js';
```
jspada, line 1: Are any of these imported from https://github.com/o1-labs/proof-systems/blob/226de4aeb11b8814327ab832e4fccdce5585f473/signer/tests/signer.rs#L63
?

jspada (approved):
Looks good to me.  Left some comments and suggestions.

src/mina-signer/signature.ts
mitschabaude: yeah I thought the same! but then I thought that this code is never ever supposed to change anyway, since we want to support the old signatures indefinitely. still would be nice to abstract out, I agree

src/mina-signer/signature.ts
mitschabaude: I think this is supposed to be handled by browsers, as they sandbox code that runs, say, in a browser extension (as this code would typically), and ensure it's not vulnerable to side channel attacks

src/mina-signer/sign-legacy.ts
```diff
@@ -0,0 +1,248 @@
...
+) {
+  return verifyUserCommand(
+    delegationFromJson(delegation),
+    signatureBase58,
+    publicKeyBase58,
+    networkId
+  );
+}
+
+function verifyUserCommand(
```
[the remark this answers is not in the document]
mitschabaude: I think invalid arguments like invalid base58 strings will just cause an error! Do you think this should change?

src/mina-signer/test-vectors/legacySignatures.ts
mitschabaude: no - they come from a test that existed on the OCaml side, which was used to check consistency of the OCaml and JS versions of mina-signer: https://github.com/MinaProtocol/mina/blob/develop/scripts/compare_test_signatures.sh

I just ran the script that generated those test signatures and copied the output in this file

src/mina-signer/test-vectors/legacySignatures.ts
mitschabaude: credit to Paul I believe for generating these test cases :D

src/mina-signer/sign-legacy.ts
mitschabaude: I think that toBase58, as opposed to fromBase58, can't crash because the output from `signLegacy` will always be a valid signature. But I'll check!

src/mina-signer/sign-legacy.ts
```diff
@@ -0,0 +1,248 @@
...
+  return signUserCommand(command, privateKeyBase58, networkId);
+}
+
+function signUserCommand(
+  command: UserCommand,
+  privateKeyBase58: string,
+  networkId: NetworkId
+) {
+  let input = toInputLegacy(command);
+  let privateKey = PrivateKey.fromBase58(privateKeyBase58);
```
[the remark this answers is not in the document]
mitschabaude: this would crash in that case

PR #666, opened 2023-01-03
Mina-signer: Transaction hash

### LINKED ISSUE #667: Mina-signer: implement transaction hash

mitschabaude:
closes #667

This PR adds three things:
* A file `transaction-hash.ts` which implements `hashPayment` and `hashStakeDelegeation`. Both functions use the old "mainnet" behaviour of hashing by default, but can be made to use berkeley-style hashing with the optional `{ berkeley: true }` argument. Therefore, this PR removes the need to publish a separate version of `mina-signer` for mainnet compatibility until the hardfork.
  * Transaction hashing is conceptually simple: First, the input is serialized into an array of bytes using `bin_prot` logic. Second, the array of bytes is fed to the blake2b hash function, and the result is base58check'd. The devil lies in the details of the `bin_prot` serialization.
* A new set of unit tests which check all kinds of transaction hashing, plus sub-algorithms like the `bin_prot` serialization, against newly exposed OCaml versions.
* Implementations of some `bin_prot` primitives like serialization of integers, strings and records. Also, `Binable` is refactored somewhat because its previous implementation relied on statically known sizes of types, which is not available in general (for example, strings should be (de)serializable without statically knowing their size). The new formulation instead introduces a `readBytes` function that has to be implemented by every type of binable, which not only deserializes the type from a given array of bytes at a given offset, but also returns the new offset, so that higher-level binables know where to continue.

### DIFF
--- a/src/snarky.d.ts
+++ b/src/snarky.d.ts
@@ -1225,6 +1225,7 @@ declare const Test: {
     examplePayment(): string;
     serializePayment(payment: string): { data: Uint8Array };
     serializeCommon(common: string): { data: Uint8Array };
+    hashPayment(payment: string): string;
   };
 };
 
--- a/src/snarky.d.ts
+++ b/src/snarky.d.ts
@@ -1224,8 +1224,10 @@ declare const Test: {
   transactionHash: {
     examplePayment(): string;
     serializePayment(payment: string): { data: Uint8Array };
+    serializePaymentV1(payment: string): string;
     serializeCommon(common: string): { data: Uint8Array };
     hashPayment(payment: string): string;
+    hashPaymentV1(payment: string): string;
   };
 };
 
--- a/src/snarky.d.ts
+++ b/src/snarky.d.ts
@@ -419,6 +419,10 @@ declare class Field {
   static toInput(x: Field): { fields: Field[] };
   static toBytes(x: Field): number[];
   static fromBytes(bytes: number[]): Field;
+  static fromBytesInternal(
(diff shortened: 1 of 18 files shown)


MartinMinkov (approved):
Awesome work!

src/snarky.d.ts
```diff
@@ -1209,6 +1220,17 @@ declare class Ledger {
   };
 }
 
+declare const Test: {
```
bkase, line 1223: Can we also use this to test the full signature logic — ie expose signPayment and signDelegstion and verifySignsture(s) and ensure we can verify signatures created with ocaml with js and vice versa

bkase (changes requested):

src/snarky.d.ts
mitschabaude: We already test that we can recreate and verify OCaml signatures, however with a different architecture: Currently we have hard-coded signature test vectors for payments / stake delegations where the signatures where created with the OCaml code (this was part of an existing test, so I just took that).

Assuming that we never change how how payments and delegations are signed in OCaml (which, AFAIU, is the goal), hard coding the reference signatures is equivalent to creating them in OCaml on the fly (and it's arguably more light-weight). However, if they should change at one point, the existing test won't detect the newly introduced inconsistency.

bkase: Ah okay! Yes we need to maintain backwards compatibility so those signatures being hard-coded is actually better I think!

PR #679, opened 2023-01-06
Mina-signer: Implement base58 in JS

### LINKED ISSUE #677: Mina-signer: implement base58 check
Remove the OCaml dependency for base58 check. Instead, either use a third-party library (if one exists) or implement base58 check on top of existing building blocks

mitschabaude:
closes #677 

Implements base58 check in JS, which was the last remaining OCaml dependency of the mina-signer code base. Base58 check is implemented using the following ingredients:
* a `changeBase` function (which was already there), to map an array of digits in one base into an array of digits in another base. In this case we go from base 256 (bytes) to base 58, and back.
* for computing the checksum, a sha256 implementation, which we take from the [js-sha256](https://www.npmjs.com/package/js-sha256) library: (700,000 downloads / week, has no runtime dependencies of its own)
  * we'd prefer to use the built-in [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest), but it is async, which would be too disruptive to our API surface

### DIFF
--- a/src/js_crypto/bigint-helpers.ts
+++ b/src/js_crypto/bigint-helpers.ts
@@ -41,7 +41,7 @@ function changeBase(digits: bigint[], base: bigint, newBase: bigint) {
  * but that is for performance reasons.
  *
  * we'll explain it for `fromBase`. this function is about taking an array of digits
- * [x0, ..., xn]
+ * `[x0, ..., xn]`
  * and returning the integer (bigint) that has those digits in the given `base`:
  * ```
  * let x = x0 + x1*base + x2*base**2 + ... + xn*base**n
@@ -58,8 +58,8 @@ function changeBase(digits: bigint[], base: bigint, newBase: bigint) {
  * ```
  *
  * in the ith step, `p = base**i` which is multiplied with `xi` and added to the sum.
- * however, note that this algorithm is O(n^2): let `l = log2(base)`. the base power `p` is a bigint of bit length `i*l`,
- * which is multiplied by a "small" number `xi` (length l), which takes O(i) time in every step.
+ * however, note that this algorithm is `O(n^2)`: let `l = log2(base)`. the base power `p` is a bigint of bit length `i*l`,
+ * which is multiplied by a "small" number `xi` (length l), which takes `O(i)` time in every step.
  * since this is done for `i = 0,...,n`, we end up with an `O(n^2)` algorithm.
  *
  * HOWEVER, it turns out that there are fast multiplication algorithms, and JS bigints have them built in!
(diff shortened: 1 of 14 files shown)


src/js_crypto/bigint-helpers.ts
```diff
@@ -0,0 +1,93 @@
+export { changeBase, bytesToBigInt, bigIntToBytes };
```
mitschabaude, line 1: this new file collects from helpers which were scattered through the code base - nothing here is new

Trivo25 (approved):
lgtm

src/js_crypto/bigint-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+    throw Error(`bigIntToBytes: input does not fit in ${length} bytes`);
+  }
+  return bytes;
+}
+
+function changeBase(digits: bigint[], base: bigint, newBase: bigint) {
+  // 1. accumulate digits into one gigantic bigint `x`
+  let x = fromBase(digits, base);
+  // 2. compute new digits from `x`
+  let newDigits = toBase(x, newBase);
```
jspada, line 35: why not just `return toBase(fromBase(digits, base), newBase);`

src/js_crypto/bigint-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+  console.assert(digits.length === 1);
+  let [digit] = digits;
+  return digit;
+}
+
+function toBase(x: bigint, base: bigint) {
+  // compute powers base, base^2, base^4, ..., base^(2^k)
+  // with largest k s.t. base^(2^k) < x
+  let basePowers = [];
+  for (let power = base; power < x; power **= 2n) {
```
jspada, line 70: Seems like this and above instance could be factored into a nice helper.

src/js_crypto/bigint-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function changeBase(digits: bigint[], base: bigint, newBase: bigint) {
+  // 1. accumulate digits into one gigantic bigint `x`
+  let x = fromBase(digits, base);
+  // 2. compute new digits from `x`
+  let newDigits = toBase(x, newBase);
+  return newDigits;
+}
+
+// NOTE: toBase / fromBase are so complicated for performance reasons
```
jspada, line 39: Would be good to give some details -- comment why and how this performs better.

src/js_crypto/bigint-helpers.ts
```diff
@@ -0,0 +1,93 @@
...
+
+function fromBase(digits: bigint[], base: bigint) {
```
jspada, line 41: When when `digits.length == 1` I think `digits` may be a zero-length array and the assertion below will fail.  Not sure if this is a case you wanted to allow/(have fail)?

jspada, line 41: Also may have to fail on `base <= 0`?

jspada (approved):
Nice code!  Approved with some feedback and suggestions.  Why not use an already existing base58 implementation?

src/js_crypto/bigint-helpers.ts
mitschabaude: I could, but I feel that the two steps that are involved would be less clear

src/js_crypto/bigint-helpers.ts
mitschabaude: yeah it's similar.. but not quite the same 🤔 gonna leave it separate for now

src/js_crypto/bigint-helpers.ts
mitschabaude: I checked! In that case, the base powers array stays empty so the big for loop doesn't run, and we'll end up with the same length 1 array as in the beginning, which doesn't fail the assertion

src/provable/base58.ts
```diff
@@ -1,8 +1,78 @@
...
+      `fromBase58Check: input version byte ${versionByte} does not match encoded version byte ${originalBytes[0]}`
+    );
+  // return result
+  return originalBytes.slice(1);
+}
+
+function toBase58(bytes: number[] | Uint8Array) {
+  // count the leading zeroes. these get turned into leading zeroes in the output
+  let z = 0;
+  while (bytes[z] === 0) z++;
```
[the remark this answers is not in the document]
mitschabaude: unfortunately that increments it one too often :D

src/provable/base58.unit-test.ts
```diff
@@ -0,0 +1,22 @@
+import { fromBase58Check, toBase58Check } from './base58.js';
```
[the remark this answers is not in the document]
mitschabaude: done!

PR #681, opened 2023-01-09
rename assertions to long form

Trivo25:
- replaces short forms of assertions with long form (`lt` -> `lessThan`..) and deprecates old ones
- adjusts Field.lt in OCaml/bindings
- fixed unknown error in berkeley live int test (fetching tx status)

https://github.com/MinaProtocol/mina/pull/12494

closes https://github.com/o1-labs/snarkyjs/issues/68
closes https://github.com/o1-labs/snarkyjs/issues/26

gotta wait for a handful of other PRs before merging


MartinMinkov (approved):
Nice! :D

src/snarky.d.ts
```diff
@@ -155,8 +155,109 @@ declare class Field {
...
   /**
+   *
+   * Assert that this {@link Field} is lower than another Field-like value.
+   *
+   * ```ts
+   * Field(1).assertLessThan(2);
+   * ```
+   *
+   * This function can only be called inside a checked computation, like a
+   * SmartContract method, and causes it to fail if the assertion fails.
```
mitschabaude, line 209: Didn't we fix that this "can only run in a checked computation"? Should remove those comments here and elsewhere

mitschabaude: Oh ok, well I hope we can merge it to main soon

mitschabaude (approved):

PR #682, opened 2023-01-09
Make mina-signer publishable

### LINKED ISSUE #678: Mina-signer: match the old interface 1:1 and prepare publishable package
Completing this will also close #611

mitschabaude:
Closes #678

Stacked on https://github.com/o1-labs/snarkyjs/pull/666, https://github.com/o1-labs/snarkyjs/pull/679

This finishes up initial work on `mina-signer`, by matching the old interface, getting old tests to run, and setting the package up to be published


run-ci-tests.sh
```diff
@@ -24,7 +24,11 @@ case $TEST_TYPE in
     "Unit tests" )
       echo "Running unit tests";
       npm run test:unit || exit 1
-      npm run test ;;
+      npm run test || exit 1
```
bkase, line 27: Can you add !#/bin/bash and set -e to the top of this script so you don’t need to think about the || exits everywhere? One of these days we should lint for this

run-jest-tests.sh
```diff
@@ -0,0 +1,3 @@
+for f in ./src/**/*.test.ts; do
+  NODE_OPTIONS=--experimental-vm-modules npx jest $f || exit 1;
```
bkase, line 2: And here

run-unit-tests.sh
```diff
@@ -1,4 +1,9 @@
 npm run build:test
+# TODO: how to capture all tests with a single glob expression?
+for f in ./dist/node/**/**/*.unit-test.js; do
+  echo "Running $f"
+  node $f || exit 1;
```
bkase, line 5: And jere

run-unit-tests.sh
```diff
@@ -1,4 +1,9 @@
...
+done
 for f in ./dist/node/**/*.unit-test.js; do
```
bkase, line 7: These two for loops look the same to me, is there a mistake here?

src/mina-signer/MinaSigner.ts
```diff
@@ -78,13 +109,16 @@ class Client {
...
-        publicKey: key.publicKey,
-        message,
+      signature: {
+        ...signString(message, key.privateKey, this.network),
+        signer: key.publicKey,
       },
+      // TODO: returning the public key, which is an input, twice seems awfully redundant
+      // this should just be `data: message`
+      // and do we really need `signer` as part of `Signature`?
+      data: { publicKey: key.publicKey, message },
```
bkase, line 121: @MartinMinkov what are the implications of making a breaking change here? We could deprécate this and introduce a new function I suppose?

src/mina-signer/src/rosetta.ts
```diff
@@ -0,0 +1,146 @@
...
+      };
+    } else if (t.kind === 'Delegation') {
+      return {
+        common,
+        body: [
+          'Stake_delegation',
+          ['Set_delegate', { delegator: source_pk, new_delegate: receiver_pk }],
+        ],
+      };
+    } else throw Error("can't happen");
```
bkase, line 117: Mind improving this error just in case?

bkase (changes requested):

run-ci-tests.sh
mitschabaude: I did that for the snarkyjs build script and for some reason it doesn't work the way I expected (script didn't stop on an error), so I came to distrust `set -e` :D maybe I just did something wrong, but I definitely don't want to risk test failures getting through. I'll try it out here though with an artificial test failure!

run-unit-tests.sh
mitschabaude: there's a double `**/**` in the first, and one `**` in the second. I originally wanted all nested subdirectories, but bash instead just treated `**` like `*` and just gave me one level of hierarchy. For the mina-signer tests I needed to include another level.

I just looked it up and learned how to make it behave the way I want (https://unix.stackexchange.com/a/197382), so I'll try that

run-ci-tests.sh
mitschabaude: it works now, I tested it out, removed all the `exit 1`s

bkase (approved):

PR #683, opened 2023-01-09
Fix/apple silicon performance

jackryanservia:
First merge: MinaProtocol/Mina#12479

Explain your changes:
* Added detect-gpu dependency for SnarkyJS
* Fixed Apple silicon performance issue.
* SnarkyJS will now detect when running on Apple silicon and use the most efficient number of workers. Required because of the issue outlined [here](https://bugs.chromium.org/p/chromium/issues/detail?id=1228686).

Explain how you tested your changes:
* Built version that console logged all relevant information.
* Ran on all environment/machine combinations (Except M1 Ultra and M2 (but it's obviouse they will work)).

Caveats:
* If the page is rendered serverside, we can't detect the platform. So we default the number of workers to a conservative guess. There is probably a way to enforce that the detect-gpu code runs client-side (in most cases), but I don't have a ton of experience with front-end frameworks and didn't think this was a high priority.
* Mobile devices aren't accounted for yet (many of them (even if they don't run Apple silicon) will suffer from the same problem (but it's easy to add support any time)).
* Values for the number of workers are hard-coded and based on limited testing.

* Closes o1-labs/snarkyjs#491


src/chrome_bindings/getEfficientNumWorkers.js
```diff
@@ -0,0 +1,25 @@
...
+  let gpuModel = gpuTier.gpu || 'SSR';
+
+  var numWorkers =
+    {
+      'apple m1': 2,
+      'apple m1 pro': numCpus === 10 ? 3 : 2,
+      'apple m1 max': 3,
+      'apple m1 ultra': 7,
+      'apple m2': 2,
+      'SSR': ((numCpus / 6) >> 0) + 1,
```
mitschabaude, line 21: could you elaborate on that case? I'd assume that if server-side rendering prevents `getGPUTier` from running on the client, then snarkyjs won't run on the client in its entirety, so this case doesn't really matter? Or am I wrong?

mitschabaude (approved):
Excellent work!

Trivo25 (approved):
This is great :) Would you mind adding the fix to the changelog?

PR #685, opened 2023-01-11
Feature/set timing permission

mrmr1993:


mitschabaude:
triggered manual workflow run: https://github.com/o1-labs/snarkyjs/actions/runs/3891134963

mitschabaude:
actually this is probably missing `./scripts/update-snarkyjs-bindings.sh`

src/lib/account_update.ts
```diff
@@ -245,6 +246,7 @@ let Permissions = {
     setTokenSymbol: Permission.signature(),
     incrementNonce: Permission.signature(),
     setVotingFor: Permission.signature(),
+    setTiming: Permission.signature(),
```
mitschabaude, line 249: These are the permissions that get applied by default when you deploy a zkApp. I think to prevent the "malicious zkApp developer" case that you outlined, the default for `setTiming` should be "proof". Note: we changed the default for `send` to "proof" as well on `main`, even though here it's still "signature"

mitschabaude (approved):
looks good to me with one comment, pre-approving. but should be made to pass the tests still (next push should trigger CI automatically)

PR #686, opened 2023-01-11
Publish the new mina-signer

### LINKED ISSUE #611: Reduce Mina Signer below <4MB
FireFox has a file size limit for extensions of 4MB. Achieving this will allow Firefox-extensions wallets to support zkApps!

Work is underway currently that should contribute to this goal.

Tagging @bitcat365 so you can track this.

Creating an issue to track on the roadmap.

mitschabaude:
closes #611

**Merging this will publish the new version of mina-signer!**

- Tweaks the `package.json` and build artifacts to support all targets: web, node (esm) and node (cjs).
  - I confirmed that the output of `npm pack` works for all these targets and that TS types are properly resolved.
- Adds GH action that auto-publishes mina-signer when its version is bumped

### DIFF
--- a/.github/workflows/build-action.yml
+++ b/.github/workflows/build-action.yml
@@ -94,3 +94,29 @@ jobs:
           token: ${{ secrets.NPM_TOKEN }}
         env:
           INPUT_TOKEN: ${{ secrets.NPM_TOKEN }}
+
+  Release-mina-signer-on-NPM:
+    timeout-minutes: 180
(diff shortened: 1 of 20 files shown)


bkase (approved):
Looks good to me assuming you run through the QA steps Martin typically uses

MartinMinkov (approved):
Just to verify, tested broadcasting payments on all current networks. Checks out for me! :D

ymekuria (approved):

PR #689, opened 2023-01-12
Fix: this === undefined in Struct static methods

mitschabaude:
fixes a bug reported by community member

### DIFF
--- a/src/lib/zkapp.ts
+++ b/src/lib/zkapp.ts
@@ -1268,16 +1268,8 @@ Use the optional \`maxTransactionsWithActions\` argument to increase this number
         // also, for each action length, compute the new state and then pick the actual one
         let newStates = actionss.map((actions) => {
           // we generate a new witness for the state so that this doesn't break if `apply` modifies the state
-          let newState = Circuit.witness(stateType, () => {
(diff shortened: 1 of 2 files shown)


src/lib/zkapp.ts
```diff
@@ -1268,16 +1268,8 @@ Use the optional \`maxTransactionsWithActions\` argument to increase this number
         // also, for each action length, compute the new state and then pick the actual one
         let newStates = actionss.map((actions) => {
           // we generate a new witness for the state so that this doesn't break if `apply` modifies the state
-          let newState = Circuit.witness(stateType, () => {
-            // TODO: why doesn't this work without the toConstant mapping?
-            let { toFields, fromFields, toAuxiliary } = stateType;
```
mitschabaude, line 1273: the problem here was that pulling those static members out of the class is not compatible with using `this` in those static methods. js rookie mistake

mitschabaude: and it turns out wee don't need that `toConstant` logic anymore, at all, because `Circuit.witness` does that under the hood anyway

Trivo25 (approved):

PR #691, opened 2023-01-12
Fix fetchMissingData: re-fetch cached accounts

### LINKED ISSUE #430: Fix `fetchMissingData` with remote blockchain

### LINKED ISSUE #599: Make cache invalidation timeout to be configurable
[This one](https://github.com/o1-labs/snarkyjs/blob/7e3323ef638900cbf3c9f5ecc8e85597fefa5448/src/lib/fetch.ts#L293).

mitschabaude:
closes #430 (let's see if this can finally stay closed!)

this is confirmed to fix one instance of stale account data (incorrect nonce) used by `Mina.transaction`. The issue was that an account that already was fecthed once, and was therefore cached, was not re-fetched when it was marked to be fetched during `Mina.transaction`. This meant that an account that was _updated_ by an earlier transaction remained stale.

The simple solution is to fetch all accounts that are marked for fetching.

closes #599

as a  side-effect, this makes the cache invalidation timeout completely unused, therefore we remove it and close the issue of making it configurable


Trivo25 (approved):
nice :)

bkase (approved):

PR #692, opened 2023-01-13
Fix `TransactionId.wait()`: do not wait if transaction failed

mitschabaude:
has been tested as part of the work on https://github.com/o1-labs/docs2/pull/234


Trivo25 (approved):
good catch!

PR #700, opened 2023-01-18
Mina-signer: Unify and simplify API

mitschabaude:
This makes some simplifying changes to the mina-signer API:
* Return signatures as base58 strings instead of `{ field, scalar }` objects, to be consistent with graphQL / snarkyjs / usage of base58 for public and private keys
* Unify the interface of `sign*` and `verify*` functions:
```ts
signSomething(data: T, privateKey: string): Signed<T>;
verifySomething(signed: Signed<T>): boolean;

type Signed<T> = { signature: string, publicKey: string, data: T };
```
* adds missing API methods `verifyZkappCommand` and `verifyTransaction`, as counter-parts to `signZkappCommand` and `signTransaction`


src/mina-signer/tests/stake-delegation.test.ts
```diff
@@ -207,20 +208,7 @@ describe('Stake Delegation', () => {
       const mainnetClient = new Client({ network: 'mainnet' });
       const invalidMessage = mainnetClient.verifyStakeDelegation(delegation);
       expect(invalidMessage).toBeFalsy();
-    });
-
-    it('verifies a signed delegation generated by signTransaction', () => {
```
mitschabaude, line 212: removed this because it was a duplicate

jasongitmail:
We should bump the major version because the function signatures for `signMessage()` and `verifymessage()` are changing

MartinMinkov (approved):

PR #703, opened 2023-01-19
Direct call type

mitschabaude:
companion of https://github.com/MinaProtocol/mina/pull/12450


src/lib/account_update.ts
```diff
@@ -780,7 +781,7 @@ class AccountUpdate implements Types.AccountUpdate {
     layout: AccountUpdatesLayout = AccountUpdate.Layout.NoDelegation
   ) {
     makeChildAccountUpdate(this, childUpdate);
-    this.isDelegateCall = Bool(false);
+    this.body.callType = AccountUpdate.CallType.ChildrenMayUseMyToken;
```
mitschabaude, line 784: API for controlling call type will be iterated on in the next PR (which adds `Blind_call`)

mitschabaude (commented):
some comments to explain what we did here!

src/lib/account_update.ts
mitschabaude: Update: went with existing names for now, since it seems the `ChildrenMayUseMyToken` semantics weren't intended

PR #705, opened 2023-01-19
Mina-signer: sign account updates

mitschabaude:
closes https://github.com/MinaProtocol/mina/issues/12014

- signs other account updates than the fee payer, if they have the same public key and need a signature
- refactors the mina-signer/zkapp unit test to hit the case of signing multiple account updates
  - compares result with snarkyjs signing output
  - implements `Transaction.fromJSON` which was useful here and recently requested as a feature


src/mina-signer/tests/zkapp.unit-test.ts
```diff
@@ -0,0 +1,113 @@
...
+expect(() => {
+  client.signZkappCommand(
+    {
+      zkappCommand: exampleZkappCommand,
+      // @ts-ignore - fee is not defined
+      feePayer: { feePayer: publicKey, nonce: '0', memo: 'test memo' },
+    },
+    privateKey
+  );
+}).toThrow('Fee must be greater than');
```
Trivo25, line 76: nit to the error that `signZkappCommand` throws: I think it would be good to explicitly differentiate between the two cases - no fee specified and fee not high enough

Trivo25 (approved):

PR #708, opened 2023-01-30
Correct faucet owner in CHANGELOG.md

jasongitmail:
Minor change to correct faucet--provided by MF, not Mina Explorer.


Trivo25 (approved):
good catch!

PR #710, opened 2023-01-30
Mina-signer: in-SNARK verifiable Signatures

mitschabaude:
closes https://github.com/MinaProtocol/mina/issues/11327

this is about letting users sign data with mina-signer (i.e., with their wallet), such that the signature can be verified by a smart contract. in other words, we need to make sure there is a signing function in mina-signer whose signatures can be verified in-snark by snarkyjs.

- adds `signFields` and `verifyFields` to mina-signer
- modifies the existing `Signature.create` and `Signature.verify` in snarkyjs, to be compatible with `signFields`
  - note that `Signature.create` is not for in-snark usage, but `Signature.verify` is
  
Making snarkyjs signatures compatible uncovered a bug/irregularity in the previous implementation:
- after hashing the message to be signed, the algorithm reinterprets the hash (a base field element) as a scalar
- in snarkyjs, this was done with `Scalar.fromBits(hash.toBits())`
- however, it turned out that `Scalar` uses [a "shifted" representation](https://github.com/MinaProtocol/mina/blob/f6756507ff7380a691516ce02a3cf7d9d32915ae/src/lib/pickles_types/shifted_value.ml#L44) and `fromBits` just interprets the bits to already be in shifted form
- so, it can be said that the existing algorithm used a non-standard way of moving from base field to scalar
- this is fixed here, by unshifting the scalar. in the case of the in-snark verify, we do the unshifting on the group element level, leveraging the existing `Group.scale` method. it's not clear to me if there's a more efficient way to do this

### DIFF
--- a/src/lib/account_update.ts
+++ b/src/lib/account_update.ts
@@ -1877,7 +1877,11 @@ function signJsonTransaction(
       accountUpdate.authorization.proof === null
     ) {
       zkappCommand = JSON.parse(
-        Ledger.signAccountUpdate(JSON.stringify(zkappCommand), privateKey, i)
+        Ledger.signOtherAccountUpdate(
+          JSON.stringify(zkappCommand),
+          privateKey,
+          i
+        )
       );
     }
   }
--- a/src/lib/core.ts
+++ b/src/lib/core.ts
@@ -2,6 +2,8 @@ import { bytesToBigInt } from '../js_crypto/bigint-helpers.js';
 import { defineBinable } from '../provable/binable.js';
 import { sizeInBits } from '../provable/field-bigint.js';
 import { Bool, Field, Scalar, Group } from '../snarky.js';
+import { Scalar as ScalarBigint } from '../provable/curve-bigint.js';
+import { mod } from '../js_crypto/finite_field.js';
 
 export { Field, Bool, Scalar, Group };
 
@@ -71,4 +73,9 @@ That means it can't be called in a @method or similar environment, and there's n
     field: Field(x & lowBitMask),
     highBit: Bool(x >> lowBitSize === 1n),
   };
+};
+
+Scalar.fromBigInt = function (scalar: bigint) {
+  scalar = mod(scalar, ScalarBigint.modulus);
+  return Scalar.fromJSON(scalar.toString());
 };
--- a/src/lib/signature.ts
+++ b/src/lib/signature.ts
@@ -1,7 +1,11 @@
 import { Group, Field, Bool, Scalar, Ledger } from '../snarky.js';
 import { prop, CircuitValue, AnyConstructor } from './circuit_value.js';
 import { hashWithPrefix } from './hash.js';
-import { Signature as SignatureBigint } from '../mina-signer/src/signature.js';
+import {
+  deriveNonce,
+  Signature as SignatureBigint,
+} from '../mina-signer/src/signature.js';
+import { Scalar as ScalarBigint } from '../provable/curve-bigint.js';
 import { prefixes } from '../js_crypto/constants.js';
 
 // external API
@@ -194,16 +198,23 @@ class Signature extends CircuitValue {
   static create(privKey: PrivateKey, msg: Field[]): Signature {
     const publicKey = PublicKey.fromPrivateKey(privKey).toGroup();
     const d = privKey.s;
-    // TODO: derive nonce deterministically
-    const kPrime = Scalar.random();
+    const kPrime = Scalar.fromBigInt(
+      deriveNonce(
+        { fields: msg.map((f) => f.toBigInt()) },
+        { x: publicKey.x.toBigInt(), y: publicKey.y.toBigInt() },
+        BigInt(d.toJSON()),
+        'testnet'
+      )
+    );
     let { x: r, y: ry } = Group.generator.scale(kPrime);
     const k = ry.toBits()[0].toBoolean() ? kPrime.neg() : kPrime;
     let h = hashWithPrefix(
       prefixes.signatureTestnet,
       msg.concat([publicKey.x, publicKey.y, r])
     );
-    // FIXME: this changes the value!
-    const e = Scalar.fromBits(h.toBits());
+    // TODO: Scalar.fromBits interprets the input as a "shifted scalar"
+    // therefore we have to unshift e before using it
+    let e = unshift(Scalar.fromBits(h.toBits()));
     const s = e.mul(d).add(k);
     return new Signature(r, s);
   }
@@ -218,13 +229,16 @@ class Signature extends CircuitValue {
       prefixes.signatureTestnet,
       msg.concat([point.x, point.y, this.r])
     );
-    // FIXME: this changes the value!
+    // TODO: Scalar.fromBits interprets the input as a "shifted scalar"
+    // therefore we have to use scaleShifted which is very inefficient
     let e = Scalar.fromBits(h.toBits());
-    let r = point.scale(e).neg().add(Group.generator.scale(this.s));
+    let r = scaleShifted(point, e).neg().add(Group.generator.scale(this.s));
     return Bool.and(r.x.equals(this.r), r.y.toBits()[0].equals(false));
   }
 
-  // TODO: doccomments
+  /**
+   * Decodes a base58 encoded signature into a {@link Signature}.
+   */
   static fromBase58(signatureBase58: string) {
     let { r, s } = SignatureBigint.fromBase58(signatureBase58);
     return Signature.fromObject({
@@ -232,9 +246,30 @@ class Signature extends CircuitValue {
       s: Scalar.fromJSON(s.toString()),
     });
   }
+  /**
+   * Encodes a {@link Signature} in base58 format.
+   */
   toBase58() {
     let r = this.r.toBigInt();
     let s = BigInt(this.s.toJSON());
     return SignatureBigint.toBase58({ r, s });
   }
 }
+
+// performs scalar multiplication s*G assuming that instead of s, we got s' = 2s + 1 + 2^255
+// cost: 2x scale by constant, 1x scale by variable
+function scaleShifted(point: Group, shiftedScalar: Scalar) {
+  let oneHalfGroup = point.scale(Scalar.fromBigInt(oneHalf));
+  let shiftGroup = oneHalfGroup.scale(Scalar.fromBigInt(shift));
+  return oneHalfGroup.scale(shiftedScalar).sub(shiftGroup);
+}
+// returns s, assuming that instead of s, we got s' = 2s + 1 + 2^255
+// (only works out of snark)
+function unshift(shiftedScalar: Scalar) {
+  return shiftedScalar
+    .sub(Scalar.fromBigInt(shift))
+    .mul(Scalar.fromBigInt(oneHalf));
+}
+
+let shift = ScalarBigint(1n + 2n ** 255n);
+let oneHalf = ScalarBigint.inverse(2n)!;
--- a/src/lib/signature.ts
+++ b/src/lib/signature.ts
@@ -1,6 +1,8 @@
-import { Group, Field, Bool, Scalar, Ledger, Circuit } from '../snarky.js';
+import { Group, Field, Bool, Scalar, Ledger } from '../snarky.js';
 import { prop, CircuitValue, AnyConstructor } from './circuit_value.js';
-import { Poseidon } from './hash.js';
+import { hashWithPrefix } from './hash.js';
+import { Signature as SignatureBigint } from '../mina-signer/src/signature.js';
+import { prefixes } from '../js_crypto/constants.js';
 
 // external API
 export { PrivateKey, PublicKey, Signature };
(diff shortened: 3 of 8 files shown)


src/lib/account_update.ts
```diff
@@ -1877,7 +1877,11 @@ function signJsonTransaction(
       accountUpdate.authorization.proof === null
     ) {
       zkappCommand = JSON.parse(
-        Ledger.signAccountUpdate(JSON.stringify(zkappCommand), privateKey, i)
+        Ledger.signOtherAccountUpdate(
```
mitschabaude, line 1880: unrelated fix

src/lib/core.ts
```diff
@@ -72,3 +74,8 @@ That means it can't be called in a @method or similar environment, and there's n
     highBit: Bool(x >> lowBitSize === 1n),
   };
 };
+
+Scalar.fromBigInt = function (scalar: bigint) {
```
mitschabaude, line 78: missing helper which was useful here

src/lib/signature.ts
```diff
@@ -192,12 +198,23 @@ class Signature extends CircuitValue {
   static create(privKey: PrivateKey, msg: Field[]): Signature {
     const publicKey = PublicKey.fromPrivateKey(privKey).toGroup();
     const d = privKey.s;
-    const kPrime = Scalar.random();
+    const kPrime = Scalar.fromBigInt(
+      deriveNonce(
```
mitschabaude, line 202: **crypto change**: instead of a random nonce, we now use the same deterministic nonce derivation as in mina-signer (and elsewhere in our protocol)

src/lib/signature.ts
```diff
@@ -192,12 +198,23 @@ class Signature extends CircuitValue {
...
+        { x: publicKey.x.toBigInt(), y: publicKey.y.toBigInt() },
+        BigInt(d.toJSON()),
+        'testnet'
+      )
+    );
     let { x: r, y: ry } = Group.generator.scale(kPrime);
     const k = ry.toBits()[0].toBoolean() ? kPrime.neg() : kPrime;
-    const e = Scalar.fromBits(
-      Poseidon.hash(msg.concat([publicKey.x, publicKey.y, r])).toBits()
+    let h = hashWithPrefix(
```
mitschabaude, line 211: **crypto change**: instead of plain hashing with Poseidon, we use a hash prefix

src/lib/signature.ts
```diff
@@ -192,12 +198,23 @@ class Signature extends CircuitValue {
...
+      prefixes.signatureTestnet,
```
mitschabaude, line 212: I chose to hard-code the signature to 'testnet', since the distinction between testnet and mainnet for signatures which aren't about transactions seemed not useful to me and just adds noise IMO

src/lib/signature.ts
```diff
@@ -192,12 +198,23 @@ class Signature extends CircuitValue {
...
+      msg.concat([publicKey.x, publicKey.y, r])
     );
+    // TODO: Scalar.fromBits interprets the input as a "shifted scalar"
+    // therefore we have to unshift e before using it
+    let e = unshift(Scalar.fromBits(h.toBits()));
```
mitschabaude, line 217: So this seems like a bug in `Scalar.fromBits`, which is somewhat hard to resolve in that function directly, since this function would need to unshift the scalar in a SNARK-compatible way. It seemed easier to unshift out of the snark (here) and do the shifting at the level of curve points when we're in the snark (see `verify` below)

src/lib/signature.ts
```diff
@@ -208,10 +225,51 @@ class Signature extends CircuitValue {
...
+    return SignatureBigint.toBase58({ r, s });
+  }
+}
+
+// performs scalar multiplication s*G assuming that instead of s, we got s' = 2s + 1 + 2^255
+// cost: 2x scale by constant, 1x scale by variable
+function scaleShifted(point: Group, shiftedScalar: Scalar) {
+  let oneHalfGroup = point.scale(Scalar.fromBigInt(oneHalf));
+  let shiftGroup = oneHalfGroup.scale(Scalar.fromBigInt(shift));
+  return oneHalfGroup.scale(shiftedScalar).sub(shiftGroup);
```
mitschabaude, line 264: this is how we undo the scalar shift inside the SNARK

src/mina-signer/MinaSigner.ts
```diff
@@ -96,6 +97,48 @@ class Client {
...
+   * let isValid: Bool = signature.verify(publicKey, fields.map(Field));
+   * ```
+   *
+   * @param fields An arbitrary list of field elements
+   * @param privateKey The private key used for signing
+   * @returns The signed field elements
+   */
+  signFields(fields: bigint[], privateKey: Json.PrivateKey): Signed<bigint[]> {
+    let privateKey_ = PrivateKey.fromBase58(privateKey);
+    let signature = sign({ fields }, privateKey_, 'testnet');
```
mitschabaude, line 118: very straight-forward new function to sign an arbitrary list of field elements. it uses the existing `sign` and hard-codes the network to 'testnet' since this is not a transaction signature

src/mina-signer/tests/verify-in-snark.unit-test.ts
```diff
@@ -0,0 +1,49 @@
+import { Field, isReady, shutdown } from '../../snarky.js';
```
mitschabaude, line 1: this file tests that the new signatures are really compatible between mina-signer and snarkyjs, and also checks that they can be verified inside a SNARK

mitschabaude (commented):
some comments to help with review:

src/snarky.d.ts
```diff
@@ -1274,7 +1279,7 @@ declare class Ledger {
   /**
    * Signs an account update.
    */
-  static signAccountUpdate(
+  static signOtherAccountUpdate(
```
bkase, line 1282: why did this name change?

bkase (changes requested):

src/lib/signature.ts
mitschabaude: That's a good point! It opens a can of worms though. When signing transactions, the signature verifier is the Mina node, which has a well-defined network - ti's either a mainnet node or a testnet node, so it knows which one it should use for verifying.
However, a smart contract currently doesn't have that. Do you think the network type should be hard-coded into the smart contract / circuit somehow? Should we leave it as a constant for the developer to define? (But then we're pushing to them the problem of switching this flag when deploying / compiling to either network).

I think, if we go this route, the network type should definitely be a constant in the circuit and not a `Bool` that the developer can make variable. Because if it can be variable, it's too easy a mistake to just make it a private user input which doesn't constrain the proofs to be valid for the respective network.

src/snarky.d.ts
mitschabaude: This is an unrelated fix, the name was actually `signOtherAccountUpdate` in the ocaml bindings, so it didn't work before

src/lib/signature.ts
bkase: Hmm I suppose it’s much less dangerous than a transaction so I think it’s not worth the complexity cost of making these signatures distinct, but we should probably run this by more people. I’ll approve but I’m going to send a link to one of our slack channels and let’s let it bake overnight if anyone disagrees before we land it

bkase (approved):
Product okay (tentatively, see comment). Still needs a crypto review from someone

src/lib/signature.ts
mimoo: am I understanding this right that the constant doesn't matter here because it's what the zkapp will verify and so you could pretty much use anything?

src/mina-signer/MinaSigner.ts
mimoo: would it makes sense to hardcode the chain id to 'zkapp' or something else then?

src/lib/core.ts
mimoo: same question as elsewhere, I'm curious how a dev would know not to use these functions if they want them to create constraints? My guess is that we're warning them that going to bigint or from bigint is not constraining anything?

src/lib/signature.ts
```diff
@@ -208,10 +225,51 @@ class Signature extends CircuitValue {
...
+// returns s, assuming that instead of s, we got s' = 2s + 1 + 2^255
+// (only works out of snark)
+function unshift(shiftedScalar: Scalar) {
+  return shiftedScalar
+    .sub(Scalar.fromBigInt(shift))
+    .mul(Scalar.fromBigInt(oneHalf));
 }
+
+let shift = ScalarBigint(1n + 2n ** 255n);
+let oneHalf = ScalarBigint.inverse(2n)!;
```
mimoo, line 275: can't you do `let oneHalf = Scalar.fromBigInt(ScalarBigint.inverse(2n));`?

src/lib/signature.ts
mimoo: why not just do

```
point.scale((shiftedScalar.sub(shift).mul(inv_2))
```

src/lib/signature.ts
```diff
@@ -208,10 +225,51 @@ class Signature extends CircuitValue {
...
-    let e = Scalar.fromBits(
-      Poseidon.hash(msg.concat([point.x, point.y, this.r])).toBits()
+    let h = hashWithPrefix(
+      prefixes.signatureTestnet,
+      msg.concat([point.x, point.y, this.r])
     );
-    let r = point.scale(e).neg().add(Group.generator.scale(this.s));
+    // TODO: Scalar.fromBits interprets the input as a "shifted scalar"
+    // therefore we have to use scaleShifted which is very inefficient
+    let e = Scalar.fromBits(h.toBits());
```
mimoo, line 234: is this going to create constraints?

src/lib/signature.ts
```diff
@@ -208,10 +225,51 @@ class Signature extends CircuitValue {
...
+  let shiftGroup = oneHalfGroup.scale(Scalar.fromBigInt(shift));
+  return oneHalfGroup.scale(shiftedScalar).sub(shiftGroup);
+}
+// returns s, assuming that instead of s, we got s' = 2s + 1 + 2^255
+// (only works out of snark)
+function unshift(shiftedScalar: Scalar) {
+  return shiftedScalar
+    .sub(Scalar.fromBigInt(shift))
+    .mul(Scalar.fromBigInt(oneHalf));
 }
```
mimoo, line 272: why does this only work out of snark? Also, isn't this a problem if I want to sign within a snark? (since it's used in the signature algo)

mimoo (approved):
LGTM!

src/lib/signature.ts
mimoo: oh I think I see, the scalars are in the wrong field :D

src/lib/core.ts
mitschabaude: For methods like these, which take a JS value and produce constant field elements, I think no warning is needed -- it's a valid and useful way to introduce constants into the circuit.

Going the other way -- from field elements to JS value -- is a very common mistake, but will throw an error as soon as you try to compile the contract. That's because extracting values, under the hood, always works like this:
```ocaml
match x with
  | Constant x -> x
  | x -> As_prover.read_var x
```
so during compile time you'll hit an error for the `As_prover.read_var x` case. The constant case is fine - no need to prove how constants are computed, since the prover can't change them

src/lib/signature.ts
mitschabaude: @mimoo yes, doesn't matter here in the sense that these signatures won't be sent to either network. we are completely in user-land: signatures are inputs to proofs created locally, where they are verified in-snark.

I think Brandon highlights an important problem: we don't provide replay protection for the user here, especially given that signatures are deterministic (nonce is derived from the message and private key). so, we currently leave it up to the application to figure out how to prevent that.

I think its too deep and broad a problem to solve in the context of this PR, but I created an issue so we don't forget to discuss this: https://github.com/o1-labs/snarkyjs/issues/720

src/lib/signature.ts
mitschabaude: yep exactly :D we don't have the foreign field arithmetic to do `sub` and `mul` on the scalar, inside the snark

src/mina-signer/MinaSigner.ts
mitschabaude: I think it does! Do you think we need actual domain separation between testnet signatures and signatures meant for zkapps? Or is it more about better naming, so that 'zkapp' is an alias for 'testnet'? (it wouldn't be hard to add the real domain separation fwiw)

src/lib/signature.ts
mitschabaude: no.. because `Scalar.fromBigInt` needs the JSOO / Wasm bundle to be loaded, and that's not the case at the top level. (only after `await isReady`)

src/lib/signature.ts
mitschabaude: yeah sure! `toBits` does the proper in-snark unpacking of a field element, and `fromBits` is a no-op, since scalars are already represented as an array of bits (unfortunately, the bits represent a shifted scalar :/)

src/lib/signature.ts
mitschabaude: only works out of snark because `sub` and `mul` are not available on scalars. not sure how efficient they would be to implement, compared to the extra scaling I have to do instead. do you know?

the signing algorithm was already (I think, deliberately) written to not work inside a snark (for example, it used a constant random field element for the nonce 😱 ; now I'm deriving the nonce as a blake2b hash. tbf, the nonce could be a prover witness). the above will throw an error in the snark. 

Inherently, signing in the snark is not as useful because then you have to provide your private key as input -- but that usually sits in some guarded place like your wallet. so, I think it's fine to only support signing outside + verifying inside which achieves the same and plays well with wallets

src/mina-signer/MinaSigner.ts
mimoo: sorry for late reply. In general I would say we should use domain separate as much as possible, because it's cheap/free : ) it's possible that it's not exploitable, but still why not distinguish it

PR #714, opened 2023-02-02
Address mina-signer audit, Part 1

mitschabaude:
This addresses the entire audit _except_ the final section, test coverage. (Will be addressed in a follow-up PR)

Here are inline comments on all items marked with `WARN` or `FAIL`:

> **Type coercion**
> * Avoid use of of == and !=

addressed by https://github.com/o1-labs/snarkyjs/pull/714/commits/3739483a07051c3cf8a1a1de747adfb321ca0ffd

> * a + b, `src/mina-signer/MinaSigner.ts`  
>   * `p.reduce((accumulatedFee, _) => accumulatedFee + fee, 0);`

addressed by simplifying the method (but the type was inferred anyway, see next item) https://github.com/o1-labs/snarkyjs/pull/714/commits/184029aa60915fe5658900b743c5598f9528def8

> * a + b, `src/provable/from-layout.ts`  
>   * Instance 1: `return keys.map((key) => object[key]).reduce((x, y) => x + y);`  
>   * Instance 2: `eturn isSome + value;`  
>   * Note: Types of are not explicit here. It would be good practice to enforce them if possible

in both cases the type is inferred, because methods are part of an object that has a type annotation:
Screenshot from 2023-02-02 10-49-15

> **Safe error handling**
> * `src/js_crypto/finite_field.ts`  
>   * Instance: `src/js_crypto/finite_field.ts:148: if (yinv === undefined) return;`  
>   * Mitigation: The div method should return undefined so that the caller knows a division by zero occurred

The `return;` statement already means that `undefined` is returned

> * `src/provable/provable-generic.ts`  
>   * Instance: `src/provable/provable-generic.ts:62: if (!complexTypes.has(typeof typeObj) || typeObj === null) return [];`  
>     (and many places in this file)
>   * Issue: Failing with an empty array may not alert the caller to an error. Preferred to throw an error.

addressed by https://github.com/o1-labs/snarkyjs/pull/714/commits/77632b38163106273c48726845c721e48491d558

> **Safe argument handling**
> * `src/mina-signer/MinaSigner.ts`
>   * Instance: `src/mina-signer/MinaSigner.ts:L418: fee: number = 0.001`  
>   * Mitigation: The `getAccountUpdateMinimumFee` function must prevent negative fee arguments

addressed by removing the method argument entirely, as it was not useful https://github.com/o1-labs/snarkyjs/pull/714/commits/184029aa60915fe5658900b743c5598f9528def8

> * `src/js_crypto/poseidon.ts`  
>   * Result: FAIL  
>   * Instance: `src/js_crypto/poseidon.ts:56: let n = Math.ceil(input.length / rate) * rate;`   
>   * Mitigation: The `createPoseidon` interface should check `rate`, `fullRounds` and `power` > 0

This was addressed by adding runtime checks: https://github.com/o1-labs/snarkyjs/pull/714/commits/5281e8d940b378fbc8df9a5208ffded34f172213
A check on the type level was not deemed necessary because `createPoseidon` is not user-facing. It is not exported and only used immediately with auto-generated Poseidon parameters as inputs.

> * `src/provable/binable.ts`
> * `src/provable/field-bigint.ts`
> * `src/mina-signer/src/memo.ts`
>   * Result: FAIL
>   * Mitigation: These function arguments should have types to restrict them to >= 0

All of these [were addressed together](https://github.com/o1-labs/snarkyjs/pull/714/commits/5281e8d940b378fbc8df9a5208ffded34f172213) by introducing a type interface for `Binable.readBytes` that only accepts non-negative integers as the `offset`. In addition to the type, this is also enforced with a runtime check, which is placed in `defineBinable`, the function that is always used to create binables. I also added some invariant checks on the output of `readBytes`.

Example of type error:
image

Example of runtime error (when the type check is bypassed with `as never`):
```
Error: readBytes: offset must be integer >= 0
    at assertNonNegativeInteger (/home/gregor/o1/mina/src/lib/snarky_js_bindings/snarkyjs/src/js_crypto/non-negative.ts:42:44)
    at Object.readBytes (/home/gregor/o1/mina/src/lib/snarky_js_bindings/snarkyjs/src/provable/binable.ts:57:7)
```

> * `src/provable/gen/transaction-json.ts`  
>   * Result: FAIL  
>   * Instance: `src/provable/gen/transaction-json.ts:84: callDepth: number;`  
>   * Mitigation: The interface should be updated to assure `callDepth` is a positive integer `>= 0`
>   * Notes: It appears that this code may be generated from the OCaml code, which should probably not represent `callDepth` as an `int`.

This was addressed by adding a runtime check for non-negativity at the only place where `callDepth` is used within mina-signer: https://github.com/o1-labs/snarkyjs/pull/714/commits/f21f8d9130620c0ea8d56f118910c69a06976f2d
The check above will be triggered for every account update with an invalid call depth. (Also, `callDepth` has no impact anywhere else in mina-signer, because it doesn't become part of the hash input for signing.)

The type auto-generated from OCaml wasn't changed.

> Furthermore, all auto-generated code should have a comment at the top of the source file stating that it is autogenerated and should not be edited.

This is already the case!

> * `src/mina-signer/MinaSigner.ts`  
>   * Result: WARN  
>   * Concern: The `feePayer` structure et al allows specification of `fee`, `nonce` and `validUntil` as either strings, integers or floating point values
>   * Instance: `src/mina-signer/tests/party.test.ts:139: fee: '1',`
>   * Mitigation: Restrict creation of `feePayer` so that the types of arguments can be enforced
>   * Notes: Although the low-level `BigInt` code will catch an error, the higher level code will
coerce strings and floating point values, which is not safe practice.

This one I found a bit tricky to address. I don't think restriction of how fee payers (and other transaction inputs) can be created is a good solution here. In practice, these inputs will be created in a web app and passed to a wallet, through an interface such as browser's `postMessage` or even an HTTP request. The wallet passes these inputs on to mina-signer. Since the wallet will receive JSON data and not a complex class, we would push an additional step to wallet maintainers: Invoke mina-signer's safe methods to generate valid fee payers etc, before passing these to the desired methods. 

I think it's simpler to keep existing interfaces as is and enable the direct flow we have now where the wallet can just pass on JSON data it receives. Therefore, I went with the following mitigation: In all functions that receive numeric transaction inputs, like fee/nonce/validUntil, I validate that these are well-formed, positive integers, right at the beginning: https://github.com/o1-labs/snarkyjs/pull/714/commits/a8c6488acabf6192c47a5c8b27cc16d07257cc74

> **Fuzzing**
> * Payment signatures `mina-signer/src/sign-legacy.ts`
>   * Memo FAIL

After discussion with @jspada, the implementation is actually fine

> * Payment signatures `mina-signer/src/sign-legacy.ts`
>   * BinableBigintInteger FAIL
>     * Detail: Sign flips when converting bigint to bytes and then back to a bigint

This was a major bug in the implementation, caused by not reading the `bin_prot` spec carefully, and implementing just the int64 version. Addressed by properly implementing integers of different bit lengths, and by adding extra wrappers for uints (which are converted to ints before serializing): https://github.com/o1-labs/snarkyjs/pull/714/commits/76f67bdeeadd6a50326eab394d5821387baf1e4b
(that commit also adds the beginning of a `binable.unit-test.ts` which tests different integers)

### DIFF
--- a/src/js_crypto/non-negative.ts
+++ b/src/js_crypto/non-negative.ts
@@ -1,0 +1,62 @@
+export {
+  Integer,
+  NonNegativeInteger,
+  PositiveInteger,
+  isInteger,
+  isNonNegativeInteger,
+  isPositiveInteger,
+  assertInteger,
+  assertNonNegativeInteger,
+  assertPositiveInteger,
+  asInteger,
+  asNonNegativeInteger,
+  asPositiveInteger,
+};
+
+function asInteger<N extends number>(n: Integer<N>) {
+  return n;
+}
+function asNonNegativeInteger<N extends number>(n: NonNegativeInteger<N>) {
+  return n;
+}
+function asPositiveInteger<N extends number>(n: PositiveInteger<N>) {
+  return n;
+}
+
+function isInteger<N extends number>(n: N): n is Integer<N> {
+  return Number.isInteger(n);
+}
+function isNonNegativeInteger<N extends number>(
+  n: N
+): n is NonNegativeInteger<N> {
+  return Number.isInteger(n) && n >= 0;
+}
+function isPositiveInteger<N extends number>(n: N): n is PositiveInteger<N> {
+  return Number.isInteger(n) && n > 0;
+}
+
+function assertInteger(n: number, message: string) {
+  if (!Number.isInteger(n)) throw Error(message);
+}
+function assertNonNegativeInteger(n: number, message: string) {
+  if (!Number.isInteger(n) || n < 0) throw Error(message);
+}
+function assertPositiveInteger(n: number, message: string) {
+  if (!Number.isInteger(n) || n < 0) throw Error(message);
+}
+
+type Integer<T extends number> = number extends T
+  ? never
+  : `${T}` extends `${string}.${string}`
+  ? never
+  : T;
+type NonNegativeInteger<T extends number> = number extends T
+  ? never
+  : `${T}` extends `-${string}` | `${string}.${string}`
+  ? never
+  : T;
+type PositiveInteger<T extends number> = number extends T
+  ? never
+  : `${T}` extends `-${string}` | `${string}.${string}` | `0`
+  ? never
+  : T;
--- a/src/js_crypto/non-negative.ts
+++ b/src/js_crypto/non-negative.ts
@@ -42,21 +42,25 @@ function assertNonNegativeInteger(n: number, message: string) {
   if (!Number.isInteger(n) || n < 0) throw Error(message);
 }
 function assertPositiveInteger(n: number, message: string) {
-  if (!Number.isInteger(n) || n < 0) throw Error(message);
+  if (!Number.isInteger(n) || n <= 0) throw Error(message);
 }
 
 type Integer<T extends number> = number extends T
   ? never
-  : `${T}` extends `${string}.${string}`
+  : `${T}` extends `${string}.${string}` | `${string}e-${string}`
   ? never
   : T;
 type NonNegativeInteger<T extends number> = number extends T
   ? never
-  : `${T}` extends `-${string}` | `${string}.${string}`
+  : `${T}` extends `-${string}` | `${string}.${string}` | `${string}e-${string}`
   ? never
   : T;
 type PositiveInteger<T extends number> = number extends T
   ? never
-  : `${T}` extends `-${string}` | `${string}.${string}` | `0`
+  : `${T}` extends
+      | `-${string}`
+      | `${string}.${string}`
+      | `${string}e-${string}`
+      | `0`
   ? never
   : T;
--- a/src/mina-signer/MinaSigner.ts
+++ b/src/mina-signer/MinaSigner.ts
@@ -441,16 +441,9 @@ class Client {
    * the sum of all account updates plus the specified fee amount. If no fee is passed in, `0.001`
    * is used (according to the Mina spec) by default.
    * @param accountUpdates A list of account updates
-   * @param fee The fee per accountUpdate amount
    * @returns  The fee to be paid by the fee payer accountUpdate
    */
-  getAccountUpdateMinimumFee(
-    accountUpdates: TransactionJson.AccountUpdate[],
-    fee: number = 0.001
-  ) {
-    return accountUpdates.reduce(
-      (accumulatedFee, _) => accumulatedFee + fee,
-      0
-    );
+  getAccountUpdateMinimumFee(accountUpdates: TransactionJson.AccountUpdate[]) {
+    return 0.001 * accountUpdates.length;
   }
 }
--- a/src/mina-signer/MinaSigner.ts
+++ b/src/mina-signer/MinaSigner.ts
@@ -124,17 +124,6 @@ class Client {
     return verifyStringSignature(data, signature, publicKey, this.network);
   }
 
-  private normalizeCommon(common: Json.Common): Json.StrictCommon {
-    return {
-      to: common.to,
-      from: common.from,
-      fee: String(common.fee),
-      nonce: String(common.nonce),
-      memo: common.memo ?? '',
-      validUntil: String(common.validUntil ?? defaultValidUntil),
-    };
-  }
-
   /**
    * Signs a payment transaction using a private key.
    *
@@ -149,8 +138,7 @@ class Client {
     payment: Json.Payment,
     privateKey: Json.PrivateKey
   ): Signed<Json.Payment> {
-    let { fee, to, from, nonce, validUntil, memo } =
-      this.normalizeCommon(payment);
+    let { fee, to, from, nonce, validUntil, memo } = validCommon(payment);
     let amount = String(payment.amount);
     let signature = signPayment(
       {
@@ -174,8 +162,8 @@ class Client {
    * @returns True if the `signedPayment` is a verifiable payment
    */
   verifyPayment({ data, signature, publicKey }: Signed<Json.Payment>): boolean {
-    let { fee, to, from, nonce, validUntil, memo } = this.normalizeCommon(data);
-    let amount = String(data.amount);
+    let { fee, to, from, nonce, validUntil, memo } = validCommon(data);
+    let amount = validNonNegative(data.amount);
     return verifyPayment(
       {
         common: { fee, feePayer: from, nonce, validUntil, memo },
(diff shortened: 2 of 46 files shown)


jspada:
> (that commit also adds the beginning of a binable.unit-test.ts which tests different integers)

Are there any another high-risk binable type/structures that you think you should test?

mitschabaude:
> Are there any other high-risk binable type/structures that you think you should test?

I might add some unit tests for primitives like string. But I don't think there is impact in further testing that area, because the binprot serializations of the high-level data types we actually use are comprehensively tested, and thanks to [our new approach of randomly generating a wide variety of test cases](https://github.com/o1-labs/snarkyjs/pull/718), these tests will hit all possible cases in the lower level types.

src/mina-signer/MinaSigner.ts
```diff
@@ -441,16 +410,46 @@ class Client {
...
+  return n0.toString();
+}
+
+function validCommon(common: Json.Common): Json.StrictCommon {
+  return {
+    to: common.to,
+    from: common.from,
+    fee: validNonNegative(common.fee),
+    nonce: validNonNegative(common.nonce),
+    memo: common.memo ?? '',
```
jspada, line 432: Check header and length of memo.

src/mina-signer/MinaSigner.ts
```diff
@@ -441,16 +410,46 @@ class Client {
...
+
+function validNonNegative(n: number | string | bigint): string {
+  let n0 = BigInt(n); // validates that string represents an integer; also throws runtime errors for nullish inputs
+  if (n0 < 0) throw Error('input must be non-negative');
+  return n0.toString();
+}
+
+function validCommon(common: Json.Common): Json.StrictCommon {
+  return {
+    to: common.to,
```
jspada, line 428: My understanding is that to and from are already checked base base58 decode/encode

src/mina-signer/MinaSigner.ts
```diff
@@ -441,16 +410,46 @@ class Client {
...
+    from: common.from,
+    fee: validNonNegative(common.fee),
+    nonce: validNonNegative(common.nonce),
+    memo: common.memo ?? '',
+    validUntil: validNonNegative(common.validUntil ?? defaultValidUntil),
+  };
+}
+
+function validFeePayer(
+  feePayer: Json.ZkappCommand['feePayer'],
```
jspada, line 438: Is this already checked by base58 decode somewhere?  It may be handy to create a base85Check function that checks the checksum and whether the pub key point is on the curve and field element checks.

src/mina-signer/MinaSigner.ts
```diff
@@ -441,16 +410,46 @@ class Client {
...
+  minimumFee: number
+): Json.StrictFeePayer {
+  if (feePayer.fee === undefined) throw Error('Missing fee in fee payer');
+  let fee = validNonNegative(feePayer.fee);
+  if (Number(fee) < minimumFee)
```
jspada, line 443: Don't you need to check minimumFee is `validNonNegative` first?

src/mina-signer/src/sign-zkapp-command.ts
```diff
@@ -111,8 +112,11 @@ function accountUpdatesToCallForest(updates: AccountUpdate[], callDepth = 0) {
   let forest: CallForest = [];
   while (remainingUpdates.length > 0) {
     let accountUpdate = remainingUpdates[0];
+    assertNonNegativeInteger(
+      accountUpdate.body.callDepth,
+      `callDepth must be positive integer, got ${callDepth}`
+    );
```
jspada, line 118: Nice

src/provable/binable.ts
```diff
@@ -37,11 +47,31 @@ function defineBinable<T>({
...
-    readBytes,
+    // spec: input offset has to be a non-negative integer;
+    // output offset has to be greater or equal input, and not exceed the bytes length
+    readBytes(bytes, offset) {
+      assertNonNegativeInteger(
+        offset,
+        'readBytes: offset must be integer >= 0'
+      );
+      let [value, end] = readBytes(bytes, offset);
+      if (end < offset)
```
jspada, line 65: Nice!

src/provable/binable.ts
```diff
@@ -188,55 +218,88 @@ const CODE_INT16 = 0xfe;
...
-    }[code];
-    if (size === undefined) {
-      throw Error('binable integer: invalid start byte');
-    }
-    let end = offset + size;
-    let x = fillInt64(bytes.slice(offset, end));
-    return [x, end];
-  },
-});
+function BinableInt(bits: number) {
```
jspada, line 221: need to check `bits` `IsPositiveInteger`?

src/provable/binable.ts
```diff
@@ -188,55 +218,88 @@ const CODE_INT16 = 0xfe;
...
+        [CODE_INT32]: 4,
+        [CODE_INT64]: 8,
+      }[code];
+      if (size === undefined) {
+        throw Error('binable integer: invalid start byte');
+      }
+      let end = offset + size;
+      let x = fillUInt(bytes.slice(offset, end), nBytes);
+      // map from uint to int range
+      if (x >= maxValue) x -= 2n * maxValue;
```
jspada, line 260: Clearer and less error prone to add curly braces around these.

For example, apple's infamous `goto fail;` vulnerability could have been avoided/spotted if curly braces were used.

src/provable/binable.ts
```diff
@@ -188,55 +218,88 @@ const CODE_INT16 = 0xfe;
...
+      if (x < -maxValue || x >= maxValue)
+        throw Error(`int${bits} out of range, got ${x}`);
+      return [x, end];
+    },
+  });
+}
 
-function fillInt64(startBytes: number[]) {
+function fillUInt(startBytes: number[], nBytes: number) {
```
jspada, line 268: need to check nBytes is positive int?

jspada (approved):
Approved with a couple things you should probably do and a few suggestions and questions.

src/js_crypto/non-negative.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!Number.isInteger(n)) throw Error(message);
+}
+function assertNonNegativeInteger(n: number, message: string) {
+  if (!Number.isInteger(n) || n < 0) throw Error(message);
+}
+function assertPositiveInteger(n: number, message: string) {
+  if (!Number.isInteger(n) || n < 0) throw Error(message);
+}
+
+type Integer<T extends number> = number extends T
```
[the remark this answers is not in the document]
mitschabaude: great catch!

jspada:
> This one I found a bit tricky to address. 
After some discussions with @mitschabaude - my understanding is that the current code actually doesn't leave anything up to the calls, but actually presents a secure API.

In particular, inside our signing method we do something like

```
function signZkappCommand({ feePayer: unvalidatedUserInput, ... }, privateKey) {
  feePayer = createValidFeePayer(unvalidatedUserInput);
  // ...
}
```

src/mina-signer/MinaSigner.ts
mitschabaude: no because `minimumFee` is internal to the library, and hard-coded

src/mina-signer/MinaSigner.ts
mitschabaude: actually in this place a string without the header is expected (it gets encoded later), so I'll only add a check for the length

src/mina-signer/MinaSigner.ts
mitschabaude: Yes it's checked by `fromBase58Check`. That's the only function in mina-signer which can decode base58, and it checks both the checksum, validity of input characters and the version byte.

Since you can't use a base58 string without decoding it, anyway, I think its fine to rely on the decoding function to assert validity. (a "checking only" function would add redundant computation)

PR #716, opened 2023-02-06
Address mina-signer audit, Pt 2: Low-level crypto tests

mitschabaude:
Successor to https://github.com/o1-labs/snarkyjs/pull/714

Addresses the following parts of the audit -- tests for low-level crypto arithmetic:

> ## 2 Test coverage
> Result: FAIL
> * js_crypto's field arithmetic methods do not have individual unit tests
> * js_crypto should have low-level tests to verify fundamental group operations
>   * Addition is commutative X + Y == Y + X
>   * Associativity X + (Y + Z) == (X + Y) z
>   * Identity I + X == X + I == X
>   * Inverse X + inv(X) == inv(X) + X == I
>   * Scaling commutes with adding scalars A * (x + y) == A * x + A * y
>   * Scaling commutes with multiplying scalars A * (x * y) == (x * y) * A
>   * Scaling commites with negation A * neg(x) == neg(A * x)
>   * Always test on curve after scaling
>   * Note: These are easy to add, so highly recommended to implement
> * js_crypto primitives sub, inverse, div, square, fromNumber and fromBigint do not have any test coverage

This PR ended up making some changes to the curve arithmetic, so in order to enable a focused review I'm leaving other missing tests for a third PR.

* Added tests for finite field and curve arithmetic
* Fixed issues that were uncovered by these tests:
  * added missing methods `Group.equal`, `Group.isOnCurve` to be able to test
  * `Group.add` was not complete addition -- failed for `Group.add(X, X)`. I added the necessary logic to fall back to doubling. Also handles the `Group.add(X, -X) = 0` case now to return zero in the default representation
* For completeness, added implementation and tests for endomorphism

### DIFF
--- a/src/mina-signer/tests/verify-in-snark.unit-test.ts
+++ b/src/mina-signer/tests/verify-in-snark.unit-test.ts
@@ -2,7 +2,7 @@ import { Field, isReady, shutdown } from '../../snarky.js';
 import { ZkProgram } from '../../lib/proof_system.js';
 import Client from '../MinaSigner.js';
 import { PrivateKey, Signature } from '../../lib/signature.js';
-import { provablePure } from '../../lib/circuit_value.js';
+import { Circuit, provablePure } from '../../lib/circuit_value.js';
 import { expect } from 'expect';
 
 let fields = [10n, 20n, 30n, 340817401n, 2091283n, 1n, 0n];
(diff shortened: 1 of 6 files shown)


src/lib/account_update.ts
```diff
@@ -1877,7 +1877,11 @@ function signJsonTransaction(
       accountUpdate.authorization.proof === null
     ) {
       zkappCommand = JSON.parse(
-        Ledger.signAccountUpdate(JSON.stringify(zkappCommand), privateKey, i)
+        Ledger.signOtherAccountUpdate(
+          JSON.stringify(zkappCommand),
```
jspada, line 1881: Do you need to worry about `stringify` safety wrt circular references?

src/lib/testing/random.ts
```diff
@@ -0,0 +1,501 @@
...
+      };
+    },
+  };
+}
+
+/**
+ * uniform distribution over range [min, max]
+ * with bias towards special values 0, 1, -1, 2, min, max
+ */
+function int(min: number, max: number): Random<number> {
```
jspada, line 351: cool :-)

src/mina-signer/tests/verify-in-snark.unit-test.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+    },
+  },
+});
+
+await MyProgram.compile();
+let proof = await MyProgram.verifySignature(null, signature);
+ok = await MyProgram.verify(proof);
+expect(ok).toEqual(true);
+
```
jspada, line 48: How about some negative tests?

src/lib/account_update.ts
mitschabaude: nope, none of these transaction data structures has any circular references (and they are kept really explicitly typed so something like an accidental circular ref in there shouldn't happen)

src/mina-signer/tests/verify-in-snark.unit-test.ts
mitschabaude: good idea!

PR #717, opened 2023-02-07
Address mina-signer audit, Pt 3: Test coverage

### LINKED ISSUE #712: Address mina-signer audit

mitschabaude:
Closes #712

The main work of this PR is addressing the following feedback:

> ### Test coverage
> Result: FAIL
> * Testing of signZkappCommand should cover more of the TransactionJson.ZkappCommand
structure and perform both positive and negative test cases.
> * Binprot implementation needs test coverage
>   * Cover binable.js
>   * And common/(high risk) use cases

The strategy to widen test coverage is to leverage random generation of transactions, as described in detail here https://github.com/o1-labs/snarkyjs/issues/721
Implementation of generative testing was started in https://github.com/o1-labs/snarkyjs/pull/718

This PR builds on that by adding
* generators for zkapp transactions, which need some custom logic to be valid
* generators for _invalid_ types -- in particular, invalid transactions of all kinds

Most of the tests for signing and hashing transactions (payments, delegations, zkapp) were refactored to rely on generated examples. Also, many negative tests were added. This caught a few remaining bugs. Notably, one of the zkapp transaction fields wasn't correctly implemented, and I had to fix the deriver for it here: https://github.com/MinaProtocol/mina/pull/12660

To get an idea for generation of invalid examples, here's a sample of invalid uint32 JSON inputs:
```js
> sample(Random.json.uint32.invalid, 20)
[
  'î',              '3.416',
  '4294967527',     '58474634.009',
  '-2871114700',    '-1',
  '4294967295.473', '-184107855397825',
  '1.639',          '4294967297',
  '587666881.542',  '4296509941',
  '4294967296',     '0.827',
  '187.875',        '-2',
  '',               '-1',
  'iË\x1A(Lúê',     '2.151'
]
```

Invalid higher-level structures are created by picking just _one_ of the subfields randomly and sampling an invalid value for it (and use valid samples for the other fields). To get a taste for how that looks in practice, in this example code we sample some invalid fee payers and print out the body along with the errors when decoding:
```ts
sample(Random.json.feePayer.invalid!, 10).forEach((feePayer) => {
  console.log(feePayer.body);
  try {
    FeePayer.fromJSON(feePayer);
  } catch (err: any) {
    console.log(err.message);
  }
});
```
The output shows that we're able to test a wide variety of failure cases:
```
{
  publicKey: 'B62qkhRyAj8Z24D2wdDkJGtyMcv2Yfv7uVUjeS2ZWgBAjBjBu8CQR4V',
  fee: '0',
  validUntil: '4294967295.974',
  nonce: '388363896'
}
Cannot convert 4294967295.974 to a BigInt
{
  publicKey: 'B62qogjJ5DikVfJfk2pTCVLaZJ4fKeJHAhyf4d3L3X9GAkUrACaCVnc',
  fee: '18446744073709551615',
  validUntil: '\x8A',
  nonce: '1697589'
}
fromJSON: expected a numeric string, got ""
{
  publicKey: 'B62qjD63VRcF4E7E1TbKQCMXd6j1oGz7N3WXTQKUBkqCe1r5vWxjFjH',
  fee: 'Ø',
  validUntil: null,
  nonce: '1'
}
fromJSON: expected a numeric string, got "Ø"
{
  publicKey: 'B62qrLQh6CGnzxKbjzaHiS1z1uRzhZNdjJ5n1XbWZswjqCgx7yCMxbh',
  fee: '3260485018054914',
  validUntil: '4294967295',
  nonce: '½\x1AÚÆó\x99\x9E\x02<)ëx\x1BXâ'
}
fromJSON: expected a numeric string, got "½ÚÆóx"
{
  publicKey: 'B62qrDQvTr5oo8ao5hHAbHHWtPbRH7cFh2dhFXcqTRBomhkdCY8j5or',
  fee: '119462',
  validUntil: '64',
  nonce: '1714608'
}
PublicKey: not a valid group element
{
  publicKey: 'B62qif8R1MCUZN4bxgeoCHubsbVKz6VTnXXRHv3ARZdUTeSH8D3hgbf',
  fee: '18446744073709551617',
  validUntil: null,
  nonce: '287251046'
}
UInt64: inputs larger than 18446744073709551615 are not allowed, got 18446744073709551617
{
  publicKey: 'B62qoHPdqNxEBem2YAXjRrxMMmmpyJudy9yvdNerFVmm5iRKxgFjYJS',
  fee: '1',
  validUntil: '94066',
  nonce: '1283013'
}
fromBase58Check: invalid checksum
{
  publicKey: 'B62qpPBM4v5tvv2Lut9RWUDxZLfwxFWEkZAw26ZZotcqYXoTCqjprni',
  fee: '436141000026480598',
  validUntil: '2.469',
  nonce: '34'
}
Cannot convert 2.469 to a BigInt
{
  publicKey: 'B62qnC9fQU5Xjb1N3qcJB1mkJQxPwPtnQ7v3SMbT2o86ngZMuTtcFam',
  fee: '\x9E',
  validUntil: null,
  nonce: '91194755'
}
fromJSON: expected a numeric string, got "'B62qqskgGdyAY2eCeJSKjpS45ZAuWvsxsVZzWWLtZKRUPkhJAnthMhW',
  fee: '63617053',
  validUntil: '22.483',
  nonce: '5084619'
}
Cannot convert 22.483 to a BigInt
```

In addition to the new transaction tests, we added unit tests covering most of `binable.ts` (https://github.com/o1-labs/snarkyjs/pull/717/commits/1e92d98a139c0c9aa42abc09fe412a3815ecc502)

> * Binprot implementation needs test coverage
>   * Unit tests for critical mina-specific types composed such as PaymentV1, DelegationV1, BodyV1, SignedCommandV1
>     * Example: Test all correct fields required are present and have correct types

The last point of unit-testing `PaymentV1` etc wasn't directly addressed. These types are not exposed directly in any way, they only serve `hashPayment` and `hashDelegation`. I think testing hashing with many random examples (also invalid) is exactly the coverage we need to also test the underlying binprot serialization (note: the results of hashes, and with it the serialization, is checked for consistency against OCaml).

Two other small points are addressed:
- npm audit fix https://github.com/o1-labs/snarkyjs/pull/717/commits/1aa25dde03b7a99e37913e16d0b419505e4bc6a5
- document assumptions in hash-legacy https://github.com/o1-labs/snarkyjs/pull/717/commits/8398c63f39775fefc47405b3a4e321f7e025caa4


src/lib/hash-input.unit-test.ts
```diff
@@ -1,148 +1,93 @@
 import {
```
mitschabaude, line 1: this test (which checks consistency of hash inputs between OCaml & JS) was refactored to a generative test

src/lib/testing/property.ts
```diff
@@ -1,40 +1,55 @@
...
-  // (minRuns + remainingRuns) * timePerRun = timeBudget
-  let remainingRuns = Math.floor(timeBudget / (time / minRuns)) - minRuns;
-  // run at most `maxRuns` times
-  if (remainingRuns > maxRuns - minRuns) remainingRuns = maxRuns - minRuns;
-  testN(remainingRuns, nexts, run);
-  return minRuns + remainingRuns;
+const test = Object.assign(testCustom(), {
+  negative: testCustom({ negative: true }),
+  custom: testCustom,
+});
```
mitschabaude, line 12: adds a primitive for negative testing

src/lib/testing/testing.unit-test.ts
```diff
@@ -15,10 +25,36 @@ test(Random.accountUpdate, (accountUpdate, assert) => {
...
+// note: test.negative asserts that _every_ sample fails
+test.negative(Random.json.uint64.invalid, UInt64.fromJSON);
+test.negative(Random.json.uint32.invalid, UInt32.fromJSON);
+test.negative(Random.json.publicKey.invalid, PublicKey.fromJSON);
+test.negative(Random.json.signature.invalid, Signature.fromBase58);
+
+test.custom({ negative: true, timeBudget: 1000 })(
+  Random.json.accountUpdate.invalid!,
+  AccountUpdate.fromJSON
+);
```
mitschabaude, line 53: lots of valuable new invalid tests here

src/mina-signer/src/consistency.unit-test.ts
```diff
@@ -1,257 +0,0 @@
-import { expect } from 'expect';
```
mitschabaude, line 1: this file was renamed to `sign-zkapp-command.unit-test.ts`

src/mina-signer/src/sign-zkapp-command.unit-test.ts
```diff
@@ -0,0 +1,297 @@
+import { expect } from 'expect';
```
mitschabaude, line 1: this is a file that was renamed. actual changes: https://github.com/o1-labs/snarkyjs/pull/717/commits/2b946d4ec035affc0503614d68c5ef91b2efc58c and https://github.com/o1-labs/snarkyjs/pull/717/commits/1fd93c63112b14787cc4af9fbab7862d112c6c84

src/provable/binable.unit-test.ts
```diff
@@ -1,32 +1,152 @@
 import { expect } from 'expect';
```
mitschabaude, line 1: lots of `binable` unit tests in this file!

src/mina-signer/src/transaction-hash.unit-test.ts
```diff
@@ -23,169 +23,131 @@ import { Memo } from './memo.js';
...
-    signature: Signature.fromBase58(delegation.signature),
-    payload,
-  };
-  let delegationBytes1 = SignedCommand.toBytes(command);
-  expect(JSON.stringify(delegationBytes1)).toEqual(
-    JSON.stringify(delegationBytes0)
-  );
+    // delegation hash
+    digest0 = Test.transactionHash.hashPayment(ocamlDelegation);
+    digest1 = hashStakeDelegation(delegation, { berkeley: true });
```
jspada, line 87: Same

src/provable/binable.unit-test.ts
```diff
@@ -1,32 +1,152 @@
...
+
+// finite fields
+
+test(Random.field, (field) => {
+  let bytes = Field.toBytes(field);
+  let result = Field.fromBytes(bytes);
+  expect(result).toEqual(field);
+});
+test.negative(Random.field.invalid, (field) =>
+  // TODO: toBytes doesn't fail. But should it?
```
jspada, line 40: How could it?

src/provable/binable.unit-test.ts
```diff
@@ -1,32 +1,152 @@
...
+
+test(Random.field, (field) => {
+  let bytes = VersionedFieldV1.toBytes(field);
+  let result = VersionedFieldV1.fromBytes(bytes);
+  expect(result).toEqual(field);
+});
+test.negative(Random.field, (field) => {
+  let bytes = VersionedFieldV2.toBytes(field);
+  // decoding with wrong version number should fail
+  let result = VersionedFieldV1.fromBytes(bytes);
```
jspada, line 85: excellent again!

src/provable/binable.unit-test.ts
```diff
@@ -1,32 +1,152 @@
...
+
+let FieldWithFailingCheck = withCheck(Field, () => {
+  throw Error('always fails');
+});
+test.negative(Random.field, (field) =>
+  // should fail to decode when check fails
+  FieldWithFailingCheck.fromBytes(FieldWithFailingCheck.toBytes(field))
+);
+
+let notABool = Random.map(Random.uint32, (x) => x + 2n);
```
jspada, line 99: what if randomly `x = uint32::max - 1`

src/provable/binable.unit-test.ts
```diff
@@ -1,32 +1,152 @@
...
+test.negative(Random.bool, (bool) => {
+  let bytes = MessedUpBool.toBytes(bool);
+  MessedUpBool.fromBytes(bytes); // should fail
+});
+
+// defineBinable - can't read from an invalid offset
+
+let fieldBytes = Field.toBytes(Field.random());
+expect(() => Field.readBytes(fieldBytes, -1 as never)).toThrow();
+expect(() => Field.readBytes(fieldBytes, 0.1 as never)).toThrow();
```
jspada, line 151: What about `3e-2` test?

src/lib/testing/property.ts
jspada: Are these "random generation" tests deterministic/reproducible?

jspada (approved):
This is excellent work that expands the test coverage significantly!  Reviewed with a few minor things to check and some small suggestions.

src/lib/testing/random.ts
```diff
@@ -364,7 +644,7 @@ function int(min: number, max: number): Random<number> {
     create: () => () => {
       // 25% of test cases are special numbers
```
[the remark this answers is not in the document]
mitschabaude: I considered that.. The reason I ended up not doing it is that as these random generators are combined into higher-level structures, it seems much more valuable to randomly combine special values, like 0, with other values. Also, we use random values for things like array sizes which determine the structure of the generated output. If, say, 0 would always be the first value that is drawn, then the _first sample_ of anything would contain just empty arrays, null / undefined and zeroes; instead of sprinkling zeroes, nulls and empty arrays throughout the samples, like it is done now.

I wouldn't worry about the less than 100% probability. In practice, we run these tests so often (on every push to a PR) that we're guaranteed to catch failing tests. (I would worry much more about missing tests than about ones that are not catching bugs in 1/100 runs)

Take this test for a point in case: https://github.com/o1-labs/snarkyjs/blob/d6b7995aa81070c1dcc78eb751ae70578ddeb97a/src/lib/testing/testing.unit-test.ts#L19-L24
I checks that the `nat` generator, when used twice to sample tuples `(x,y)`, generates at least one sample where `x === y`. In our current system where special values are created with 25% probability, and there are only a few special values, this should work most of the time but fail with low probability. However, it hasn't failed a single time in CI in about a month since it has been added, so this slight imperfection hasn't been a problem in practice

src/mina-signer/src/sign-zkapp-command.unit-test.ts
```diff
@@ -0,0 +1,297 @@
...
+  packed: { field: string; size: number }[];
+}) {
+  return {
+    fields,
+    packed: packed.map(({ field, size }) => [field, size] as [string, number]),
+  };
+}
+
+function fixVerificationKey(accountUpdate: AccountUpdate) {
+  // TODO we set vk to null since we can't generate a valid random one
```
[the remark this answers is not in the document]
mitschabaude: Actually I can now :D Since I did this PR, we've exposed a dummy vk to snarkyjs, which I can use here. I'll do that!

src/lib/account_update.ts
```diff
@@ -1017,6 +1017,10 @@ class AccountUpdate implements Types.AccountUpdate {
   }
   static fromJSON(json: Types.Json.AccountUpdate) {
     let accountUpdate = Types.AccountUpdate.fromJSON(json);
+    // TODO remove empty permissions hack
```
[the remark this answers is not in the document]
mitschabaude: This has since been fixed by merging current `develop` into snarkyjs, where this hack is no longer needed. Will remove this when merging `main` here

src/lib/testing/property.ts
mitschabaude: no, they aren't. I think this would come at two significant costs which IMO make it not worth pursuing:
- large engineering costs of refactoring to RNGs that take a seed
- significant cost on sample diversity. some tests are so slow that they are only run with 15 samples - the minimum. if the samples are different every time, we get diversity by running this test many times (which we do, since it runs in CI all the time). By making it deterministic, we would lose that
- In practice, bugs that occur with a few % probability are exposed in almost every test run. Concretely, for all the bugs that I uncovered when testing this, I could reproduce these bugs in _every_ test run, which made them easy to find despite the test cases not being the same every time

mitschabaude: Also I should add that the testing framework prints out the failing samples when a test fails! This enables you to debug the failure no matter how elusive it may be

jspada: My second option was that it dumps all the details required to reproduce the test.  This satisfies my requirement on reproducibility. 👍🏼

src/lib/testing/random.ts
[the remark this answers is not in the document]
jspada: Consistency is extremely important for the developer and it concerns me something could sneak by.  Without updating your generic framework, I would strongly suggest that the zero/(special case) tests can always be hard-coded to always happen.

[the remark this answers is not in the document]
jspada: (as a separate kind of test)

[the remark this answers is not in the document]
mitschabaude: That's a good suggestion! It's actually possible to add hard-coded values (which are always tested first) to any random generator by using `Random.withHardCoded(generator, ...<list of hard-coded values>)`. I use this in few places currently but will check if I can use it in more places to add important test cases that we want to test with 100% certainty

src/lib/account_update.ts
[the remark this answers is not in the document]
mitschabaude: merged base branch & removed the hack!

src/lib/hash-input.unit-test.ts
```diff
@@ -201,3 +146,8 @@ function inputToOcaml({
     packed: packed.map(([field, size]) => ({ field, size })),
   };
 }
+
+function fixVerificationKey(accountUpdate: Json.AccountUpdate) {
+  // TODO we set vk to null since we can't generate a valid random one
+  accountUpdate.body.update.verificationKey = null;
```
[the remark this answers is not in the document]
mitschabaude: this doesn't necessarily refer to a null vk, but to an account update which doesn't change the verification key.
the test here just checks hashing of account updates.
testing that verification fails when an account has no verification key is more in scope for the Mina repo (since verification happens on-chain / snarkyjs isn't really concerned with verification)

src/mina-signer/src/sign-zkapp-command.ts
```diff
@@ -177,6 +183,35 @@ function accountUpdateFromFeePayer({
...
+  return { body, authorization: { signature } };
+}
+
+function isCallDepthValid(zkappCommand: ZkappCommand) {
+  let callDepths = zkappCommand.accountUpdates.map((a) => a.body.callDepth);
+  let current = callDepths.shift() ?? 0;
+  if (current !== 0) return false;
+  for (let callDepth of callDepths) {
+    if (callDepth < 0) return false;
+    if (callDepth - current > 1) return false;
```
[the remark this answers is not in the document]
mitschabaude: no -- call depth can stay the same from one account update to another, which would mean they are siblings in the tree

src/mina-signer/src/sign-zkapp-command.unit-test.ts
```diff
@@ -0,0 +1,297 @@
...
+    );
+    expect(Signature.toBase58(sigTestnet)).toEqual(sigTestnetOcaml);
+    expect(Signature.toBase58(sigMainnet)).toEqual(sigMainnetOcaml);
+
+    let verify = (s: Signature, id: NetworkId) =>
+      verifyFieldElement(s, fullCommitment, feePayerAddress, id);
+    expect(verify(sigTestnet, 'testnet')).toEqual(true);
+    expect(verify(sigTestnet, 'mainnet')).toEqual(false);
+    expect(verify(sigMainnet, 'testnet')).toEqual(false);
+    expect(verify(sigMainnet, 'mainnet')).toEqual(true);
```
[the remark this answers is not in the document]
mitschabaude: I already checked that the signatures are equal to the ones created from OCaml. So OCaml verification will yield the same unless the OCaml implementation has a bug. I'd like to leave that to the OCaml-side unit tests because OCaml verification is not, at the moment, exposed to snarkyjs

src/mina-signer/src/transaction-hash.unit-test.ts
```diff
@@ -23,169 +23,131 @@ import { Memo } from './memo.js';
...
-  let command = {
-    signer: PublicKey.fromBase58(payment.data.body.source),
-    signature: Signature.fromBase58(payment.signature),
-    payload,
-  };
-  let paymentBytes1 = SignedCommand.toBytes(command);
-  expect(JSON.stringify(paymentBytes1)).toEqual(JSON.stringify(paymentBytes0));
+    // payment hash
+    let digest0 = Test.transactionHash.hashPayment(ocamlPayment);
+    let digest1 = hashPayment(payment, { berkeley: true });
```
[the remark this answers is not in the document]
mitschabaude: non-berkeley is tested further below ("payment v1")

src/mina-signer/src/transaction-hash.unit-test.ts
```diff
@@ -23,169 +23,131 @@ import { Memo } from './memo.js';
...
+  hashPayment(payment, { berkeley: true });
+  // for "berkeley" hashing, it's fine if the signature is invalid because it's not part of the hash
+  // => make invalid signatures fail independently
+  Signature.fromBase58(payment.signature);
+});
+test.negative(RandomTransaction.signedDelegation.invalid!, (delegation) =>
+  hashStakeDelegation(delegation)
+);
+test.negative(RandomTransaction.signedDelegation.invalid!, (delegation) => {
+  hashStakeDelegation(delegation, { berkeley: true });
```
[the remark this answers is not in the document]
mitschabaude: tested 3 lines further above :D `berkeley: false` is the default

src/provable/binable.unit-test.ts
mitschabaude: the + is not modulo something, so `x + 2n` would then just exceed the max uint32, which is fine for this test

src/lib/testing/random.ts
```diff
@@ -6,62 +6,103 @@ import {
...
     'None',
     'Proof',
     'Signature',
     'Either',
     'Impossible'
   ),
-  Bigint.AuthRequired.fromJSON
+  AuthRequired.fromJSON
+);
+// TODO non ascii strings in zkapp uri and token symbol fail
```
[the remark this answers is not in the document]
mitschabaude: I was able to fix the issue with non ascii strings! https://github.com/o1-labs/snarkyjs/pull/717/commits/7fa2ff1098884f9a26c0bf36007171259de4a042

PR #718, opened 2023-02-08
Property testing for account updates

mitschabaude:
This adds a [quickcheck-style](https://en.wikipedia.org/wiki/QuickCheck) testing lib based on random generation of many examples. The main goal is being able to generate many non-trivial examples of complex structures like account updates. To this end, the testing lib is integrated with the deriver code, so the combinators for account updates don't need to be specified manually.

### Why not use an existing property-based testing framework, like [fast-check](https://github.com/dubzzz/fast-check)?

* integration with deriver code would've been a pain
* we would've had to write many generators ourselves, anyway, to create things like random fields or public keys
* if you strip away the fancier stuff like shrinking failure cases, it's pretty simple. it's all about creating good random values.
* if this is going to be useful outside of mina-signer -- for example, creating transactions that are actually valid on the ledger -- it will need some _very_ custom tooling, which is always simpler to build from scratch than on top of an opinionated lib

### Which issue does this solve right now?

* testing a much greater variety of zkapp transactions was one of the feedback items form the mina-signer audit, and this will make it very simple and nice to maintain

### DIFF
--- a/src/lib/testing/random.ts
+++ b/src/lib/testing/random.ts
@@ -16,7 +16,7 @@ import { alphabet } from '../../provable/base58.js';
 import { bytesToBigInt } from '../../js_crypto/bigint-helpers.js';
 import { Memo } from '../../mina-signer/src/memo.js';
 
-export { Random };
+export { Random, withHardCoded };
 
 type Random<T> = {
   create(): () => T;
@@ -43,7 +43,7 @@ const AuthorizationKind = reject(
   (t) => !!t.isProved && !!t.isSigned
 );
 const AuthRequired = map(
-  oneof<Json.AuthRequired[]>(
+  oneOf<Json.AuthRequired[]>(
     'None',
     'Proof',
     'Signature',
@@ -58,7 +58,7 @@ const Events = map(
   Bigint.Events.fromList
 );
 const SequenceEvents = Events;
-const SequenceState = oneof(Bigint.SequenceState.emptyValue(), Field);
+const SequenceState = oneOf(Bigint.SequenceState.emptyValue(), Field);
 const ZkappUri = map(string(nat(50)), Bigint.ZkappUri.fromJSON);
 
 const PrimitiveMap = primitiveTypeMap<bigint>();
@@ -115,7 +115,8 @@ const Random = Object.assign(Random_, {
   array,
   record,
   map,
-  oneof,
+  oneOf,
+  withHardCoded,
   field: Field,
   bool: Bool,
   uint32: UInt32,
@@ -245,10 +246,10 @@ function string(size: number | Random<number>) {
 }
 
 function base58(size: number | Random<number>) {
-  return map(array(oneof(...alphabet), size), (a) => a.join(''));
+  return map(array(oneOf(...alphabet), size), (a) => a.join(''));
 }
 
-function oneof<Types extends readonly any[]>(
+function oneOf<Types extends readonly any[]>(
   ...values: { [K in keyof Types]: Types[K] | Random<Types[K]> }
 ): Random<Types[number]> {
   type T = Types[number];
@@ -311,6 +312,19 @@ function reject<T>(rng: Random<T>, isRejected: (t: T) => boolean): Random<T> {
           let t = next();
           if (!isRejected(t)) return t;
         }
+      };
+    },
+  };
+}
+
+function withHardCoded<T>(rng: Random<T>, ...hardCoded: T[]): Random<T> {
+  return {
+    create() {
+      let next = rng.create();
+      let i = 0;
+      return () => {
+        if (i < hardCoded.length) return hardCoded[i++];
+        return next();
       };
     },
   };
--- a/src/lib/testing/random.ts
+++ b/src/lib/testing/random.ts
@@ -1,0 +1,259 @@
+import {
+  customTypes,
+  Layout,
+  TypeMap,
+  Json,
+  AccountUpdate,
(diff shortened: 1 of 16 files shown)


src/lib/testing/random.ts
```diff
@@ -0,0 +1,454 @@
+import {
```
mitschabaude, line 1: most of the love went into this file

src/provable/field-bigint.ts
```diff
@@ -110,7 +116,7 @@ const Sign = pseudoClass(
   function Sign(value: 1 | -1): Sign {
     if (value !== 1 && value !== -1)
       throw Error('Sign: input must be 1 or -1.');
-    return (BigInt(value) % Fp.modulus) as Sign;
+    return mod(BigInt(value), Fp.modulus) as Sign;
```
mitschabaude, line 119: fixes a bug that was already uncovered by the one simple test I added :O

src/provable/from-layout.ts
```diff
@@ -409,66 +389,6 @@ function ProvableFromLayout<
     );
   }
 
-  type FoldSpec<T, R> = {
```
mitschabaude, line 412: this code was just moved out of a function to use it elsewhere, but not really changed

mitschabaude (commented):
some comments for reviewers

src/provable/base58.unit-test.ts
```diff
@@ -1,28 +1,26 @@
...
   // check roundtrip
   let recoveredBytes = fromBase58Check(base58, version);
   expect(recoveredBytes).toEqual(bytes);
-}
-
-check([0, 1, 2, 3, 4, 5], 10);
-check([250, 200, 150, 100, 50, 0], 1);
-check([0, 0], 0);
-check([1], 100);
-check([], 0x01);
```
bkase, line 25: Can you add optional argument to specify examples that must be tested? The Jane street ocaml Quickcheck library has this and it’s nice for these kinds of situations where you have specific test cases you want to include with 100% probability.

bkase (commented):
We probably need quick check tests for your quick check framework 🔃

I’m torn because this is clearly needed and you have good reasons for building out something from scratch (and it’s not much code either), but, on the other hand, this feels maybe big enough that it could use some design doc outlining the approach/alternatives before committing to this direction. On the other other hand, we can always do that later and rewrite or port to another library if necessary.

mitschabaude:
> do we treat this as a prototype and follow up with something more refined? Or just leave it as is? (Which could just mean writing about it or it could entail you or others discovering some different approach taken)

Excellent question. My plan was mainly to use this for some immediate use cases:
* generate random valid payments / stake delegations. this is possible already
* generate random valid zkapp transactions. this needs just a bit more work because the account update `callDepth` is a global property of the transaction (so can't be set randomly locally) and because I want to hit cases where the same public key is used several times (won't happen currently because public keys are completely random)
* help with negative tests by generating _invalid_ cases.

all of those are easy to solve with minor additions to the current framework. as such, I'm planning to use this as is, with minor changes, and add additional conveniences / use it in more tests as needed in the future.

given all this, I think there's an implicit commitment to this architecture and I should write up a small RFC?

src/provable/base58.unit-test.ts
mitschabaude: good idea! I'll add a combinator that takes a generator and enhances it to create some constant values unconditionally, as the first outputs of `next()`

nicc:
It looks very nicely implemented! I do have a similar concern to Brandon's, namely that we'd be committing a litle too casually to some fairly consequential decisions. I'd like a clearer sense of where we'd use this (all tests, only ones that simulate transactions, ?), and what we'd trade off with other approaches. Something like a smallish RFC would probably be the way forward from here. You up for putting that together?

An interesting thing just happened at the process level. I hope you don't mind me commenting on it here. It's always worth noting instances where work seems to appear unexpectedly like this sudden need for an RFC. It doesn't make the work wrong or bad necessarily, but it does make it surprising. So how could this be less surprising? I like to think of the act of committing to work as occuring in stages that present specific resistance barriers. Quick example: First we add an issue to the backlog, which requires very little work or commitment. We only put in the effort to properly write out clear intentions and value if we believe in that work. Then we only force the tough priority choices if we feel strongly enough to advocate for it happening soon. Only then do we put together design docs and elicit feedback, which is a necessary but fairly expensive process. There are useful barriers to commitment at each stage that focus us on the most important things and mamnage the amount of upfront work required (a clear expression of Just-In-Time from lean production principles). So how does that relate here? We find ourselves at a fairly late stage, committed to fairly expensive work (an RFC) without having benefitted from the preceeding gating steps. By skipping some friction we have lost the opportunity to choose how much effort we should spend. 

It may well be that we'd have ended up here anyway and I'm up for quickly putting together a design doc and moving on. Just wanted to point out that we take a bigger gamble than necessary with our effort if we skip those moments of friction along the way. Hope this lands okay and checks out for you! Feel free to query or challenge as always.

src/provable/base58.unit-test.ts
mitschabaude: done! https://github.com/o1-labs/snarkyjs/pull/718/commits/4082beab9f8eaac915cd76884e49683dc3eab636

nicc (approved):
We had some concerns that were highlighted and discussed in [this RFC](https://github.com/o1-labs/snarkyjs/issues/721). Moving forward based on the understanding we arrived at there.

👍

PR #736, opened 2023-02-21
add vk hash to graphql account query 

Trivo25:
We weren't fetching the hash of the vk with `fetchAccount`, resulting in the `parseFetchedAccount` function trying to parse `Field(hash)` unsuccessfully, because hash was undefined https://github.com/o1-labs/snarkyjs/blob/berkeley/src/lib/fetch.ts#L249


Trivo25:
closing, rebasing off of different branch

PR #737, opened 2023-02-21
add vk hash to account query

Trivo25:
We weren't fetching the hash of the vk with `fetchAccount`, resulting in the `parseFetchedAccount` function trying to parse `Field(hash)` unsuccessfully, because hash was undefined https://github.com/o1-labs/snarkyjs/blob/berkeley/src/lib/fetch.ts#L249

Just realized there exists an issue for that https://github.com/o1-labs/snarkyjs/issues/725
this PR is just a quick short term fix so I can get the integration test up and running


mitschabaude (approved):
Yay, this was simpler than I thought :D

PR #741, opened 2023-02-21
Account deriver

mitschabaude:
closes https://github.com/o1-labs/snarkyjs/issues/725

* this replaces the existing main account type in snarkyjs with one that is auto-generated from ocaml. it does not replace the `FetchedAccount` type which reflects the structure of the account query result.
* new file `src/lib/mina/account.ts` which contains all account types, plus logic to convert from `FetchedAccount` to `Account`.
* Some missing fields in `FetchedAccount` were added to be able to fill them in when converting. Incidentally, this means that the `provedState` precondition is now supported (that was one of the missing fields)
* Because it was a quick fix, also resolves https://github.com/o1-labs/snarkyjs/issues/638 (see Mina PR)


src/provable/gen/transaction-bigint.ts
```diff
@@ -490,3 +511,51 @@ type AccountUpdate = {
...
+    setVerificationKey: AuthRequired;
+    setZkappUri: AuthRequired;
+    editSequenceState: AuthRequired;
+    setTokenSymbol: AuthRequired;
+    incrementNonce: AuthRequired;
+    setVotingFor: AuthRequired;
+  };
+  zkapp?: {
+    appState: Field[];
+    verificationKey?: {
```
Trivo25, line 547: Just a question: can the vk even be undefined if the zkapp field exists on an account? Or maybe to phrase it different: what makes an account have a zkapp field? Is it the vk that deployed to it? just a tiny bit of curiosity :D

src/provable/gen/transaction.ts
```diff
@@ -490,3 +511,51 @@ type AccountUpdate = {
 let AccountUpdate = provableFromLayout<AccountUpdate, Json.AccountUpdate>(
   jsLayout.AccountUpdate as any
 );
+
+type Account = {
```
Trivo25, line 515: I think i've seen this type be define like 3 times, couldn't we export it once and re use it?

Trivo25 (approved):
Awesome! Just some minor questions :)

src/lib/fetch.ts
```diff
@@ -101,6 +101,8 @@ async function fetchAccountInternal(
       },
     };
   }
+  console.log(fetchedAccount);
```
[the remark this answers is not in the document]
mitschabaude: good catch!

src/provable/gen/transaction.ts
mitschabaude: It's auto-generated in three places :D Once in a "JSON" version, once in a "bigint" version (where field elements are bigints, for mina-signer) and once the "normal snarkyjs" version where field elements are `Field` instances
We could make it generic, but I think since it's auto-generated its fine (also easier to read than generic)

src/provable/gen/transaction-bigint.ts
mitschabaude: yes, you can set app state even without setting the vk. I think the `zkapp` field exists as soon as one of those fields is written to, and they aren't really coupled to each other (except `provedState` which can only become true when a VK exists and appState is written to)

src/provable/gen/transaction.ts
Trivo25: okay true, thats fine :D thanks!

PR #746, opened 2023-02-23
Release 0.9

mitschabaude:
- updates the changelog with changes from develop
- makes the new validWhile precondition actually work
- renames `validWhile` to `globalSlot`


CHANGELOG.md
```diff
@@ -17,12 +17,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+## [0.9.0](https://github.com/o1-labs/snarkyjs/compare/c5a36207...HEAD)
+
 ### Added
 
 - `Transaction.fromJSON` to recover transaction object from JSON https://github.com/o1-labs/snarkyjs/pull/705
 - New precondition: `provedState`, a boolean which is true if the entire on-chain state of this account was last modified by a proof https://github.com/o1-labs/snarkyjs/pull/741
   - Same API as all preconditions: `this.account.provedState.assertEquals(Bool(true))`
   - Can be used to assert that the state wasn't tampered with by the zkApp developer using non-contract logic, for example, before deploying the zkApp
+- New on-chain value `globalSlot`, to assert that a method can only in a certain time range https://github.com/o1-labs/snarkyjs/pull/649
```
nicc, line 30: ... can only _what_ in a certain time range? Perform account updates? I think there's a word or two missing :)

src/lib/account_update.ts
```diff
@@ -632,7 +632,7 @@ class AccountUpdate implements Types.AccountUpdate {
     undefined;
   account: Precondition.Account;
   network: Precondition.Network;
-  validWhile: Precondition.ValidWhile;
+  globalSlot: Precondition.ValidWhile;
```
nicc, line 635: It's a bit confusing to use `globalSlot` as the public term but keep `ValidWhile` as the internal term in the code. Should this be updated everywhere? Seems like one of those things that doesn't appear to be a big deal but gets annoying and confusing in time.

src/lib/account_update.ts
```diff
@@ -660,7 +660,7 @@ class AccountUpdate implements Types.AccountUpdate {
     );
     this.account = account;
     this.network = network;
-    this.validWhile = validWhile;
+    this.globalSlot = validWhile;
```
nicc, line 663: As above.

src/lib/precondition.ts
```diff
@@ -309,6 +326,10 @@ function assertPreconditionInvariants(accountUpdate: AccountUpdate) {
 
     // we accessed a precondition field but not constrained it explicitly - throw an error
     let hasAssertBetween = isRangeCondition(precondition);
+    // TODO: maybe the "validWhile" should be changed to "globalSlot" in a more global way?
```
nicc, line 329: +1. If we're in a hurry it could be deferred I guess but this is the kind of debt that's hard to prioritise at a later date so I'm in favour of doing it now.

nicc (commented):
Looks good I think. Basically refactoring validWhile precondition subclass into its own function and renaming to globalSlot? Would it make sense to call it something like `globalSlotRange` or `globalSlotBetween` to be a bit more explicit? And I'm in favour of doing the rename more globally now. What do you think? I know it's more effort but I think this kind of debt is annoying to pay back later.

nicc:
Do we need a matching cli version bump?

ymekuria:
> Do we need a matching cli version bump?

We also need to update the CLI, the CLI examples/template contacts and test all the flows with the `SnarkyJS  0.9.0`

mitschabaude:
I addressed the feedback by renaming validWhile to globalSlot in more places. It's not possible to do everywhere, though, because the original OCaml code still uses "validWhile" and some of this TS code relies on structures derived from OCaml, so has to use the OCaml names. I think I want as far as possible under that constraint.

> Would it make sense to call it something like globalSlotRange or globalSlotBetween to be a bit more explicit?

I'm sticking with globalSlot because the intention was to make the field sound more like a "value", so that an expression like `this.globalSlot.get()` makes sense -- you're getting the current global slot. You can assert that the global slot lies in a range, but that doesn't make the global slot itself a range - it's just one value.

Trivo25 (approved):

ymekuria (approved):

PR #754, opened 2023-02-27
Fix error on zkapp.state.get()

mitschabaude:
- avoid 'can only run in checked computation' error
- 0.9.1


ymekuria (approved):
LGTM! Thanks for fixing this!

PR #787, opened 2023-03-10
Add field type for tokenId in fetchAccount

rpanic:


src/lib/fetch.ts
```diff
@@ -67,8 +68,12 @@ async function fetchAccount(
     accountInfo.publicKey instanceof PublicKey
       ? accountInfo.publicKey.toBase58()
       : accountInfo.publicKey;
+  let tokenIdBase58 = 
+    accountInfo.tokenId instanceof Field
```
mitschabaude, line 72: unfortunately we can't use `instanceof Field` (because the Field class is proxied, so this actually won't be true for fields)

but `typeof tokenId === "string"` will do it!

mitschabaude (changes requested):
thanks for this 🙌🏻  just one comment then this should be good to go

mitschabaude (approved):

PR #788, opened 2023-03-13
Add get actions to network API

MartinMinkov:
## Description
Adds `getActions` to the network API. Uses a GraphQL server implements the [following schema](https://github.com/o1-labs/Archive-Node-API/blob/main/schema.graphql)

`getActions` is used in `methods,` so it cannot be asynchronous. Instead, we use the caching system already wired up to fetch account and network data. We create a new action cache and add it to the promises executed in `fetchMissingData`. When `fetchMissingData` is called, `fetchActions` queries the GraphQL schema and then caches the results, which can then be used by `method`s of a smart contract in a non-asynchronous way.

## Tested
Manual Testing at these addresses:
https://berkeley.minaexplorer.com/wallet/B62qnL4b3wuAPwFb5E7ChYgDJFdxZgmud2kkx2N8RzNaq4ykagfzumy
https://berkeley.minaexplorer.com/wallet/B62qp7bmbYVpuhqRFmmMDGWcZQBavf9T5Lo1Jfj7yjSaPx2awYcTTvF
https://berkeley.minaexplorer.com/wallet/B62qp2SfYeLBo3p1ZwMS33qtqn6qCnAjAMXzTy6ZAYzYG8DLTr8xqhS

## Discussion
Do we want to change the return types for actions to include the extra data returned from the GraphQL server? I left the data inside the query, but if it's not essential, I will remove it to save on network traffic.

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -536,7 +536,7 @@ function sendZkappQuery(json: string) {
 }
 `;
 }
-type FetchedEventActionBase = {
+type FetchedEvents = {
   blockInfo: {
     distanceFromMaxBlockHeight: number;
     globalSlotSinceGenesis: number;
@@ -550,18 +550,16 @@ type FetchedEventActionBase = {
     memo: string;
     status: string;
   };
-};
-type FetchedEvents = {
   eventData: {
     data: string[];
   }[];
-} & FetchedEventActionBase;
+};
 type FetchedActions = {
+  actionState: string;
   actionData: {
     data: string[];
   }[];
-  actionState: string;
-} & FetchedEventActionBase;
+};
 
 type EventActionFilterOptions = {
   to?: UInt32;
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -21,12 +21,14 @@ export {
   parseFetchedAccount,
   markAccountToBeFetched,
   markNetworkToBeFetched,
+  markActionsToBeFetched,
   fetchMissingData,
   fetchTransactionStatus,
   TransactionStatus,
   EventActionFilterOptions,
   getCachedAccount,
   getCachedNetwork,
+  getCachedActions,
   addCachedAccount,
   defaultGraphqlEndpoint,
   archiveGraphqlEndpoint,
@@ -36,6 +38,7 @@ export {
   sendZkapp,
   removeJsonQuotes,
   fetchEvents,
+  fetchActions,
 };
 
 let defaultGraphqlEndpoint = 'none';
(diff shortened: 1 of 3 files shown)


Trivo25:
> Do we want to change the return types for actions to include the extra data returned from the GraphQL server? I left the data inside the query, but if it's not essential, I will remove it to save on network traffic.

I would approach this similar to `fetchEvents`. When someone invokes `Mina.fetchEvents`, they get some more information whereas when they invoke it within a smart contract, we would just get the action data

src/lib/fetch.ts
```diff
@@ -610,6 +715,54 @@ async function fetchEvents(
...
+    ),
+    graphqlEndpoint
+  );
+  if (error) throw Error(error.statusText);
+  let fetchedActions = response?.data.actions as FetchedActions[];
+  if (fetchedActions === undefined) {
+    return {
+      error: {
+        statusCode: 404,
+        statusText: `fetchActions: Account with public key ${publicKey} does not exist.`,
```
Trivo25, line 742: should we specify the tokenId in the error message here as well? This could cause some confusion if the address exists but just the particular token account not(?)

Trivo25 (approved):
lgtm!

src/lib/fetch.ts
MartinMinkov: Will do!

src/lib/fetch.ts
```diff
@@ -188,6 +219,17 @@ async function fetchMissingData(graphqlEndpoint: string) {
...
     }
   );
+  let actionPromises = Object.entries(actionsToFetch).map(
+    async ([key, { publicKey, tokenId }]) => {
+      let response = await fetchActions(
+        { publicKey, tokenId },
+        archiveEndpoint
+      );
+      if ('error' in response && response.error === undefined)
+        delete actionsToFetch[key];
```
mitschabaude, line 229: this logic seems wrong to me: we want to delete from `actionsToFetch` if and only if the fetch was successful. that means checking for `if (response.error === undefined)`, or, if TS doesn't like that, `if (!('error' in response) || response.error === undefined)`

src/lib/mina.ts
```diff
@@ -797,9 +797,21 @@ function Network(input: { mina: string; archive: string } | string): Mina {
...
+      if (
+        !currentTransaction.has() ||
+        currentTransaction.get().fetchMode === 'cached'
+      ) {
+        let actions = Fetch.getCachedActions(publicKey, tokenId);
+        if (actions !== undefined) return actions;
+      }
       throw Error(
-        'fetchEvents() is not implemented yet for remote blockchains.'
+        `getActions: Could not find actions for the for public key ${publicKey}`
```
mitschabaude, line 814: typo: "Could not find actions for the ~~for~~ public key"

src/lib/fetch.ts
```diff
@@ -610,6 +715,54 @@ async function fetchEvents(
...
+        statusCode: 404,
+        statusText: `fetchActions: Account with public key ${publicKey} with tokenId ${tokenId} does not exist.`,
+      },
+    };
+  }
+
+  const actionData = fetchedActions
+    .map((action) => {
+      return {
+        hash: Ledger.fieldToBase58(Field(action.actionState)),
```
mitschabaude, line 750: doesn't need to be fixed here, but it seems to me the field => base58 conversion is done both and in the local blockchain version, and would save computation to just leave the hash as a field element in both cases

mitschabaude (approved):
Look great! Approving in advance with one comment that needs to be addressed

src/lib/fetch.ts
MartinMinkov: I think that's a good idea, but I have to change a bunch of type definitions to make that work :/ I'll create an issue so it can be addressed later.

PR #801, opened 2023-03-20
Support actions being dispatched from multiple AccountUpdates

MartinMinkov:
## Description:
Fixes the format from fetched actions to support multiple account updates with events in the same transaction.

Addresses https://github.com/o1-labs/snarkyjs/issues/798

## Tested
All listed zkApp accounts have successful transactions. To test, two transactions were issued with varying AccountUpdates and actions emitted. Then a `reduce` was given, with success on Berkeley.

- zkApp emitting single action in a single Account Update:
https://berkeley.minaexplorer.com/wallet/B62qp2R75kU97Sovki9987LNiDDcJkivRhPnpizvpzHWnhkXh3vqjt4
`[ [ '1' ], [ '1' ] ]` -- getAction return value for two zkApp transactions with 1 Account Update each

- zkApp emitting two actions in a single AccountUpdate:
https://berkeley.minaexplorer.com/wallet/B62qqNfTahass2YQfvgCs4L595gRnraadMGuK8bYsBK5VG2BdYEqhVw
`[ [ '1', '1' ], [ '1', '1' ] ]` -- getAction return value for two zkApp transactions with 1 Account Update each

- zkApp emitting single action with 3 Account Updates in a single transaction
https://berkeley.minaexplorer.com/wallet/B62qpYzTFqB7FcF79dyytvJ21jRUCxRespJVU2ZewV4CAEVAneop2BM
`[ [ '1' ], [ '1' ], [ '1' ], [ '1' ], [ '1' ], [ '1' ] ]` -- getAction return value for two zkApp transactions with 3 Account Updates each

- zkApp emitting two actions in 3 AccountUpdates in a single transaction:
https://berkeley.minaexplorer.com/wallet/B62qreiFZdVDNuXKVdujzJnqERfuxS5Wy4RQeNkJqhkjZCFibm25Qja
`[
  [ '1', '1' ],
  [ '1', '1' ],
  [ '1', '1' ],
  [ '1', '1' ],
  [ '1', '1' ],
  [ '1', '1' ]
]` -- getAction return value for two zkApp transactions with 3 Account Updates each

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -796,7 +796,11 @@ async function fetchActions(
 
     if (currentActionHash !== expectedActionHash) {
       throw new Error(
-        `Failed to derive correct actions hash for ${publicKey}. Derived hash: ${currentActionHash}, expected hash: ${expectedActionHash})`
+        `Failed to derive correct actions hash for ${publicKey}.
+        Derived hash: ${currentActionHash}, expected hash: ${expectedActionHash}).
+        All action hashes derived: ${JSON.stringify(actionsList, null, 2)}
+        Please try a different Archive Node API endpoint.
+        `
       );
     }
   });
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -731,27 +731,34 @@ async function fetchActions(
     };
   }
 
-  const actionData = fetchedActions
-    .map((action) => {
-      const actionMap = action.actionData
-        .sort((a, b) => {
-          return Number(a.accountUpdateId) - Number(b.accountUpdateId);
-        })
-        .reduce((actionMap, action) => {
-          if (actionMap.has(action.accountUpdateId)) {
-            actionMap.get(action.accountUpdateId)?.concat(action.data);
-          } else {
-            actionMap.set(action.accountUpdateId, action.data);
-          }
-          return actionMap;
-        }, new Map<string, string[]>());
+  let actionMap = new Map<
+    string,
+    {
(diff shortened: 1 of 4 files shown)


src/lib/fetch.ts
```diff
@@ -730,44 +730,61 @@ async function fetchActions(
...
+    });
+
+    const currentActionHash = Ledger.fieldToBase58(Field(latestActionsHash));
+    const expectedActionHash = Ledger.fieldToBase58(Field(actionState));
 
-  // Reverse the order of actions since the API returns in descending order of block height while Localblockchain pushes new actions to end of array.
-  const actionData = [...actionMap.values()].reverse();
+    if (currentActionHash !== expectedActionHash) {
+      throw new Error(
+        `Failed to derive correct actions hash for ${publicKey}. Derived hash: ${currentActionHash}, expected hash: ${expectedActionHash})`
```
Trivo25, line 799: I wonder if we can tell the user what to do in that case: like "try a different archive API" or maybe the previous tx hasnt been included yet?

Trivo25: Maybe also let the user know what actions we found so its easier to debug

Trivo25 (approved):
Just confirmed it works with a contract that didn't work yesterday, good job!

mitschabaude (approved):
Lgtm!

src/lib/fetch.ts
MartinMinkov: Will do!

PR #812, opened 2023-03-27
Fix self-updating verification key

### LINKED ISSUE #813: Wrong verification key hash when updating the verification key from within a zkApp method

mitschabaude:
fixes #813 - updating of the zkapp verification key
there was a bug where the update was using the new (not old) verification key hash inside `body.authorizationKind`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/snarkyjs/compare/21de489...HEAD)
 
-> No unreleased changes yet
+### Fixed
+
+- Update the zkApp from within one of its own methods, via proof https://github.com/o1-labs/snarkyjs/pull/808
 
 ## [0.9.4](https://github.com/o1-labs/snarkyjs/compare/9acec55...21de489)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixed
 
-- Update the zkApp from within one of its own methods, via proof https://github.com/o1-labs/snarkyjs/pull/812
+- Update the zkApp verification key from within one of its own methods, via proof https://github.com/o1-labs/snarkyjs/pull/812
 
 ## [0.9.4](https://github.com/o1-labs/snarkyjs/compare/9acec55...21de489)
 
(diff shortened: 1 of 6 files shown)


src/lib/account_update.ts
```diff
@@ -1732,13 +1732,24 @@ const Authorization = {
...
   ) {
     body.authorizationKind.isSigned = Bool(false);
     body.authorizationKind.isProved = Bool(true);
     let hash = Circuit.witness(Field, () => {
-      priorAccountUpdates ??= zkAppProver.getData().transaction.accountUpdates;
+      let proverData = zkAppProver.getData();
+      let isProver = proverData !== undefined;
+      if (!isProver && priorAccountUpdates === undefined) {
+        throw Error(
+          'invariant violation: only the prover can call `setProofAuthorizationKind()` without passing in `priorAccountUpdates`'
```
MartinMinkov, line 1745: How can the user run into this error? I tried to reproduce it but I don't quite understand how to trigger it :sweat_smile:  

`setProofAuthorizationKind` seems only to happen when we analyze `method`s. Could we come up with a better error message that gives guidance on what the potential issues could be? The error message has some internal parameters used, which could be confusing without knowing the code of SnarkyJS. 

`priorAccountUpdates` is confusing since that mostly seems like an internal thing?

CHANGELOG.md
```diff
@@ -17,7 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/snarkyjs/compare/21de489...HEAD)
 
-> No unreleased changes yet
+### Fixed
+
+- Update the zkApp from within one of its own methods, via proof https://github.com/o1-labs/snarkyjs/pull/812
```
MartinMinkov, line 22: nit: `Update the zkApp from within` -> `Update the zkApp verification key from within`

MartinMinkov (approved):
I tested the script, runs excellently. LGTM :D

Trivo25 (approved):

src/lib/account_update.ts
mitschabaude: The error is not supposed to ever happen :D I just added this as an 'assert' to make sure that we notice when the assumptions here are violated

MartinMinkov: Got it! Do you think it would be of any value to indicate that this error should never happen and means something is really broken in a more obvious way? 

Maybe decreasing unknown error fatigue would be nice, and give a message that lets users know that it's not the fault of their zkApp?

CHANGELOG.md
MartinMinkov, line 22:
```suggestion
- Update the zkApp verification key from within one of its own methods, via proof https://github.com/o1-labs/snarkyjs/pull/812
```

PR #817, opened 2023-03-28
Remove multiple best tip blocks for actions, if they exist

MartinMinkov:


Trivo25 (approved):
Nice! Quick and simple fix
> This fix will be removed once the related issue in the Archive-Node-API repository (https://github.com/o1-labs/Archive-Node-API/issues/7) is resolved

I wonder if it's worth waiting, whats the estimate on that?

MartinMinkov:
> Nice! Quick and simple fix
> 
> > This fix will be removed once the related issue in the Archive-Node-API repository ([o1-labs/Archive-Node-API#7](https://github.com/o1-labs/Archive-Node-API/issues/7)) is resolved
> 
> I wonder if it's worth waiting, whats the estimate on that?

Good question! To implement the best-tip selection algorithm that's used, we need a redeployment of Berkeley to happen. The Archive Node schema needed two required fields to compute the algorithm, so currently, we can't do it. Once we redeploy Berkeley and migrate the block data for the old blocks, we can implement the algorithm.

PR #825, opened 2023-04-03
Change fetchEvents to return transactionInfo inside events type

MartinMinkov:
# Description
🔗 Related Issue: https://github.com/o1-labs/snarkyjs/issues/806

This PR addresses an issue with the internal event type in SnarkyJS. Previously, the event handling assumed that all events were contained within the same zkApp transaction, which isn't always true. Events can be emitted in different zkApp transactions, and multiple zkApp transactions can be present within a block.

To resolve this, we have updated the internal event type to include event data and transaction information as separate objects. This allows us to provide accurate information about each event, taking into account the specific transaction it belongs to.

Please note that there is a related PR for the Archive Node API, which can be found here: https://github.com/o1-labs/Archive-Node-API/pull/68

Both PRs should be merged simultaneously to ensure proper coordination and integration.


mitschabaude (approved):
Nice change!

PR #826, opened 2023-04-03
[WIP] Fix `isSuccess` always returning true, even if transaction fails

MartinMinkov:
## Description
:link: Related Issue: https://github.com/o1-labs/snarkyjs/issues/819

Refactors some of the response and error checking logic when sending GraphQL requests. 

At the time of this PR, the Berkeley network is not including transactions in blocks due to a SNARK issue, so this has not been tested yet. This will be tested when Berkeley is accepting transactions again.

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -134,7 +134,7 @@ async function fetchAccountInternal(
 }
 
 type FetchConfig = { timeout?: number };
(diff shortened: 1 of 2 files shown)


nicc:
IRI: @Trivo25

Trivo25 (approved):

PR #828, opened 2023-04-04
Add endActionState and fromActionState to fetchActions

MartinMinkov:
# Description
:link: Resolves https://github.com/o1-labs/snarkyjs/issues/802

This PR introduces `fromActionState` and `endActionState` parameters to the `fetchActions` function in SnarkyJS. This enhancement allows us to fetch only the necessary actions to compute the latest actions state, instead of retrieving the entire actions history of a zkApp as done previously.

With these improvements, we no longer need to start from an empty action state and apply hashes to derive the current action state by analyzing each block from the beginning of a zkApp's history. Instead, we can utilize the new `actionStateTwo` field returned by the Archive Node API. As `actionStateTwo` represents the second latest action state calculated, it serves as a safe starting point for deriving the most recent action state. We can now apply AccountUpdates to the `actionStateTwo` value to compute the current action state.

Furthermore, users can now provide a `fromActionState` parameter to the Archive Node API, which will return actions only after the specified action state. This feature offers a filtering option in SnarkyJS that was previously unavailable.

Please note that this PR relies on https://github.com/o1-labs/Archive-Node-API/pull/71.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Improve number of constraints needed for Merkle tree hashing https://github.com/o1-labs/snarkyjs/pull/820
   - This breaks deployed zkApps which use `MerkleWitness.calculateRoot()`, because the circuit is changed
   - You can make your existing contracts compatible again by switching to `MerkleWitness.calculateRootSlow()`, which has the old circuit
+- Renamed Function Parameters: The `getAction` function now accepts a new object structure for its parameters. https://github.com/o1-labs/snarkyjs/pull/828
+  - The previous object keys, `fromActionHash` and `endActionHash`, have been replaced by `fromActionState` and `endActionState`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - The internal event type now includes event data and transaction information as separate objects, allowing for more accurate information about each event and its associated transaction.
 - Removed multiple best tip blocks when fetching action data https://github.com/o1-labs/snarkyjs/pull/817
   - Implemented a temporary fix that filters out multiple best tip blocks, if they exist, while fetching actions. This fix will be removed once the related issue in the Archive-Node-API repository (https://github.com/o1-labs/Archive-Node-API/issues/7) is resolved.
-- New `fromActionHash` and `endActionHash` parameters for fetchActions function in SnarkyJS https://github.com/o1-labs/snarkyjs/pull/828
-  - Allows fetching only necessary actions to compute the latest actions hash
+- New `fromActionState` and `endActionState` parameters for fetchActions function in SnarkyJS https://github.com/o1-labs/snarkyjs/pull/828
+  - Allows fetching only necessary actions to compute the latest actions state
   - Eliminates the need to retrieve the entire actions history of a zkApp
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - The internal event type now includes event data and transaction information as separate objects, allowing for more accurate information about each event and its associated transaction.
 - Removed multiple best tip blocks when fetching action data https://github.com/o1-labs/snarkyjs/pull/817
   - Implemented a temporary fix that filters out multiple best tip blocks, if they exist, while fetching actions. This fix will be removed once the related issue in the Archive-Node-API repository (https://github.com/o1-labs/Archive-Node-API/issues/7) is resolved.
+- New `fromActionHash` and `endActionHash` parameters for fetchActions function in SnarkyJS https://github.com/o1-labs/snarkyjs/pull/828
```
mitschabaude, line 33: could we rename to `fromActionState` and `endActionState`?

src/lib/mina.ts
```diff
@@ -148,6 +149,11 @@ type DeprecatedFeePayerSpec =
     })
   | undefined;
 
+type ActionHashes = {
```
mitschabaude, line 152: rename to `ActionStates`?

src/lib/mina.ts
```diff
@@ -340,6 +346,7 @@ interface Mina {
   ) => ReturnType<typeof Fetch.fetchEvents>;
   getActions: (
     publicKey: PublicKey,
+    actionHashes: ActionHashes,
```
mitschabaude, line 349: I would make this argument optional, otherwise users have to pass in an empty object

src/lib/mina.ts
```diff
@@ -581,6 +588,7 @@ function LocalBlockchain({
     },
     getActions(
       publicKey: PublicKey,
+      _actionHashes: ActionHashes,
```
mitschabaude, line 591: I'm afraid that ignoring the `_actionHashes` input here leads to inconsistent behavior between the `Mina.Network` and `Mina.LocalBlockchain` interfaces. Maybe you could move the logic in `SmartContract.getActions` which filters the actions here instead?

mitschabaude (approved):
Mostly looks good, although I would change the 'actionHash' terminology to 'actionState', and make sure the local and remote Mina interfaces are consistent.

A general comment is that we don't have good test coverage for `getActions`. Ideally, there'd be a unit test which confirms expected behavior on LocalBlockchain (for cases like multiple actions in an account update / in a block / computing action states from multiple blocks with actions). Plus a manual test which confirms it with an archive node, for the time being where the Berkeley test doesn't work in CI

src/lib/mina.ts
MartinMinkov: Moved the logic over! :D

CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - The internal event type now includes event data and transaction information as separate objects, allowing for more accurate information about each event and its associated transaction.
 - Removed multiple best tip blocks when fetching action data https://github.com/o1-labs/snarkyjs/pull/817
   - Implemented a temporary fix that filters out multiple best tip blocks, if they exist, while fetching actions. This fix will be removed once the related issue in the Archive-Node-API repository (https://github.com/o1-labs/Archive-Node-API/issues/7) is resolved.
+- New `fromActionState` and `endActionState` parameters for fetchActions function in SnarkyJS https://github.com/o1-labs/snarkyjs/pull/828
```
mitschabaude, line 33: could we also, under the "breaking changes" section, mention the API change in `getActions` from `fromActionHash` etc to `fromActionState`?

Issue #827, opened 2023-04-04
Add analyzeMethods() to ZkProgram


maht0rz:
Since i began using ZkProgram i have been trying to be cautious about not going over the 2^16 circuit size limit. The only way i knew how to figure out my circuit size, was to use `SmartContract.analyzeMethods()`. I could not find the same for ZkProgram, so i've decided to follow how it is done for SmartContract.

For my own purposes, i've tried adding the following to the [`ZkProgram`](https://github.com/o1-labs/snarkyjs/blob/main/src/lib/proof_system.ts#L270):

```typescript
function analyzeMethods() {
  return methodIntfs.map((methodEntry, i) =>
    analyzeMethod(publicInputType, methodEntry, methodFunctions[i])
  );
}

return Object.assign(
  selfTag,
  { compile, verify, digest, publicInputType, analyzeMethods },
  provers
);
```

^ the code above might not be 100% correct, but when i injected something along these lines into my built snarkyjs dependency, i was able to inspect the amount of 'rows' in my ZkProgram.

Does this approach make sense? If yes, i can open a PR.

mitschabaude:
Yes this makes sense and a PR would be highly appreciated!

mitschabaude:
done with the linked PR

PR #829, opened 2023-04-05
Add analyzeMethods() to ZkProgram

maht0rz:
This PR implements #827 and adds a simple unit test for the `proof_system.ts` to test it.

I have encountered a few issues when implementing this:
- i couldn't find an appropriate type/interface definition for `analyzeMethod`, so i've used its `ReturnType` instead.
- `run-unit-tests.sh` does not work on Mac due to using features available on bash v4+ (mac runs bash v3). I have installed bash v5 with brew and ran unit tests by hand like this to workaround this issue: `/opt/homebrew/bin/bash ./run-unit-tests.sh`. This could be solved for the long term by replacing `/bin/bash` in the shell scripts to respect the user env, [more information here](https://stackoverflow.com/a/56491222)

~~Question on the side - `.github/workflows/build-action.yml` does not seem to run unit tests. When/where do the unit tests run, assuming they're part of the CI?~~ Answer: `run-ci-tests.sh`

### DIFF
--- a/src/lib/proof_system.unit-test.ts
+++ b/src/lib/proof_system.unit-test.ts
@@ -1,0 +1,58 @@
+import { Field, isReady, shutdown } from '../snarky.js';
+import { Struct } from './circuit_value.js';
+import { UInt64 } from './int.js';
+import { ZkProgram } from './proof_system.js';
+import { expect } from 'expect';
+
+await isReady;
+
+const EmptyProgram = ZkProgram({
+  publicInput: Field,
+  methods: {
+    run: {
+      privateInputs: [],
+      method: (publicInput: Field) => {},
+    },
+  },
+});
+
+const emptyMethodsMetadata = EmptyProgram.analyzeMethods();
+emptyMethodsMetadata.forEach((methodMetadata) => {
+  expect(methodMetadata).toEqual({
+    rows: 0,
+    digest: '4f5ddea76d29cfcfd8c595f14e31f21b',
+    result: undefined,
+    gates: [],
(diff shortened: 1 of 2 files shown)


src/lib/proof_system.unit-test.ts
```diff
@@ -0,0 +1,58 @@
...
+      },
+    },
+  },
+});
+
+const incrementMethodMetadata = CounterProgram.analyzeMethods()[0];
+expect(incrementMethodMetadata).toEqual(
+  expect.objectContaining({
+    rows: 18,
+    digest: '62d893f727b12d540bdc483427cbd70b',
```
mitschabaude, line 54: can we remove the digest here and in the other test? otherwise this will break every time pickles internal circuits change

mitschabaude: ah, never mind, this just gets the circuit created from the method directly, without extra stuff added by pickles. should be fine

mitschabaude (approved):
Very cool, thanks for this addition!

mitschabaude:
@maht0rz one ask before I merge this - can you add your addition to CHANGELOG.md?

src/lib/proof_system.unit-test.ts
maht0rz: Can you please shine a bit more light in what you meant by ‘extra stuff added by pickles’? Do you mean the wrap circuit in SmartContract or something? 🧐

maht0rz:
@mitschabaude i will add it to the changelog and ping you once i’m done thanks!

PR #833, opened 2023-04-05
Rename all references of actionsHash to actionState

MartinMinkov:
# Summary
This pull request introduces breaking changes to improve naming consistency with the Mina protocol implementation. We've renamed all references of `actionsHash` to `actionState`. These changes affect variables, function parameters, and object keys throughout the codebase.

:link: Associated changes to the docs are located here: https://github.com/o1-labs/docs2/pull/345


Trivo25 (approved):
I wonder if we plan on releasing a minor version before the next big one, if that's the case imo we should wait with merging cc @jasongitmail

MartinMinkov:
> I wonder if we plan on releasing a minor version before the next big one, if that's the case imo we should wait with merging cc @jasongitmail
> edit: ah nvm, I dont think this touches any public APIs +1

The intention behind this PR was to be consistent with this PR too :P https://github.com/o1-labs/snarkyjs/pull/828

It renames the input parameters fromStateHash and endStateHash to fromStateAction and endStateAction. As a follow-up, we've also renamed actionHash to actionState for the reducer.

This update does affect the public API, but no worries - the new release is already planned as a breaking change due to the `getActions` renaming.

jasongitmail (approved):

mitschabaude:
reopened this because it isn't merged yet :D

mitschabaude (approved):

PR #807, opened 2023-03-23
Do not fail on assertions in test mode

mitschabaude:
Mina PR: https://github.com/MinaProtocol/mina/pull/12911
Snarky PR: https://github.com/o1-labs/snarky/pull/821

Fixes https://github.com/o1-labs/snarkyjs/issues/799
Fixes https://github.com/o1-labs/snarkyjs/issues/530

PR #840, opened 2023-04-10
Reintroduce #807

mitschabaude:
This reintroduces the changes from https://github.com/o1-labs/snarkyjs/pull/807 which were reverted for the last snarkyjs release.

In addition, it changes how `Circuit.runAndCheck` works: it now creates witnesses and checks constraints, similar to the prover. Previously, all of `runAndCheck` was run in an asProver block, which was an error in how this function was exposed to snarkyjs.

Fixes https://github.com/o1-labs/snarkyjs/issues/799
Fixes https://github.com/o1-labs/snarkyjs/issues/530


Trivo25 (approved):
yay :D

PR #844, opened 2023-04-11
Fix fetchActions

Comdex:
- Fix the issue mentioned in #843 
This problem is because the actionData array returned by the archive node is in descending order according to accoutUpdateId, which will cause problems in the order of using actions to calculate actionHash, and the actionData array needs to be reversed.

- Removed redundant code

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -805,7 +805,7 @@ async function fetchActions(
   let actionsList: { actions: string[][]; hash: string }[] = [];
 
   fetchedActions.forEach((fetchedAction) => {
-    const { actionData } = fetchedAction;
+    let { actionData } = fetchedAction;
     let latestActionsHash = Field(fetchedAction.actionState.actionStateTwo);
     let actionState = Field(fetchedAction.actionState.actionStateOne);
 
@@ -813,9 +813,12 @@ async function fetchActions(
       throw new Error(
         `No action data was found for the account ${publicKey} with the latest action state ${actionState}`
       );
-
+    
+    actionData.reverse();
     let { accountUpdateId: currentAccountUpdateId } = actionData[0];
     let currentActionList: string[][] = [];
+
+
 
     actionData.forEach((action, i) => {
       const { accountUpdateId, data } = action;
@@ -824,7 +827,6 @@ async function fetchActions(
 
       if (isSameAccountUpdate && !isLastAction) {
         currentActionList.push(data);
-        currentAccountUpdateId = accountUpdateId;
         return;
       } else if (isSameAccountUpdate && isLastAction) {
         currentActionList.push(data);
@@ -837,7 +839,7 @@ async function fetchActions(
           actions: currentActionList,
           hash: Ledger.fieldToBase58(Field(latestActionsHash)),
         });
-        currentAccountUpdateId = accountUpdateId;
+        
         currentActionList = [data];
       }
 


src/lib/fetch.ts
```diff
@@ -837,7 +839,7 @@ async function fetchActions(
           actions: currentActionList,
           hash: Ledger.fieldToBase58(Field(latestActionsHash)),
         });
-        currentAccountUpdateId = accountUpdateId;
```
mitschabaude, line 840: are we sure we want to remove this? if yes, why? in this `else` branch, we don't have `accountUpdateId === currentAccountUpdateId` so the line is not redundant

Comdex: Yes, sure, because the code after the end of this branch will also perform this operation. 
You can see the code comments below: 
```typescript
fetchedActions.forEach((fetchedAction) => {
    const { actionData } = fetchedAction;
    let latestActionsHash = Field(fetchedAction.actionState.actionStateTwo);
    let actionState = Field(fetchedAction.actionState.actionStateOne);

    if (actionData.length === 0)
      throw new Error(
        `No action data was found for the account ${publicKey} with the latest action state ${actionState}`
      );

    let { accountUpdateId: currentAccountUpdateId } = actionData[0];
    let currentActionList: string[][] = [];

    actionData.forEach((action, i) => {
      const { accountUpdateId, data } = action;
      const isLastAction = i === actionData.length - 1;
      const isSameAccountUpdate = accountUpdateId === currentAccountUpdateId;

      if (isSameAccountUpdate && !isLastAction) {
        currentActionList.push(data);
        currentAccountUpdateId = accountUpdateId;
        return;
      } else if (isSameAccountUpdate && isLastAction) {
        currentActionList.push(data);
      } else if (!isSameAccountUpdate && isLastAction) {
        // branch 3
        latestActionsHash = processActionData(
          currentActionList,
          latestActionsHash
        );
        actionsList.push({
          actions: currentActionList,
          hash: Ledger.fieldToBase58(Field(latestActionsHash)),
        });
        // The following line of code is redundant
        currentAccountUpdateId = accountUpdateId;
        currentActionList = [data];
      }

      latestActionsHash = processActionData(
        currentActionList,
        latestActionsHash
      );
      actionsList.push({
        actions: currentActionList,
        hash: Ledger.fieldToBase58(Field(latestActionsHash)),
      });
      // After the above branch 3 is executed, the currentAccoutUpdateId will also be updated here
      currentAccountUpdateId = accountUpdateId;
      currentActionList = [data];
    });
```

mitschabaude: thanks!

mitschabaude (approved):
The fix makes sense to me -- previously, we reversed the blocks of fetched actions, now we're also reversing them within each block. But would be nice @MartinMinkov if you could also approve this

Comdex:
I just found another issue. The behavior of zkapp.reducer.getActions under LocalBlockChain and Berkeley is different. Under the Berkeley network, executing the getActions method will also get the action corresponding to fromActionState, but it does not under LocalBlockChain. In my personal opinion, the handling of LocalBlockChain is more user-friendly because the action that corresponds to fromActionState is typically already processed. If we want to solve this problem, there are two feasible solutions: handling it in the 'fetchActions' or processing it on the archive node side. What do you think about this? @mitschabaude @MartinMinkov

PR #854, opened 2023-04-12
Fix fetchActions, part 2

### LINKED ISSUE #795: Change internal types stored for actions data to be Fields rather than strings
Instead of doing the conversion to field => base58 format, we should keep the data as a field to save on the extra work that is being done.

here: https://github.com/o1-labs/snarkyjs/blob/9acec551e1cc51c3356d5bc3118fc026e104ea05/src/lib/fetch.ts#L735
here: https://github.com/o1-labs/snarkyjs/blob/9acec551e1cc51c3356d5bc3118fc026e104ea05/src/lib/mina.ts#L513

This means updating some type definitions used for actions to make TypeScript happy.

### LINKED ISSUE #852: `fetchActions` returns one action too many
See https://github.com/o1-labs/snarkyjs/pull/844#issuecomment-1503411842

mitschabaude:
* fixes #852
  * if `fromActionState` is used, the archive node returns one block too many (namely, the one that ended in `fromActionState`). we check for this in snarkyjs and correct if necessary
* refactored `fetchActions` / `getActions` functions throughout to improve readability and do less back and forth conversion.
  * don't store action states with base58 tokenId encoding. Instead, store them as decimal strings which is cheaper to convert to and more consistent with other APIs like graphql
  * fixes #795
  * throw an error in `LocalBlockchain.getActions` when `fromActionState` or `endActionState` aren't found; filed issue for archive node to do the same: https://github.com/o1-labs/Archive-Node-API/issues/77
* reverts https://github.com/o1-labs/snarkyjs/pull/844 which fixed the order that actions are processed inside reduce, but broke hashing which rely on the reversed order; instead, reverse back inside reduce and otherwise keep the reversed order that is consistent with how actions have to be stored on the account update

### DIFF
--- a/src/examples/zkapps/voting/test.ts
+++ b/src/examples/zkapps/voting/test.ts
@@ -79,16 +79,12 @@ export async function testSet(
     );
     console.log('checking that the tx is valid using default verification key');
 
+    let m = Member.from(PrivateKey.random().toPublicKey(), UInt64.from(15));
+    verificationKeySet.Local.addAccount(m.publicKey, m.balance.toString());
+
     await assertValidTx(
       true,
       () => {
-        let m = Member.from(
-          PrivateKey.random().toPublicKey(),
-
-          UInt64.from(15)
-        );
-        verificationKeySet.Local.addAccount(m.publicKey, m.balance.toString());
-
         verificationKeySet.voting.voterRegistration(m);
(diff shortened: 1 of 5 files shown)


shimkiv (approved):
LGTM, Thank you

PR #863, opened 2023-04-21
Refactor DEX to fit tx limits & misc improvements

### LINKED ISSUE #847: Update DEX zkApp to use events and actions
In order to serve as a test case for the Archive Node API, we want it to emit both actions and events.

Question: are we retrieving actions and events from within an app on Berkeley? Should we add this?

### LINKED ISSUE #848: Deploy DEX sample zkApp to Berkeley
This serves to test the live network. Depends on issue #847.

Are we within account update limits? Any other constraints?

Once deployed, we should run the app to exercize the Archive Node API on Berkeley.

mitschabaude:
fixes #847
fixes #848

* add variant of the DEX example which makes use of a 2-step pattern (1. dispatch action - 2. reduce) to implement its methods within the account update limit.
  - [x] supplyLiquidity (this didn't need actions, but some changes to `this.token` APIs to enable reusing AUs
  - [x] redeemLiquidity
  - [x] swap (same as supplyLiqudity, didn't need actions)
* also uses events in the new example
  - [x] emit some events
* create & successfully run a `run-berkeley.ts` script which runs through the happy path of setting up all zkApps, supplying & redeeming liquidity, and performing a swap;
  - DEX account (on-chain state = total supply of liquidity token): https://berkeley.minaexplorer.com/wallet/B62qpeQuY8d97Gc9p2CJbnCBs4eeuqgcncLHnj2Spu519y6fzkz5UUw
  - user account which holds 3 different tokens: https://berkeley.minaexplorer.com/wallet/B62qigeb4rz3UJu7Xs4XQY4qdrsWcQg1ouU5rh3a92aDNA2PCxTYqh5/tokens

Various snarkyjs improvements that helped me achieve this DEX refactor:
* fix the way that transaction limits are counted (the protocol counts the fee payer as signed account update, while snarkyjs didn't count it before)
* add `state.getAndAssertEquals()` and `state.fromAppState()`, doccomment all state methods
* make token APIs (mint, burn, transfer) accept account updates or smart contracts as sender/receiver (to avoid creating unnecessary account updates)
* add an advanced option to `reduce` to skip adding the action state precondition (necessary for when we want to reduce actions on a different account than our own and need permissions for doing so)
* add `reducer.forEach(actions, ...)` which is just `reduce(actions, ...)` without a `state`
* Make the `{ fromActionState, .. }` config optional in all versions of `getActions`
* Make much heavier use of `AccountUpdate.label` in many methods that create account udpates, to make `tx.toPretty()` output easier to understand. Also, add a label for the fee payer.
* Fix to the `@method` wrapper: use cloned method inputs for hashing the function signature for `callData`, so that the hash is not changed by the method mutating the inputs
* Fix to the `@method` wrapper: in the composability case, no longer assume that `MayUseToken.InheritFromParent` is the correct default for the caller method (this was based on an earlier version of `mayUseToken` plus a misunderstanding about how it was supposed to work)
* Export `TokenId` directly instead of a never-used `Token` class which contained the `TokenId` module as (hard to find) static `Token.Id`
* add `Permissions.allImpossible()` for a set of permissions where nothing is allowed (more convenient than `Permissions.default()` when you want to make most actions impossible)


mitschabaude:
TODO: still have to update the changelog

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+    let dl = dy.add(dx);
+    this.token.mint({ address: user, amount: dl });
+
+    // update l supply
+    let l = this.totalSupply.get();
+    this.totalSupply.assertEquals(l);
+    this.totalSupply.set(l.add(dl));
+
+    // emit event
+    let key = 'supply-liquidity' as const satisfies keyof Dex['events'];
```
MartinMinkov, line 114: Interesting use of the typing here! I'm wondering if we chose this method because the event structs are defined inline or if it's due to concerns about type safety when emitting events and the inference isn't great here. Since this pattern appears a few times in the file, do you think it might be helpful to create a wrapper type for events that simplifies these statements for users to use if they run into this? I'd love to hear your thoughts on this!

MartinMinkov (approved):
Additional Question: Would it be worth hooking this up to CI and the Voter app? It doesn't have to be in this PR or anything cc @jasongitmail @nicc

LGTM! :partying_face:

src/lib/state.ts
```diff
@@ -21,11 +21,57 @@ export { assertStatePrecondition, cleanStatePrecondition };
  * Gettable and settable state that can be checked for equality.
  */
 type State<A> = {
+  /**
```
MartinMinkov, line 24: Calling for the help of @barriebyron here for these doc comments :smile: Would you happen to have any suggestions for the wording of these doc comments? Just for context, these comments will be shown to developers when they inspect the methods and in our API Reference docs.

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
+/**
+ * This DEX implementation differs from ./dex.ts in two ways:
+ * - More minimal & realistic; stuff only designed for testing protocol features was removed
```
barriebyron, line 3:
```suggestion
 * - More minimal & realistic; stuff designed only for testing protocol features was removed
```

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+ * - Uses an "async" pattern with actions that lets users claim funds later & reduces account updates
```
barriebyron, line 4:
```suggestion
 * - Uses an async pattern with actions that lets users claim funds later and reduces account updates
```

not sure we enclose "async" with double quotation marks. Is this an API convention? we can remove

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+
+class RedeemAction extends Struct({ address: PublicKey, dl: UInt64 }) {}
+
+class Dex extends SmartContract {
+  // addresses of token contracts are constants
+  tokenX = addresses.tokenX;
+  tokenY = addresses.tokenY;
+
+  /**
+   * state which keeps track of total lqXY supply -- this is needed to calculate what to return when redeeming liquidity
```
barriebyron, line 37:
```suggestion
   * state that keeps track of total lqXY supply -- this is needed to calculate what to return when redeeming liquidity
```

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+   *
+   * total supply is zero initially; it increases when supplying liquidity and decreases when redeeming it
```
barriebyron, line 39:
```suggestion
   * total supply is initially zero; it increases when supplying liquidity and decreases when redeeming it
```

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+   */
+  reducer = Reducer({ actionType: RedeemAction });
+
+  events = {
+    'supply-liquidity': Struct({ address: PublicKey, dx: UInt64, dy: UInt64 }),
+    'redeem-liquidity': Struct({ address: PublicKey, dl: UInt64 }),
+  };
+
+  /**
+   * Initialization. We set _all_ permissions to impossible except the ones we explicitly need.
```
barriebyron, line 54:
```suggestion
   * Initialization. _All_ permissions are set to impossible except the explicitly required permissions.
```

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  /**
+   * Swap X tokens for Y tokens
+   * @param dx input amount of X tokens
+   * @return output amount Y tokens
+   *
+   * The transaction needs to be signed by the user's private key.
+   *
+   * Note: this is not a `@method`, since it doesn't do anything beyond
+   * the called methods which requires proof authorization.
```
barriebyron, line 189:
```suggestion
   * the called methods that require proof authorization.
```

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -0,0 +1,367 @@
...
+
+  /**
+   * Swap Y tokens for X tokens
+   * @param dy input amount of Y tokens
+   * @return output amount Y tokens
+   *
+   * The transaction needs to be signed by the user's private key.
+   *
+   * Note: this is not a `@method`, since it doesn't do anything beyond
+   * the called methods which requires proof authorization.
```
barriebyron, line 207:
```suggestion
   * the called methods that require proof authorization.
```

src/lib/state.ts
```diff
@@ -21,11 +21,57 @@ export { assertStatePrecondition, cleanStatePrecondition };
...
+   * Get the current on-chain state.
+   *
+   * Beware: If you use this method inside a smart contract, calling it alone will not prove that your contract uses the current on-chain state.
+   * To achieve that, either add an additional `.assertEquals()` statement or use `.getAndAssertEquals()`:
```
barriebyron, line 28:
```suggestion
   * Caution: If you use this method alone inside a smart contract, it does not prove that your contract uses the current on-chain state.
   * To successfully prove that your contract uses the current on-chain state, you must add an additional `.assertEquals()` statement or use `.getAndAssertEquals()`:
```

barriebyron (commented):
I suggested some text changes, lmk if helpful @MartinMinkov and teach me where feedback is ineffective, ty!

src/examples/zkapps/dex/dex.ts
```diff
@@ -425,6 +424,17 @@ class TokenContract extends SmartContract {
     zkapp.requireSignature();
   }
 
+  @method approveUpdate(zkappUpdate: AccountUpdate) {
+    this.approve(zkappUpdate);
+    let balanceChange = Int64.fromObject(zkappUpdate.body.balanceChange);
+    balanceChange.assertEquals(Int64.from(0));
+  }
+
+  // FIXME: remove this
```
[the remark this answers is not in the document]
mitschabaude: we need it until we solved https://github.com/o1-labs/snarkyjs/issues/706

src/examples/zkapps/dex/dex-with-actions.ts
mitschabaude: This complicated typing is not strictly necessary. What it does: it will cause a type error when changing the types in the `events` property on the smart contract. In the future, I want to enable such type safety by default via https://github.com/o1-labs/snarkyjs/issues/664, at which point we can remove this code

src/examples/zkapps/dex/dex-with-actions.ts
mitschabaude: this is not what I meant - not the methods require proof authorization, but "it doesn't to anything which requires proof authorization". So, the "s" in requires is correct

Trivo25 (approved):

PR #871, opened 2023-04-26
Add fallback URLs to GraphQL network requests

MartinMinkov:
# Description
This pull request makes several updates to the fetch.ts module, including supporting fallback endpoints and refactoring the code.

Firstly, we now allow users to provide an array of endpoints for GraphQL network requests. The input URLs are checked for validity, and an error is thrown if they're invalid. Instead of issuing multiple requests at once using `Promise.race`, which may lead to unintended spamming, we use a "waterfall" approach to loop through the fallback endpoints. If a request is successful, it returns immediately; otherwise, we collect error messages and display helpful information if all requests fail.

We've also replaced the two standalone string variables with a `NetworkConfig` object, making the network information more organized within the `Fetch` module.

### DIFF
--- a/src/lib/fetch.ts
+++ b/src/lib/fetch.ts
@@ -32,10 +32,13 @@ export {
   getCachedNetwork,
   getCachedActions,
   addCachedAccount,
-  defaultGraphqlEndpoint,
-  archiveGraphqlEndpoint,
+  networkConfig,
+  // defaultGraphqlEndpoint,
+  // archiveGraphqlEndpoint,
   setGraphqlEndpoint,
+  setMinaGraphqlFallbackEndpoints,
   setArchiveGraphqlEndpoint,
+  setArchiveGraphqlFallbackEndpoints,
   sendZkappQuery,
   sendZkapp,
   removeJsonQuotes,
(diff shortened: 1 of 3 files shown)


ymekuria (approved):

src/lib/mina.ts
```diff
@@ -647,20 +647,42 @@ LocalBlockchain satisfies (...args: any) => Mina;
...
   } else if (input && typeof input === 'object') {
-    graphqlEndpoint = input.mina;
-    archiveEndpoint = input.archive;
-    Fetch.setGraphqlEndpoint(graphqlEndpoint);
-    Fetch.setArchiveGraphqlEndpoint(archiveEndpoint);
+    if (!input.mina || !input.archive)
+      throw new Error(
+        "Network: malformed input. Please provide an object with 'mina' and 'archive' endpoints."
+      );
+    if (Array.isArray(input.mina) && input.mina.length > 1) {
```
mitschabaude, line 669: I think you should remove the `&& input.mina.length > 1` check here -- there's no reason to fail on an input like `["some-url"]`, it should be treated the same as just a string `"some-url"`

src/lib/mina.ts
```diff
@@ -647,20 +647,42 @@ LocalBlockchain satisfies (...args: any) => Mina;
...
+      minaGraphqlEndpoint = input.mina[0];
+      Fetch.setGraphqlEndpoint(minaGraphqlEndpoint);
+      Fetch.setMinaGraphqlFallbackEndpoints(input.mina.slice(1));
+    } else if (typeof input.mina === 'string') {
+      minaGraphqlEndpoint = input.mina;
+      Fetch.setGraphqlEndpoint(minaGraphqlEndpoint);
+    }
+
+    if (Array.isArray(input.archive) && input.archive.length > 1) {
```
mitschabaude, line 678: same comment about `.length > 1` as above

src/lib/fetch.ts
```diff
@@ -32,24 +32,56 @@ export {
...
+  minaEndpoint: '',
+  minaFallbackEndpoints: [] as string[],
+  archiveEndpoint: '',
+  archiveFallbackEndpoints: [] as string[],
+} satisfies NetworkConfig;
+
+function checkForValidUrl(url: string) {
+  try {
+    new URL(url);
+    return true;
```
mitschabaude, line 64: nice 💯

mitschabaude (changes requested):
Hey @MartinMinkov, I appreciate that you made this more future proof than in the RFC, by not letting devs spam the network with an arbitrary number of requests. Also, I get that with our current APIs which let you specify a single `graphqlEndpoint`, it was more natural to single out one endpoint in the `NetworkConfig` which is tried first.

However, I liked about the `Promise.race` idea that it would make temporary node unresponsiveness basically invisible to the user -- users would always get the fastest response. If we have two endpoints which both have a chance of `p` (`0 < p < 1`) of being unresponsive at any given time, trying both endpoints at the same time decreases the "unresponsiveness chance" to `p^2`. I think we'd be missing a UX improvement opportunity by not using multiple endpoints on the first try.

As a middle ground between arbitrary spamming and no speedup, I propose that you change the logic as follows:
* send requests to _two endpoints at a time_, so use the default endpoint plus the first fallback endpoint with `Promise.race` on the initial try
* subsequent tries (> 2) have diminishing returns probably so you can do whatever is easiest to implement with those

ok?

mitschabaude (approved):
sweet!

PR #872, opened 2023-04-27
Start workers only when needed

### LINKED ISSUE #609: Reduce SnarkyJS memory requirements
Creating an issue to track for the roadmap. Goal to make SnarkyJS and zkApp work on more run times, such as free-tiers of serverless functions like Vercel and Cloudflare Pages.

To discuss and plan approach.

Targeted Q1 2023.

### LINKED ISSUE #674: Get rid of `shutdown` by running workers only when needed
IS: Workers are started when snarkyjs gets loaded and they keep running throughout the JS process, so that we need to shut them down at the end.

SHOULD: Workers should be started on demand when we start to compile, prove or verify (these are the functions that currently make use of workers). They should be shut down again when those functions terminate.

IMPACT: We can remove `shutdown` which is inconvenient, especially in Jest tests, and is a source of CI failures (forgetting to add `shutdown` to a script means the script will run indefinitely in CI). We also reduce the time to `isReady`, to zero in the case of nodejs (the web version still needs `isReady` since the Wasm bundle can only be loaded asynchronously. We could discuss loading it synchronously with top-level await, to get rid of `isReady` altogether)

### LINKED ISSUE #696: Replace panics with Results in the Rust bindings for much more descriptive failures
There's a 2 line change to the WASM bindings which will give you a descriptive error instead of panicking when proving fails.

Make https://github.com/MinaProtocol/mina/blob/a3661e44e01b2c400d6eb5ac254c6ab7a91b3623/src/lib/crypto/kimchi_bindings/wasm/src/plonk_proof.rs#L551 return a Result<WasmProverProof, JSValue>
Change https://github.com/MinaProtocol/mina/blob/a3661e44e01b2c400d6eb5ac254c6ab7a91b3623/src/lib/crypto/kimchi_bindings/wasm/src/plonk_proof.rs#L612 to return maybe_proof.map_err(|str| str.into()) (or something like that), which will make it return the error as a JS exception

Additionally, https://github.com/MinaProtocol/mina/blob/a3661e44e01b2c400d6eb5ac254c6ab7a91b3623/src/lib/crypto/kimchi_bindings/wasm/src/plonk_proof.rs#L585

mitschabaude:
This PR refactors how SnarkyJS uses web workers / Node.js worker threads. It thereby significantly **reduces memory consumed by SnarkyJS**. As a welcome side effect, we get rid of having to call `await isReady` and `shutdown()` before / after using SnarkyJS.

Examples of memory reduction with 16 threads in Node.js:
- From **3.2GB** down to **300MB** when not compiling or proving
- From **3.6GB** down to **1.7GB** when proving a small contract (`src/examples/simple_zkapp.ts`)
- From **4.9GB** down to **3.2GB** when proving complex, multi-contract zkApp (`src/examples/zkapps/dex/happy-path-with-proofs.ts`)

This was achieved because workers are no longer running by default -- they are only started when compiling or proving, and shut down after that. In addition, we achieved a reduction in memory consumption of each worker in the Node.js version, by reducing the JS dependency graph of each worker down to a minimum.

closes #674 
closes #696
closes #609

mina: https://github.com/MinaProtocol/mina/pull/13102
snarkyjs-bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/1

main changes:
* move `src/snarky` down into snarkyjs-bindings
* snarkyjs-bindings introduces a new `withThreadPool()` function which wraps all operations that need workers. it takes care of starting and shutting down workers before/after the wrapped function.
* `withThreadPool()` is used to wrap compile / prove / verify both in `proof_system.ts` (wrapping Pickles / SmartContract methods) and in the new `circuit.ts` (wrapping plain kimchi / `Circuit` methods)
* to wrap `Circuit`, we had to lift its API surface from OCaml to TS; lives in a new `circuit.ts` file; prove/compile/verify become async, because `withThreadPool()` is async
* deprecate both `shutdown()` and `isReady` which become no-ops

more details are in the PR comments below!


package.json
```diff
@@ -33,12 +33,13 @@
     "type-check": "tsc --noEmit",
     "dev": "npx tsc -p tsconfig.node.json && cp src/snarky.d.ts dist/node/snarky.d.ts",
     "make": "make -C ../../.. snarkyjs",
+    "bindings": "cd ../../.. && ./scripts/update-snarkyjs-bindings.sh && cd src/lib/snarkyjs",
```
mitschabaude, line 36: handy new script to update the bindings

package.json
```diff
@@ -81,7 +82,6 @@
   "dependencies": {
     "blakejs": "1.2.1",
     "detect-gpu": "^5.0.5",
-    "env": "^0.0.2",
```
mitschabaude, line 84: finally getting rid of this stupid dependency

src/lib/caller.unit-test.ts
```diff
@@ -30,6 +30,6 @@ let tx = await Mina.transaction(privateKey, () => {
 });
 
 // according to this test, the child doesn't get token permissions
-expect(tx.send()).rejects.toThrow('Token_owner_not_caller');
-
-shutdown();
+await expect(tx.send()).rejects.toThrow(
+  'can not use or pass on token permissions'
+);
```
mitschabaude, line 35: these tests were actually not working 🤦🏻‍♂️ because `expect.rejects.toThrow` is an async function, and wasn't awaited before calling shutdown, it never finished

mitschabaude (commented):
some comments to help with reviewing!

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+   * Serializes an element into {@link Field} elements.
+   */
+  static toFields = SnarkyCircuit.toFields;
+  /**
+   * Checks if the circuit is in prover mode.
+   */
+  static inProver = SnarkyCircuit.inProver;
+  /**
+   * Checks if the circuit is in checked computation mode.
+   */
```
MartinMinkov, line 124:
```suggestion
  /**
   * Checks if the circuit is in checked computation mode.
   * @example
   * ```ts
   * if (Circuit.inCheckedComputation()) {
   *   // Checked computation-specific code
   * }
   * ```
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+  static inCheckedComputation = SnarkyCircuit.inCheckedComputation;
+  /**
+   * Interface to log elements within a circuit. Similar to `console.log()`.
+   */
```
MartinMinkov, line 128:
```suggestion
  /**
   * Interface to log elements within a circuit. Similar to `console.log()`.
   * @example
   * ```ts
   * const element = Field(42);
   * Circuit.log(element);
   * ```
   */
```

package.json
VivekTyagi528: Hey, I also want to contribute to this project, Any pre-requisites, I have decent experience with Solidity and Javascript

mitschabaude: Yeah sure, JS experience is enough to start contributing! Maybe you find something you can do labeled with `good first issue`.
But if you notice something else that should be improved, were always happy to take suggestions! Also, what snarkyjs needs are more community-maintained libraries

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+
+// external API
+export { public_, circuitMain, Circuit };
+
+class Circuit {
+  // circuit-writing interface
+
+  /**
+   * Generates a proving key and a verification key for this circuit.
+   */
```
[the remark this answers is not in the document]
mitschabaude: just changing this to async:
```suggestion
 /**
   * Generates a proving key and a verification key for this circuit.
   * @example
   * ```ts
   * const keypair = await Circuit.generateKeypair();
   * ```
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+  static prove(privateInput: any[], publicInput: any[], keypair: Keypair) {
+    return withThreadPool(async () =>
+      SnarkyCircuit.prove(this as any, privateInput, publicInput, keypair)
+    );
+  }
+
+  /**
+   * Verifies a proof using the public input, the proof and the initial {@link Keypair} of the circuit.
+   */
```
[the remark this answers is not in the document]
mitschabaude:
```suggestion
  /**
   * Verifies a proof using the public input, the proof, and the initial {@link Keypair} of the circuit.
   * @example
   * ```ts
   * const keypair = await Circuit.generateKeypair();
   * const proof = await Circuit.prove(privateInput, publicInput, keypair);
   * const isValid = await Circuit.verify(publicInput, keypair.vk, proof);
   * ```
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+  static runAndCheck = SnarkyCircuit.runAndCheck;
+  /**
+   * Returns information about the constraint system in the callback function.
+   */
+  static constraintSystem = SnarkyCircuit.constraintSystem;
+  /**
+   * Returns a low-level JSON representation of the `Circuit` from its {@link Keypair}:
+   * a list of gates, each of which represents a row in a table, with certain coefficients and wires to other (row, column) pairs
+   */
```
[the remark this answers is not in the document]
mitschabaude:
```suggestion
  /**
   * Returns a low-level JSON representation of the `Circuit` from its {@link Keypair}:
   * a list of gates, each of which represents a row in a table, with certain coefficients and wires to other (row, column) pairs
   * @example
   * ```ts
   * const keypair = await Circuit.generateKeypair();
   * const jsonRepresentation = Circuit.constraintSystemFromKeypair(keypair);
   * ```
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+   * Creates a {@link Provable} for a generic array.
+   */
+  static array = SnarkyCircuit.array;
+  /**
+   * Asserts that two values are equal.
+   */
+  static assertEqual = SnarkyCircuit.assertEqual;
+  /**
+   * Checks if two elements are equal.
+   */
```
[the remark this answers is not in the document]
mitschabaude: I'm going to change this so that it shows how to use for Structs (for which it is most useful -- on `Field` we already have `.equals()`)

```suggestion
  /**
   * Checks if two elements are equal.
   * @example
   * ```ts
   * class MyStruct extends Struct({ a: Field, b: Bool }) {};
   * const a: MyStruct = { a: Field(0), b: Bool(false) };
   * const b: MyStruct = { a: Field(1), b: Bool(true) };
   * const isEqual = Circuit.equal(MyStruct, a, b);
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+  static constraintSystemFromKeypair =
+    SnarkyCircuit.constraintSystemFromKeypair;
+  /**
+   * Creates a {@link Provable} for a generic array.
+   */
+  static array = SnarkyCircuit.array;
+  /**
+   * Asserts that two values are equal.
+   */
```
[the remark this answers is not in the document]
mitschabaude:
```suggestion
  /**
   * Asserts that two values are equal.
   * @example
   * ```ts
   * class MyStruct extends Struct({ a: Field, b: Bool }) {};
   * const a: MyStruct = { a: Field(0), b: Bool(false) };
   * const b: MyStruct = { a: Field(1), b: Bool(true) };
   * Circuit.assertEqual(MyStruct, a, b);
   * ```
   */
```

src/lib/circuit.ts
```diff
@@ -0,0 +1,221 @@
...
+  static constraintSystem = SnarkyCircuit.constraintSystem;
+  /**
+   * Returns a low-level JSON representation of the `Circuit` from its {@link Keypair}:
+   * a list of gates, each of which represents a row in a table, with certain coefficients and wires to other (row, column) pairs
+   */
+  static constraintSystemFromKeypair =
+    SnarkyCircuit.constraintSystemFromKeypair;
+  /**
+   * Creates a {@link Provable} for a generic array.
+   */
```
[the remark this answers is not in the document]
mitschabaude:
```suggestion
  /**
   * Creates a {@link Provable} for a generic array.
   * @example
   * ```ts
   * const ProvableArray = Circuit.array(Field, 5);
   * ```
   */
```

MartinMinkov (approved):

package.json
barriebyron: thank you for your interest @0xvicky 
see https://github.com/o1-labs/snarkyjs/blob/main/CONTRIBUTING.md

PR #878, opened 2023-04-28
Update signMessage to pass private key

garethtdavies:
The sample code no longer works as `signMessage` now takes a private key and not a keypair.


mitschabaude (approved):
Thanks!

PR #882, opened 2023-05-01
add nullifier to snarkyjs and mina-signer

Trivo25:
implements the `Nullifier` as a new primitive in both mina-signer and SnarkyJS (in-snark variant) and demonstrates its usage in `src/examples/nullifier.ts`

see https://eprint.iacr.org/2022/1255.pdf

bindings https://github.com/o1-labs/snarkyjs-bindings/pull/10

closes https://github.com/o1-labs/snarkyjs/issues/816

RFC: https://github.com/o1-labs/snarkyjs/issues/756


Trivo25:
The "helper" functions for the Nullifier are somewhat limited, I didn't want to put too many constraints on the developers (e.g. limiting them to use our Merkle tree/map, etc)

src/mina-signer/src/nullifier.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  const nullifier = Group.scale(h_m_pk, sk);
+  const h_m_pk_r = Group.scale(h_m_pk, r);
+
+  const g_r = Group.scale(G, r);
+
+  const c = Hash2([
+    ...Group.toFields(G),
+    pk.x,
+    pk.isOdd ? 1n : 0n,
```
mitschabaude, line 36: nit: `PublicKey.toFields()`

src/mina-signer/src/nullifier.ts
```diff
@@ -0,0 +1,58 @@
...
+*/
+function createNullifier(message: Field, sk: PrivateKey): Nullifier {
+  const Hash2 = Poseidon.hash;
+  const Hash = Poseidon.hashToGroup;
+
+  const pk = PrivateKey.toPublicKey(sk);
+
+  const G = Group.generatorMina;
+
+  const r = Field.random();
```
mitschabaude, line 22: `r` is a scalar, so this should be `Scalar.random()`

src/mina-signer/src/nullifier.ts
```diff
@@ -0,0 +1,58 @@
...
+    ...Group.toFields(G),
+    pk.x,
+    pk.isOdd ? 1n : 0n,
+    ...Group.toFields(h_m_pk),
+    ...Group.toFields(nullifier),
+    ...Group.toFields(g_r),
+    ...Group.toFields(h_m_pk_r),
+  ]);
+
+  const s = Fp.add(r, Fp.mul(sk, c));
```
mitschabaude, line 43: these are operations on scalars, so should be `Fq.add`, `Fq.mul`; very important! otherwise, the math in the exponent doesn't work

only `c` is in `Fp` since it is the Poseidon result. This is unavoidable, since we can only do Fp Poseidon efficiently in the SNARK.
So, we need to interpret $c \in F_p$ as an element of $F_q$. Luckily this is always possible, since Fq is larger than Fp. There should be a comment explaining that

mitschabaude (changes requested):
Some initial review, will look at the snark verifier now to gain a fuller understanding

src/lib/nullifier.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+}) {
+  static fromJSON(json: JsonNullifier): Nullifier {
+    return super.fromJSON(json as any) as Nullifier;
+  }
+
+  /**
+   * Verifies the correctness of the Nullifier. Throws an error if the Nullifier is incorrect.
+   */
+  verify() {
```
mitschabaude, line 38: we definitely need a unit test for `createNullifier()` -> `Nullfier.fromJSON()` -> `Nullifier.verify()`

should also include negative tests (i.e., I can't make `Nullifier.verify()` work without knowing `sk`)

src/mina-signer/src/nullifier.ts
Trivo25: I see, interesting :thinking:

src/mina-signer/src/nullifier.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Field } from '../../provable/field-bigint.js';
+import { Nullifier } from './TSTypes.js';
+
+export { createNullifier };
+/*
+PLUME: An ECDSA Nullifier Scheme for Unique
+Pseudonymity within Zero Knowledge Proofs
+https://eprint.iacr.org/2022/1255.pdf chapter 3 page 14
+*/
+function createNullifier(message: Field, sk: PrivateKey): Nullifier {
```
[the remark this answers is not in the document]
Trivo25: I agree it would improve usability, but I think then we would have to specify the static arrays size in advance, or using some workarounds for the in-snark variant, I made it a one Field argument so both implementations (out-of-snark and in-snark) can be uniform. Or can we pass a dynamic-size array in-snark into Poseidon? :thinking:

src/lib/nullifier.ts
```diff
@@ -0,0 +1,84 @@
...
+    } = Poseidon.hashToGroup([message, ...publicKey.toFields()]);
+
+    Poseidon.hash([
+      ...Group.toFields(G),
+      ...publicKey.toFields(),
+      x,
+      x0,
+      ...Group.toFields(nullifier),
+      ...Group.toFields(g_r),
+      ...Group.toFields(h_m_pk_r),
```
mitschabaude, line 60: The way you use `g_r` and `h_m_pk_r` here makes the check not sufficient: You don't prove that there's a connection between `g_r`, `h_m_pk_r` on the one hand and `pk`, `c`, `nullifier` and `s` on the other hand.

So, at the moment, I can create a **valid nullifier without owning the private key**, by doing the following:
* create any random group elements `g_r` and `h_m_pk_r`
* computing `hash[m, pk] = x, x0` from the message and public key, which I have access to
* recreate the `c` hash
* put all of those together into a `Nullifier`

So, it's not secure yet!

I know that the paper says that `g_r` and `h_m_pk_r` are
> ...included here to increase prover efficiency via precomputation

I'm not sure how using them to increase efficiency without losing security works, yet.
But I would ask that we first try to make it work without those values, and re-compute everything, like `g^s / pk^c` etc. We can worry about introducing efficiency tricks later

src/examples/nullifier.ts
```diff
@@ -0,0 +1,126 @@
...
+
+import { createNullifier } from '../mina-signer/src/nullifier.js';
+
+class PayoutOnlyOnce extends SmartContract {
+  @state(Field) nullifierRoot = State<Field>();
+  @state(Field) nullifierMessage = State<Field>();
+
+  @method payout(nullifier: Nullifier) {
+    let nullifierRoot = this.nullifierRoot.getAndAssertEquals();
+    let nullifierMessage = this.nullifierMessage.getAndAssertEquals();
```
mitschabaude, line 24: We should also have an example that verifies the nullfier inside a SNARK

mitschabaude (commented):
Done - the SNARK verifier still needs a bit of work!

src/mina-signer/src/nullifier.ts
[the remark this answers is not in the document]
mitschabaude: you can pass any size array into Poseidon, but you're fixing the circuit to always use that size by doing that. I think that's ok

src/lib/nullifier.ts
Trivo25: Ooof that part confused me a bit, I thought by "optimizing the computation" it meant we could just take these values directly :D good catch

src/provable/curve-bigint.ts
```diff
@@ -108,6 +111,10 @@ const PublicKey = {
   toInputLegacy({ x, isOdd }: PublicKey): HashInputLegacy {
     return { fields: [x], bits: [!!isOdd] };
   },
+
+  toFields({ x, isOdd }: PublicKey) {
+    return [x, isOdd ? 1n : 0n];
```
Trivo25, line 116: `PublicKey.toFields` was returning `[y, x]` instead of `[x, y]` like SnarkyJS does. It's probably better to reverse the order on the SnarkyJS side, to not introduce any breaking changes into mina-signer or do you think thats not the case @mitschabaude ?

mitschabaude: ph -- I think this is a bug, it should be `[x, y]`

this is the change you probably need (a couple of lines above this):
```diff
- ...provable({ x: Field, isOdd: Bool }),
+ ...provable({ x: Field, isOdd: Bool }, { customObjectKeys: ['x', 'isOdd'] }),
```

btw, since I wrote the binable code I learned that the order in JS objects is deterministic and basically the insertion order, so I might be able to get rid of `customObjectKeys`

src/lib/nullifier.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  isUnused(witness: MerkleMapWitness, root: Field) {
+    return witness.computeRootAndKey(Field(0))[0].equals(root);
+  }
+
+  /**
+   * Sets the Nullifier, returns the new Merkle root.
+   */
+  setUsed(witness: MerkleMapWitness) {
+    return witness.computeRootAndKey(Field(1))[0];
```
mitschabaude, line 104: same comment as above about computing `this.key()` here to prove that the witness is correct

src/lib/nullifier.unit-test.ts
```diff
@@ -0,0 +1,59 @@
...
+
+// trying to manipulate the nullifier to take a real pk that it doesnt know the sk to
+jsonNullifier2.publicKey = priv.toPublicKey().toBase58();
+
+let nullifier2 = Nullifier.fromJSON(jsonNullifier2);
+try {
+  nullifier2.verify(message);
+  console.log('incorrect nullifier was verified');
+  console.log(JSON.stringify(nullifier2));
+  process.exit(1);
```
mitschabaude, line 38: this is clever :D

src/lib/nullifier.ts
```diff
@@ -0,0 +1,106 @@
...
+    let G = Group.generator;
+
+    // serialize public key into fields once
+    let pk_fields = publicKey.toFields();
+
+    // x and y of hash(msg, pk), it doesn't return a Group because y is split into x0 and x1, both two roots of a field element
+    let {
+      x,
+      y: { x0 },
+    } = Poseidon.hashToGroup([...message, ...pk_fields]);
```
mitschabaude, line 56: your scheme seems to be different from the original scheme in one minor aspect: you hash the public key in compressed `(x, isOdd)` form (`publicKey.toFields()`) instead of in the uncompressed curve point form `(x, y)` (`publicKey.toGroup().toFields()`). I'm sure this has no security impact, but I think it would be cleaner to use the uncompressed curve point form, because that seems to be closer to an ideal "standard implementation". Usage of the public key in compressed form is really more a Mina implementation detail that should be kept out of here.

Decompressing with `publicKey.toGroup()` is an expensive operation in the proof, so it might make sense to make this operation _optional_ for the prover by returning the _uncompressed_ publicKey (the `Group`) from the wallet. Then the public compressed pk plays no role in the entire protocol.

For smart contracts it would then be convenient to have a method `getPublicKey(): PublicKey` on `Nullifier`, so it becomes

```ts
Nullifier extends Struct({ publicKey: Group, ... }) {
   getPublicKey() {
     return PublicKey.fromGroup(this.publicKey);
   }
   // ...
}
```

mitschabaude: well - decide for yourself if you think this is worth it

mitschabaude (commented):
Awesome work, I have a few comment left

src/lib/nullifier.ts
```diff
@@ -0,0 +1,106 @@
...
+  key() {
+    return Poseidon.hash(Group.toFields(this.public.nullifier));
+  }
+
+  /**
+   * Checks if the Nullifier has been used before.
+   */
+  isUnused(witness: MerkleMapWitness, root: Field) {
+    return witness.computeRootAndKey(Field(0))[0].equals(root);
+  }
```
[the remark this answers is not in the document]
Trivo25: ooof, good point with the merkle witness, completely forgot about that

src/lib/nullifier.ts
Trivo25: I would rather keep it close to the original protocol, I'll change it to use the uncompressed point instead!

mitschabaude (approved):
perfect!

MartinMinkov (commented):
Awesome work! Do you mind making some changes to the doc comments in `nullifier.ts`?

Could you include a short summary of the paper in the class declaration comment? I think even something like this would be incredibly helpful (which is taken directly from the abstract):

```
Nullifiers are used as a public commitment to a specific anonymous account, to forbid actions like double spending, or allow a consistent identity between anonymous actions.
```

Could you also include an `@example` section in each method? I think it'd be really helpful to see these in your intellisense as you're exploring the API to give some hints to a developer. This will also really help our API reference.

mrmr1993 (approved):

PR #883, opened 2023-05-01
Bump Actions/Events limit

MartinMinkov:
# Description
Defines a new `constants.ts` file in the `mina` directory. Moves the locally defined constants used for actions/events (and a few other constants) into its own file with descriptions on where to find them in the Mina repo.


Trivo25 (approved):

mitschabaude (approved):
love the new constants file!

PR #884, opened 2023-05-01
Remove duplicate user command fields

nholland94:
As part of https://github.com/MinaProtocol/mina/pull/13048, a bug was discovered with how the signed command transaction logic would handle inputs where duplicate fields must be equal (this bug only existed in the berkeley version of the code, not what is released to mainnet). As part of fixing this, I opted to remove the duplicate fields to ensure that we aren't missing checks anywhere. The duplicate fields only existed to support user command tokens before mainnet, a feature which was cut from scope for mainnet, and has now been superseded by the zkapps token design.

This PR propagates these changes into SnarkyJS.


mitschabaude (commented):
This looks very thorough, thanks @nholland94!

Unit tests seem to fail because the js_of_ocaml build artifact doesn't reflect these changes yet. Mina-signer doesn't depend on OCaml by itself, but the tests that check mina-signer for correctness do.
Which brings up the question: Is there a Mina branch this corresponds to? This probably won't be compatible with the current testnet, so we shouldn't merge this PR to `main`. I'll switch the base branch to `berkeley`, which is where we land upcoming changes that correspond to Mina's `berkeley` branch

PR #889, opened 2023-05-03
Move Provable (formerly Circuit) to JS

mitschabaude:
bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/9

closes https://github.com/o1-labs/snarkyjs/issues/675

* re-implement `Circuit` on top of a new, simpler and less JavaScript-y OCaml interface called `Snarky`
* move utility functions such as `Circuit.if` and `Circuit.witness` to a new namespace `Provable`, which lives in `provable.ts`
* much of `provable.ts` is existing code that was moved around.
* significant changes:
  - `Provable.{if, assertEqual, equal}` are re-implemented in JS on top of lower-level circuit gadgets, and equipped with more precise typing and better-defined error behavior
  - The `Circuit` circuit-writing API has some logic (defining input types, & the snarky circuit function) moved to JS, to get rid of `Provable<T>` handling in OCaml altogether
  
 see comments below for details. I recommend to focus review on the re-written primitives in `provable.ts`
 
 note:  tests are failing because this doesn't have updated bindings yet


src/lib/circuit_value.ts
```diff
@@ -276,97 +259,12 @@ function prop(this: any, target: any, key: string) {
   }
 }
 
-function circuitArray<A extends FlexibleProvable<any>>(
```
mitschabaude, line 279: this is moved to `provable.ts` with no changes

src/lib/provable.ts
```diff
@@ -0,0 +1,596 @@
...
+}
+function equalExplicit<T>(type: Provable<T>, x: T, y: T) {
+  let xs = type.toFields(x);
+  let ys = type.toFields(y);
+  return xs.map((x, i) => x.equals(ys[i])).reduce(Bool.and);
+}
+
+function if_<T>(condition: Bool, type: FlexibleProvable<T>, x: T, y: T): T;
+function if_<T extends ToFieldable>(condition: Bool, x: T, y: T): T;
+function if_(condition: Bool, typeOrX: any, xOrY: any, yOrUndefined?: any) {
```
mitschabaude, line 278: `if` - new implementation!

src/lib/provable.ts
```diff
@@ -0,0 +1,596 @@
...
+  if (!('fromFields' in type && 'toFields' in type)) {
+    throw Error(
+      'Provable.if: Invalid argument type. Try using an explicit type argument:\n' +
+        `Provable.if(bool, MyType, x, y)`
+    );
+  }
+  return ifExplicit(condition, type as any as Provable<T>, x, y);
+}
+
+function switch_<T, A extends FlexibleProvable<T>>(
```
mitschabaude, line 334: `switch` is moved here from `circuit_value.ts` without changes

src/lib/provable.ts
```diff
@@ -0,0 +1,596 @@
...
+    let i = mask.findIndex((b) => b.toBoolean());
+    if (i === -1) return undefined;
+    return values[i];
+  });
+  return (type as Provable<T>).fromFields(fields, aux);
+}
+
+// runners for provable code
+
+function log(...args: any) {
```
mitschabaude, line 375: `log` is moved here from `circuit_value.ts` without changes

src/lib/provable.ts
```diff
@@ -0,0 +1,596 @@
...
+        } catch {
+          prettyArgs.push(arg);
+        }
+      }
+    }
+    console.log(...prettyArgs);
+  });
+}
+
+function asProver(f: () => void) {
```
mitschabaude, line 392: everything below in this file is moved here without changes, except the new `checkLength` and `clone` helpers

mitschabaude (commented):
some comments to help reviewing!

src/lib/provable.ts
```diff
@@ -0,0 +1,596 @@
...
+  if (inCheckedComputation()) {
+    Snarky.asProver(f);
+  } else {
+    f();
+  }
+}
+
+function runAndCheck(f: () => void) {
+  let [, result] = snarkContext.runWith({ inCheckedComputation: true }, () =>
```
Trivo25, line 401: `result` is void, do we want to explicitly return that? And why is it even returning void in the first place :thinking:

Trivo25 (approved):
Nice! Left some questions for my understanding

src/lib/provable.ts
mitschabaude: good question, I should pick one :D if we allow to return something we need to know the shape of the return value as a `Provable`, and have to add an `asProver` block where any variables in the return value are converted to constants. Otherwise we get variables leaking out into the environment. Seemed easier to just not allow that and require the function to be `void`

src/lib/circuit.ts
```diff
@@ -230,6 +191,68 @@ function public_(target: any, _key: string | symbol, index: number) {
...
+  publicInputType: ProvablePure<P>;
+  privateInputType: ProvablePure<W>;
+};
+
+function mainFromCircuitData<P, W>(
+  data: CircuitData<P, W>,
+  privateInput?: W
+): (publicInput: Field[]) => void {
+  return function main(publicInputFields: Field[]) {
+    let id = snarkContext.enter({ inCheckedComputation: true });
```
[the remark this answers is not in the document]
mitschabaude: I used to `enter` / `leave` contexts with `context.runWith(state, () => {...))`, but found that this pollutes stack traces with lots of `runWith` statements. So I decided to flatten this into `enter` & `try { ... } finally { leave() }` patterns

PR #894, opened 2023-05-07
Feature/update snarkyjs docs

yunus433:


src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * 
+   * @return A {@link Field} array of length 1 created from this {@link Field}.
+   * 
    */
-  static fromJSON(x: string): Field;
+  static toInput(value: Field): { fields: Field[] };
 
-  static check(x: Field): void;
+  /**
+   * Create an array of digits equal to the [little-endian](https://betterexplained.com/articles/understanding-big-and-little-endian-byte-order/) byte order of the given {@link Field} element.
```
MartinMinkov, line 778: Do you mind replacing the link with a website that could be more "permanent"? I just worry about the chance of referencing a website and then having the link 404.  Wikipedia is probably good enough to link: https://en.wikipedia.org/wiki/Endianness

What do you think?

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+declare function Field(value: Field | number | string | boolean | bigint): Field;
 declare class Field {
   /**
-   * Coerces anything field-like to a {@link Field}.
+   * Coerce anything `field-like` (bigint, boolean, number, string, and {@link Field}) to a {@link Field}.
+   * A {@link Field} is an element of a prime order field. Every other provable type is build using the {@link Field} type.
+   * The field is the [pasta fp field](https://electriccoin.co/blog/the-pasta-curves-for-halo-2-and-beyond/) of order 28948022309329048855892746252171976963363056481941560715954676764349967630337.
+   * **Warning: You cannot create a {@link Field} from a non-integer number.
+   * 
+   * ```ts
```
MartinMinkov, line 66: For all examples, do you mind using the `@example` doclet?
https://jsdoc.app/tags-example.html

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
    *
    * ```ts
-   * let a = Field(3);
-   * let sum = a.add(5)
+   * const value1 = Field(1);
+   * const value2 = Field(-7);
+   * 
+   * const sum = value1.add(value2);
+   * 
+   * // If you try to print sum - `console.log(sum.toBigInt())` - you will realize that it prints a very big integer, so you cannot use arithmetic operations on snarkyJS as in regular programming languages.
```
mitschabaude, line 144:
```suggestion
   * // If you print sum - `console.log(sum.toBigInt())` - you will see a very big integer, because this is modular arithmetic and 1 + (-7) circles around the field to become p - 6
```
I think its wrong to say that "you cannot use arithmetic operations on snarkyJS as in regular programming languages", for several reasons:
* snarkyjs isn't a programming language. JavaScript is the programming language.
* you can use Field arithmetic, and it works fine, it's just different
* there is no "default arithmetic in regular programming languages". In many languages, you have int32 and int64 arithmetic which is modulo ranges like [-2^31, 2^31 - 1]. Then you have floats which are different again, etc

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * // However, all the operations always stay provable: This means you can use the reverse operation of addition (substraction) to prove the sum is calculated correctly.
```
mitschabaude, line 145:
```suggestion
   * // You can use the reverse operation of addition (substraction) to prove the sum is calculated correctly.
```

src/snarky.d.ts
mitschabaude: @yunus433 can you change the wording across all comments?

I'm curious what others think about this though, @MartinMinkov @bkase

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
-   * Squares this {@link Field} element.
+   * Square this {@link Field} element.
    *
    * ```typescript
-   * const x2 = x.square();
-   * x2.assertEquals(x.mul(x));
+   * const someField = Field(7);
+   * const square = someField.square();
+   * 
+   * square.assertEquals(someField.mul(someField)); // This statement is always true regardless of the value of `someField`
```
mitschabaude, line 257: I'm really not sure about changes like this, where everything is made more verbose and comments are added. To me, the original comment is cleaner, and therefore easier to understand.

changing `x` to `someField`, and showing the `Field(7)` construction, and the long comment all just add visual noise IMO

Since  this is done throughout this PR, I'd be interested in other's opinions about this, maybe it's just me, @MartinMinkov @bkase

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
    *
    * ```typescript
-   * x.square().sqrt().assertEquals(x);
+   * const someField = Field(42);
+   * const squareRoot = someField.sqrt();
+   * 
+   * squareRoot.mul(squareRoot).assertEquals(someField); // This statement is always true regardless of the value of `someField`
    * ```
+   * 
+   * **Warning: This is a modular square root. Please see `div()` function for more details.
```
mitschabaude, line 276: @yunus433 please fix `**Warning: ..` and `** Warning: ..` and `**Important` without a closing `**` throughout the PR. Also I recommend that you look at these changes in VS code's intellisense to double-check that they look as you intended.

The `**double-star syntax**` is Markdown for **bold text** and will show as bold text in the editor

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * @return A string equivalent to the JSON representation of the {@link Field}.
    */
-  sizeInFields(): number;
+  toJSON(): string;
 
   /**
-   * Serializes this data structure into {@link Field} elements.
+   * This function is the implementation of {@link Provable.toFields} in {@link Field} type.
+   * Serialize this {@link Field} element into an array of {@link Field} elements.
+   * You can use this array to calculate the {@link Poseidon} hash of a {@link Field}.
```
mitschabaude, line 327: Delete this line? It makes no sense to me. You calculate the Poseidon hash of a Field with `Poseidon.hash([x])`

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * @return A {@link Field} array of length 1 created from this {@link Field}.
    */
   toFields(): Field[];
 
   /**
-   *
-   * Check if this {@link Field} is lower than another Field-like value.
-   * Returns a {@link Bool}.
+   * Check if this {@link Field} is lower than another `field-like` value.
+   * Returns a {@link Bool}, which is a `provable` type and can be used prove to the validity of this statement.
```
mitschabaude, line 336: can you remove the backticks on non-code like `field-like` and `provable` everywhere?

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
    * ```
-   */
-  lessThan(y: Field | number | string | boolean): Bool;
-  /**
-   *
-   * Check if this {@link Field} is lower than or equal to another Field-like value.
-   * Returns a {@link Bool}.
-   *
+   * 
+   * **Warning: As this function compares the bigint value of a {@link Field}, it can result in an unexpected behaviour while comparing negative numbers or floating point operations.
```
mitschabaude, line 342: This is a good call-out @ negative numbers / division! "floating point operations" is not correct phrasing for our modular division IMO.

```suggestion
   * **Warning**: As this method compares the bigint value of a {@link Field}, it can result in unexpected behaviour when used with negative inputs or modular division.
  * ```ts
  * Field(0).lessThanOrEqual(-1).assertTrue(); // works, because -1 becomes the large field element p - 1 
  * ```
```

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * 
+   * @return A {@link Bool} representing if this {@link Field} equals 0.
+   */
   isZero(): Bool;
 
   /**
-   * Little endian binary representation of the field element.
+   * Returns an array of {@link Bool} elements representing [little endian binary representation](https://betterexplained.com/articles/understanding-big-and-little-endian-byte-order/) of this {@link Field} element.
+   * 
+   * **Warning: Binary operations on snarkyJS are currently really costy on chain. Do not use this function if it is not absolutely necessary.
```
mitschabaude, line 542:
```suggestion
   * **Warning**: This is a costly operation in a zk proof, because a Field can have 255 bits, each of which has to be constrained to prove that the bits equal the original field element. Prefer to use this method with the optional `length` argument to reduce the maximum amount of bits that the input can have.
```

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
    */
   toBits(): Bool[];
 
   /**
-   * Little endian binary representation of the field element.
-   * Fails if the field element cannot fit in `length` bits.
+   * Returns an array of {@link Bool} elements representing [little endian binary representation](https://betterexplained.com/articles/understanding-big-and-little-endian-byte-order/) of this {@link Field} element.
+   * Throws an error if the element cannot fit in `length` bits.
+   * 
+   * **Warning: Binary operations on snarkyJS are currently really costy on chain. Do not use this function if it is not absolutely necessary.
```
mitschabaude, line 552:
```suggestion
   * **Warning**: The cost of this operation in a zk proof depends on the `length` you specify, so prefer a smaller value if possible.
```

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
 
-  // TODO: Izzy to document
+  /**
+   * **Warning: This function is mainly for internal use. Normally it is not intended to be used by a zkApp developer.
+   * 
+   * In snarkyJS, addition and scaling (multiplication of variables by a constant) of variables is represented as an AST - [abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree). For example, the expression `x.add(y).mul(2)` is represented as `Scale(2, Add(x, y))`.
+   * 
+   * Only when the variable is needed in a multiplicative or any higher level constraint (for example multiplication of two {@link Field} elements) a new internal variable is created to represent the operation.
+   * 
+   * The `seal()` function tells snarkyJS to stop building an AST and create a new variable right away.
```
mitschabaude, line 581: Awesome comment! Just a small thing: Can you spell it SnarkyJS instead of snarkyJS everywhere?
```suggestion
   * The `seal()` function tells SnarkyJS to stop building an AST and create a new variable right away.
```

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   */
   rangeCheckHelper(numBits: number): Field;
 
   /**
-   * Checks whether this is a hard-coded constant in the Circuit.
+   * Check whether this {@link Field} element is a hard-coded constant in the Circuit.
+   * If a {@link Field} is constructed outside a zkApp method, it is a constant.
+   * 
+   * ```ts
+   * console.log(Field(42).isConstant()); // False
```
mitschabaude, line 602:
```suggestion
   * console.log(Field(42).isConstant()); // True
```

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * ```
+   * 
+   * ```ts
+   * @method(x: Field) {
+   *    console.log(x.isConstant()); // True
```
mitschabaude, line 607:
```suggestion
   *    console.log(x.isConstant()); // False
```
you mixed them up :D

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
-   * Deserialize a JSON structure into a {@link Field}.
-   * This operation does NOT affect the circuit and can't be used to prove anything about the string representation of the Field.
+   * **Warning: This function is mainly for internal use. Normally it is not intended to be used by a zkApp developer.
+   * 
+   * This function is the implementation of {@link Provable.toInput} in {@link Field} type.
+   * This function acts identical to {@link Field.toFields} for {@link Field} type.
+   * 
+   * @param value - The {@link Field} element to get the `input` array.
+   * 
+   * @return A {@link Field} array of length 1 created from this {@link Field}.
```
mitschabaude, line 772: it's not quite true what you say here - the output is an object of type `{ fields: [x] }`, and not just an array

mitschabaude (changes requested):
Thanks for these great comments @yunus433! I have a couple of suggestions for improvement

src/snarky.d.ts
```diff
@@ -28,6 +28,8 @@ export { Test };
  * `Provable<T>` is the general circuit type interface. It describes how a type `T` is made up of field elements and auxiliary (non-field element) data.
  *
  * You will find this as the required input type in a few places in snarkyjs. One convenient way to create a `Provable<T>` is using `Struct`.
+ * 
+ * The properties and methods on the `Provable` type exist in all base snarkyjs types as well (aka. {@link Field}, {@link Bool}, etc.). In most of the cases, a ZK app developer does not need these functions to create dApps.
```
barriebyron, line 32:
```suggestion
 * The properties and methods on the `Provable` type exist in all base SnarkyJS types as well (aka. {@link Field}, {@link Bool}, etc.). In most cases, a ZK app developer does not need these functions to create Dapps.
```

barriebyron: I really like this description, thank you @yunus433 

See the word list in the docs style guide for spelling and capitalization guidance. We have inconsistencies in our docs and are working to establish consistency. Dapps is one example of how we have landed on the Dapp spelling even when that spelling might not be the most common

src/snarky.d.ts
```diff
@@ -28,6 +28,8 @@ export { Test };
  * `Provable<T>` is the general circuit type interface. It describes how a type `T` is made up of field elements and auxiliary (non-field element) data.
  *
  * You will find this as the required input type in a few places in snarkyjs. One convenient way to create a `Provable<T>` is using `Struct`.
```
barriebyron, line 30:
```suggestion
 * You will find `Provable<T>` as the required input type in a few places in SnarkyJS. One convenient way to create a `Provable<T>` is using `Struct`.
```

barriebyron: I am always on the lookout for adding a clear subject to replace "this" and "it"
see https://developers.google.com/style/pronouns#ambiguous-pronoun-references

src/snarky.d.ts
barriebyron: TIL `@example` doclet!

src/snarky.d.ts
```diff
@@ -51,489 +53,767 @@ declare interface ProvablePure<T> extends Provable<T> {
...
+   * 
+   * ```typescript
+   * const field_negative = new Field(-1); // Valid Field contruction, equivalent to `new Field(28948022309329048855892746252171976963363056481941560715954676764349967630336n)`as the order of the pasta field is 28948022309329048855892746252171976963363056481941560715954676764349967630337n
+   * ```
+   * 
+   * **Important: All the functions defined on a Field (arithmetic, logic, etc.) takes their arguments as `field-like`. A {@link Field} itself is also defined as a field-like element.
+   * 
+   * @param value - the value to coerce to a {@link Field}
+   * 
+   * @return A {@link Field} element which the value coerced from the argument in the pasta field.
```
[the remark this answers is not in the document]
barriebyron: fix the subject-verb agreement "take"

**Important**: All the functions defined on a Field (arithmetic, logic, etc.) take their arguments as "field-like". A Field itself is also defined as a field-like element.

barriebyron (commented):
@yunus433  thank you for this PR! I learned a lot reviewing the suggested updates. I do not have technical context for the many suggestions and appreciate the detailed feedback from @mitschabaude

barriebyron:
new PR for this work is in https://github.com/o1-labs/snarkyjs/pull/917

PR #902, opened 2023-05-08
Move Field to JS

mitschabaude:
- handle proof public inputs in js
- changelog
- use Empty / undefined as the empty provable
- bindings
- fixup


src/examples/zkapps/voting/test.ts
```diff
@@ -468,10 +468,10 @@ export async function testSet(
         voting.voterRegistration(newVoter1);
       },
       feePayer,
-      'assert_equal'
+      'Member already exists!'
     );
```
mitschabaude, line 472: some changes to errors required us to update some of the tests

src/lib/core.ts
```diff
@@ -1,44 +1,39 @@
...
 import { Scalar as ScalarBigint } from '../provable/curve-bigint.js';
 import { mod } from '../bindings/crypto/finite_field.js';
 
 export { Field, Bool, Scalar, Group };
 
-type InternalConstantField = { value: [0, Uint8Array] };
+/**
+ * An element of a finite field.
+ */
+const Field = toFunctionConstructor(InternalField);
```
mitschabaude, line 14: Here is where `Field` is turned from a class into a "class that can also be used as a function", to support `Field(x)`

src/snarky.d.ts
```diff
@@ -170,22 +265,22 @@ declare class Field {
    * let sum = a.add(5)
    * ```
    */
-  add(y: Field | number | string | boolean): Field;
+  add(y: Field | number | string | bigint): Field;
```
mitschabaude, line 268: I think nobody ever used a `boolean` to specify a field element, anywhere. I removed this feature and made sure you can use `bigint` everywhere instead

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+    if (Array.isArray(x)) {
+      this.value = x;
+      return;
+    }
+    // TODO this should handle common values efficiently by reading from a lookup table
+    this.value = FieldVar.constant(Fp(x));
+  }
+
+  // helpers
+  static #isField(x: bigint | number | string | Field | FieldVar): x is Field {
```
mitschabaude, line 93: I'm using the new JS private class fields with `#` to encapsulate logic. Right now, these are compiled to a polyfill by TS, but once we switch to es2022 target the builtin JS feature is used

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+      Object.assign(Object.create(new Field(1).constructor.prototype), {
+        value: [0, uint8array],
+      }) as Field,
+      offset + 32,
+    ];
+  },
+});
+
+function isField(x: unknown): x is Field {
+  return x instanceof Field || (x as any) instanceof SnarkyFieldConstructor;
```
mitschabaude, line 769: in some places we still have to deal with the possibility of field classes created in OCaml

src/lib/field.unit-test.ts
```diff
@@ -0,0 +1,317 @@
...
+});
+
+// special generator
+let SmallField = Random.reject(
+  Random.field,
+  (x) => x.toString(2).length > Fp.sizeInBits - 2
+);
+
+// arithmetic, both in- and outside provable code
+equivalent2((x, y) => x.add(y), Fp.add);
```
mitschabaude, line 45: the idea of these tests is that you have two implementations - the TS finite field `Fp`, and `Field` - and you test that they are equivalent, both in- and outside provable code. "Equivalent" also means that one throws an error if and only if the other throws an error

mitschabaude (commented):
some comments for reviewers!

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+   */
+  assertEquals(y: Field | bigint | number | string, message?: string) {
+    try {
+      if (this.isConstant() && isConstant(y)) {
+        if (this.toBigInt() !== toFp(y)) {
+          throw Error(`Field.assertEquals(): ${this} != ${y}`);
+        }
+        return;
+      }
+      Snarky.field.assertEqual(this.value, Field.#toVar(y));
```
Trivo25, line 159: am I correct in my assumption that `Snarky` only deals with variables (in-circuit variables?) and everything that is constant related has to be dealt with in JS now?

mitschabaude: snarky has methods that deal with both, but we don't use them because handling all constant arithmetic on the JS side is an important milestone of the whole refactor. it's the main step towards `snarkyjs/light`

src/snarky.d.ts
```diff
@@ -43,11 +45,19 @@ declare interface ProvablePure<T> extends Provable<T> {
   check: (x: T) => void;
 }
 
+// ocaml types
+type MlTuple<X, Y> = [0, X, Y];
```
Trivo25, line 49: is there a reason why OCaml arrays always start with a leading 0?

Trivo25 (commented):
first pass looks great!

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+  /**
+   * @deprecated use `x.equals(0)` which is equivalent
+   */
+  isZero() {
+    if (this.isConstant()) {
+      return Bool(this.toBigInt() === 0n);
+    }
+    // create witnesses z = 1/x, or z=0 if x=0,
+    // and b = 1 - zx
+    let [, b, z] = Snarky.exists(2, () => {
```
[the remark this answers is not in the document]
mitschabaude: Izzy explained it to me, it's like saying (in a mathematical proof), "there exists x, such that ...". That's a good analogy for how you use `exists` in provable code

src/snarky.d.ts
dannywillems: I would guess it is to mimic the memory layout of the native compiler. The native Caml runtime will represent tuples, records and arrays with block with a tag `0`, documentation [here](https://dev.realworldocaml.org/runtime-memory-layout.html).
> Tuples, records, and arrays are all represented identically at runtime as a block with tag 0. Tuples and records have constant sizes determined at compile time, whereas arrays can be of variable length. While arrays are restricted to containing a single type of element in the OCaml type system, this is not required by the memory representation.     

And in the JS runtime, it seems it is mimic (tuple, records and arrays are identical, i.e. a JS array with the first element set to `0`.

src/lib/field.test.ts
```diff
@@ -4,7 +4,7 @@ describe('Field constructor', () => {
   beforeAll(() => isReady);
```
Trivo25, line 4: We have a `field.test.ts` and a `field.unit-test.ts`, although one uses Jest, I think we should consider combining both into `field.unit-test.ts`

Trivo25 (approved):
awesome! lgtm

src/lib/field.test.ts
mitschabaude: yeah you're right, the old test does almost nothing

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+
+  /**
+   * Squares this {@link Field} element.
+   *
+   * ```typescript
+   * const x2 = x.square();
+   * x2.assertEquals(x.mul(x));
+   * ```
+   */
+  square() {
```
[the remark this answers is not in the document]
mitschabaude: Very useful, we should do it!

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+  },
+  // TODO: handle (special) constants
+  add(x: FieldVar, y: FieldVar): FieldVar {
+    return [FieldType.Add, x, y];
+  },
+  // TODO: handle (special) constants
+  scale(x: FieldVar, c: FieldConst): FieldVar {
+    return [FieldType.Scale, c, x];
+  },
+  [0]: [0, FieldConst[0]] satisfies ConstantFieldVar,
```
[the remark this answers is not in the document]
mitschabaude: It acts as a type hint, that's why I used it here. Without it, TS thinks `[0, FieldConst[0]]` is of type `(number | FieldConst)[]`. With it, TS understands it is of type `[number, FieldConst]`.

The only other way to change the type of an inline expression is with `as type`, but that force-overrides the type, so it's not type-safe. `satisfies` on the other hand is type-safe (it won't let it pass when the expression _doesn't_ satisfy the type)

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   */
+  toConstant(): ConstantField {
+    if (this.isConstant()) return this;
+    // TODO: fix OCaml error message, `Can't evaluate prover code outside an as_prover block`
+    let value = Snarky.field.readVar(this.value);
+    return new Field(FieldVar.constant(value)) as ConstantField;
+  }
+
+  /**
+   * Serialize the {@link Field} to a bigint, e.g. for printing. Trying to print a {@link Field} without this function will directly stringify the Field object, resulting in an unreadable output.
```
bkase, line 196: nit: `in an unreadable` -> `in unreadable`

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   *
+   * @example
+   * ```ts
+   * const someField = Field(42);
+   * console.log(someField.toBigInt());
+   * ```
+   *
+   * @return A bigint equivalent to the bigint representation of the Field.
+   */
+  toBigInt() {
```
bkase, line 208: can we change this `debugToBigInt()` or `toBigIntDebug()` something 
 like that to make it clear in the name that it's only for debugging?

We've seen so many people make this mistake, it's probably worth it.

I imagine tactically, you'd need to deprecate these since they're in-use already and we can remove them fully later.

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   *
+   * @example
+   * ```ts
+   * const someField = Field(42);
+   * console.log(someField.toString());
+   * ```
+   *
+   * @return A string equivalent to the string representation of the Field.
+   */
+  toString() {
```
bkase, line 226: same here ^^

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   *
+   * **Warning**: This is a modular negation. For details, see the {@link sub} method.
+   *
+   * @return A {@link Field} element that is equivalent to the element multiplied by -1.
+   */
+  neg() {
+    if (this.isConstant()) {
+      return new Field(Fp.negate(this.toBigInt()));
+    }
+    // return new AST node Scale(-1, x)
```
bkase, line 314: did you mean the flip the parameters in the comment? Should we flip them in the actual representation to match this?

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   * const someField = Field(42);
+   * const inverse = someField.inv();
+   * inverse.assertEquals(Field(1).div(example)); // This statement is always true regardless of the value of `someField`
+   * ```
+   *
+   * **Warning**: This is a modular inverse. See {@link div} method for more details.
+   *
+   * @return A {@link Field} element that is equivalent to one divided by this element.
+   */
+  inv() {
```
bkase, line 401: Can we add in the comment that this throws on invert by zero? Should we have some warning about division by zero in-snark vars too?

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   * // You can use the reverse operation of division (multiplication) to prove the quotient is calculated correctly.
+   *
+   * quotient.mul(y).assertEquals(x);
+   * ```
+   *
+   * @param value - a "field-like" value to divide with the {@link Field}.
+   *
+   * @return A {@link Field} element equivalent to the modular division of the two value.
+   */
+  div(y: Field | bigint | number | string) {
```
bkase, line 447: same here ^^ note on what errors are thrown and div-by-zero warnings as appropriate

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   * const squareRoot = someField.sqrt();
+   *
+   * squareRoot.mul(squareRoot).assertEquals(someField); // This statement is always true regardless of the value of `someField`
+   * ```
+   *
+   * **Warning**: This is a modular square root. See `div()` method for more details.
+   *
+   * @return A {@link Field} element equivalent to the square root of the {@link Field} element.
+   */
+  sqrt() {
```
bkase, line 495: same as above

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+  }
+
+  // internal base method for all comparisons
+  #compare(y: FieldVar) {
+    // TODO: support all bit lengths
+    let maxLength = Fp.sizeInBits - 2;
+    asProver(() => {
+      let actualLength = this.toBigInt().toString(2).length;
+      if (actualLength > maxLength)
+        throw Error(
```
bkase, line 583: can you add in the doccomments that the comparison methods can throw this?

src/lib/field.ts
```diff
@@ -0,0 +1,1194 @@
...
+   *
+   * @example
+   * ```ts
+   * const someField = Field(42);
+   * console.log(someField.toJSON());
+   * ```
+   *
+   * @return A string equivalent to the JSON representation of the {@link Field}.
+   */
+  toJSON() {
```
bkase, line 1049: same as the note on the debug representation -- feel free to defer all this to another issue though, it's looking like it won't be a small change on top of what you have

bkase (changes requested):
Amazing! Excited for `snarkyjs/light`!

src/lib/field.ts
mitschabaude: I'm flipping them in the `Snarky.field.scale` function to match the type!

src/lib/field.ts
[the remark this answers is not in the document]
mitschabaude: Good idea, I added it to the doccomment of the (internal) `exists` function!

src/lib/field.ts
```diff
@@ -0,0 +1,794 @@
...
+   * ```
+   */
+  add(y: Field | bigint | number | string): Field {
+    if (this.isConstant() && isConstant(y)) {
+      return new Field(Fp.add(this.toBigInt(), toFp(y)));
+    }
+    // return new AST node Add(x, y)
+    let z = Snarky.field.add(this.value, Field.#toVar(y));
+    return new Field(z);
+  }
```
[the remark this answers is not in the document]
mitschabaude: We could do that, but I don't feel it would make the code better. IMO, patterns are often better than abstractions, as in this case. 

Abstractions use specialized concepts (`binop(x, y, tsImpl, snarkyImpl)`) that have to be learned, while patterns use plain language features like if conditions that everyone already knows

Abstractions get leaky, or are only partially applicable. Patterns are flexible to adapt to any use case.

Unrelated reason why I'm not in favor is that I try to avoid all wrapper functions, because they add a line to our stack traces and our stack traces are too long.

src/lib/field.ts
bkase: > Instead, everything that's not a variable should just be a plain JS type, like bigint or number, and it should be possible to use that in place of constant Field elements everywhere

I like this ^^, especially because you can monkey-patch things in JavaScript (though of course we should be careful)

bkase (approved):

PR #908, opened 2023-05-10
Fix `init()` by removing optional key argument

mitschabaude:
fixes https://github.com/o1-labs/zkapp-cli/issues/406

This removes the optional `zkappKey` argument to `smartContract.init()`

For context, this argument was originally added for the case that you make `init()` a `@method`, to prove that it was called and to get the `provedState = true` property on the zkapp account.

If `init()` is a `@method`, it means everybody can call it after deploy (since it doesn't require a signature!). That's bad of course, because we want `init()` to only be callable once; we don't want everyone to be able to reset the state.

The solution to this is to add a precondition on `provedState`: to call `init()`, `provedState` must be false. Since `init()` causes `provedState` to become true, and assuming that the state is henceforth only updated with proofs, nobody will ever be able to call it again.

At the time of introducing `init()`, the proved state precondition was not available. Therefore, as a workaround, we added an optional `privateKey` argument to `init()`, and made it prove that `privateKey.toPublicKey()` equals the account's address. With this workaround, only the owner of the zkapp private key could call `init()` (but they could do so as often as they wanted). We no longer need this workaround since the `provedState` precondition is available now.

Removing the optional argument also solves a bug that occured when you didn't annotate the optional argument as a `@method` argument, but still passed in the private key. It's not supported to leave out a `@method` argument, but the previous `deploy()` behaviour made it extremely easy to accidentally cause such a bug. (see https://github.com/o1-labs/zkapp-cli/issues/406, where the Sudoku zkApp didn't specify the zkappKey argument). In general, optional arguments to a zkApp method are an absolute anti-pattern


CHANGELOG.md
```diff
@@ -17,11 +17,20 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+
 ### Changes
 
 - New decorator `CatchAndPrettifyStacktrace` for printing better stack traces thrown from SnarkyJS https://github.com/o1-labs/snarkyjs/pull/890
   - Cleans up stack traces by filtering out extra details, making them more readable
 
+### Fixes
+
+- Fix running SnarkyJS in Node.js on Windows https://github.com/o1-labs/snarkyjs-bindings/pull/19 (@wizicer)[https://github.com/wizicer]
+
```
mitschabaude, line 33: Unrelated: this points to latest `snarkyjs-bindings` with a fix for windows https://github.com/o1-labs/snarkyjs-bindings/pull/19

Trivo25 (approved):

PR #911, opened 2023-05-10
Fix for actions cache key string.

shimkiv:
We are fetching actions and putting them into the cache using `Archive-Node-API` GraphQL endpoint (as one of the cache key string entities), but then we're getting data from cache using Mina's Daemon GraphQL endpoint (as one of the cache key string entities), because `getCachedActions` is called without the GQL endpoint specified and hence uses defaults one (which is Mina's Daemon GQL endpoint).


mitschabaude (approved):
thanks!

PR #916, opened 2023-05-22
Improve stack traces

### LINKED ISSUE #928: Make stack trace prettifying cover >90% of dev flows

mitschabaude:
closes #928 

bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/23

This PR does 3 things to improve stack traces:
1. Modify `prettifyStacktrace()` so that it handles error arrays from OCaml (`exception` type in OCaml = `[0, .., Error]`)
2. Apply `prettifyStacktrace()` everywhere we call into OCaml, in particular all circuit runners
3. Replace the `context.runWith(ctx, () => {})` pattern with a pattern that instead enters and leaves the context separately. This is done to reduce the number of entries in the stack trace, which used to contain a lot of `runWith` calls

We also remove the stack trace prettifying in a couple of places, in particular the decorator for int classes. Here is the reasoning for why we don't need it on those classes:
- with the in-progress refactor, all non-provable operations will be done in TS
- therefore, we only call into OCaml when we run provable code
- therefore, we only need to prettify the stack trace at the entry points of running provable code, which this PR does


MartinMinkov (approved):
LGTM! :D

PR #932, opened 2023-05-28
Move Group to JavaScript

Trivo25:
This PR aims to move the Group implementation from OCaml to JavaScript, reducing OCaml dependencies as best as possible. The new Group implementation covers the same API surface as the old one and produces the exact same constraints (important for compatibility with already existent projects)

This will allow us to easier maintain more of the SnarkyJS elements by relying less on OCaml as well as giving us more control over what we are doing. Additionally, in the future this will enable a `snarkyjs/light` version by relying on native JS arithmetic instead of requiring users to load a lot of bindings for simple calculations -- builds on top of https://github.com/o1-labs/snarkyjs/pull/902 and is motivated by the same reasons

https://github.com/o1-labs/snarkyjs-bindings/pull/25

closes https://github.com/o1-labs/snarkyjs/issues/924


src/examples/primitive_constraint_system.ts
```diff
@@ -0,0 +1,66 @@
+import { Field, Group, Poseidon, Provable, Scalar } from 'snarkyjs';
+
+function mock(obj: { [K: string]: (...args: any) => void }, name: string) {
```
Trivo25, line 3: We didn't really have any regression tests that tested `Group` and its constraints, I just hacked this mock contract-like thing together to integrate it into the contract regression tests

src/examples/primitive_constraint_system.ts
```diff
@@ -0,0 +1,66 @@
...
+      return {
+        verificationKey: { data: '', hash: '' },
+      };
+    },
+    name,
+    digest: () => name,
+  };
+}
+
+const GroupMock = {
```
Trivo25, line 35: this one adds some "tests" for Group constraints

src/examples/vk_regression.ts
```diff
@@ -1,22 +1,39 @@
...
-
-await isReady;
+import { GroupCS } from './primitive_constraint_system.js';
 
 // usage ./run ./src/examples/vk_regression.ts --bundle --dump ./src/examples/regression_test.json
 let dump = process.argv[4] === '--dump';
 let jsonPath = process.argv[dump ? 5 : 4];
 
-const Contracts: (typeof SmartContract)[] = [
+type MinimumConstraintSystem = {
```
Trivo25, line 12: some minor changes to accommodate the Group constraint tests, see above

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+   */
+  static get generator() {
+    return new Group(Pallas.one.x, Pallas.one.y);
+  }
+
+  /**
+   * Coerces anything group-like to a {@link Group}.
+   */
+  constructor(g: {
+    x: FieldVar | Field | number | string | bigint;
```
Trivo25, line 33: Using the overloaded constructor so that the new implementation is equivalent to the old OCaml one, which allowed both `(x, y)` as well as `{x, y}` as arguments. I am not using the `FieldLike` or `GroupLike` type from above so users arent confused by it and see "the actual type".

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
+import { Scalar, withMessage } from './core.js';
+import { Field, FieldVar, isField } from './field.js';
+import { Bool, Snarky } from '../snarky.js';
+import { Field as Fp } from '../provable/field-bigint.js';
+import { Pallas } from '../bindings/crypto/elliptic_curve.js';
+
+export { Group };
+
+type FieldLike = FieldVar | Field | number | string | bigint;
```
Trivo25, line 9: just to make the internal code a bit more readable

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+    } else {
+      x = arg1 as FieldLike;
+      y = arg2 as FieldLike;
+    }
+
+    this.x = isField(x) ? x : new Field(x);
+    this.y = isField(y) ? y : new Field(y);
+
+    /*
+    technically elements are only group elements if they are on the curve (aka for two points y, x the following equation must hold y^2 = x^3 + 5 over our field)
```
Trivo25, line 61: I briefly talked to Gregor about this; the constructor doesn't check if a (constant) element is on the curve - it only does so if its a witness. Adding this would add constraints that could break existing contracts. That could be something for a future PR

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+      let { x: x2, y: y2 } = g;
+
+      return Bool(x1.equals(x2).and(y1.equals(y2)));
+    } else {
+      /*
+    let equal_x = x1.equals(x2);
+    let equal_y = y1.equals(y2);
+    return equal_x.and(equal_y);*/
+
+      let z = Snarky.group.equals(Group.#toTuple(this), Group.#toTuple(g));
```
Trivo25, line 219: I am still trying to figure out why this isn't producing the same set of constraints:

I tried implementing it with `x1.equals(x2).and(y1.equals(y2))` - which is (at least to my knowledge) equaivalent to the OCaml implementation of 

`Boolean.all [ Impl.Field.equal x1 x2; Impl.Field.equal y1 y2 ]`
where `Boolean.all` just calls `b1 && b2` for tuples. it produces the same amount of rows/gates but with slightly different coefficients?

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+      if (this.x.toBigInt() === 0n) {
+        return g;
+      } else if (g.x.toBigInt() === 0n) {
+        return this;
+      } else {
+        let g_proj = Pallas.add(this.#toAffine(), g.#toAffine());
+        return Group.#fromProjective(g_proj);
+      }
+    } else {
+      let [, x, y] = Snarky.group.add(Group.#toTuple(this), Group.#toTuple(g));
```
Trivo25, line 136: I would like to move this entirely to JavaScript, but wanted to establish a baseline first

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+      typeof s === 'bigint' || typeof s === 'number'
+        ? Scalar.fromBigInt(BigInt(s))
+        : s;
+    let fields = scalar.toFields();
+
+    if (this.#isConstant() && fields.every((f) => f.isConstant())) {
+      let g_proj = Pallas.scale(this.#toAffine(), BigInt(scalar.toJSON()));
+      return Group.#fromProjective(g_proj);
+    } else {
+      let [, x, y] = Snarky.group.scale(Group.#toTuple(this), [
```
Trivo25, line 175: similar to the comment above

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+   * ```ts
+   * g1.assertEquals(g2);
+   * ```
+   */
+  assertEquals(g: Group, message?: string) {
+    let { x: x1, y: y1 } = this;
+    let { x: x2, y: y2 } = g;
+
+    x1.assertEquals(x2, message);
+    // y1.assertEquals(y2, message); need to enable this later on, but it breaks constraint backwards compatibility
```
Trivo25, line 196: The OCaml implementation only checked if the `x` coordinates are equal, allowing devs to prove `g1 === g2` even if `g1.y !== g2.y`

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+    x: FieldVar | Field | number | string | bigint,
+    y: FieldVar | Field | number | string | bigint
+  );
+  constructor(
+    arg1: GroupLike | FieldLike,
+    arg2: FieldLike | undefined = undefined
+  ) {
+    let x, y: FieldLike;
+
+    if (isGroupLike(arg1)) {
```
Trivo25, line 49: i feel like this is a bit hacky, if anyone has recommendations let me know :D

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+      if (this.x.toBigInt() === 0n) {
+        return g;
+      } else if (g.x.toBigInt() === 0n) {
+        return this;
+      } else {
+        let g_proj = Pallas.add(this.#toAffine(), g.#toAffine());
+        return Group.#fromProjective(g_proj);
+      }
+    } else {
+      let [, x, y] = Snarky.group.add(this.#toTuple(), g.#toTuple());
```
Trivo25, line 136: I will move that implementation to pure JS as well, just wanted to establish a baseline first. 
I will stack another PR with the detailed implementation on top of this one soon

src/lib/group.ts
```diff
@@ -0,0 +1,394 @@
...
+      typeof s === 'bigint' || typeof s === 'number'
+        ? Scalar.fromBigInt(BigInt(s))
+        : s;
+    let fields = scalar.toFields();
+
+    if (this.#isConstant() && fields.every((f) => f.isConstant())) {
+      let g_proj = Pallas.scale(this.#toAffine(), BigInt(scalar.toJSON()));
+      return Group.#fromProjective(g_proj);
+    } else {
+      let [, x, y] = Snarky.group.scale(this.#toTuple(), [
```
Trivo25, line 175: same as above

src/lib/group.test.ts
```diff
@@ -10,6 +10,10 @@ import {
 } from 'snarkyjs';
 
 describe('group', () => {
+  let g = Group({
+    x: -1,
+    y: 2,
+  });
```
mitschabaude, line 16: love it!! I didn't know there was a neat example group element. we should make this the generator

src/lib/group.test.ts
```diff
@@ -25,8 +29,18 @@ describe('group', () => {
       it('g+g does not throw', () => {
         expect(() => {
           Provable.runAndCheck(() => {
-            const x = Provable.witness(Group, () => Group.generator);
-            const y = Provable.witness(Group, () => Group.generator);
+            const x = Provable.witness(Group, () =>
+              Group({
+                x: -1,
+                y: 2,
+              })
```
mitschabaude, line 36: nit: would be more readable to reuse `g` here

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+      y: this.y.toBigInt(),
+      infinity: false,
+    });
+  }
+
+  /**
+   * Adds this {@link Group} element to another {@link Group} element.
+   *
+   * ```ts
+   * let g1 = Group({ x: 1, y: 1})
```
mitschabaude, line 89: would be good to make the example code valid (I guess this one throws because it's not a valid group element)

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+
+    if (this.#isConstant()) {
+      const { add, mul, square } = Fp;
+
+      let x_bigint = this.x.toBigInt();
+      let y_bigint = this.y.toBigInt();
+
+      let onCurve =
+        add(mul(x_bigint, mul(x_bigint, x_bigint)), Pallas.b) ===
+        square(y_bigint);
```
mitschabaude, line 50: I notice there's no way to pass in a `zero` / `infinity` element here. however, not all methods enforce this restriction in the same way. for example, you can to `g.add(g.neg())` and will end up with the group element `(1, 1)` which supposedly stands for `zero`

I think the whole class should pick one consistent way of treating `zero`. It should be consistent between in-snark and out of snark, and could be one of the following:
1. we never ever represent `zero` as a Group. methods that would result in `zero` will throw
2. we have one unique way to represent `zero`. something like `(0, 0)` would make sense.

it's ok if you can use `zero` in some but not all operations on group elements. however, if the constructor and `Group.check()` doesn't allow `zero`, then I'd argue no method should allow you to create `zero`. In order to preserve constraints, I think this might be what you have to do

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+        throw Error(
+          `(x: ${x_bigint}, y: ${y_bigint}) is not a valid group element`
+        );
+      }
+    }
+  }
+
+  // helpers
+  static #fromAffine({ x, y }: { x: bigint; y: bigint; infinity: boolean }) {
+    return new Group({ x, y });
```
mitschabaude, line 62: this helper seems a bit dangerous since it's ignoring (and thus, allowing) `infinity` without guaranteeing a canonical `(x, y)` representation of the zero element. I recommend one of two simple options:
- since this seems to be only used once, inline it
- make this method throw when is encounters `infinity: true`

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+  #toTuple(): [0, FieldVar, FieldVar] {
+    return [0, this.x.value, this.y.value];
+  }
+
+  #isConstant() {
+    return this.x.isConstant() && this.y.isConstant();
+  }
+
+  #toAffine() {
+    return Pallas.fromAffine({
```
mitschabaude, line 78: the name here is confusing! it should be called `#toProjective()`, right?

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+   */
+  add(g: Group) {
+    if (this.#isConstant() && g.#isConstant()) {
+      if (this.x.toBigInt() === 0n) {
+        return g;
+      } else if (g.x.toBigInt() === 0n) {
+        return this;
+      } else {
+        let g_proj = Pallas.add(this.#toAffine(), g.#toAffine());
+        return Group.#fromProjective(g_proj);
```
mitschabaude, line 101: see my other comments about consistent treatment of `zero`. This seems to accept `(0, y)` and treats that as zero element. at the same time, adding `g` to `g.neg()` (which should result in zero) returns `(1, 1)` IIUC

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+   * ```
+   */
+  scale(s: Scalar | number | bigint) {
+    let scalar =
+      typeof s === 'bigint' || typeof s === 'number'
+        ? Scalar.fromBigInt(BigInt(s))
+        : s;
+    let fields = scalar.toFields();
+
+    if (this.#isConstant() && fields.every((f) => f.isConstant())) {
```
mitschabaude, line 139: TODO (not in this PR): when `Scalar` move merged, use `scalar.isConstant()`

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+      let g_proj = Pallas.scale(this.#toAffine(), BigInt(scalar.toJSON()));
+      return Group.#fromProjective(g_proj);
+    } else {
+      let [, x, y] = Snarky.group.scale(this.#toTuple(), [
+        0,
+        ...fields.map((f) => f.value).reverse(),
```
mitschabaude, line 145: while I appreciate keeping the `Snarky` interface low-level and having logic like reversing the scalar bits on the JS side, it would be good if a doccomment on `Snarky.group.scale` documented this unexpected behaviour

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+    if (this.#isConstant() && g.#isConstant()) {
+      let { x: x1, y: y1 } = this;
+      let { x: x2, y: y2 } = g;
+
+      return Bool(x1.equals(x2).and(y1.equals(y2)));
+    } else {
+      /*
+    let equal_x = x1.equals(x2);
+    let equal_y = y1.equals(y2);
+    return equal_x.and(equal_y);*/
```
mitschabaude, line 185: this lgtm! if it results in a constraint change, I'd also try to switch the two equals constraints (it's OCaml so stuff is sometimes happening in reverse order :D)

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+   * ```ts
+   * g1.equals(g1); // Bool(true)
+   * ```
+   */
+  equals(g: Group) {
+    if (this.#isConstant() && g.#isConstant()) {
+      let { x: x1, y: y1 } = this;
+      let { x: x2, y: y2 } = g;
+
+      return Bool(x1.equals(x2).and(y1.equals(y2)));
```
mitschabaude, line 180: nit: does this need a `Bool` constructor?

src/lib/group.ts
```diff
@@ -0,0 +1,364 @@
...
+    x: FieldVar | Field | number | string | bigint,
+    y: FieldVar | Field | number | string | bigint
+  ) {
+    return new Group({ x, y });
+  }
+
+  /**
+   * Adds a {@link Group} element to another one.
+   */
+  static add(g1: Group, g2: Group) {
```
mitschabaude, line 229: IMO all these extra static methods are weird and they should either be deprecated or be present consistently on all classes, i.e. added to `Field` for example

src/snarky.d.ts
```diff
@@ -177,6 +176,28 @@ declare const Snarky: {
...
+  group: {
+    /**
+     * Addition of two group elements, handles only variables.
+     */
+    add(
+      p1: MlTuple<FieldVar, FieldVar>,
+      p2: MlTuple<FieldVar, FieldVar>
+    ): MlTuple<FieldVar, FieldVar>;
+
+    onCurve(p1: MlTuple<FieldVar, FieldVar>): void;
```
mitschabaude, line 188: nit: can we rename this to `assertOnCurve()` to make it more self-documenting?

mitschabaude (commented):
Awesome work, I like how this keeps the OCaml API slim.

A few comments, most notably regarding consistent treatment of the `zero` Group element. I haven't looked up myself how the in-snark OCaml methods treat zero, so I'll leave it to you to identify a consistent way of doing it (and implement + ideally document it)

src/lib/group.ts
Trivo25: ah yes, I did that before I found `(-1, 2)` :laughing:

src/lib/group.ts
Trivo25: Ohh that was actually it, it was checking it in reverse

src/lib/group.ts
Trivo25: I am also not a massive fan, I would rather deprecate them

src/lib/group.ts
Trivo25: Oh nope, this actually still behaves strangely :thinking:

src/lib/group.ts
Trivo25: That's a good catch - I'll have to take another look at this.
OCaml returns `(0, 0)` for constants, whereas `(12664759760331458874453076485325239921471337210849432813230171084403110838275, 19449452489080454700052938888178047022259553573804486106032048451047634501628)` for variables

mitschabaude:
note: this PR has some conflicts with the Scalar move. I prepared a branch `refactor/group-and-scalar` (snarkyjs + bindings) which combines the two, that can be merged into whichever of the two PRs gets merged second

src/lib/group.ts
Trivo25: although allowing certain functions to handle the `zero` element would require less constraints, I feel like we are limiting certain low-level operations that may require a `zero` element for certain use case, wdyt?

src/lib/group.ts
bkase: We can always recreate them with `bind` anyway -- JS devs are used to this; so I agree we should deprecate

bkase (changes requested):
This great!

There are a lot of subtle security gotchas in this sort of code -- as a side-effect of moving more of our logic to TypeScript, it becomes more likely that auditors will be productive reviewing this kind of code. Maybe we should be making a queue of std library code that we should get third-party audited and figuring out when to prioritize it.

src/examples/vk_regression.ts
```diff
@@ -1,22 +1,39 @@
 import fs from 'fs';
```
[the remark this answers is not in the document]
Trivo25: it already is running in the CI! The PR gets blocked if a constrained mismatch is detected

src/lib/group.ts
Trivo25: Gregor and I had a quick huddle, we thought about leaving it as simple as possible but making it uniform for now. I am about to move the entire 'add` algorithm to JS as well to gain a better understanding, because right now, the OCaml implementation also returns something "weird" for the `zero element (see comment above), We also had the idea of using a `Safe` and `Unsafe` namespace, similar to `Bool`

mitschabaude (approved):

src/lib/group.ts
bkase: safe and unsafe is a great idea 👍 , but yeah can come after

bkase (approved):

PR #935, opened 2023-05-31
Move Scalar to JS

### LINKED ISSUE #925: Move Scalar to JS
Same as https://github.com/o1-labs/snarkyjs/pull/902 but for the `Scalar` class

mitschabaude:
closes #925

bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/30

This PR continues the work of https://github.com/o1-labs/snarkyjs/pull/902 by moving the `Scalar` class from OCaml to JS. The OCaml class is removed entirely. Apart from minor API changes like introduction of `Scalar.toConstant()` (useful for tests), this keeps the same API intact. The internal representation of `Scalar` stays the same as before because the `Group` class relies on the structure and the PR to move that is in flight.

Some other things we also did, which stick to the theme of cleaning up the OCaml side:
- Remove back and forth conversion of private keys from the OCaml side
- Changed `PrivateKey.to/fromBase58` to be the pure TS implementation
- Moved private key conversions from the `Ledger` module to the `Test` module, because they are only used for tests now
- Added a file where we will add tests for consistency that are no longer tested elsewhere as we move to pure TS implementations in snarkyjs
- Remove the `signFeePayer` method from the global exports and a couple of exposed functions that it depended on


src/lib/signature.ts
```diff
@@ -39,6 +46,20 @@ class PrivateKey extends CircuitValue {
...
+   * Convert this {@link PrivateKey} to a bigint
+   */
+  toBigInt() {
+    return this.s.toBigInt();
+  }
+
+  /**
+   * Create a {@link PrivateKey} from a bigint
+   */
+  static fromBigInt(sk: PrivateKeyBigint) {
```
Trivo25, line 59: I think we should add to the comment that creating a private key from a `bigint` may not be secure enough if the bigint wasn't sampled correctly

src/lib/scalar.ts
```diff
@@ -0,0 +1,352 @@
...
+  neg() {
+    let x = this.#assertConstant('neg');
+    let z = Fq.negate(x);
+    return Scalar.from(z);
+  }
+
+  /**
+   * Add scalar field elements.
+   *
+   * **Warning**: This method is not available for provable code.
```
Trivo25, line 127: Interesting - is this because Scalars are over Vesta and we do our provable operations over Pallas?

Trivo25 (approved):
very nice!

src/lib/core.ts
```diff
@@ -1,11 +1,9 @@
-import { bytesToBigInt } from '../bindings/crypto/bigint-helpers.js';
 import { defineBinable } from '../bindings/lib/binable.js';
 import { sizeInBits } from '../provable/field-bigint.js';
-import { Bool, Scalar } from '../snarky.js';
+import { Bool } from '../snarky.js';
 import { Field as InternalField } from './field.js';
 import { Group as InternalGroup } from './group.js';
-import { Scalar as ScalarBigint } from '../provable/curve-bigint.js';
-import { mod } from '../bindings/crypto/finite_field.js';
+import { Scalar } from './scalar.js';
```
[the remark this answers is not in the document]
mitschabaude: Good point. I think the need is not pressing and requires refactoring the current constructor, so would save it for a future PR

src/lib/scalar.ts
```diff
@@ -0,0 +1,360 @@
...
+   */
+  div(y: Scalar) {
+    let x = this.#assertConstant('div');
+    let y0 = Scalar.#assertConstantStatic(y, 'div');
+    let z = Fq.div(x, y0);
+    if (z === undefined) throw Error('Scalar.div(): Division by zero');
+    return Scalar.from(z);
+  }
+
+  // TODO don't leak 'shifting' to the user and remove these methods
```
bkase, line 182: +1

bkase (approved):
Looks great! Exploring these files are going to be a great way for SnarkyJS devs to learn more about our crypto stack

PR #967, opened 2023-06-06
Move low-level operations to JS

Trivo25:
This PR is a follow up to https://github.com/o1-labs/snarkyjs/pull/932. It further reduces dependencies to OCaml by re-using more of the existing clue as well as moving some additional functionality to JavaScript. The PR also addresses handling of the `zero`/`infinity` element by adding further constraints (it breaks existing smart contracts, whereas the previous PR tried to establish a baseline without any major constraint changes).

bindings https://github.com/o1-labs/snarkyjs-bindings/pull/35


src/lib/group.ts
```diff
@@ -20,6 +23,25 @@ class Group {
...
+   *
+   * ```typescript
+   * // g + -g = 0
+   * g.add(g.neg()).assertEquals(zero);
+   * // g + 0 = g
+   * g.add(zero).assertEquals(g);
+   * ```
+   */
+  static get zero() {
+    return new Group({
```
Trivo25, line 39: `zero` is `(1, 1)`, similar to the projective zero in the crypto implementation - `(0, 0)` may be a little more obvious but would have to make sure the native crypto code handles it the same way

src/lib/group.ts
```diff
@@ -86,33 +114,106 @@ class Group {
...
+      // and the implementation above (original OCaml implementation) returns something wild -> g + 0 != g where it should be g + 0 = g
+      let gIsZero = g.#isZero();
+      let thisIsZero = this.#isZero();
+
+      // if either one is the negation of the other, we just return the zero element since g + (-g) = 0 - but the OCaml implementation doesn't pick that up
+      let isNegation = g.neg().equals(this);
+
+      let isNewElement = gIsZero.or(thisIsZero).not().and(isNegation.not());
+
+      return Provable.switch(
```
Trivo25, line 187: similar to the constant implementation; if one of the operands is `zero`, we just return the original element `g`. If no operand is `zero`, we return `g1 + g2`, we also have to handle negation `g + (-g) = 0` since the original implementation didn't do so (it returned a seemingly random element)

src/lib/group.ts
```diff
@@ -167,15 +270,10 @@ class Group {
...
-      let { x: x2, y: y2 } = g;
+    let { x: x1, y: y1 } = this;
+    let { x: x2, y: y2 } = g;
 
-      return Bool(x1.equals(x2).and(y1.equals(y2)));
-    } else {
-      let z = Snarky.group.equals(this.#toTuple(), g.#toTuple());
-      return Bool.Unsafe.ofField(new Field(z));
-    }
+    return x1.equals(x2).and(y1.equals(y2));
```
Trivo25, line 276: if we already break the constraints, might as well move this part to JS entirely

src/snarky.d.ts
```diff
@@ -178,24 +179,20 @@ declare const Snarky: {
...
+      p2: MlGroup,
+      p3: MlGroup,
+      inf: FieldVar,
+      same_x: FieldVar,
+      slope: FieldVar,
+      inf_z: FieldVar,
+      x21_inv: FieldVar
+    ): MlGroup;
+
+    scale(p: MlGroup, s: MlArray<BoolVar>): MlGroup;
```
Trivo25, line 195: `Group` now only relies on 1) JS-Field implementations and 2) the pure `ec_add` gate (moved helper code to JS) and the `scale` gate + helpers in OCaml

src/lib/group.ts
```diff
@@ -86,32 +116,102 @@ class Group {
...
+        x21_inv.value
+      );
+
+      // similarly to the constant implementation, we check if either operand is zero
+      // and the implementation above (original OCaml implementation) returns something wild -> g + 0 != g where it should be g + 0 = g
+      let gIsZero = g.#isZero();
+      let thisIsZero = this.#isZero();
+
+      // if either one is the negation of the other, we just return the zero element since g + (-g) = 0 - but the OCaml implementation doesn't pick that up
+      let isNegation = g.neg().equals(this);
```
mitschabaude, line 190: this should just be set to `inf` if it's not hard-coded anymore

afaict, `inf` will be 0, and constrained to 0, if either `this` of `g` is 0

src/lib/group.ts
mitschabaude: I expect this currently to throw an error if `this = g = (0,0)`, because `Provable.switch` expects a boolean vector with only one `1` set (this isn't part of the constraints, though)

mitschabaude: `Provable.switch` adds up the inputs times the booleans. So, constraint-wise, it's all good, because it's fine to add `0 + 0` if you want `0`

src/lib/group.ts
```diff
@@ -128,7 +228,9 @@ class Group {
 
     if (this.#isConstant() && scalar.isConstant()) {
       let g_proj = Pallas.scale(this.#toProjective(), scalar.toBigInt());
-      return Group.#fromProjective(g_proj);
+      // in the JS code, zero is denoted with (1, 1) - but here we want to convert it to (0, 0) (its less constraints that way)
+      let isZero = g_proj.x === 1n && g_proj.y === 1n;
+      return isZero ? new Group({ x: 0, y: 0 }) : Group.#fromProjective(g_proj);
```
mitschabaude, line 233: same comment as before: fix `#fromProjective()` instead of making this change

src/lib/group.ts
```diff
@@ -345,7 +442,14 @@ class Group {
...
     try {
-      Snarky.group.assertOnCurve(g.#toTuple());
+      const { x, y } = g;
+
+      let x2 = x.square();
+      let x3 = x2.mul(x);
+      let ax = x.mul(Pallas.a); // this will obviously be 0, but just for the sake of correctness
+
+      // we also check the zero element (0, 0) here
+      g.#isZero().or(x3.add(ax).add(Pallas.b).equals(y.square())).assertTrue();
```
mitschabaude, line 452: I wonder if in this one place, we should actually check both `x` and `y` to be zero instead of just `x`. or at least carefully think through the implications of leaving `y` unconstrained

mitschabaude (commented):
Great work, I have a few comments!

src/lib/group.test.ts
```diff
@@ -35,6 +51,46 @@ describe('group', () => {
...
+
+      it('(-g)+g = zero', () => {
+        expect(() => {
+          Provable.runAndCheck(() => {
+            const x = Provable.witness(Group, () => g);
+            const zero = Provable.witness(Group, () => Group.zero);
+            x.neg().add(x).assertEquals(zero);
+          });
+        }).not.toThrow();
+      });
```
mitschabaude, line 93: Please add a check for `0 + 0`

mitschabaude (changes requested):

src/lib/group.ts
Trivo25: Good point, I think we definitely should check that otherwise we will be able to create an invalid element with `(0, y)` and still prove, for example, `(0, y) === zero`

src/lib/group.ts
```diff
@@ -86,32 +116,102 @@ class Group {
...
+
+      let zero = new Field(0);
+
+      let same_x = Provable.witness(Field, () => x1.equals(x2).toField());
+
+      let inf = checkFinite
+        ? zero
+        : Provable.witness(Field, () =>
+            x1.equals(x2).and(y1.equals(y2).not()).toField()
+          );
```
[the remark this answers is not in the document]
Trivo25: that's super helpful, didn't know where to find the gate implementation!

src/lib/group.test.ts
```diff
@@ -1,30 +1,46 @@
...
-    await isReady;
-  });
-
-  afterAll(async () => {
-    setTimeout(async () => {
-      await shutdown();
-    }, 0);
-  });
 
   describe('Inside circuit', () => {
```
bkase, line 9: can we use the quickcheck generator Gregor made earlier to check some of the algebraic properties for edge cases?

bkase (commented):
we should probably get a once-over by someone on the crypto team since we're changing the constraints here -- maybe @jspada ?

mitschabaude (approved):
the PR looks good to me now but I agree with Brandon's suggestion to create more test cases using the random generators

and we should cut down the constraints used in `Group.add`

src/lib/group.test.ts
Trivo25: I was thinking about that as well, however, I think there is no easy way to sample valid Group elements other than brute forcing (or is there?) that's why the existing tests mostly work with the generator and `(-1, 2)`

mitschabaude: I used to sample public keys (= group elements) by picking a random private key (= scalar) and scaling the group generator with it.
That's not super efficient but good enough.

However, now we have hashToCurve which should be faster, especially if you skip the hash and only do random field element + mapToCurve

Trivo25: Oh right, I didn't consider that at all :face_with_spiral_eyes:

src/examples/regression_test.json
```diff
@@ -140,24 +140,24 @@
...
-        "rows": 106,
-        "digest": "1296dd40b232fb0bc4cd908d47fd88a9"
+        "rows": 114,
+        "digest": "17d1f373892d2c437dc9ce340e17e3d6"
       },
       "equals": {
-        "rows": 27,
-        "digest": "adbcdc756d41853e24612e50b99be285"
+        "rows": 42,
+        "digest": "8d5ab248aa4602e6f77a8ab1015615df"
```
mitschabaude, line 156: can you explain this blowup in `Group.equals()` constraints @Trivo25?

mitschabaude: is it just caused by the added constraints in `check()`?

Trivo25: The `Group Primitive.equals` check bundles a couple additional equality checks than just a simple `equals()`, it does the following:

```ts
    let g1 = Provable.witness(Group, () => Group.generator);
    let g2 = Provable.witness(Group, () => Group.generator);
    g1.equals(g2).assertTrue();
    g1.equals(g2).assertFalse();
    g1.equals(g2).assertEquals(true);
    g1.equals(g2).assertEquals(false);
```

So its calling `equals` multiple times, which blows up the constraints a bit more. The single `equals` constraints aren't that bad, just inflated by the batch. The additional constraints come from `zero`-related checks (such as `.check`) as well as the changes to the `add` method

mitschabaude: yeah I understand that it has more, but the only thing that changed in that circuit should be `check()` and `equals()`, so I was wondering what explains the increase in circuit size compared to the previous version

Trivo25: yea, its the change in `check`

PR #974, opened 2023-06-07
Remove uses of JS classes in OCaml, part 2

mitschabaude:
bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/36

follow up to #956 

* Removes the old `Field` and `Bool`
* Moves OCaml side of Poseidon into the `Snarky` namespace, so that we only have 4 namespaces to manage - `Snarky`, `Pickles`, `Ledger`, `Test`
* Move some of the `Ledger` static methods into `Test`
* Removed some unused stuff


src/lib/bool.ts
```diff
@@ -311,37 +310,31 @@ class Bool {
     return 1;
   }
 
-  // TODO
-  static count(x: Bool | boolean[]): Field {
```
mitschabaude, line 315: removed unimplemented function

src/lib/bool.ts
```diff
@@ -311,37 +310,31 @@ class Bool {
...
-      if (typeof x === 'number') {
-        return new Bool(x === 1);
-      } else if (typeof x === 'string') {
-        return new Bool(x === '1');
-      } else if (typeof x === 'boolean') {
-        return new Bool(x);
-      } else {
-        return new Bool(x.value);
-      }
+    ofField(x: Field) {
```
mitschabaude, line 326: I refactored this function as I felt the additional options to pass in non-`Field`s were never used and made it less safe, and I added a runtime error

mitschabaude (commented):
some comments!

MartinMinkov (approved):
Looks great! :partying_face:

PR #977, opened 2023-06-07
Make CI run pull requests against `develop`

mrmr1993:


psteckler (approved):

mitschabaude (approved):
Good catch!

mitschabaude:
I think you need this change in the branch where you want CI to run on

mrmr1993:
> I think you need this change in the branch where you want CI to run on

I'm pretty sure it needs to be main, but you can check for yourself that https://github.com/o1-labs/snarkyjs/pull/975 still won't run CI.

PR #983, opened 2023-06-12
Fix SnarkyJS API docs generation

MartinMinkov:
# Description
Adds a `typedoc` configuration file to specify how we generate API reference documentation. Additionally, we add the needed dependencies as dev dependencies in the project. Finally, we add a package.json script to generate docs for us.

Note: We have to manually specify a few files as entry points due to some complexities with how we are exporting certain classes/types. 

:link: Related docs PR: https://github.com/o1-labs/docs2/pull/422

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ o1labs*.tgz
 tests/report/
 tests/test-artifacts/
 profiling.md
+snarkyjs-reference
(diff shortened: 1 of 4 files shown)


Trivo25 (approved):
this is great! Thanks for fixing it so fast :)

ymekuria (approved):
Nice!

PR #985, opened 2023-06-13
Foreign field operations

### LINKED ISSUE #961: Expose FFAdd
Here @querolita can help you if you need to clarify anything about this primitive's interface or behavior.

### LINKED ISSUE #963: Expose FFMul
Here @jspada can help you if you need to clarify anything about this primitive's interface or behavior.

mitschabaude:
~~mina branch: `feature/foreign-field-develop` (currently no changes but depends on an open PR)~~
~~bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/41~~

closes #961 
closes #963 

This PR adds a `ForeignField` class, which is created by the `createForeignField` class factory (taking the field modulus as parameter).

`ForeignField` aims for a DX similar to what developers are used to from `Field`, with a subset of methods that should be powerful enough for most use cases. ~~So far, all methods are "safe" and fully constrain their outputs, which allows to use the class with confidence and no inside knowledge. An `Unsafe` API which allows advanced developers to save some constraints will be added in a separate PR.~~

EDIT: More info about recent changes in https://github.com/o1-labs/o1js/pull/985#issuecomment-1833498392


src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+    // bit packing
+
+    /**
+     * Unpack a field element to its bits, as a `Bool[]` array.
+     *
+     * This method is provable!
+     */
+    toBits(length = sizeInBits) {
+      checkBitLength('ForeignField.toBits()', length, sizeInBits);
+      let [l0, l1, l2] = this.toFields();
```
querolita, line 304: Just for clarity, `l0` is the least significant limb?

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+      let [, ...limbs] = this.value;
+      return limbs.every(FieldVar.isConstant);
+    }
+
+    /**
+     * Convert this field element to a constant.
+     *
+     * See {@link FieldVar} to understand constants vs variables.
+     *
+     * **Warning**: This function is only useful in `Provable.witness()` or `Provable.asProver()` blocks,
```
MartinMinkov, line 129: Nit: Is it possible to add a link to `Provable.witness` or `Provable.asProver` in the doc comments here?

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+    static fromBits(bits: Bool[]) {
+      let length = bits.length;
+      checkBitLength('ForeignField.fromBits()', length, sizeInBits);
+      let limbSize = Number(limbBits);
+      let l0 = Field.fromBits(bits.slice(0 * limbSize, 1 * limbSize));
+      let l1 = Field.fromBits(bits.slice(1 * limbSize, 2 * limbSize));
+      let l2 = Field.fromBits(bits.slice(2 * limbSize, 3 * limbSize));
+      let x = ForeignField.fromFields([l0, l1, l2]);
+      // TODO: can this be made more efficient? since the limbs are already range-checked
+      if (length === sizeInBits) x.assertValidElement();
```
querolita, line 330: In particular, it could be the case that the output is 256 bits for example, but the actual value is greater than the foreign modulus (of secp256k1 for example)

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+// helpers
+
+const limbMax = (1n << limbBits) - 1n;
+
+// the max foreign field modulus is f_max = floor(sqrt(p * 2^t)), where t = 3*limbBits = 264 and p is the native modulus
+// see RFC: https://github.com/o1-labs/proof-systems/blob/1fdb1fd1d112f9d4ee095dbb31f008deeb8150b0/book/src/rfcs/foreign_field_mul.md
+// since p = 2^254 + eps for both Pasta fields with eps small, a fairly tight lower bound is
+// f_max >= sqrt(2^254 * 2^264) = 2^259
+const foreignFieldMaxBits = (BigInt(Fp.sizeInBits - 1) + 3n * limbBits) / 2n;
+const foreignFieldMax = 1n << foreignFieldMaxBits;
```
querolita, line 414: Actually this is not super super accurate.

We computed with sage the real maximum possible field we can support with pallas and vesta which is the one with prime order `926336713898529563388567880069503262826888842373627227613104999999999999999607 `

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+     */
+    static from(x: ForeignField | ForeignFieldVar | bigint | number | string) {
+      if (x instanceof ForeignField) return x;
+      return new ForeignField(x);
+    }
+
+    /**
+     * Checks whether this field element is a constant.
+     *
+     * See {@link FieldVar} tp understand constants vs variables.
```
MartinMinkov, line 117:
```suggestion
     * See {@link FieldVar} to understand constants vs variables.
```

src/lib/foreign-field.ts
MartinMinkov, line 129:
```suggestion
     * **Warning**: This function is only useful in {@link Provable.witness}` or {@link Provable.asProver}` blocks,
```

querolita (commented):
Very well done! I would accept, but let's have @jspada take a look as well.

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+      if (this.isConstant()) return;
+      Snarky.foreignField.assertValidElement(this.value, pMl);
+    }
+
+    // arithmetic with full constraints, for safe use
+
+    /**
+     * Finite field addition
+     *
+     * ```ts
```
MartinMinkov, line 161:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+        return new ForeignField(z);
+      }
+      let z = Snarky.foreignField.add(this.value, toVar(y), pMl);
+      return new ForeignField(z);
+    }
+
+    /**
+     * Finite field negation
+     *
+     * ```ts
```
MartinMinkov, line 177:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+        return new ForeignField(z);
+      }
+      let z = Snarky.foreignField.sub(ForeignFieldVar[0], this.value, pMl);
+      return new ForeignField(z);
+    }
+
+    /**
+     * Finite field subtraction
+     *
+     * ```ts
```
MartinMinkov, line 194:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+        return new ForeignField(z);
+      }
+      let z = Snarky.foreignField.sub(this.value, toVar(y), pMl);
+      return new ForeignField(z);
+    }
+
+    /**
+     * Finite field multiplication
+     *
+     * ```ts
```
MartinMinkov, line 210:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+        return new ForeignField(z);
+      }
+      let z = Snarky.foreignField.mul(this.value, toVar(y), pMl);
+      return new ForeignField(z);
+    }
+
+    /**
+     * Multiplicative inverse in the finite field
+     *
+     * ```ts
```
MartinMinkov, line 226:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+
+      return z;
+    }
+
+    // convenience methods
+
+    /**
+     * Assert equality with a ForeignField-like value
+     *
+     * ```ts
```
MartinMinkov, line 262:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+        return Provable.assertEqual(ForeignField, this, ForeignField.from(y));
+      } catch (err) {
+        throw withMessage(err, message);
+      }
+    }
+
+    /**
+     * Check equality with a ForeignField-like value
+     *
+     * ```ts
```
MartinMinkov, line 284:
```suggestion
     * @example
     * ```ts
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+      if (this.isConstant() && isConstant(y)) {
+        return new Bool(this.toBigInt() === toFp(y));
+      }
+      return Provable.equal(ForeignField, this, ForeignField.from(y));
+    }
+
+    // bit packing
+
+    /**
+     * Unpack a field element to its bits, as a `Bool[]` array.
```
MartinMinkov, line 298:
```suggestion
     * Unpack a field element to its bits, as a {@link Bool}[] array.
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+      return new ForeignField([0, ...limbs]);
+    }
+
+    /**
+     * `Provable<ForeignField>.check`, see {@link Provable.check}
+     *
+     * This will check that the field element is in the range [0, p),
+     * where p is the foreign field modulus.
+     *
+     * Exception: If {@link createForeignField} is called with `{ unsafe: true }`,
```
MartinMinkov, line 379:
```suggestion
     * **Exception**: If {@link createForeignField} is called with `{ unsafe: true }`,
```

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+  let sizeInBits = p.toString(2).length;
+
+  class ForeignField {
+    static modulus = p;
+    value: ForeignFieldVar;
+
+    /**
+     * Create a new {@link ForeignField} from a bigint, number, string or another ForeignField.
+     *
+     * ```ts
```
MartinMinkov, line 88:
```suggestion
     * @example
     * ```ts
```

CHANGELOG.md
```diff
@@ -17,7 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/snarkyjs/compare/3fbd9678e...HEAD)
 
-> no unreleased changes yet
+### Added
+
+- **Foreign field arithmetic** exposed through the `createForeignField()` class factory https://github.com/o1-labs/snarkyjs/pull/985
```
MartinMinkov, line 22: Nit: Do you think it's helpful to have a more descriptive entry in the changelog for such an awesome feature? :D 

Something like:
`
Introduced 'ForeignField', a class for prime order finite fields distinct from the native Field. Supports arithmetic operations, helper methods, and allows optional bypass of automatic validity checks for optimized code. Supports prime moduli up to 259 bits. https://github.com/o1-labs/snarkyjs/pull/985
`

MartinMinkov (approved):
I am approving on my end! Great work as always :D

mitschabaude (commented):
forgot to send these comments 😓

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+ * Create a class representing a prime order finite field, which is different from the native {@link Field}.
+ *
+ * ```ts
+ * const SmallField = createForeignField(17n); // the finite field F_17
+ * ```
+ *
+ * `createForeignField(p)` takes the prime modulus `p` of the finite field as input, as a bigint.
+ * We support prime moduli up to a size of 259 bits.
+ *
+ * The returned {@link ForeignField} class supports arithmetic modulo `p` (addition and multiplication),
```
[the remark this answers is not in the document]
mitschabaude: I think `p` is more typical to use for a prime number than `f`, would prefer to stick with it for clarity

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+ * - `unsafe: boolean` determines whether `ForeignField` elements are constrained to be valid on creation.
+ */
+function createForeignField(modulus: bigint, { unsafe = false } = {}) {
+  const p = modulus;
+  const pMl = ForeignFieldConst.fromBigint(p);
+
+  if (p <= 0) {
+    throw Error(`ForeignField: expected modulus to be positive, got ${p}`);
+  }
+  if (p > foreignFieldMax) {
```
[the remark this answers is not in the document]
mitschabaude: Thanks! I actually read the RFC and wanted to use the exact number, but sadly I couldn't compute it exactly with JS floats, and I also didn't want to hard-code a number when everything else is derived from simple constants.
So I looked at the difference between 2^259 and the exact max modulus and figured that they are so close that probably no one will ever use a prime modulus larger than 259 bits.

In fact, this is how `926336713898529563388567880069503262826888842373627227613104999999999999999607` looks in hex:
`80000000000000000000000000000000224698fc094cf91b9908b7c12f823fe77`
you can see how close it is to 2^259 which is `80000...000`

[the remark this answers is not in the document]
mitschabaude: Also, documenting "259 bits" felt simpler / easier to to use than some random large number

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+    /**
+     * Finite field subtraction
+     *
+     * ```ts
+     * x.sub(1); // x - 1 mod p
+     * ```
+     */
+    sub(y: ForeignField | bigint | number) {
+      if (this.isConstant() && isConstant(y)) {
+        let z = mod(this.toBigInt() - toFp(y), p);
```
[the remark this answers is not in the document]
mitschabaude: yeah, bigint subtraction works and can be negative, and z will always be positive. the modulo operation `%` in JS can return a negative number, that's why we use our own `mod()` function which is written so that the result is always in [0, p): https://github.com/o1-labs/snarkyjs-bindings/blob/6b069e49b9c80b6004262333e83c33047a8293eb/crypto/finite_field.ts#L26-L30

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+      } catch (err) {
+        throw withMessage(err, message);
+      }
+    }
+
+    /**
+     * Check equality with a ForeignField-like value
+     *
+     * ```ts
+     * let isXZero = x.equals(0);
```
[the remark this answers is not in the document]
mitschabaude: yep just an example

src/lib/foreign-field.ts
mitschabaude: yes!

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+  }
+
+  return ForeignField;
+}
+
+// helpers
+
+const limbMax = (1n << limbBits) - 1n;
+
+// the max foreign field modulus is f_max = floor(sqrt(p * 2^t)), where t = 3*limbBits = 264 and p is the native modulus
```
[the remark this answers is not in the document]
mitschabaude: good point!

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,496 @@
...
+     * ```ts
+     * x.add(2); // x + 2 mod p
+     * ```
+     */
+    add(y: ForeignField | bigint | number) {
+      if (this.isConstant() && isConstant(y)) {
+        let z = mod(this.toBigInt() + toFp(y), p);
+        return new ForeignField(z);
+      }
+      let z = Snarky.foreignField.add(this.value, toVar(y), pMl);
```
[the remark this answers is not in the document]
mitschabaude: Thanks for calling that out! (Actually I think `Foreign_field.add` adds only one zero gate)

Yeah, the idea was to land this safe & simple API as a first step, and figure out a more flexible, advanced API in a future PR, after we had the chance to experiment with implementing an optimized gadget ourselves. I think the advanced API will probably live under an `Unsafe` or `Advanced` namespace and expose all possible options.

Providing the `full` flag in the current API didn't feel right to me: the `~full:false` option is not usable by itself without also knowing how to chain additions and without access to the zero gate / the correct bounds check. Apart from that, it makes the full method more error-prone to use. I think `~full:false` is a much lower-level building block that IMO should be a separate method, instead of being mixed with the full `add()` method.

src/lib/foreign-field.ts
mitschabaude:
```suggestion
     * **Warning**: This function is only useful in {@link Provable.witness} or {@link Provable.asProver} blocks,
```

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+import { ZkProgram } from './proof_system.js';
+import { Circuit, circuitMain } from './circuit.js';
+import { Scalar } from './scalar.js';
+
+// toy example - F_17
+
+class SmallField extends createForeignField(17n) {}
+let x = new SmallField(16);
+x.assertEquals(-1); // 16 = -1 (mod 17)
+x.mul(x).assertEquals(1); // 16 * 16 = 15 * 17 + 1 = 1 (mod 17)
```
jspada, line 19: cool

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+// invalid example - modulus too large
+
+expect(() => createForeignField(1n << 260n)).toThrow(
```
jspada, line 23: very good

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+  Random.scalar
+);
+
+// equality
+equivalentBool2(
+  (x, y) => x.equals(y),
+  (x, y) => x === y,
+  Random.scalar
+);
+equivalentVoid2(
```
jspada, line 71: what's the `Void2` mean?

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+function scaleShifted(point: Group, shiftedScalar: Scalar) {
+  let oneHalfGroup = point.scale(oneHalf);
+  let shiftGroup = oneHalfGroup.scale(scalarShift);
+  return oneHalfGroup.scale(shiftedScalar).sub(shiftGroup);
+}
+
+let scalarBigint = Fq.random();
+let pointBigint = G.scale(G.generatorMina, scalarBigint);
+
+// perform a "scalar unshift" in foreign field arithmetic,
```
jspada, line 105: very cool

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+// check provable and non-provable versions are correct
+main0();
+main1();
+Provable.runAndCheck(main0);
+Provable.runAndCheck(main1);
+
+// using foreign field arithmetic should result in much fewer constraints
+let { rows: rows0 } = Provable.constraintSystem(main0);
+let { rows: rows1 } = Provable.constraintSystem(main1);
+expect(rows0 + 100).toBeLessThan(rows1);
```
jspada, line 144: nice test!

src/lib/foreign-field.unit-test.ts
```diff
@@ -0,0 +1,186 @@
...
+
+// TODO: add when proving works
```
jspada, line 146: Is there an issue for doing this already?

CHANGELOG.md
jspada: I think you want to be more restrictive than just checking less than 259 bits, but also check it's less than the max modulus specifically, which is commented as `926336713898529563388567880069503262826888842373627227613104999999999999999607` here: https://github.com/MinaProtocol/mina/blob/cf5664de9cdd8459edc87cc4c9759b3cbe534cf3/src/lib/crypto/kimchi_backend/gadgets/foreign_field.ml#L42

jspada (approved):
nice tests; added some questions and a suggestion

src/lib/foreign-field.unit-test.ts
mitschabaude: the idea was to not land this PR until proving works and we can enable these tests
this is the Mina PR that I'm waiting on: https://github.com/MinaProtocol/mina/pull/12524

src/lib/foreign-field.unit-test.ts
mitschabaude: that we are testing a function which takes two arguments (`2`) and returns nothing (`Void`)

src/lib/foreign-field.ts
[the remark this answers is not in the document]
querolita: > you can see how close it is to 2^259 which is 80000...000

Yeah exactly, that's why I said 259.5 bits, because that number is more or less sqrt of 2^519

I am fine with restricting it further, but it is good to know for us that actually we do have some margin in case we want to use it in another scenario where 259.5 bits are needed.

src/lib/foreign-field.ts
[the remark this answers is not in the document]
querolita: Okay! In any case, if you want to provide an easy method to chain, there's the `sum_chain` function which does precisely that. The user basically has to include all of the inputs to be added/subtracted, and a list of operands (+/-), and then it chains it for you, with some final checks at the end.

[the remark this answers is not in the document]
mitschabaude: yeah I thought about your comments more and decided I want to include a simple to use method which exposes `sum_chain`! will add that today

src/lib/foreign-field.ts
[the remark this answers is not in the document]
mitschabaude: Cool! The max modulus is not that close to the sqrt of 2^519 though. It's much closer to the sqrt of 2^518, which is 2^259, because the pasta moduli are very close to 2^254 (not 2^255)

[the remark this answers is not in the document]
querolita: Well, just for curiosity, the exact number is computed performing `sqrt(2^264*pallas)` which is about `sqrt(2^519)`, that's why I say 259.5 bits. Then that modulus is the maximum prime number that still satisfies that.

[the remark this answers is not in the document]
mitschabaude: `pallas` is about 2^254

mitschabaude:
Update: Motivated by @querolita's comments I added a new method `ForeignField.sum()` which uses `sum_chain` to chain any number of additions and subtractions efficiently. `add()`, `sub()` and `neg()` are now implemented using `sum()`, so we reduce the OCaml API surface while also exposing a more efficient API which is not `Unsafe`

querolita (approved):
Great!

src/lib/testing/equivalent.ts
```diff
@@ -0,0 +1,242 @@
...
+          () => op2(x0, y0),
+          (a, b) =>
+            Provable.asProver(() =>
+              assert(a.toBoolean() === b, 'equal results')
+            )
+        );
+      });
+    });
+  }
+  function equivalentVoid1(
```
bkase, line 147: nit: `unit` not `void`, see 11.3.1 of https://www.cs.cmu.edu/~rwh/pfpl/2nded.pdf . C and Java made a mistake

rahulghangas:
@mitschabaude any update on this? We're currently blocked by it and if possible, would love to help get this through the finish line

mitschabaude:
@rahulghangas the latest on this is that
* we finally have all the dependencies for this in  `o1js/main`
* but, decided to rewrite the foreign field gadgets (and others) in TS before releasing them
the rewrite relevant to this PR will likely start later this week and may take a week at most.

so 2-3 weeks is a realistic ETA until you can use this

mitschabaude:
This PR is officially revived! For the reviewer's convenience, here is the diff from the last review: https://github.com/o1-labs/o1js/compare/d462e7c...4330a98

I made a bunch of changes to the original `ForeignField` class, the most impactful of what are:
* the bulk of the class was moved _outside_ the class factory function `createForeignField()`. The reason for that is that TS is not good at all about understanding the types of dynamically created classes (most methods end up having `any` inputs in the compiled `d.ts` file, even as they are fully typed  in the source code). By moving the class outside we can use the precise types specified by it. The base class is of course not functional (it doesn't know its field modulus yet). So the pattern I use is to "inject" the dynamic parts into static properties that are possibly undefined, and then lift those to always defined properties by wrapping them with getters that throw an error.
* second big change: instead of just having one class, I introduce three different variants of every foreign field class: `Unreduced`, `AlmostReduced` and `Canonical`. that way, I can have fine-grained control over which method expects which kind of range check on its inputs, and what range check is proved about its return value. to understand this detail, I recommend to read the doccomment on `createForeignField()`, and also to read through the new example: `src/examples/crypto/foreign-field.ts`

src/lib/testing/equivalent.ts
mitschabaude: fixed 🫡

Trivo25 (approved):
nice!

src/lib/foreign-field.ts
```diff
@@ -0,0 +1,683 @@
...
+}
+
+/**
+ * Create a class representing a prime order finite field, which is different from the native {@link Field}.
+ *
+ * ```ts
+ * const SmallField = createForeignField(17n); // the finite field F_17
+ * ```
+ *
+ * `createForeignField(p)` takes {@link AlmostForeignField} the prime modulus `p` of the finite field as input, as a bigint.
```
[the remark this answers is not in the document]
mitschabaude: huh, this might have been a misplaced `Ctrl+V`

PR #988, opened 2023-06-14
Add bitwise `AND` to Fields

MartinMinkov:
:link: Bindings PR: https://github.com/o1-labs/snarkyjs-bindings/pull/44


src/lib/field.unit-test.ts
```diff
@@ -55,6 +55,22 @@ test(Random.field, Random.int(-5, 5), (x, k) => {
   deepEqual(Field(x + BigInt(k) * Field.ORDER), Field(x));
 });
 
+// AND with some common and odd lengths
+[2, 4, 8, 16, 32, 64, 3, 5, 10, 15, Fp.sizeInBits].forEach((length) => {
+  test(Random.field, Random.field, (x_, y_, assert) => {
+    let x = x_ % BigInt(length);
+    let y = y_ % BigInt(length);
```
mitschabaude, line 62: take inputs modulo 2^length, see comment on the XOR PR: https://github.com/o1-labs/snarkyjs/pull/989/files#r1241966777

src/lib/field.ts
```diff
@@ -518,6 +518,42 @@ class Field {
...
+   * let a = Field(5);    // ... 000101
+   * let b = Field(3);    // ... 000011
+   *
+   * let c = a.and(b);    // ... 000001
+   * c.assertEquals(1);
+   * ```
+   */
+  and(
+    y: Field | bigint | number | string,
+    length: number = Fp.sizeInBits
```
mitschabaude, line 543: the XOR PR uses a default bit length of 32. I think either 32 or 255 kinda make sense as a default, but the choice should be the same for XOR and AND

src/lib/field.ts
```diff
@@ -518,6 +518,42 @@ class Field {
...
+  ): Field {
+    if (this.isConstant() && isConstant(y)) {
+      let y_ = toFp(y);
+      let thisBigint = this.toBigInt();
+      if (y_ > 2 ** length - 1 || thisBigint > 2 ** length - 1) {
```
mitschabaude, line 548: this should use bigints for comparisons, see comment on the XOR PR: https://github.com/o1-labs/snarkyjs/pull/989/files#r1241942673

mitschabaude (approved):
Great work @MartinMinkov, love the doccomments! I have similar comments as on https://github.com/o1-labs/snarkyjs/pull/989. Also, would be good to add a changelog entry

src/lib/field.ts
MartinMinkov: Sounds good, fixed!

src/lib/field.ts
```diff
@@ -545,8 +545,9 @@ class Field {
     if (this.isConstant() && isConstant(y)) {
       let y_ = toFp(y);
       let thisBigint = this.toBigInt();
-      if (y_ > 2 ** length - 1 || thisBigint > 2 ** length - 1) {
-        throw Error(`${y} and ${thisBigint} need to fit into ${length} bits.`);
+      let max = 1n << BigInt(length);
+      if (y_ >= max || thisBigint >= max) {
+        throw Error(`${y} and ${thisBigint} need to fit into ${max} bits.`);
```
mitschabaude, line 550: actually should be `need to fit into ${max} bits.`!

MartinMinkov: fixed :sweat_smile:

src/lib/field.ts
```diff
@@ -536,6 +536,43 @@ class Field {
     return new Field(z);
   }
 
+  /**
+   * Bitwise AND gate on {@link Field} elements. Equivalent to the [bitwise AND `&` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_AND).
+   * An AND gate works by comparing two bits and returning `1` if both bits are `1`, and `0` otherwise.
+   *
+   * The `length` parameter lets you define how many bits should be compared. It defaults to the size of the field in bits ({@link Fp.sizeInBits}).
```
barriebyron, line 543:
```suggestion
   * The `length` parameter lets you define how many bits to compare. It defaults to the size of the field in bits ({@link Fp.sizeInBits}).
```

barriebyron (commented):
grammatical feedback: let's avoid "should" and avoid ambiguous language, even in comments

thank you!

PR #989, opened 2023-06-14
add XOR and ROT gadget to Field

Trivo25:
bindings https://github.com/o1-labs/snarkyjs-bindings/pull/45

### DIFF
--- a/src/lib/field.ts
+++ b/src/lib/field.ts
@@ -522,7 +522,7 @@ class Field {
    * Bitwise XOR gate on {@link Field} elements. Equivalent to the [bitwise XOR `^` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR).
    * A XOR gate works by comparing two bits and returning `1` if two bits differ, and `0` if two bits are equal.
    *
-   * The `length` parameter lets you define how many bits should be compared. By default it is set to `32`.
+   * The `length` parameter lets you define how many bits should be compared. By default it is set to `32`. The output is not constrained to the length.
    *
    * **Note:** Specifying a larger `length` parameter adds additional constraints.
    *
@@ -539,9 +539,10 @@ class Field {
    */
   xor(y: Field | bigint | number | string, length: number = 32) {
     if (this.isConstant() && isConstant(y)) {
+      let max = 1n << BigInt(length);
       let y_ = toFp(y);
       let thisBigint = this.toBigInt();
-      if (y_ > 2 ** length - 1 || thisBigint > 2 ** length - 1) {
+      if (y_ >= max || thisBigint >= max) {
         throw Error(`${y} and ${thisBigint} need to fit into ${length} bits.`);
       }
 
@@ -553,24 +554,26 @@ class Field {
 
   /**
    * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
-   * `direction` is a boolean, defining the direction of the rotation - `left - true` and `right = false`
+   * `direction` is a string which accepts either `'left'` or `'right'`, defining the direction of the rotation.
    *
-   * **Note:** You can not rotate {@link Field} elements that exceed 64 bits.
+   * **Note:** You can not rotate {@link Field} elements that exceed 64 bits. For elements that exceed 64 bits this operation will fail.
    *
    * ```typescript
    * let a = Field(12);
-   * let b = a.rot(2, true);  // left rotation by 2 bit
+   * let b = a.rot(2, 'left');  // left rotation by 2 bit
    * c.assertEquals(20);
    * ```
    *
    * @param bits amount of bits to rotate this {@link Field} element with.
    * @param direction (true) left or (false) right rotation direction.
    */
-  rot64(bits: number, direction: boolean = true) {
+  rot64(bits: number, direction: 'left' | 'right' = 'left') {
     if (this.isConstant()) {
-      return new Field(Fp.rot64(this.toBigInt(), bits, direction));
+      return new Field(Fp.rot64(this.toBigInt(), bits, direction === 'left'));
     } else {
-      return new Field(Snarky.field.rot64(this.value, bits, direction));
+      return new Field(
+        Snarky.field.rot64(this.value, bits, direction === 'left')
+      );
     }
   }
 
--- a/src/lib/field.ts
+++ b/src/lib/field.ts
@@ -550,6 +550,24 @@ class Field {
   }
 
   /**
+   * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
+   *
+   * *Note:* You can not rotate {@link Field} elements that exceed 64 bits.
+   *
+   * ```typescript
+   * let a = Field(12);
+   * let b = a.rot(2, true);  // left rotation by 2 bit
+   * c.assertEquals(20);
+   * ```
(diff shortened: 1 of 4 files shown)


src/lib/field.ts
```diff
@@ -518,6 +518,62 @@ class Field {
...
+   *
+   * let c = a.xor(b);    // ... 000110
+   * c.assertEquals(6);
+   * ```
+   */
+  xor(y: Field | bigint | number | string, length: number = 32) {
+    if (this.isConstant() && isConstant(y)) {
+      let y_ = toFp(y);
+      let thisBigint = this.toBigInt();
+      if (y_ > 2 ** length - 1 || thisBigint > 2 ** length - 1) {
```
mitschabaude, line 544: it would be better to use bigints for the comparisons here, because large values aren't represented exactly with JS floats
 ```suggestion
      let max = 1n << BigInt(length);
      if (y_ >= max || thisBigint >= max) {
```

File: src/lib/field.ts [hunk 2]
```diff
@@ -518,6 +518,62 @@ class Field {
...
+   * ```typescript
+   * let a = Field(12);
+   * let b = a.rot(2, true);  // left rotation by 2 bit
+   * c.assertEquals(20);
+   * ```
+   *
+   * @param bits amount of bits to rotate this {@link Field} element with.
+   * @param direction (true) left or (false) right rotation direction.
+   */
+  rot64(bits: number, direction: boolean = true) {
```
mitschabaude, line 569: I think calling the parameter `direction` and making it a boolean is a bit confusing. I think it should either:
* be a boolean and be called `isLeft` or `isDirectionLeft` or something
* be called `direction` and be of type `"left" | "right"`

I have a slight preference for the second option since it makes the code using this method more readable:
`x.rot64(2, false)` vs `x.rot64(2, "right")`

src/lib/int.ts
```diff
@@ -202,6 +203,38 @@ class UInt64 extends CircuitValue {
...
+   * ```typescript
+   * let a = UInt64.from(12);
+   * let b = a.rot(2, true);  // left rotation by 2 bit
+   * c.assertEquals(20);
+   * ```
+   *
+   * @param bits amount of bits to rotate this {@link UInt64} element with.
+   * @param direction (true) left or (false) right rotation direction.
+   */
+  rot64(bits: number, direction: boolean = true) {
```
mitschabaude, line 234: nit: I would call this just `rot`, not `rot64`, since the 64 bit length is already implicit in using a `UInt64`

src/lib/int.ts
```diff
@@ -202,6 +203,38 @@ class UInt64 extends CircuitValue {
...
+   * let a = UInt64.from(5);    // ... 000101
+   * let b = UInt64.from(3);    // ... 000011
+   *
+   * let c = a.xor(b);    // ... 000110
+   * c.assertEquals(6);
+   * ```
+   */
+  xor(y: UInt64) {
+    return new UInt64(this.value.xor(y.value, UInt64.NUM_BITS));
+  }
```
mitschabaude, line 219: I think it would be useful to have the same method on `UInt32`! even if it's a bit redundant with the bit length, but the mathod kinda fits better on a bit-constrained type than on general field elements (not saying we should not also have it on Field)

src/lib/field.ts
```diff
@@ -518,6 +518,62 @@ class Field {
...
+  /**
+   * Bitwise XOR gate on {@link Field} elements. Equivalent to the [bitwise XOR `^` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR).
+   * A XOR gate works by comparing two bits and returning `1` if two bits differ, and `0` if two bits are equal.
+   *
+   * The `length` parameter lets you define how many bits should be compared. By default it is set to `32`.
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional constraints.
+   *
+   * **Note:** Both {@link Field} elements need to fit into `2^length - 1`, or the operation will fail.
+   * For example, for `length = 2` ( 2² = 4), `.xor` will fail for any element that is larger than `> 3`.
```
mitschabaude, line 530: Are both inputs actually proven to be only `length` bits? And the output? I'd like to know that from the doccomment

src/lib/field.ts
```diff
@@ -518,6 +518,62 @@ class Field {
...
+    } else {
+      return new Field(Snarky.field.xor(this.value, Field.#toVar(y), length));
+    }
+  }
+
+  /**
+   * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
+   * `direction` is a boolean, defining the direction of the rotation - `left - true` and `right = false`
+   *
+   * **Note:** You can not rotate {@link Field} elements that exceed 64 bits.
```
mitschabaude, line 558: It would be helpful to also explain how the method behaves on inputs > 64 bits

src/lib/field.unit-test.ts
```diff
@@ -34,6 +34,39 @@ test(Random.field, Random.json.field, (x, y, assert) => {
...
 
+// rotation
+test(
+  Random.uint64,
+  Random.nat(64),
+  Random.boolean,
+  (x, n, direction, assert) => {
+    let z = Field(x);
+
+    let r1 = Fp.rot64(BigInt(x), n, direction);
```
mitschabaude, line 45: nit: I think `x` is already a bigint

src/lib/field.unit-test.ts
```diff
@@ -34,6 +34,39 @@ test(Random.field, Random.json.field, (x, y, assert) => {
   assert(z.toJSON() === y);
 });
 
+// rotation
+test(
+  Random.uint64,
+  Random.nat(64),
+  Random.boolean,
```
mitschabaude, line 41: nice usage of the testing framework! :D

src/lib/field.unit-test.ts
```diff
@@ -34,6 +34,39 @@ test(Random.field, Random.json.field, (x, y, assert) => {
...
+      Provable.asProver(() => assert(r1 === r2.toBigInt()));
+    });
+  }
+);
+
+// XOR with some common and odd lengths
+[2, 4, 8, 16, 32, 64, 3, 5, 10, 15].forEach((length) => {
+  test(Random.field, Random.field, (x_, y_, assert) => {
+    let x = x_ % BigInt(length);
+    let y = y_ % BigInt(length);
```
mitschabaude, line 58: I think what you wanted to do here is take the inputs modulo `2**length`:
```suggestion
    let modulus = 1n << BigInt(length);
    let x = x_ % modulus;
    let y = y_ % modulus;
```

mitschabaude: btw, not important here but an equivalent and very efficient way to take a number modulo 2^n is to use a bitwise AND:
```ts
let x = x_ & (modulus - 1n);
```

src/lib/int.ts
```diff
@@ -3,6 +3,7 @@ import { AnyConstructor, CircuitValue, prop } from './circuit_value.js';
 import { Types } from '../bindings/mina-transaction/types.js';
 import { HashInput } from './hash.js';
 import { Provable } from './provable.js';
+import { dir } from 'console';
```
mitschabaude, line 6: whoops, delete

mitschabaude (approved):
Great work @Trivo25, I really like the love that you put into doccomments and testing

mitschabaude:
Would be nice to add a changelog!

Trivo25:
closing in favor of https://github.com/o1-labs/o1js/pull/1259

PR #997, opened 2023-06-20
Blocking SnarkyJS CI tests

mitschabaude:
companion of https://github.com/MinaProtocol/mina/pull/13430

* uses the existing `./run --bundle` script to build & run TS scripts on the fly
* in order to make `./run` handle our entire code base, and don't mix the worker code into the bundle, we stop bundling at `wrapper.js` (which has the compiled wasm/jsoo artifacts as dependencies)
* we make sure that artifacts are actually at the place where they are imported from (`src/bindings/compiled/_node_bindings`). The `make` script now puts compiled artifacts into that folder instead of into `/dist`.
* other build scripts are modified so that they copy additional files into `/dist` after invoking the TS compiler
* our old hack to ensure that TS doesn't type-check the jsoo output (importing a file name that doesn't exist within `src`) is no longer valid. We instead add an empty declaration file `snarky_js_bindings.bc.d.cts` which also stops TS from looking at that file

All in all, the build setup is a bit more logical and less hacky now - in particular, files are located where they are imported from and it's actually possible to bundle the entire TS code base without copying around files before doing that


src/build/utils.js
```diff
@@ -0,0 +1,18 @@
+import fse from 'fs-extra';
+
+export { copyFromTo };
+
+async function copyFromTo(files, srcDir = undefined, targetDir = undefined) {
+  let promises = [];
+  for (let source of files) {
```
bkase, line 7: nit, this can be a one-liner: 

```typescript
return Promise.all(
files.map(source =>  
  fse.copy(source,
         source.replace(`${srcDir}/`, `${targetDir}/`), {
        recursive: true,
        overwrite: true,
        dereference: true,
      })
    )
    )
```

src/tests/test.ts
```diff
@@ -0,0 +1,4 @@
+import { Field } from '../index.js';
```
bkase, line 1: when is this run?

src/tests/inductive-proofs.ts
```diff
@@ -0,0 +1,156 @@
+import {
```
bkase, line 1: when is this run?

bkase (approved):

src/tests/inductive-proofs-small.ts
```diff
@@ -0,0 +1,88 @@
...
+});
+
+tic('compiling program');
+await MaxProofsVerifiedOne.compile();
+toc();
+
+await testRecursion(MaxProofsVerifiedOne, 1);
+
+async function testRecursion(
+  Program: typeof MaxProofsVerifiedOne,
```
mrmr1993, line 43: Nit: should this be explicit? I think we want to have to change things if the backend changes, right?

mrmr1993 (approved):

src/tests/test.ts
mitschabaude: never :D I can delete it!

src/tests/inductive-proofs.ts
mitschabaude: never, but I hope to reconcile some Mina tests which are currently in snarkyjs/tests/integration and are written in raw JS w/o types, with this directory

PR #999, opened 2023-06-21
Add new hashing functions (SHA & Keccak)

MartinMinkov:
# Description
Implements https://github.com/o1-labs/rfcs/pull/9

This PR adds support for hashing using various SHA algorithms and Keccak. The OCaml bindings expose the base functionality, which calls into Kimchi.

To see example usage, see the following tests:
Unit tests: https://github.com/o1-labs/snarkyjs/pull/999/files#diff-772a0ee5ba611c22e73c36051c7139a50869713a53c357d2f6abb49abfbfa497

Integration test: https://github.com/o1-labs/snarkyjs/pull/999/files#diff-1f506bf5891efbbfba747df80700012d212be51e402f9f406e21155992858664

**Note**: Hashing only works for Field variables. No hashing is implemented for constant fields, which can lead to errors if users try to hash outside of a provable method. We can add a lightweight JS dependency that implements SHA/Keccak hashing and convert our constant Field values into something the JS dependency expects. 

Currently waiting on :link: https://github.com/MinaProtocol/mina/pull/12524 to be merged for proving to be enabled.

:link: Bindings https://github.com/o1-labs/snarkyjs-bindings/pull/53


src/lib/hash.ts
```diff
@@ -192,3 +194,40 @@ class TokenSymbol extends Struct(TokenSymbolPure) {
 function emptyReceiptChainHash() {
   return emptyHashWithPrefix('CodaReceiptEmpty');
 }
+
+function buildSHA(length: 224 | 256 | 384 | 512, nist: boolean) {
+  return {
+    hash(message: (Field | UInt8)[]) {
```
Trivo25, line 200: I think we should only allow `UInt8` as arguments to make it clear SHA only works with bytes

src/lib/hash.ts
```diff
@@ -192,3 +194,40 @@ class TokenSymbol extends Struct(TokenSymbolPure) {
...
+      if (message.length === 0) {
```
Trivo25, line 201: technically `H('')` should work https://en.wikipedia.org/wiki/SHA-3#Examples_of_SHA-3_variants

src/lib/hash.ts
```diff
@@ -192,3 +194,40 @@ class TokenSymbol extends Struct(TokenSymbolPure) {
...
+    },
+  };
+}
+
+const Hash = {
+  default: Poseidon.hash,
+
+  Poseidon: Poseidon.hash,
+
+  SHA224: buildSHA(224, true).hash,
```
Trivo25, line 224: nit: we can make `buildSHA` return the function directly, rather than an object containing a single `hash` function - I'll change that in the RFC if we decide to go this route

src/lib/hash.ts
MartinMinkov: great point!

src/examples/zkapps/hashing/hash.ts
```diff
@@ -0,0 +1,92 @@
...
+  State,
+  method,
+  Permissions,
+  Struct,
+} from 'snarkyjs';
+
+let initialCommitment: Field = Field(0);
+
+// 32 UInts
+export class HashInput extends Struct({
```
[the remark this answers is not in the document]
MartinMinkov: Ah, much more convienient :D

mitschabaude:
Update: I'm reviving this branch to augment the keccak API introduced in https://github.com/o1-labs/o1js/pull/1291 and https://github.com/o1-labs/o1js/pull/1307

mitschabaude:
new bindings: https://github.com/o1-labs/o1js-bindings/pull/220

mitschabaude:
Another idea for `Hash.hash` is that it takes any provables as input and hashes them with fewer constraints -- i.e., use the "chunked" method to convert to fields instead ot `toFields`

src/examples/crypto/ecdsa/ecdsa.ts
```diff
@@ -4,37 +4,34 @@ import {
   createEcdsa,
   createForeignCurve,
   Bool,
-  Struct,
-  Provable,
-  Field,
   Keccak,
-  Gadgets,
+  Bytes,
```
Trivo25, line 8: Nice!

src/lib/hashes-combined.ts
```diff
@@ -0,0 +1,41 @@
+import { Poseidon } from './hash.js';
+import { Keccak } from './keccak.js';
+import { Bytes } from './provable-types/provable-types.js';
+
+export { Hash };
+
+// TODO do we want this?
```
Trivo25, line 7: I am not sure anymore, tbh :D we proposed this in the RFC many many months ago, which is now severely outdated. I think it doesn't hurt, at least for now

src/lib/int.test.ts
```diff
@@ -2150,4 +2137,897 @@ describe('int', () => {
       });
     });
   });
+
```
Trivo25, line 2140: one day i will refactor this file to save us 3k of lines :D

src/lib/keccak-old.unit-test.ts
```diff
@@ -0,0 +1,272 @@
+import { test, Random } from './testing/property.js';
```
Trivo25, line 1: should we remove this? whats the benefit of keeping both this and `keccak.unit-test.ts`? Can we combine the two?

src/snarky.d.ts
```diff
@@ -525,6 +525,7 @@ declare const Snarky: {
     };
   };
 
+  // TODO: implement in TS
```
Trivo25, line 528: this would be fun :eyes:

Trivo25 (approved):

src/lib/keccak-old.unit-test.ts
mitschabaude: on it

PR #1007, opened 2023-06-28
Foreign curve and ECDSA

### LINKED ISSUE #966: Expose ECDSA
Functor wrappers must have already been expose by the Crypto team before taking on this task. Here @jspada can help you if you need to clarify anything about this primitive's interface or behavior.

mitschabaude:
stacked behind both #985 and #1240

closes #966, #1156 , #1158

Main changes:
* Adds several missing EC gadgets in `gadgets/elliptic-curve.ts`: `assertOnCurve()`, `assertInSubgroup()`, `negate()`, `scale()`
  * unifies the API of all EC gadgets, to take inputs first and then a `Curve` and possibly other config
* Adds unit tests for EC gadgets
* New export: `ForeignCurve` class and `createForeignCurve()` factory, see `foreign-curve.ts`
* New export: `EcdsaSignature` class and `createEcdsa()` factory, see `foreign-ecdsa.ts`
* _Removes_ the lower-level `Gadgets.Ecdsa` namespace. I concluded that it wasn't very usable without a sibling low-level EC gadgets namespace, and exposing the full EC + ECDSA API in two different forms seemed complete overkill to me. Better to have one way to do things
* Some additions and changes to `ForeignField`

bindings: https://github.com/o1-labs/o1js-bindings/pull/215


src/lib/foreign-curve.ts
```diff
@@ -128,6 +135,9 @@ function createForeignCurve(curve: CurveParams) {
 
     static #curveParamsMlVar: unknown | undefined;
 
+    /**
+     * Initialize usage of the curve. This function has to be called oncel per provable method to use the curve.
```
barriebyron, line 139: once

(typo)

package.json
mitschabaude, line 1: I simplified the scripts as I was finding myself running `npm run build:test` everytime -- now both `npm run dev` and `npm run build` include building test dependencies, and only the production build doesn't

src/examples/crypto/ecdsa/ecdsa.ts
mitschabaude, line 1: this example was moved here from `examples/zkprogram/` and now uses the new class APIs

src/lib/foreign-field.ts
```diff
@@ -497,6 +494,18 @@ class AlmostForeignField extends ForeignFieldWithMul {
...
+
+  /**
+   * Check equality with a constant value.
+   *
+   * @example
+   * ```ts
+   * let isXZero = x.equals(0);
+   * ```
+   */
+  equals(y: bigint | number) {
```
mitschabaude, line 506: I noticed (when I wanted to use it) that the original `ForeignField.equals()` is very error-prone, since it checked for exact equality instead of equality mod p. I replaced it with this constant version (which assumes an almost reduced input) and the full version on `CanonicalForeignField`

src/lib/foreign-field.ts
```diff
@@ -186,7 +191,7 @@ class ForeignField {
    *
    * Returns the field element as a {@link CanonicalForeignField}.
    */
-  assertCanonicalFieldElement() {
+  assertCanonical() {
```
mitschabaude, line 194: old name was too long

src/lib/foreign-field.ts
```diff
@@ -661,7 +696,7 @@ type Constructor<T> = new (...args: any[]) => T;
 
 function provable<F extends ForeignField>(
   Class: Constructor<F> & { check(x: ForeignField): void }
-): ProvablePure<F> {
+): ProvablePureExtended<F, string> {
```
mitschabaude, line 699: making this support `toJSON` etc. makes a nicer inferred `Provable` for `ForeignCurve` and `EcdsaSignature` as well

src/lib/gadgets/ecdsa.unit-test.ts
mitschabaude, line 1: I realized that the ECDSA unit tests weren't actually testing provable code (because the `record` combinator previously didn't forward the `provable` property of its fields). There are a bunch of changes here that fix the test, and in `equivalent.ts` to fix the behavior there

src/lib/gadgets/elliptic-curve.ts
```diff
@@ -258,6 +319,11 @@ function multiScalarMul(
     for (let i = 0; i < n; i++) {
       sum = Curve.add(sum, Curve.scale(P[i], s[i]));
     }
+    if (mode === 'assert-zero') {
+      assert(sum.infinity, 'scalar multiplication: expected zero result');
+      return Point.from(Curve.zero);
+    }
+    assert(!sum.infinity, 'scalar multiplication: expected non-zero result');
```
mitschabaude, line 326: added a different mode for the scalar mul gadget so that we can also support the subgroup check (q*P = 0) with it

src/lib/provable-context.ts
```diff
@@ -105,6 +105,15 @@ function constraintSystem<T>(f: () => T) {
       print() {
         printGates(gates);
       },
+      summary() {
+        let gateTypes: Partial<Record<GateType | 'Total rows', number>> = {};
+        gateTypes['Total rows'] = rows;
```
mitschabaude, line 110: I had this code manually copied in a lot of files, find it useful to expose

src/lib/testing/constraint-system.ts
```diff
@@ -389,18 +389,22 @@ function drawFieldVar(): FieldVar {
...
+      return FieldVar.constant(1n);
     }
     case FieldType.Var: {
       return [FieldType.Var, 0];
     }
     case FieldType.Add: {
       let x = drawFieldVar();
       let y = drawFieldVar();
+      // prevent blow-up of constant size
+      if (x[0] === FieldType.Constant && y[0] === FieldType.Constant) return x;
```
mitschabaude, line 401: changes here prevent rare cases where `constraintSystem` tests would be flaky on a very large constant input (for example, if the circuit asserts that the input is less than 256)

src/lib/testing/equivalent.ts
```diff
@@ -226,10 +237,18 @@ function equivalentProvable<
...
         );
       });
     });
+    if (verbose) {
+      let ms = (performance.now() - start).toFixed(1);
+      let runs = nRuns.toString().padStart(2, ' ');
+      console.log(
+        `${label.padEnd(20, ' ')}    success on ${runs} runs in ${ms}ms.`
+      );
+    }
```
mitschabaude, line 251: new argument `verbose` prints a nice summary of the test run. I used this to debug why the new EC unit tests were taking so long (turns out that `scale()` just takes a pretty long time, like a few seconds, I'm planning to look into how it can be made faster, as a follow-up)

src/lib/testing/equivalent.ts
```diff
@@ -183,19 +186,27 @@ function equivalentAsync<
...
+  return spec.specs.some((spec) => spec.provable);
+}
+
 function equivalentProvable<
   In extends Tuple<OrUnion<any, any>>,
   Out extends ToSpec<any, any>
->({ from: fromRaw, to }: { from: In; to: Out }) {
+>({ from: fromRaw, to, verbose }: { from: In; to: Out; verbose?: boolean }) {
   let fromUnions = fromRaw.map(toUnion);
+  assert(fromUnions.some(isProvable), 'equivalentProvable: no provable input');
```
mitschabaude, line 198: this prevents the accident with ecdsa, where we weren't testing it in provable code, in the future

tests/vk-regression/vk-regression.json
```diff
@@ -203,16 +203,16 @@
...
   },
   "ecdsa": {
-    "digest": "6e4078c6df944db119dc1656eb68e6c272c3f2e9cab6746913759537cbfcfa9",
+    "digest": "36c90ec17088e536562b06f333be6fc1bed252b47a3cfb4e4c24aa22e7a1247b",
     "methods": {
       "verifyEcdsa": {
-        "rows": 38846,
-        "digest": "892b0a1fad0f13d92ba6099cd54e6780"
+        "rows": 38888,
+        "digest": "70f148db469f028f1d8cb99e8e8e278a"
```
mitschabaude, line 210: the example circuit changed: it previously did no validity checks on the input public key, now it does (automatically, when the input is witnessed)

src/lib/gadgets/elliptic-curve.ts
```diff
@@ -405,6 +477,22 @@ function initialAggregator(Curve: CurveAffine) {
   // use that as x coordinate
   const F = Curve.Field;
   let x = F.mod(bytesToBigInt(bytes));
+  return simpleMapToCurve(x, Curve);
+}
+
+function random(Curve: CurveAffine) {
+  let x = Curve.Field.random();
+  return simpleMapToCurve(x, Curve);
```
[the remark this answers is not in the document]
mitschabaude: no, map to curve is faster. EC scale is a fairly heavy operation - much heavier than square root

nicc:
Reminder to remove "coming soon" from [the docs](https://github.com/o1-labs/docs2/pull/774/files#diff-00d1313e51b0fa79b29fd9fbbc7dd064ef5a6c8862c4e0caf9b9db1f76e33cceR23) when this is merged

PR #1009, opened 2023-06-28
Update CONTRIBUTING.md for new RFC

barriebyron:
I was looking at the SnarkyJS contributing file as a model, and noticed that it didn't include the new RFC workflow. 

This PR adds information about the new RFC process and links to examples in that repo.

Question: Do we want to include our RFC in the contributing file? or shall we be clear about using a GitHub issue for community contributions?


barriebyron:
closing, per decision to have community start with issues first, thanks @bkase for guidance!

PR #1013, opened 2023-07-04
Improve construction of Merkle tree

chiro-hiro:
You might notice that, `.push()` isn't free at all since it need to resize the capacity of the array. If we compared to fixed size of `Array`, it's 50% slower at height `256`.

### DIFF
--- a/src/lib/merkle_tree.ts
+++ b/src/lib/merkle_tree.ts
@@ -36,9 +36,10 @@ class MerkleTree {
    * @returns A new MerkleTree
    */
   constructor(public readonly height: number) {
-    this.zeroes = [Field(0)];
-    for (let i = 1; i < height; i++) {
-      this.zeroes.push(Poseidon.hash([this.zeroes[i - 1], this.zeroes[i - 1]]));
+    this.zeroes = new Array(height);
+    this.zeroes[0] = Field(0);
+    for (let i = 1; i < height; i+=1) {
+      this.zeroes[i] = Poseidon.hash([this.zeroes[i - 1], this.zeroes[i - 1]]);
     }
   }
 


mitschabaude:
Hey @chiro-hiro I really appreciate you looking into performance! But this particular change will make no difference at all -- the `Poseidon.hash()` is many orders of magnitude more costly than `Array.push()`.

For improving perf I recommend to use the `./run-in-browser.js` script in snarkyjs, and then going into a browser and investigating performance using the Chrome dev tools "Performance" tab. This will give you a detailed into where the bottlenecks are, and might give you ideas for changes that make a real difference

chiro-hiro:
> Hey @chiro-hiro I really appreciate you looking into performance! But this particular change will make no difference at all -- the `Poseidon.hash()` is many orders of magnitude more costly than `Array.push()`.
> 
> For improving perf I recommend to use the `./run-in-browser.js` script in snarkyjs, and then opening localhost in Chrome and investigating performance using the dev tools "Performance" tab. This will give you a detailed view into where the bottlenecks are, and might give you ideas for changes that make a real difference

Hi @mitschabaude, 

I just wonder, why don't we precompute the digests to minimize the cost of performing Poseidon hash?.

```ts
const precomputed = [
  '0',
  '21565680844461314807147611702860246336805372493508489110556896454939225549736',
  '2447983280988565496525732146838829227220882878955914181821218085513143393976',
  '544619463418997333856881110951498501703454628897449993518845662251180546746',
  '20468198949394563802460512965219839480612000520504690501918527632215047268421',
  '16556836945641263257329399459944072214107361158323688202689648863681494824075',
  '15433636137932294330522564897643259724602670702144398296133714241278885195605',
  '14472842460125086645444909368571209079194991627904749620726822601198914470820',
  '21614416876217972474084851109688329000791937035724439964738173004620435920527',
  '23396673455667782815008357063662227432928854130481827049177088579579506912772',
  '16799216270319797546551726730220821530700130944535729528697378284641302758053',
  '13496994890596928254174830027007320908142597452643688487140586680795895052589',
  '3136367688481366987314253891173247447839122679172869317967104414474412425595',
  '16414894720763442886261603851925762864778244212151669304308726942427436045416',
  // May be height = 256 would be enough for now
 ];
 
 class MerkleConstructionTest {
  private zeroes: Field[];
  constructor(public readonly height: number) {
    this.zeroes = precomputed.slice(0, height).map((x) => Field(x));
  }
}
```

It's a trade-off between storage and performance, I think it would make a big different for browser.

mitschabaude:
> It's a trade-off between storage and performance, I think it would make a big different for browser.

How long does the `MerkeTree` constructor take? I'm not enthusiastic about adding 256 field elements = about 77 * 256 byte = 19kB to the JS bundle size just to speed up a single library function which is called once.  

I'm pretty sure we can speed up Poseidon by 3-5x if we want

mitschabaude (approved):

PR #1014, opened 2023-07-05
release mina-signer

Trivo25:
this bumps the mina-signer version and release the newly added `createNullifier` method (which is now also released in SnarkyJS directly, but wasn't accessible because devs couldn't create the nullifier


mitschabaude (approved):

barriebyron:
thank you @Trivo25  I recently learned about the Mina Signer. 
Does this update impact SnarkyJS? Should we track these updates for the What's New? (or put into changelog? or?)
While we are talking about impact, do we need more docs for Mina Signer?

PR #1021, opened 2023-07-10
Fix SnarkyJS to be compatible with new Berkeley

MartinMinkov:
# Description
Merges `berkeley` into `main`

## Notable Changes

- Because the Out-Of-Snark dummy verification key hash changed in [this PR](https://github.com/MinaProtocol/mina/pull/13394), the old verification key hash defaults we used (`Field(0)`) were incorrect. There are several places where this value is hard-coded, so we change each place to use the new dummy value. All changes are contained in this commit: https://github.com/o1-labs/snarkyjs/pull/1021/commits/9b4805dc28b1892187dd9c4c442cbb7a1aaf2571

- Update the vk regression tests by creating a new dump.

:link: Accompanying Bindings PR https://github.com/o1-labs/snarkyjs-bindings/pull/80


ymekuria (approved):
LGTM!

src/mina-signer/src/sign-zkapp-command.ts
```diff
@@ -17,6 +17,7 @@ import {
   signFieldElement,
   verifyFieldElement,
 } from './signature.js';
+import { Pickles } from '../../snarky.js';
```
mitschabaude, line 20: this makes the JSOO output a dependency of mina-signer, which we don't want (it's supposed to be small)

To fix this, I recommend to autogenerate the verification key hash as a constant in `snarky_js_constants.ml`

src/lib/account_update.ts
```diff
@@ -1288,6 +1289,12 @@ class AccountUpdate implements Types.AccountUpdate {
       other
     );
   }
+  static emptyValue() {
+    let empty = Types.AccountUpdate.emptyValue();
+    empty.body.authorizationKind.verificationKeyHash =
+      dummyVerificationKeyHash();
+    return empty;
+  }
```
mitschabaude, line 1297: oof, this makes `AccountUpdate` differ from the auto-generated one. It seems fine as a quick fix, but medium term I'd want to add an annotation in OCaml which allows us to specify a custom `emptyValue()` at the lower level where it also ends up in the autogenerated type. We already do the same for `ActionState`

mitschabaude (changes requested):
awesome fix 🙌🏻 

Requesting changes because mina-signer should be kept free from OCaml dependencies!

src/lib/account_update.ts
MartinMinkov: Gotcha! I was trying to figure out how to override `emptyValue` but couldn't see how it was done. I'll take a look for `ActionState` :+1:

src/mina-signer/src/sign-zkapp-command.ts
MartinMinkov: Gotcha, will do!

src/examples/regression_test.json
```diff
@@ -1,14 +1,14 @@
 {
   "Voting_": {
-    "digest": "be2fb91778c112dba345230d33f53b60d6aebc9d21c34af42d2d06169c03521",
+    "digest": "1ea9c9b3a7a4a496b7ff475c22703afc1d76e1d43a083f158a8e0fcdf5c38299",
     "methods": {
       "voterRegistration": {
-        "rows": 1259,
-        "digest": "252726e585fdd28607537e32214069ab"
+        "rows": 1260,
```
Trivo25, line 6: that's interesting, I wonder where that extra constraint comes from 🤔

mitschabaude: One more constant that wasn't used before takes one half generic gate!

mitschabaude:
Two more small change requests:
- Make this a minor version bump -- we want to release a version that works on berkeley asap
- Call out in the changelog that all existing zkApp verification keys change

tsconfig.mina-signer.json
```diff
@@ -10,6 +10,11 @@
     "moduleResolution": "nodenext", // comply with node + "type": "module"
     "esModuleInterop": true, // to silence jest
 
+    "experimentalDecorators": true, // needed for decorators
```
MartinMinkov, line 13: I kept getting errors related too:

```
../lib/signature.ts:28:9 - error TS1219: Experimental support for decorators is a feature that is subject to change in a future release. Set the 'experimentalDecorators' option in your 'tsconfig' or 'jsconfig' to remove this warning.

28   @prop s: Scalar;
           ~

../lib/signature.ts:28:9 - error TS2564: Property 's' has no initializer and is not definitely assigned in the constructor.

28   @prop s: Scalar;
```
Because of these errors, I had to add these rules to the `tsconfig` file for `mina-signer`. Is this incorrect?

mitschabaude: yeah I think the problem is that you're now importing `proof_system.ts` (and with it, most of snarkyjs) in `transaction-leaves-bigint.ts` which mina-signer depends on. better to use the verification key hash that's exposed from constants everywhere, not the one from JSOO

MartinMinkov: yep, that was it! I have it changed locally and building works without these changes :+1:

PR #1026, opened 2023-07-11
Nullifier improvements 

Trivo25:
- I've got some feedback that a `createNullifier` method on the Nullifier class directly in SnarkyJS would be useful for testing and demo'ing. I added a method named `createTestNullifier` which creates a nullifier, additionally, preventing developers from calling that method within provable code. 
- adds a fix to prevent non-unique Nullifiers https://github.com/o1-labs/snarkyjs/issues/1022


src/lib/field.ts
```diff
@@ -376,6 +376,63 @@ class Field {
     return this.add(Field.from(y).neg());
   }
 
+  #parity() {
+    if (this.isConstant()) return new Field(this.toBigInt() % 2n);
+
+    let [, isOddVar, xDiv2Var] = Snarky.exists(2, () => {
+      let bits = Fp.toBits(this.toBigInt());
+      let isOdd = bits.shift()! ? 1n : 0n;
```
MartinMinkov, line 384: Why do you need the `!` on `bits.shift()`, that method doesn't return undefined, as far as I understand

src/lib/field.ts
```diff
@@ -376,6 +376,63 @@ class Field {
...
+      ];
+    });
+
+    let isOdd = new Field(isOddVar);
+    let xDiv2 = new Field(xDiv2Var);
+
+    // range check for 253 bits
+    // WARNING: this makes use of a special property of the Pasta curves,
+    // namely that a random field element is < 2^254 with overwhelming probability
+    // TODO use 88-bit RCs to make this more efficient
```
MartinMinkov, line 399: Can you file an issue for this TODO and reference it here? :D We should always aim to track these sorts of things!

src/lib/nullifier.ts
```diff
@@ -165,4 +170,63 @@ class Nullifier extends Struct({
   getPublicKey() {
     return PublicKey.fromGroup(this.publicKey);
   }
+
+  /**
+   *
+   * _Note_: This is *not* the recommended way to create a Nullifier in production. Please use mina-signer to create Nullifiers. Also, this function cannot be run within provable code.
```
MartinMinkov, line 176: Could you explain briefly why you can't run this in provable code, if it's possible? What are the limitations so devs can know why they shouldn't do this.

MartinMinkov (approved):
Had some nits, pre approving! Great work!! :D

src/lib/field.ts
Trivo25: it returns `undefined` if the array is empty, but since we can be sure `bits` will never be empty, we can just "ignore" it like that

mitschabaude (commented):
Very nice! Regarding the API, I think that `parity()` would be a universally useful function, so I would make it public. Since `assertEven()` and `assertOdd()` are only trivial wrappers of `parity()`, I'd actually remove them and use `x0.parity().asserEquals(0)` in the nullifier code.

If 'parity' is not clear enough, we could also call it `isEven()` and make it return `1 - parity` as a `Bool`

Trivo25:
> I think that parity() would be a universally useful function, so I would make it public

Right! I prefer `isEven` over `parity`, as this might sound more familiar to developers. Wdyt @MartinMinkov ?

MartinMinkov:
> > I think that parity() would be a universally useful function, so I would make it public
> 
> Right! I prefer `isEven` over `parity`, as this might sound more familiar to developers. Wdyt @MartinMinkov ?

I'm in favour of keeping the API as simple to use as possible, so `isEven` makes sense to me!

PR #1029, opened 2023-07-12
README-dev.md: add tips when updating the bindings

dannywillems:


README-dev.md
```diff
@@ -23,6 +23,9 @@ To see the test running in a web browser, go to `http://localhost:8000/`.
 
 ## Run tests
 
+If you are updating the bindings, you might need to clean first your
+environment using `npm run clean`.
```
mitschabaude, line 27:
```suggestion
If you are have recently changed branches, you might need to clean first your
environment using `npm run clean`.
```
how about this? it's more general and doesn't have to do with updating the bindings. Actually, `npm run bindings` does run `npm run clean`

dannywillems: Your suggestion makes more sense as it is what certainly happened to me.

dannywillems:
I think closing https://github.com/o1-labs/snarkyjs-bindings/pull/83 is ok to do. I thought duplicating the info would be useful as the dev might have a look at the different readmes, but if they want to run the tests, it is the readme from the repo they will have a look at.

PR #1032, opened 2023-07-14
Fix: Revert vk hash to make snarkyjs compatible

mitschabaude:
mina: https://github.com/MinaProtocol/mina/pull/13602
bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/86

* Sets the hard-coded dummy vk hash to 0 to be compatible with berkeley again
* Fixes the recently introduced `Field.isEven()` (it was returning `isOdd`)
* Release patch version

I manually tested the `zkapps/hello_world/run_berkeley.ts` script to work with these changes


src/lib/field.ts
```diff
@@ -416,7 +416,7 @@ class Field {
     // check composition
     xDiv2.mul(2).add(isOdd).assertEquals(this);
 
-    return new Bool(isOddVar);
+    return new Bool(isOddVar).not();
```
Trivo25, line 419: damn :D

Trivo25 (approved):

PR #1034, opened 2023-07-14
Merge main back to berkeley

mrmr1993:
This PR merges main to berkeley, and reverts the reversion of the vk fix.

Bindings PR: https://github.com/o1-labs/snarkyjs-bindings/pull/88


mitschabaude (changes requested):
I'm not a fan of reverting the entire snarkyjs PR here. It contains important changes that are easy to get lost this way, and I don't think any of those actually break the tests -- I think that just reverting the bindings part should be enough to make the tests work on berkeley.

Let's try by undoing the revert commit and just bumping the submodule to the reverted state, and if the tests don't work with that, add any necessary incremental changes 

Reverting the bindings PR is fine because it's a two line change

mitschabaude (approved):

PR #1045, opened 2023-07-23
[berkeley] Expose and use overrideWrapDomain

mrmr1993:
This PR is the counterpart to https://github.com/MinaProtocol/mina/pull/13546, using the binding changes in https://github.com/o1-labs/snarkyjs-bindings/pull/91.


mitschabaude:
I wonder if we can tweak this so that `overrideWrapDomain` isn't exposed to users but figured out by snarkyjs

mitschabaude (approved):

PR #1048, opened 2023-07-24
[berkeley] Pickles encoding and verification hardening

mrmr1993:
This is the counterpart of https://github.com/MinaProtocol/mina/pull/13601 via https://github.com/o1-labs/snarkyjs-bindings/pull/94.


mitschabaude:
I don't think we need this PR (and bindings counter-part) since it literally only rebuilds the artifacts

mitschabaude:
on the other hand, this PR seems like a welcome opportunity to update vks once more and ensure `snarkyjs/berkeley` is compatible with the current berkeley

mitschabaude (approved):

PR #1073, opened 2023-08-16
Reduce loading time

### LINKED ISSUE #1065: Remove bottlenecks in snarkyjs loading time

mitschabaude:
PR to reduce the startup time of SnarkyJS
Closes #1065

Bindings: https://github.com/o1-labs/snarkyjs-bindings/pull/116
Mina: https://github.com/MinaProtocol/mina/pull/13925
Snarky: https://github.com/o1-labs/snarky/pull/829

Benchmark (node):
```sh
./run src/examples/benchmarks/import.ts
```

Time to import at various steps:
* e09772d: 1763ms - initial
* https://github.com/MinaProtocol/mina/pull/13878/commits/8ee87dc5d75e475a1c34352d0cb4cc318be3988c: 1183ms - remove expensive poseidon param conversion
* https://github.com/MinaProtocol/mina/pull/13878/commits/d232a0d984860fdae7950cd1bb0732d04124a933: 1061ms - compute vk hash lazily
* https://github.com/MinaProtocol/mina/pull/13878/commits/78de73ece274713f25c35138c88e58ce8a4b6006: 950ms - compute dummy tx lazily
* https://github.com/MinaProtocol/mina/pull/13878/commits/7d9fa28971da68bf0162692e3bf82ffd9a5ad809: 877ms - compute Pickles dummy data lazily
* https://github.com/o1-labs/snarkyjs-bindings/pull/116/commits/7e2696bb2ba619e2e0a3495b6ee5c1b65f0617e4: delay getting number of workers; only affects web (reduction by >100ms)
* afee400: 749ms - precompute and bundle hash prefix states instead of computing them at the top level
* latest: 585ms - avoid a bunch of other top level computation / avoid going through zarith when converting fields

Web flame graph of the first ~2 seconds, before:
image

Same flame graph after:
image

Note: There's basically no way to reduce this further within the current architecture. After this PR, most of the loading time of ~0.5sec comes from compiling the JS bundle, which is large. Of the time spent executing that bundle, most happens in core JSOO / OCaml libs. There's only about 50ms of top-level computation caused by Mina code. Some of which is just instantiating modules that we need. There's almost nothing left that we can get rid of.


src/snarky.js
```diff
@@ -1,3 +1,4 @@
+import './bindings/crypto/bindings.js';
```
mitschabaude, line 1: so, here's how this is supposed to work:

the new `bindings.ts` file imported here is written in TS, and contains code integrated with our normal codebase. It is exposed to JSOO by populating a global variable with an object. It gets imported as the first thing here so that the global variable is ready by the time JSOO code gets executed.

It may not be the cleanest thing to have an implicit dependency like that, but it's a huge improvement over the status quo, where all JS code used in JSOO has to be part of an ugly file containing outdated syntax which can use neither TS nor normal imports.

mitschabaude (commented):
Note for reviewers: all the non-trivial changes are in the linked PRs (Mina, bindings, snarky)

MartinMinkov (approved):

ymekuria (approved):

Trivo25 (approved):

L-as:
Can't you do the parsing lazily? You could use a dynamic import instead. If you use the separate jsoo compilation, you can maybe even import each separately lazily.

mitschabaude:
> Can't you do the parsing lazily? You could use a dynamic import instead. If you use the separate jsoo compilation, you can maybe even import each separately lazily.

we can (we do) but then you can't use the lib directly after import. we'd get back `await isReady`

PR #1117, opened 2023-09-12
Bigint backend contd: Move conversion to TS

mitschabaude:
Auxiliary PR to run CI for https://github.com/o1-labs/o1js-bindings/pull/139 and its successors

Every commit bumps the submodule by one bindings PR, so we're checking that each bindings PR passes CI


mitschabaude:
this was just for running CI, closing

Issue #1074, opened 2023-08-16
Live tests improvements (dynamic accounts)


shimkiv:
- https://github.com/o1-labs/snarkyjs/pull/1057#discussion_r1280178805
- After implementation of https://github.com/o1-labs/quality-and-testing/issues/121

PR #1167, opened 2023-10-08
An API for interaction with the lightnet accounts manager.

shimkiv:
This PR attempts to provide the simple API integrated into `o1js` in order to retrieve and release accounts managed by [lightnet Mina network](https://hub.docker.com/r/o1labs/mina-local-network).
`Note`: Keep in mind that `accounts manager` synchronises access to the endpoints in order to guarantee the proper accounts management for concurrent requests.

As an example the corresponding live test was updated as well.

1. Run the local Mina:

  ```shell
  docker run -id --env NETWORK_TYPE="single-node" --env PROOF_LEVEL="none" -p 3085:3085 -p 5432:5432 -p 8080:8080 -p 8181:8181 o1labs/mina-local-network:rampup-latest-lightnet
  ```

2. Build `o1js`:

  ```shell
  npm run build && npm run build:node
  ```

3. Run live tests:

  ```shell
  USE_CUSTOM_LOCAL_NETWORK=true TEST_TYPE="Live integration tests" ./run-ci-tests.sh
  ```

In result test will pass and in Docker container you should see the output like:

```shell
[2023-10-08 12:54:13] [INFO   ] Checking the 'http://localhost:8080/graphql' availability...
[2023-10-08 12:54:13] [INFO   ] An attempt to acquire non-zkApp account...
[2023-10-08 12:54:13] [INFO   ] Acquired account with Index #636 and Public Key: B62qrYhoJRyTtQ7RdU3KutCBQHs69BgYnMBjSY3E72QTNH6CkneM84B
[2023-10-08 12:56:10] [INFO   ] Account with public key B62qrYhoJRyTtQ7RdU3KutCBQHs69BgYnMBjSY3E72QTNH6CkneM84B is set to be released.
```

---

We have linter dependency but we don't provide scripts nor automatic linting on commit, hence some imports/exports were reformatted.
Am I missing something or we need to enable linter properly? 

---

Closes https://github.com/o1-labs/quality-and-testing/issues/121
Closes https://github.com/o1-labs/o1js/issues/1074

### DIFF
--- a/src/examples/zkapps/hello_world/run_live.ts
+++ b/src/examples/zkapps/hello_world/run_live.ts
@@ -1,28 +1,35 @@
 // Live integration test against real Mina network.
-import { AccountUpdate, Field, Mina, PrivateKey, fetchAccount } from 'o1js';
+import {
+  AccountUpdate,
+  Field,
+  Mina,
+  PrivateKey,
+  acquireKeyPair,
+  fetchAccount,
+  releaseKeyPair,
+} from 'o1js';
 import { HelloWorld, adminPrivateKey } from './hello_world.js';
 
-const useLocalNetwork = process.env.USE_LOCAL_NETWORK === 'true';
-const senderKey = useLocalNetwork
-  ? PrivateKey.fromBase58(
-      'EKEnVLUhYHDJvgmgQu5SzaV8MWKNfhAXYSkLBRk5KEfudWZRbs4P'
-    )
-  : PrivateKey.random();
-const sender = senderKey.toPublicKey();
+const useCustomLocalNetwork = process.env.USE_CUSTOM_LOCAL_NETWORK === 'true';
 const zkAppKey = PrivateKey.random();
 const zkAppAddress = zkAppKey.toPublicKey();
 const transactionFee = 100_000_000;
 
 // Network configuration
-const network = Mina.Network(
-  useLocalNetwork
+const network = Mina.Network({
+  mina: useCustomLocalNetwork
     ? 'http://localhost:8080/graphql'
-    : 'https://proxy.berkeley.minaexplorer.com/graphql'
-);
+    : 'https://proxy.berkeley.minaexplorer.com/graphql',
+  accountsManager: 'http://localhost:8181',
+});
 Mina.setActiveInstance(network);
 
 // Fee payer setup
-if (!useLocalNetwork) {
+const senderKey = useCustomLocalNetwork
+  ? (await acquireKeyPair(true)).privateKey
+  : PrivateKey.random();
+const sender = senderKey.toPublicKey();
+if (!useCustomLocalNetwork) {
   console.log(`Funding the fee payer account.`);
   await Mina.faucet(sender);
 }
@@ -91,3 +98,6 @@ try {
   );
 }
 console.log('Success!');
+
+// Tear down
+await releaseKeyPair(senderKey.toPublicKey().toBase58());
--- a/src/examples/zkapps/hello_world/run_live.ts
+++ b/src/examples/zkapps/hello_world/run_live.ts
@@ -20,13 +20,13 @@ const network = Mina.Network({
   mina: useCustomLocalNetwork
     ? 'http://localhost:8080/graphql'
     : 'https://proxy.berkeley.minaexplorer.com/graphql',
-  accountsManager: 'http://localhost:8181',
+  lightnetAccountManager: 'http://localhost:8181',
 });
 Mina.setActiveInstance(network);
 
 // Fee payer setup
 const senderKey = useCustomLocalNetwork
-  ? (await acquireKeyPair(true)).privateKey
+  ? (await acquireKeyPair()).privateKey
   : PrivateKey.random();
 const sender = senderKey.toPublicKey();
 if (!useCustomLocalNetwork) {
@@ -100,4 +100,7 @@ try {
 console.log('Success!');
 
 // Tear down
-await releaseKeyPair(senderKey.toPublicKey().toBase58());
+const keyPairReleaseMessage = await releaseKeyPair({
+  publicKey: sender.toBase58(),
+});
+if (keyPairReleaseMessage) console.info(keyPairReleaseMessage);
--- a/src/examples/zkapps/hello_world/run_live.ts
+++ b/src/examples/zkapps/hello_world/run_live.ts
@@ -2,11 +2,10 @@
 import {
   AccountUpdate,
   Field,
+  Lightnet,
   Mina,
   PrivateKey,
-  acquireKeyPair,
   fetchAccount,
-  releaseKeyPair,
 } from 'o1js';
 import { HelloWorld, adminPrivateKey } from './hello_world.js';
 
@@ -26,7 +25,7 @@ Mina.setActiveInstance(network);
 
 // Fee payer setup
 const senderKey = useCustomLocalNetwork
-  ? (await acquireKeyPair()).privateKey
+  ? (await Lightnet.acquireKeyPair()).privateKey
   : PrivateKey.random();
 const sender = senderKey.toPublicKey();
 if (!useCustomLocalNetwork) {
@@ -100,7 +99,7 @@ try {
 console.log('Success!');
 
 // Tear down
-const keyPairReleaseMessage = await releaseKeyPair({
+const keyPairReleaseMessage = await Lightnet.releaseKeyPair({
   publicKey: sender.toBase58(),
 });
 if (keyPairReleaseMessage) console.info(keyPairReleaseMessage);
--- a/src/index.ts
+++ b/src/index.ts
@@ -65,7 +65,6 @@ export {
   setGraphqlEndpoints,
   setArchiveGraphqlEndpoint,
   sendZkapp,
-  KeyPair,
   acquireKeyPair,
   releaseKeyPair
 } from './lib/fetch.js';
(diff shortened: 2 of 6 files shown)


mitschabaude:
Thanks @shimkiv for the PR!

> We have linter dependency but we don't provide scripts nor automatic linting on commit, hence some imports/exports were reformatted.
Am I missing something or we need to enable linter properly?

We have a prettier config, and prettier handles auto-formatting of yaml files. However we don't apply this in CI, we just rely on developers to have `prettier` installed locally.

However, it seems that you have a different auto-formatter active which
- overrides the `prettier` style of yaml formatting
- adds additional sorting of imports which isn't done by `prettier`

In the interest of providing a clean and easy-to-review PR, and not make changes that would be reverted when the next developer touches the yaml file, I'd prefer if you revert all of the auto-formatting done here (both the yaml file and the import reordering; you can submit a separate PR that reorders imports, but I think it's better to not have auto-formatting enabled that does more than what is enforced by the repo configs and done by other developers. Otherwise everything you touch will create a huge diff.)

shimkiv:
> I'd prefer if you revert all of the auto-formatting done here

Got it, thank you, then will do revert in next commit.

src/index.ts
```diff
@@ -65,6 +65,9 @@ export {
   setGraphqlEndpoints,
   setArchiveGraphqlEndpoint,
   sendZkapp,
+  KeyPair,
```
Trivo25, line 68: What do you think about providing the new functions under a namespace, similarly to `Mina` for everything Mina/tx related? Something like `Lightnet.KeyPair`, `Lightnet.acquireKeyPair`, ...

src/lib/fetch.ts
```diff
@@ -38,25 +38,35 @@ export {
...
 type NetworkConfig = {
   minaEndpoint: string;
   minaFallbackEndpoints: string[];
   archiveEndpoint: string;
   archiveFallbackEndpoints: string[];
+  accountsManagerEndpoint: string;
+};
+type KeyPair = {
+  publicKey: PublicKey;
+  privateKey: PrivateKey;
```
mitschabaude, line 61: can you remove the `KeyPair` type and just use an inline type for the function which returns a keypair? I think we shouldn't export more things than necessary.

src/index.ts
mitschabaude: +1, I don't want more top-level exports (we already have way too many)

src/lib/fetch.ts
```diff
@@ -38,25 +38,35 @@ export {
...
+  acquireKeyPair,
+  releaseKeyPair
 };
 
 type NetworkConfig = {
   minaEndpoint: string;
   minaFallbackEndpoints: string[];
   archiveEndpoint: string;
   archiveFallbackEndpoints: string[];
+  accountsManagerEndpoint: string;
```
mitschabaude, line 57: I'd advocate for making this name even more explicit to avoid confusing people. Like `lightnetAccountManagerEndpoint`

src/lib/fetch.ts
```diff
@@ -993,6 +1017,76 @@ async function fetchActions(
...
+ * If an error is returned by the specified endpoint, an error is thrown. Otherwise,
+ * the data is returned.
+ *
+ * @param isRegularAccount Whether to acquire regular or zkApp account (one with already configured verification key)
+ * @param accountsManagerEndpoint Accounts manager endpoint to fetch from
+ * @returns Key pair
+ */
+async function acquireKeyPair(
+  isRegularAccount = true,
+  accountsManagerEndpoint = networkConfig.accountsManagerEndpoint
```
mitschabaude, line 1033: to avoid having a unnamed boolean in the API, I think you should make this function take an object, `{ isRegularAccount, accountManagerEndpoint }`. object should be optional because both properties are

src/lib/fetch.ts
```diff
@@ -993,6 +1017,76 @@ async function fetchActions(
...
+/**
+ * Releases previously acquired key pair from accounts manager by public key.
+ *
+ * @param publicKey Public key of previously acquired key pair to release
+ * @param accountsManagerEndpoint Accounts manager endpoint to fetch from
+ * @returns Void (since we don't really care about the success of this operation)
+ */
+async function releaseKeyPair(
+  publicKey: string,
+  accountsManagerEndpoint = networkConfig.accountsManagerEndpoint
```
mitschabaude, line 1070: this function should take an object like the other one, with `accountManagerEndpoint` as optional property

src/lib/fetch.ts
```diff
@@ -993,6 +1017,76 @@ async function fetchActions(
...
+    },
+    body: JSON.stringify({
+      pk: publicKey,
+    }),
+  });
+
+  if (response.ok) {
+    const data = await response.json();
+    if (data) {
+      console.info(data.message);
```
mitschabaude, line 1085: I don't think it's good practice to log an info message here, as a general rule of thumb a library should not touch the console

src/lib/mina.ts
```diff
@@ -677,22 +677,34 @@ LocalBlockchain satisfies (...args: any) => Mina;
...
 function Network(graphqlEndpoint: string): Mina;
-function Network(graphqlEndpoints: {
+function Network(endpoints: {
+  mina: string | string[];
+  accountsManager?: string;
+}): Mina;
+function Network(endpoints: {
   mina: string | string[];
   archive: string | string[];
+  accountsManager?: string;
```
mitschabaude, line 687: same comment as before: I'd advocate for making this name even more explicit to avoid confusing people. Like `lightnetAccountManager`

Also, can you add an explanation of the accounts manager argument to the doccomment on this function?

src/lib/mina.ts
```diff
@@ -677,22 +677,34 @@ LocalBlockchain satisfies (...args: any) => Mina;
  * Represents the Mina blockchain running on a real network
  */
 function Network(graphqlEndpoint: string): Mina;
-function Network(graphqlEndpoints: {
+function Network(endpoints: {
+  mina: string | string[];
+  accountsManager?: string;
+}): Mina;
```
mitschabaude, line 683: I don't think it makes sense to have a separate function signature which includes `accountsManager` but excludes `archive` -- please delete. Instead, make `archive` optional in the one below

src/lib/mina.ts
```diff
@@ -677,22 +677,34 @@ LocalBlockchain satisfies (...args: any) => Mina;
...
+  let accountsManagerEndpoint: string;
 
   if (input && typeof input === 'string') {
     minaGraphqlEndpoint = input;
     Fetch.setGraphqlEndpoint(minaGraphqlEndpoint);
   } else if (input && typeof input === 'object') {
-    if (!input.mina || !input.archive)
+    if (!input.mina)
       throw new Error(
         "Network: malformed input. Please provide an object with 'mina' and 'archive' endpoints."
```
mitschabaude, line 709: please adjust this error message since you changed the logic!

src/lib/mina.ts
```diff
@@ -705,13 +717,20 @@ function Network(
...
     }
 
-    if (Array.isArray(input.archive) && input.archive.length !== 0) {
-      archiveEndpoint = input.archive[0];
-      Fetch.setArchiveGraphqlEndpoint(archiveEndpoint);
-      Fetch.setArchiveGraphqlFallbackEndpoints(input.archive.slice(1));
-    } else if (typeof input.archive === 'string') {
-      archiveEndpoint = input.archive;
-      Fetch.setArchiveGraphqlEndpoint(archiveEndpoint);
+    if (input.archive) {
```
mitschabaude, line 720: please use `if (input.archive !== undefined)`

src/lib/mina.ts
```diff
@@ -705,13 +717,20 @@ function Network(
...
+        archiveEndpoint = input.archive[0];
+        Fetch.setArchiveGraphqlEndpoint(archiveEndpoint);
+        Fetch.setArchiveGraphqlFallbackEndpoints(input.archive.slice(1));
+      } else if (typeof input.archive === 'string') {
+        archiveEndpoint = input.archive;
+        Fetch.setArchiveGraphqlEndpoint(archiveEndpoint);
+      }
+    }
+
+    if (input.accountsManager && typeof input.accountsManager === 'string') {
```
mitschabaude, line 731: please use `if (input.accountsManager !== undefined)`. We don't use runtime typeguards since we rely on TS (`accountsManager` can only be `string | undefined`)

src/examples/zkapps/hello_world/run_live.ts
```diff
@@ -91,3 +98,6 @@ try {
   );
 }
 console.log('Success!');
+
+// Tear down
+await releaseKeyPair(senderKey.toPublicKey().toBase58());
```
mitschabaude, line 103: nit: use `sender.toBase58()`

mitschabaude (commented):
I have a few commens!

mitschabaude:
@shimkiv another general comment: I think "account manager" is better as a name than "accounts manager". Would be nice if you could remove the "s" everywhere :D

src/lib/mina.ts
mitschabaude: hey @shimkiv here I also meant remove the `typeof input.accountsManager === 'string'` check. we don't do checks like that since the type enforces it already.

mitschabaude (approved):
added one more comment, otherwise looks good!

mitschabaude:
@shimkiv actually I have one more ask -- could you update the CHANGELOG?

PR #1176, opened 2023-10-11
Activate custom gates in JS and add rangeCheck64 gadget

### LINKED ISSUE #1172: Set feature flags in Pickles and Kimchi depending on custom gates used

mitschabaude:
closes #1172

bindings: https://github.com/o1-labs/o1js-bindings/pull/179

This PR
* adds logic to the Wasm bindings to infer feature flags from a verifier index  
 => makes Kimchi verification work for proofs that use custom gates
* dynamically passes the right feature flags to `Pickles.compile` based on the gates used in a circuit  
  => makes Pickles proofs work with custom gates
* to test that this works, exposes the `rangeCheck0` gate to JS, and wraps it in a 64-bit range check gadget
* uses that gadget in `UInt64`, which is very well covered by existing tests

TODOs:
* [x] understand performance impact on example zkapps and possibly remove custom gate from `UInt64` again, or move it behind a customization flag
* [x] changelog and documentation about the new gate

**UPDATE**
Following internal discussion, I didn't use the `rangeCheck64` gate in `UInt64` for now. Thus, there is no more test in this PR that everything works (but CI for a [previous commit ](https://github.com/o1-labs/o1js/pull/1176/commits/a8f06d1b9b5bcd7925052bc320cbbad275dffac8) shows it working)
However, there is now a follow-up PR which exposes the new gate directly (for advanced users), and also adds a unit test: https://github.com/o1-labs/o1js/pull/1181


src/lib/gadgets/range-check.ts
```diff
@@ -0,0 +1,14 @@
+import { type Field } from '../field.js';
+import * as Gates from '../gates.js';
+
+export { rangeCheck64 };
+
+function rangeCheck64(x: Field) {
```
MartinMinkov, line 6: Same comment from here: https://github.com/o1-labs/o1js/pull/1181#discussion_r1357115891

src/lib/proof_system.ts
```diff
@@ -822,6 +846,57 @@ function dummyBase64Proof() {
...
+    rangeCheck0: false,
+    rangeCheck1: false,
+    foreignFieldAdd: false,
+    foreignFieldMul: false,
+    xor: false,
+    rot: false,
+    lookup: false,
+    runtimeTables: false,
+  };
+  for (let gate of gates) {
```
MartinMinkov, line 860: nit: This is a superficial suggestion, but you could add some type safety here instead of the switch statement

```ts
const gateTypeToFlagKey: Record<GateType, keyof FeatureFlags> = {
  RangeCheck0: 'rangeCheck0',
  RangeCheck1: 'rangeCheck1',
  ForeignFieldAdd: 'foreignFieldAdd',
  ForeignFieldMul: 'foreignFieldMul',
  Xor16: 'xor',
  Rot64: 'rot',
  Lookup: 'lookup',
  ...
};
  
for (let gate of gates) {
  const flagKey = gateTypeToFlagKey[gate.type];
  if (flagKey) {
    flags[flagKey] = true;
  }
};

File: src/lib/gates.ts [hunk 3]
```diff
@@ -0,0 +1,49 @@
...
+      getBits(xx, 28, 12),
+      getBits(xx, 40, 12),
+      getBits(xx, 52, 12),
+    ];
+  });
+  Snarky.gates.rangeCheck0(
+    x.value,
+    [0, FieldVar[0], FieldVar[0], x52, x40, x28, x16],
+    [0, x14, x12, x10, x8, x6, x4, x2, x0],
+    // not using compact mode
```
MartinMinkov, line 40: `// not using compact mode`
why not? :sweat_smile:

MartinMinkov (approved):
Looks great to me! :tada:

Trivo25 (approved):

src/lib/proof_system.ts
mitschabaude: fwiw I think the switch statement is perfectly type-safe
image

but your suggestion is definitely less spaghetti-like than the long switch statement

mitschabaude: I went with the object approach. This is the type signature you need to encode that _some but not all_ gate types cause a feature flag to be set (same as the switch statement with the default branch):
```ts
const gateToFlag: Partial<Record<GateType, keyof FeatureFlags>> = {
  RangeCheck0: 'rangeCheck0',
  RangeCheck1: 'rangeCheck1',
  ```

src/lib/gates.ts
mitschabaude: this is just a different "mode" of the gate which is used in specific circumstances (when combining this with other gates to do a  2x88 + 1x88-bit range check where the 2x88-bit value is what is called "compact")

src/lib/gadgets/range-check.ts
mitschabaude: will address this in the next PR, thanks!

PR #1177, opened 2023-10-11
implements XOR gadget 

### LINKED ISSUE #1138: Port XOR gadget to TypeScript

Trivo25:
implements the `XOR` gadget - original implementation https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L205

bindings https://github.com/o1-labs/o1js-bindings/pull/180

closes #1138


src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,192 @@
...
+  lengthXor: number
+) {
+  // if inputs are zero and length is zero, add the zero check
+  if (padLength === 0) {
+    Gates.zeroCheck(a, b, expectedOutput);
+
+    let zero = new Field(0);
+    zero.assertEquals(a);
+    zero.assertEquals(b);
+    zero.assertEquals(expectedOutput);
```
Trivo25, line 92: I am actually unsure why we need two "zero checks" here, 1) the zero check gate, and 2) asserting that both elements equal zero. Isn't that the same thing?

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -0,0 +1,59 @@
...
+    let r1 = Fp.xor(BigInt(x), BigInt(y));
+
+    Provable.runAndCheck(() => {
+      let zz = Provable.witness(Field, () => z);
+      let yy = Provable.witness(Field, () => new Field(y));
+      let r2 = Gadgets.xor(zz, yy, length);
+      Provable.asProver(() => assert(r1 === r2.toBigInt()));
+    });
+  });
+});
```
mitschabaude, line 40: I tried to rewrite this test in equivalence testing style:

```ts
let uint = (length: number) => fieldWithRng(Random.biguint(length));

[2, 4, 8, 16, 32, 64, 3, 5, 10, 15].forEach((length) => {
  equivalent({ from: [uint(length), uint(length)], to: field })(
    Fp.xor,
    (x, y) => Gadgets.xor(x, y, length)
  );
});

// helper that should be added to `equivalent.ts`
function fieldWithRng(rng: Random<bigint>): Spec<bigint, Field> {
  return { ...field, rng };
}
```

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,195 @@
...
+function witnessSlices(f: Field, start: number, stop = -1) {
+  if (stop !== -1 && stop <= start)
+    throw Error('stop offset must be greater than start offset');
+
+  return Provable.witness(Field, () => {
+    let bits = f.toBits();
+    if (stop > bits.length) throw Error('stop must be less than bit-length');
+    if (stop === -1) stop = bits.length;
+
+    return Field.fromBits(bits.slice(start, stop));
```
mitschabaude, line 170: bitwise operations `>>` and `&` on `f.toBigint()` would be more efficient than slicing the bits.
also, since `f` is a variable, this probably goes through the provable `toBits` logic and snarky, which is always slow

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,27 @@ const Gadgets = {
...
+  /**
+   * Bitwise XOR gadget on {@link Field} elements. Equivalent to the [bitwise XOR `^` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR).
+   * A XOR gate works by comparing two bits and returning `1` if two bits differ, and `0` if two bits are equal.
+   *
+   * The `length` parameter lets you define how many bits should be compared. The output is not constrained to the length.
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional constraints.
+   *
+   * **Note:** Both {@link Field} elements need to fit into `2^length - 1`, or the operation will fail.
+   * For example, for `length = 2` ( 2² = 4), `.xor` will fail for any element that is larger than `> 3`.
```
mitschabaude, line 47: It is super important to document how this fails when the inputs don't fit in the bit length. Does this gate _constrain_ the inputs to that bit length? Or is this the caller's job, and this just fails because we have an `asProver` check for the bit length?

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,27 @@ const Gadgets = {
   rangeCheck64(x: Field) {
     return rangeCheck64(x);
   },
+
+  /**
+   * Bitwise XOR gadget on {@link Field} elements. Equivalent to the [bitwise XOR `^` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR).
+   * A XOR gate works by comparing two bits and returning `1` if two bits differ, and `0` if two bits are equal.
+   *
+   * The `length` parameter lets you define how many bits should be compared. The output is not constrained to the length.
```
mitschabaude, line 42: "The output is not constrained to the length" is interesting. Because further below you write that the inputs need to be of that bit length. And the `xor` of two n-bit values is again an n-bit value. So, what is happening here?

The following might be the correct answer (but I don't know and you should check @Trivo25):
"The gate does not constrain the inputs to `length` bits and if they don't fit in `length` bits, the method fails to properly constrain the operation. However, if the inputs do fit in `length` bits, the output is guaranteed to do so as well"

mitschabaude (changes requested):
Great job @Trivo25! There are some questions open for me

src/examples/regression_test.json
```diff
@@ -164,5 +164,18 @@
       "data": "",
       "hash": ""
     }
+  },
+  "Bitwise Primitive": {
+    "digest": "Bitwise Primitive",
+    "methods": {
+      "xor": {
+        "rows": 15,
```
mitschabaude, line 172: 15 seems correct to me:
* we can do 16 bits per xor gate
* so it's `length/16` xor gates, plus 1 zero gate
* since the circuit does 16, 32, 48, 64, we get 1+1, 2+1, 3+1, 4+1 which sums to 14
* the 15th row is caused by encoding the 0 constant (we use `assertEquals(0)` multiple times)

src/lib/gadgets/gadgets.ts
Trivo25: Yea thats right, I believe we had this discussion during reviewing the API PR, which I just took the doc comment from - let me know if the explanation makes more sense now

src/lib/field.ts
```diff
@@ -1245,13 +1246,24 @@ class Field {
...
+   * As all {@link Field} elements have 31 bytes, this function returns 31.
    *
    * @return The size of a {@link Field} element - 31.
    */
   static sizeInBytes() {
     return Fp.sizeInBytes();
   }
+
+  /**
+   * **Warning**: This function is mainly for internal use. Normally it is not intended to be used by a zkApp developer.
```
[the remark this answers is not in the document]
Trivo25: I agree, but I also think this is may be a bigger API change and should be handled separately by another PR. wdyt?

mitschabaude:
After readng through the gate again, I'm pretty sure it doesn't constrain the xor operation at all. The way that `a`, `b`, `expectedOutput` change from one iteration to the next is not constrained. So after all we probably do need the [code I asked you to remove here](https://github.com/o1-labs/o1js/pull/1177#discussion_r1362253560). However, it needs to become part constrained, not part of a witness block

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+let maybeUint64: Spec<bigint, Field> = {
+  ...field,
+  rng: Random.map(Random.oneOf(Random.uint64, Random.uint64.invalid), (x) =>
+    mod(x, Field.ORDER)
+  ),
+};
+
+// do a couple of proofs
+equivalentAsync({ from: [maybeUint64, maybeUint64], to: field }, { runs: 3 })(
```
mitschabaude, line 45: this test is missing an `await` and I tried it without -- it doesn't actually succeed

I made the same error in the range check unit test (but that one seems to work) -- shoutout to @MartinMinkov for spotting this mistake in his PR!

The test fails because
1) you use numbers for `2**64` which is not exact; use bigints
2) because you test for `x >`, `y >` but it should be `>=`

mitschabaude: @Trivo25 can you fix the range check unit test as well?

Trivo25: yep!

src/examples/regression_test.json
```diff
@@ -164,5 +164,18 @@
       "data": "",
       "hash": ""
     }
+  },
+  "Bitwise Primitive": {
+    "digest": "Bitwise Primitive",
+    "methods": {
+      "xor": {
+        "rows": 75,
```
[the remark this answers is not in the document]
mitschabaude: recalling that your test circuit does 1+2+3+4=10 rounds of 16-bit XOR, and it takes 60 more rows than before, id sounds like 6 additional constraints per 16-bit Xor round.

so this calculation, which is done 3 times per round, seems to take 2 rows:
```
(x - (x0 + x1 2^4 + x2 2^8 + x3 2^12)) / 2^16
```
this makes sense, because we can do 2 additions per row (double generic gate) and there are 4 additions (the multiplications by constants can be included at no additional costs)

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,130 @@
...
+  if (a.isConstant() && b.isConstant()) {
+    assert(
+      a.toBigInt() < 2 ** length,
+      `${a.toBigInt()} does not fit into ${length} bits`
+    );
+
+    assert(
+      b.toBigInt() < 2 ** length,
+      `${b.toBigInt()} does not fit into ${length} bits`
+    );
```
mitschabaude, line 28: To match the behaviour of the provable version, `2 ** length` should be replace by `2 ** padLength`. Also, it should be done with bigints to get an exact result. To make this work you'll need to move the definition of `padLength` to above this.
```suggestion
   let max = 1n << BigInt(padLength);
    assert(
      a.toBigInt() < max,
      `${a.toBigInt()} does not fit into ${padLength} bits`
    );

    assert(
      b.toBigInt() < max,
      `${b.toBigInt()} does not fit into ${padLength} bits`
    );
```

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,130 @@
...
+  let outputXor = Provable.witness(
+    Field,
+    () => new Field(Fp.xor(a.toBigInt(), b.toBigInt()))
+  );
+
+  // obtain pad length until the length is a multiple of 16 for n-bit length lookup table
+  let l = 16;
+  let padLength = Math.ceil(length / l) * l;
+
+  // recursively build xor gadget chain
```
mitschabaude, line 43: nit: not recursive

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,130 @@
...
+  buildXor(a, b, outputXor, padLength);
+
+  // return the result of the xor operation
+  return outputXor;
+}
+
+// builds xor chain recursively
```
mitschabaude, line 50: nit: not recursive

mitschabaude (commented):
More comments!

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,126 @@
...
+  let l = 16;
+  let padLength = Math.ceil(length / l) * l;
+
+  // handle constant case
+  if (a.isConstant() && b.isConstant()) {
+    let max = 1n << BigInt(padLength);
+
+    assert(a.toBigInt() < max, `${a.toBigInt()} does not fit into ${max} bits`);
+
+    assert(b.toBigInt() < max, `${b.toBigInt()} does not fit into ${max} bits`);
```
mitschabaude, line 28:
```suggestion
    assert(a.toBigInt() < max, `${a.toBigInt()} does not fit into ${padLength} bits`);

    assert(b.toBigInt() < max, `${b.toBigInt()} does not fit into ${padLength} bits`);
```

mitschabaude (approved):

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,126 @@
...
+  assert(length > 0, `Input lengths need to be positive values.`);
+
+  // check that length does not exceed maximum field size in bits
+  assert(
+    length <= Field.sizeInBits(),
+    `Length ${length} exceeds maximum of ${Field.sizeInBits()} bits.`
+  );
+
+  // obtain pad length until the length is a multiple of 16 for n-bit length lookup table
+  let l = 16;
```
jspada, line 19: maybe just `let padLength = Math.ceil(length / 16) * 16;` since `l` is not used anywhere else...

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,126 @@
...
+    let in1_2 = witnessSlices(a, 8, 4);
+    let in1_3 = witnessSlices(a, 12, 4);
+
+    // slices of b
+    let in2_0 = witnessSlices(b, 0, 4);
+    let in2_1 = witnessSlices(b, 4, 4);
+    let in2_2 = witnessSlices(b, 8, 4);
+    let in2_3 = witnessSlices(b, 12, 4);
+
+    // slice of expected output
```
jspada, line 68: typo: slices

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,126 @@
...
+}
+
+function assert(stmt: boolean, message?: string) {
+  if (!stmt) {
+    throw Error(message ?? 'Assertion failed');
+  }
+}
+
+function witnessSlices(f: Field, start: number, length: number) {
+  if (length <= 0) throw Error('Length must be a positive number');
```
jspada, line 116: do you want to support zero length?

jspada (commented):
lgtm

src/examples/regression_test.json
querolita: Does the above means that this primitive performs all of the 4 xor variants together? Instead of one per variant (one Xor for 16 of 3 rows, another for 64 with 6 rows, etc). Or is this circuit just being used for tests?

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+import * as Gates from '../gates.js';
+
+export { xor };
+
+function xor(a: Field, b: Field, length: number) {
+  // check that both input lengths are positive
+  assert(length > 0, `Input lengths need to be positive values.`);
+
+  // check that length does not exceed maximum field size in bits
+  assert(
```
querolita, line 13: ~I wonder if you might also want to check that the given inputs fit in the desired length and fail otherwise?~ Never mind, I saw this is checked inside the case where `a` and `b` are constants below.

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
+import { Provable } from '../provable.js';
+import { Field as Fp } from '../../provable/field-bigint.js';
+import { Field } from '../field.js';
+import * as Gates from '../gates.js';
+
+export { xor };
+
+function xor(a: Field, b: Field, length: number) {
```
querolita, line 8: Maybe clarify that length shall be given in bits instead of bytes.

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+      a.toBigInt() < max,
+      `${a.toBigInt()} does not fit into ${padLength} bits`
+    );
+
+    assert(
+      b.toBigInt() < max,
+      `${b.toBigInt()} does not fit into ${padLength} bits`
+    );
+
+    return new Field(Fp.xor(a.toBigInt(), b.toBigInt()));
```
querolita, line 35: I am not familiar with the types in o1js so this might be a basic question. Is converting to BigInt the desired behavior here? Since the inputs are `Field`, I wonder why this conversion needs to take place.

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // calculate expect xor output
```
querolita, line 38: *expected?

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+// builds a xor chain
+function buildXor(
+  a: Field,
+  b: Field,
+  expectedOutput: Field,
+  padLength: number
+) {
+  // construct the chain of XORs until padLength is 0
+  while (padLength !== 0) {
+    // slices the inputs 4 4bit-sized chunks
```
querolita, line 60: maybe "x4 4bit-sized chunks" reads better

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,195 @@
...
+    let in2_2 = witnessSlices(b, second, third);
+    let in2_3 = witnessSlices(b, third, fourth);
+
+    // slice of expected output
+    let out0 = witnessSlices(expectedOutput, 0, first);
+    let out1 = witnessSlices(expectedOutput, first, second);
+    let out2 = witnessSlices(expectedOutput, second, third);
+    let out3 = witnessSlices(expectedOutput, third, fourth);
+
+    // assert that the xor of the slices is correct, 16 bit at a time
```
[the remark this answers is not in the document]
querolita: Good point. Actually, there's an unmerged branch generalizing the Xor gate for whatever xor table length available, in the proof-systems repo -> https://github.com/o1-labs/proof-systems/compare/master...xor/coeff

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    // update the values for the next loop iteration
+    a = witnessNextValue(a);
+    b = witnessNextValue(b);
+    expectedOutput = witnessNextValue(expectedOutput);
+    padLength = padLength - 16;
+  }
+
+  // inputs are zero and length is zero, add the zero check - we reached the end of our chain
+  Gates.zero(a, b, expectedOutput);
```
querolita, line 106: Maybe this is too low level, but o1js is sure that this line, together with the following 4, makes sure that all 3 values are zero right? Does it use a Zero gate and then a wiring to a Generic zero constant gate? Otherwise, what is it using beneath? Thank you

src/lib/gadgets/bitwise.ts
querolita: I don't think they do. Even if you are xoring zero, it still needs to be padded to some amount of bits.

querolita: Actually, if I remember correctly, they only call `witnessSlices` from within the gadget, so I understand this error will not show up in "legit" executions.

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+    throw Error(message ?? 'Assertion failed');
+  }
+}
+
+function witnessSlices(f: Field, start: number, length: number) {
+  if (length <= 0) throw Error('Length must be a positive number');
+
+  return Provable.witness(Field, () => {
+    let n = f.toBigInt();
+    return new Field((n >> BigInt(start)) & ((1n << BigInt(length)) - 1n));
```
querolita, line 125: lgtm

src/lib/gadgets/bitwise.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+}
+
+function witnessNextValue(current: Field) {
+  return Provable.witness(Field, () => new Field(current.toBigInt() >> 16n));
```
querolita, line 130: lgtm

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,29 @@ const Gadgets = {
...
+
+  /**
+   * Bitwise XOR gadget on {@link Field} elements. Equivalent to the [bitwise XOR `^` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_XOR).
+   * A XOR gate works by comparing two bits and returning `1` if two bits differ, and `0` if two bits are equal.
+   *
+   * This gadget builds a chain of XOR gates recursively. Each XOR gate can verify 16 bit at most. If your input elements exceed 16 bit, another XOR gate will be added to the chain.
+   *
+   * The `length` parameter lets you define how many bits should be compared. `length` is rounded to the nearest multiple of 16, `paddedLength = ceil(length / 16) * 16`, and both input values are constrained to fit into `paddedLength` bits. The output is guaranteed to have at most `paddedLength` bits as well.
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional constraints.
```
querolita, line 46: Also, note that using a smaller length can let the verifier infer the length of the inputs. So even if the zkapp developer knew that the inputs to some program will be <32bits, they should still consider if they truly want a smaller gadget or a larger one that leaks less side information about the data.

querolita (approved):
Looks good to me! Left a bunch of comments and doubts for clarification since this is the first piece of o1js code that I ever read. But other than that, everything looks tidy and safe to my eyes.

src/lib/gadgets/bitwise.ts
Trivo25: This case only happens if both elements are constants (not variables that the prover needs to provide). Because of that, we can just convert them to bigint and do our computation there and convert them back. Doing so is more efficient than going through snarky and letting snarky do the calculation on constants. This is a change we did somewhat recently: Computations on constants are done on native javascript types

src/lib/gadgets/gadgets.ts
Trivo25: Uhhh this is actually a super good point, I love that!

src/lib/gadgets/bitwise.ts
querolita: Oh very cool and useful! It is not that simple from Snarky!

PR #1181, opened 2023-10-12
Expose and test rangeCheck64 gadget

mitschabaude:
expose the range check gate added in #1176 and add a unit test

Notes:
Most of this PR consists of refactoring the `equivalence()` testing framework that I wanted use for the new test.
I really liked the easy-to-write and type-safe way of how I used this framework in the [bindings unit test](https://github.com/o1-labs/o1js-bindings/blob/8247786182780abf729b313b8c026d5fa9aed8fd/crypto/bindings/bindings.unit-test.ts), that's why I decided to move it from the bindings repo to o1js and consolidate it with the equivalence testing logic that was already there. For a good idea of how these tests are done, see the field unit test (which changed a lot in this PR)

bindings: https://github.com/o1-labs/o1js-bindings/pull/183


src/lib/gadgets/range-check.ts
```diff
@@ -3,6 +3,10 @@ import * as Gates from '../gates.js';
 
 export { rangeCheck64 };
 
+/**
```
MartinMinkov, line 6: I'm assuming that this function will be what users will see for the exported comments, I would love to see a more helpful doc comment here, do you think something like this would be okay?

```ts
/**
 * Validates if the given field value fits within the range [0, 2^64).
 * 
 * This function checks if the provided field value can be represented within 64 bits.
 * If the field value exceeds 64 bits, an error is thrown.
 *
 * @param x - The field value to be checked.
 * 
 * @throws Will throw an error if the field value exceeds 64 bits.
 * 
 * @example
 * ```ts
 * const field = new Field(12345678n);
 * rangeCheck64(field);  // Validates without throwing an error
 * 
 * const invalidField = new Field(12345678901234567890123456789012345678n);
 * rangeCheck64(invalidField);  // Throws an error since it exceeds 64 bits
 * ```
 */
 ```

MartinMinkov (approved):
This is awesome! :tada: Testing changes look great too :+1:

src/lib/gadgets/range-check.ts
MartinMinkov: Or maybe if we have another import that's used, we can mention the TSDoc there too?

Trivo25 (approved):

PR #1182, opened 2023-10-12
Add rot gadget to o1js

MartinMinkov:
# Summary
This PR adds the implementation for a `ROT` gate to the existing `Gadgets` namespace. For context of what a bitwise rotation is, a rotation (often referred to as a "bitwise rotation") is an operation that shifts the bits of a binary number either to the left or to the right, but unlike a standard shift operation, the bits that "fall off" the end are not discarded. Instead, they "wrap around" to the other end.

The implementation is based on the original OCaml implementation which is specified [here](https://github.com/MinaProtocol/mina/blob/4b2e023eec39718813fbacc38661558a3193a433/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L154)

In our `ROT` implementation, we handle the constant case by using the added functionality in the bindings, and for the prover case, we specify the implementation in this PR.

**Note**: The `rotate` function that is specified for the `ROT` gate will also be used for the upcoming shift gates.

# Tested
This PR adds e2e tests, unit tests, and verification key tests for the ROT gate.

:link: bindings: https://github.com/o1-labs/o1js-bindings/pull/184


Trivo25 (commented):
Great job! Some quick initial comments, I will take a deeper look tomorrow :)

src/examples/gadgets.ts
```diff
@@ -0,0 +1,93 @@
+import { Field, Provable, Experimental, Gadgets } from 'o1js';
+
+function testRot(
```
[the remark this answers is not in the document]
MartinMinkov: Will do!

src/examples/gadgets.ts
```diff
@@ -0,0 +1,93 @@
...
+    let output = Gadgets.rot(w, bits, mode);
+    Provable.asProver(() => {
+      Provable.log(`rot(${word}, ${bits}, ${mode}) = ${output}`);
+    });
+    output.assertEquals(r, `rot(${word}, ${bits}, ${mode})`);
+  });
+}
+
+console.log('Running positive tests...');
+testRot(Field('0'), 0, 'left', Field('0'));
```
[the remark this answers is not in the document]
MartinMinkov: Yep! The unit tests in the OCaml implementation follow a similar style for [its tests](https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L608), and I wanted to keep things consistent with the OCaml implementation. I also ran into a failing test when using numbers. Specifically, [this test](https://github.com/MinaProtocol/mina/blob/f16e20a908ef0beccdef77e4128853ed8cbc2dab/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L613) wasn't working, but the string version does, so just to keep things simple and used strings.

src/examples/gadgets.ts
```diff
@@ -0,0 +1,93 @@
...
+  4,
+  'right',
+  Field('11936128518282651045')
+);
+console.log('🎉 Passed positive tests');
+
+let cs = Provable.constraintSystem(() => {
+  let res1 = Provable.witness(Field, () => {
+    let f = Field(12);
+    return Gadgets.rot(f, 2, 'left');
```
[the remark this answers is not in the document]
MartinMinkov: Got it, will fix this

src/examples/gadgets.ts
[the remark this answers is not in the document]
mitschabaude: Isn't the reason that other test didn't work that the numbers were too large to represent as JS numbers exactly? IMO for such a case we should use bigints, strings feel dirty somehow (they have to be parsed into bigints)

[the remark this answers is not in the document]
MartinMinkov: That's a good point that I missed! Will try bigints instead!

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,121 @@
...
+  );
+  // Compute next row
+  Gates.rangeCheck64(shifted);
+  // Compute following row
+  Gates.rangeCheck64(excess);
+  Gates.rangeCheck64(rotated);
+  return [rotated, excess, shifted];
+}
+
+function witnessSlices(f: Field, start: number, stop = -1) {
```
MartinMinkov, line 99: One thing for the reviewer: I tried using t[his implementation for this function instead](https://github.com/o1-labs/o1js/pull/1177/files#diff-549b20f5f9b5702170289d8f7fe81901e79f801c2e07cfb35d1c9bf3b01a35e1R115), but I get issues with unit tests. Working on a solution to make this not an error but address the original comments.

Specifically I run into the following stacktrace:
```sh
Error: the lookup failed to find a match in the table: row=12
    at module.exports.__wbindgen_error_new (/home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/bindings/compiled/_node_bindings/plonk_wasm.cjs:9293:17)
    at <wasm_bindgen::JsError as core::convert::From<E>>::from::h745e731c0bba10da (wasm://wasm/01112d6e:wasm-function[4037]:0x3bd9f8)
    at caml_pasta_fp_plonk_proof_create (wasm://wasm/01112d6e:wasm-function[718]:0x24d985)
    at module.exports.caml_pasta_fp_plonk_proof_create (/home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/bindings/compiled/_node_bindings/plonk_wasm.cjs:2756:14)
    at async withThreadPool (file:///home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/bindings/js/node/node-backend.js:47:18)
    at async prettifyStacktracePromise (file:///home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/lib/errors.js:104:16)
    at async prove_ (file:///home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/lib/proof_system.js:165:26)
    at async file:///home/martin/Code/o1/mina/src/lib/snarkyjs/dist/node/lib/gadgets/gadgets.unit-test.js:56:17
``

Comment (Trivo25) [reply src/lib/gadgets/rot.ts:99]: The function from the other PR has the `length` as input, rather than an `end` index - maybe that was the reason and the specified parameters exceeded that? Curious to see what it fails, can you point me to a PR?

File: src/lib/gadgets/gadgets.ts [hunk 5]
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
...
+   * const y = rot(x, 2, 'left'); // left rotation by 2 bits
+   * const z = rot(x, 2, 'right'); // right rotation by 2 bits
+   * y.assertEquals(48);
+   * z.assertEquals(3)
+   *
+   * const xLarge = Provable.witness(Field, () => Field(12345678901234567890123456789012345678n));
+   * rot(xLarge, 32, "left"); // throws an error since input exceeds 64 bits
+   * ```
+   */
+  rot(word: Field, bits: number, direction: 'left' | 'right' = 'left') {
```
MartinMinkov, line 62: One thought we could do here instead is instead of taking `word` as a `Field`, we specify its type to only be `UInt64`. Any thoughts?

src/lib/gadgets/rot.ts
MartinMinkov: Ah, great point, I missed that :sweat: Fixed it and it passes the failing tests I had

src/lib/gadgets/gadgets.ts
Trivo25: My preference would be to keep the gadgets "pure" and only accept field elements, one reason for that is gadgets and gates naturally feel low level to me, so I think it would be a good idea to only feed them with field elements. Another reason is that algorithms built on top of it might operate on field element by default (e.g. because its better DX or other gates that aren't constrained by bit size do) - `UInt64 <> Field` conversions might become a bit tedious in the future

src/examples/gadgets.ts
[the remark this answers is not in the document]
Trivo25: No problem! I have to admit that changing the original API was my fault :D

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,113 @@
...
+function rotate(
+  word: Field,
+  bits: number,
+  direction: 'left' | 'right' = 'left'
+): [Field, Field, Field] {
+  // Check as the prover, that the input word is at most 64 bits.
+  Provable.asProver(() => {
+    if (word.toBigInt() > 2 ** MAX_BITS) {
+      throw Error(
+        `rot: expected word to be at most 64 bits, got ${word.toBigInt()}`
```
Trivo25, line 38: I think we should also check for this case when the inputs are constants (or before, to catch the constant-case)

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+function rotate(
+  word: Field,
+  bits: number,
+  direction: 'left' | 'right' = 'left'
+): [Field, Field, Field] {
+  // Check as the prover, that the input word is at most 64 bits.
+  Provable.asProver(() => {
+    if (word.toBigInt() > 2 ** MAX_BITS) {
```
Trivo25, line 36: `word.toBigInt()` is of type `bigint`, but `2 ** MAX_BITS` is a `number` - I would prefer it if both operands are `bigint`

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,113 @@
...
+      witnessSlices(bound, 2, 2), // bits 2-4
+      witnessSlices(bound, 0, 2), // bits 0-2
+    ],
+    Field.from(big2PowerRot)
+  );
+  // Compute next row
+  Gates.rangeCheck64(shifted);
+  // Compute following row
+  Gates.rangeCheck64(excess);
+  Gates.rangeCheck64(rotated);
```
Trivo25, line 95: i think the `rotated` value might already be constrained in the gate, we should check this - if that's the case, we can save this constraint. But I am not sure, need to check myself

Trivo25 (approved):
Great job! looks good to me :)

src/examples/gadgets.ts
```diff
@@ -0,0 +1,49 @@
...
+      method: () => {
+        let a = Provable.witness(Field, () => Field(48));
+        let actualLeft = Gadgets.rot(a, 2, 'left');
+        let actualRight = Gadgets.rot(a, 2, 'right');
+
+        let expectedLeft = Field(192);
+        actualLeft.assertEquals(expectedLeft);
+
+        let expectedRight = 12;
+        actualRight.assertEquals(expectedRight);
```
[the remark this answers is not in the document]
MartinMinkov: Good point, I'll change that

src/lib/gadgets/rot.ts
MartinMinkov: Is that necessary since we check for it [here](https://github.com/o1-labs/o1js/pull/1182/files#diff-50db730c4f28e956a4c9fdf2c7c3a12a1ed77ea519468ceb8d66428afa678b1dR10)?

I figured we wouldn't expose `rotate` anywhere external, so the only entry point is through `rot`, which does that check.

Trivo25: maybe I am mistaken but I don't think `rot` checks for the size of  `word`, it only checks that `!(bits < 0 || bits > MAX_BITS)` and `!(direction !== 'left' && direction !== 'right')` or does it?

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
...
     return rangeCheck64(x);
   },
+
+  /**
+   * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
+   * `direction` is a string which accepts either `'left'` or `'right'`, defining the direction of the rotation.
+   *
+   * **Note:** You can not rotate {@link Field} elements that exceed 64 bits. For elements that exceed 64 bits this operation will fail.
+   *
+   * @param word {@link Field} element to rotate.
```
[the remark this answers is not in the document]
MartinMinkov: Open to changing it. We could do:

1.`rot(field: Field, ...)`
2.`rot(f: Field, ...)`
3.`rot(x: Field, ...)`
4.`rot(value: Field, ...)`
5.`rot(bitField: Field, ...)`
6.`rot(bits: Field, ...)`

Open to any suggestions!

src/lib/gadgets/rot.ts
MartinMinkov: Fixed!

src/lib/gadgets/rot.ts
MartinMinkov: Oh yeah, you're right. Sorry, I misunderstood!

src/lib/gadgets/gadgets.ts
MartinMinkov: Sounds like great reasoning!

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
...
+   * @example
+   * ```ts
+   * const x = Provable.witness(Field, () => Field(12));
+   * const y = rot(x, 2, 'left'); // left rotation by 2 bits
+   * const z = rot(x, 2, 'right'); // right rotation by 2 bits
+   * y.assertEquals(48);
+   * z.assertEquals(3)
+   *
+   * const xLarge = Provable.witness(Field, () => Field(12345678901234567890123456789012345678n));
+   * rot(xLarge, 32, "left"); // throws an error since input exceeds 64 bits
```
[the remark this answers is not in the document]
MartinMinkov: I think that would be useful! Would probably change the algorithm too much so I agree that it's probably something for later and not in this PR

src/lib/gadgets/rot.ts
Trivo25: The conditional check in OCaml is for `word`, here it is for `rotated`. Can you double check? Maybe I am confused :D

src/lib/gadgets/gadgets.ts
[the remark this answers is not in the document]
Trivo25: either works!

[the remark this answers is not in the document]
MartinMinkov: Went with `field` for simplicity :D

src/lib/gadgets/rot.ts
MartinMinkov: Oh no, you're right. Incredibly sloppy on this PR, sorry! I switched the `rangeCheck` for the input value instead.

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -45,3 +48,71 @@ equivalentAsync({ from: [maybeUint64], to: boolean }, { runs: 3 })(
...
+
+function testRot(
+  field: Field,
+  bits: number,
+  mode: 'left' | 'right',
+  result: Field
+) {
+  Provable.runAndCheck(() => {
+    let w = Provable.witness(Field, () => field);
+    let r = Provable.witness(Field, () => result);
```
mitschabaude, line 87: nit: there's nothing gained by turning `result` into a witness here

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -45,3 +48,71 @@ equivalentAsync({ from: [maybeUint64], to: boolean }, { runs: 3 })(
...
+}
+
+testRot(Field(0), 0, 'left', Field(0));
+testRot(Field(0), 32, 'right', Field(0));
+testRot(Field(1), 1, 'left', Field(2));
+testRot(Field(1), 63, 'left', Field('9223372036854775808'));
+testRot(Field(256), 4, 'right', Field(16));
+testRot(Field(1234567890), 32, 'right', Field(5302428712241725440));
+testRot(Field(2651214356120862720), 32, 'right', Field(617283945));
+testRot(Field(1153202983878524928), 32, 'right', Field(268500993));
```
mitschabaude, line 100: nit: these two values are larger than the max safe integer, and aren't captured exactly as JS numbers, i.e.
```
$ node
> 1153202983878524928
1153202983878525000
```

the test still works because it removes the lowest 32 bits with the right shift, and for that the values are easily close enough. I'd still use bigints

mitschabaude: Oh I just saw the expected result on the line above is also larger than the max

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+  // Check that the rotation bits are in range
+  if (bits < 0 || bits > MAX_BITS) {
+    throw Error(`rot: expected bits to be between 0 and 64, got ${bits}`);
+  }
+
+  if (direction !== 'left' && direction !== 'right') {
+    throw Error(
+      `rot: expected direction to be 'left' or 'right', got ${direction}`
+    );
+  }
```
mitschabaude, line 20: nit: according to my philosophy, we don't do runtime checks for cases that are impossible according to the types. wdyt?

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+
+function rotate(
+  field: Field,
+  bits: number,
+  direction: 'left' | 'right' = 'left'
+): [Field, Field, Field] {
+  // Check as the prover, that the input is at most 64 bits.
+  Provable.asProver(() => {
+    checkMaxBits(field);
+  });
```
mitschabaude, line 38: I think this check should be removed, reasoning here: https://github.com/o1-labs/o1js/pull/1177#discussion_r1362119516

src/lib/gates.ts
```diff
@@ -42,6 +43,24 @@ function rangeCheck64(x: Field) {
   );
 }
 
+function rot(
+  field: Field,
+  rotated: Field,
+  excess: Field,
+  limbs: [Field, Field, Field, Field],
+  crumbs: [Field, Field, Field, Field, Field, Field, Field, Field],
+  two_to_rot: Field
```
mitschabaude, line 52: since this is a constant gate parameter, I think it should be passed as a bigint (to avoid ambiguity with variables),
```suggestion
  two_to_rot: bigint
```

mitschabaude (changes requested):
First pass looks great! A few comments. Want to have a second look after studying the gate, to understand the constraints used here.

General suggestion: As I'm reading a lot of `rot`, I think `rotate` might be a better name. Interested in your thoughts on this @MartinMinkov and @Trivo25

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as Gates from '../gates.js';
+
+export { rot, rotate };
+
+const MAX_BITS = 64 as const;
+
+function rot(field: Field, bits: number, direction: 'left' | 'right' = 'left') {
+  // Check that the rotation bits are in range
+  if (bits < 0 || bits > MAX_BITS) {
+    throw Error(`rot: expected bits to be between 0 and 64, got ${bits}`);
```
querolita, line 13: Actually, I would have said `<= 0 and => MAX_BITS`, given that rotation in those cases is trivial, and those could be excluded from the circuit wlog. The current implementation in Kimchi does not support =0 nor =64, for efficiency. Thus, I would say we should either adapt the Kimchi rotation witness generation code to support those edge cases (just some types adaptation to prevent overflows, the actual constraints remain unchanged), or else this interface should behave just like the other one. In any case, I just created a [Kimchi PR](https://github.com/o1-labs/proof-systems/pull/1299) to support this functionality, so both implementations do the same.

src/lib/gadgets/rot.ts
querolita: I guessed this was implicit from the input type being either `'left'` or `'right'`, no?

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+      // f * 2^rot = excess * 2^64 + shifted
+      const { quotient: excess, remainder: shifted } = divideWithRemainder(
+        f * big2PowerRot,
+        big2Power64
+      );
+
+      // Compute rotated value as:
+      // rotated = excess + shifted
+      const rotated = shifted + excess;
+      // Compute bound that is the right input of FFAdd equation
```
querolita, line 59: I wouldn't call it FFAdd equation. Even if we use this trick inside FFAdd, this is just an inequality to check excess < 2^rot.

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+      witnessSlices(bound, 2, 2), // bits 2-4
+      witnessSlices(bound, 0, 2), // bits 0-2
+    ],
+    Field.from(big2PowerRot)
+  );
+  // Compute next row
+  Gates.rangeCheck64(shifted);
+  // Compute following row
+  Gates.rangeCheck64(excess);
+  Gates.rangeCheck64(field);
```
querolita, line 92: Oh, so you're also checking that the input word is 64 bits? I guess this makes sense for the general case though... But maybe this could be disabled by the zkapp developer if the circuit already implied that the value is 64bits. For example, inside the Keccak gadget, inputs to rotation come from a previous Xor64 gadget, which implicitly check that the words are 64 bits due to decomposition into 4 chunks of 4 bits each. Otherwise this would incur in unneeded rows inside Keccak, making it less efficient.

src/lib/gadgets/rot.ts
```diff
@@ -0,0 +1,118 @@
...
+  return [rotated, excess, shifted];
```
querolita, line 93: Why do you return these three values instead of just `rotated`?

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
   rangeCheck64(x: Field) {
     return rangeCheck64(x);
   },
+
+  /**
+   * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
+   * `direction` is a string which accepts either `'left'` or `'right'`, defining the direction of the rotation.
```
querolita, line 40: Maybe clarify what left and right mean exactly in terms of endianness.

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
...
+   *
+   * @throws Throws an error if the input value exceeds 64 bits.
+   *
+   * @example
+   * ```ts
+   * const x = Provable.witness(Field, () => Field(12));
+   * const y = rot(x, 2, 'left'); // left rotation by 2 bits
+   * const z = rot(x, 2, 'right'); // right rotation by 2 bits
+   * y.assertEquals(48);
+   * z.assertEquals(3)
```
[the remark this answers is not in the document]
querolita: Agree

src/lib/gadgets/gadgets.ts
```diff
@@ -33,4 +34,32 @@ const Gadgets = {
   rangeCheck64(x: Field) {
     return rangeCheck64(x);
   },
+
+  /**
+   * A (left and right) rotation is similar to the shift operation, `<<` and `>>` in JavaScript, just that bits are being appended to the other side.
+   * `direction` is a string which accepts either `'left'` or `'right'`, defining the direction of the rotation.
+   *
+   * **Note:** You can not rotate {@link Field} elements that exceed 64 bits. For elements that exceed 64 bits this operation will fail.
```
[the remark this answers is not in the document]
querolita: Agree

src/lib/gadgets/gadgets.ts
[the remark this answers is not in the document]
querolita: Something larger than 64 is desirable for sure, but for now we kept it to 64 because other amount meant changing the Kimchi constraints themselves. But I agree that we should do this at some point.

src/examples/regression_test.json
```diff
@@ -164,5 +164,18 @@
       "data": "",
       "hash": ""
     }
+  },
+  "Bitwise Primitive": {
+    "digest": "Bitwise Primitive",
+    "methods": {
+      "rot": {
+        "rows": 17,
```
querolita, line 172: Where did this 17 come from? I am guessing this uses at least 1 row for Rot64, another 3 for RangeCheck0 (meaning excess, shifted, and the word), and a constant row with value 0 to check some cells were zero in the RangeCheck0's. But what about the rest?

querolita: Oh, is this the BitwiseMock thing? I am seeing now that one produced 4 rotations, so maybe that's where 4*4+1 came from?

querolita (approved):
Looks good to me! Left some comments and suggestions throughout the code. Well done!

mitschabaude (changes requested):
Finished review! Main comment from the second round is to not range-check the input.

src/lib/gadgets/gadgets.ts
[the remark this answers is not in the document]
mitschabaude: In light of https://github.com/o1-labs/o1js/pull/1182#discussion_r1371876017 this should be formulated differently, something like this:
```suggestion
   * **Important:** The gadgets assumes that its input is at most 64 bits in size.
   * If the input exceeds 64 bits, the gadget is invalid and does not prove correct execution of the rotation.
   * Therefore, to safely use `rot()`, you need to make sure that the values passed in are range checked to 64 bits.
   * For example, this can be done with {@link Gadgets.rangeCheck64}.
```

src/lib/gadgets/rot.ts
MartinMinkov: >Oh, I see now, this is to support shifts right?

Yep, that's right! :D

src/lib/gadgets/gadgets.unit-test.ts
MartinMinkov: Fixed with https://github.com/o1-labs/o1js/pull/1182/commits/b9ef95e90edf8959f7af0df1d0e2ed506a4a08ce

src/lib/gadgets/rot.ts
MartinMinkov: Fixed with https://github.com/o1-labs/o1js/pull/1182/commits/33aff5c6848aa2421b27a6bf21c70137271e8954

My original reasoning is that TypeScript stops you from passing in a different value, but you can still "force" it with JS. I think you're right, and the check was removed :+1:

src/lib/gates.ts
MartinMinkov: fixed in https://github.com/o1-labs/o1js/pull/1182/commits/00530afef51180d9e10fac8a619c7c6c478597d3

src/lib/gadgets/rot.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1182/commits/b3959506db7be28bbf4952484cb93e3042f77286

src/lib/gadgets/rot.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1182/commits/1948ea41cc9ba1814a3b1874f854668b471672ff

src/lib/gadgets/gadgets.ts
MartinMinkov: Good call! Fixed in https://github.com/o1-labs/o1js/pull/1182/commits/fbe8aa59fb2c41e1713924298a5c296b68249a93

src/lib/gadgets/rot.ts
MartinMinkov: Ah, nice! Thank you for making the Kimchi PR. I don't mind changing the checks; I guess I'll just wait until the Kimchi PR lands before merging this.

mitschabaude:
Re-reviewed, only thing left from my side is https://github.com/o1-labs/o1js/pull/1182#discussion_r1370720022

src/lib/gadgets/rot.ts
querolita: The PR has just landed. Feel free to include the support for =0 and =64. Nonetheless, it would be a good idea to point out in a comment or the docs, that rotation by those amounts is trivial and they should reason about whether they really want to include those rows in the circuit or not.

src/lib/gadgets/rot.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1182/commits/36415ad666933783b8a3596dd196c75b31b68c69

mitschabaude (approved):

PR #1187, opened 2023-10-19
Cache prover keys

### LINKED ISSUE #87: Finalize SnarkyJS build artifact

mitschabaude:
closes #87 

mina: https://github.com/MinaProtocol/mina/pull/14396 and https://github.com/MinaProtocol/mina/pull/14412
bindings: https://github.com/o1-labs/o1js-bindings/pull/187

This PR enables caching of prover and verifier keys in Node.js. The caching interface is kept flexible and will allow a variety of other caching strategies -- for example, bundling snark keys with your web app, or downloading them from some CDN, etc.

**Results** for calling `compile()` on the `simple_zkapp.ts` example (my machine)
* before this PR / on the first compile: **20s**
* on subsequent compiles when no circuit changed: **13s** 

The majority of the remaining 13s is spent in recreating SRS and Lagrange bases (= long lists of elliptic curve points which are the same every time). Caching of those is added in a follow-up PR.

After caching a bunch of `SmartContract`s and `ZkProgram`s, your cache directory will look like this:
image

Note that per contract, several different keys are cached:
* a step prover key (`step-pk`) and verification key (`step-vk`) _for every method_.
* a wrap prover key (`wrap-pk`) and verification key (`wrap-vk`) for the entire contract
* a `.header` file for each cache item. this just contains a unique string which is used to determine whether we can use the cached data

If only one of the methods changes, then only the wrap keys and _one_ of the step keys will be regenerated.

### DIFF
--- a/src/examples/simple_zkapp.ts
+++ b/src/examples/simple_zkapp.ts
@@ -77,7 +77,9 @@ let zkappKey = PrivateKey.random();
 let zkappAddress = zkappKey.toPublicKey();
 
 // a special account that is allowed to pull out half of the zkapp balance, once
-let privilegedKey = PrivateKey.random();
+let privilegedKey = PrivateKey.fromBase58(
+  'EKEeoESE2A41YQnSht9f7mjiKpJSeZ4jnfHXYatYi8xJdYSxWBep'
+);
 let privilegedAddress = privilegedKey.toPublicKey();
 
 let initialBalance = 10_000_000_000;
--- a/src/examples/simple_zkapp.web.ts
+++ b/src/examples/simple_zkapp.web.ts
@@ -1,0 +1,131 @@
+import {
+  Field,
+  state,
+  State,
+  method,
+  UInt64,
+  PrivateKey,
+  SmartContract,
+  Mina,
+  AccountUpdate,
+  Bool,
+  PublicKey,
+} from 'o1js';
+
+const doProofs = true;
+
+const beforeGenesis = UInt64.from(Date.now());
+
+class SimpleZkapp extends SmartContract {
+  @state(Field) x = State<Field>();
+
+  events = { update: Field, payout: UInt64, payoutReceiver: PublicKey };
+
+  @method init() {
+    super.init();
+    this.x.set(initialState);
+  }
+
+  @method update(y: Field): Field {
+    this.account.provedState.assertEquals(Bool(true));
+    this.network.timestamp.assertBetween(beforeGenesis, UInt64.MAXINT());
+    this.emitEvent('update', y);
+    let x = this.x.get();
+    this.x.assertEquals(x);
+    let newX = x.add(y);
+    this.x.set(newX);
+    return newX;
+  }
+
+  /**
+   * This method allows a certain privileged account to claim half of the zkapp balance, but only once
+   * @param caller the privileged account
+   */
+  @method payout(caller: PrivateKey) {
+    this.account.provedState.assertEquals(Bool(true));
+
+    // check that caller is the privileged account
+    let callerAddress = caller.toPublicKey();
+    callerAddress.assertEquals(privilegedAddress);
+
+    // assert that the caller account is new - this way, payout can only happen once
+    let callerAccountUpdate = AccountUpdate.create(callerAddress);
+    callerAccountUpdate.account.isNew.assertEquals(Bool(true));
+    // pay out half of the zkapp balance to the caller
+    let balance = this.account.balance.get();
+    this.account.balance.assertEquals(balance);
+    let halfBalance = balance.div(2);
+    this.send({ to: callerAccountUpdate, amount: halfBalance });
+
+    // emit some events
+    this.emitEvent('payoutReceiver', callerAddress);
+    this.emitEvent('payout', halfBalance);
+  }
+}
+
+let Local = Mina.LocalBlockchain({ proofsEnabled: doProofs });
+Mina.setActiveInstance(Local);
+
+// a test account that pays all the fees, and puts additional funds into the zkapp
+let { privateKey: senderKey, publicKey: sender } = Local.testAccounts[0];
+
+// the zkapp account
+let zkappKey = PrivateKey.random();
+let zkappAddress = zkappKey.toPublicKey();
+
+// a special account that is allowed to pull out half of the zkapp balance, once
+let privilegedKey = PrivateKey.fromBase58(
+  'EKEeoESE2A41YQnSht9f7mjiKpJSeZ4jnfHXYatYi8xJdYSxWBep'
+);
+let privilegedAddress = privilegedKey.toPublicKey();
+
+let initialBalance = 10_000_000_000;
+let initialState = Field(1);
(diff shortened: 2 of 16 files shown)


mitschabaude:
Update: Not ready for review, I'm going to refactor the default caching strategy to not fill the developer's disk so quickly. This also depends on a planned addition to ZkProgram to get a `name` argument, so that we have a persistent identifier for caching

src/lib/proof-system/cache.ts
```diff
@@ -0,0 +1,140 @@
...
+} from '../util/fs.js';
+import { jsEnvironment } from '../../bindings/crypto/bindings/env.js';
+
+export { Cache, CacheHeader, cacheHeaderVersion };
+
+/**
+ * Interface for storing and retrieving values, for caching.
+ * `read()` and `write()` can just throw errors on failure.
+ */
+type Cache = {
```
mitschabaude, line 16: this file contains the main new interface -- `Cache`! plus a default implementation of `Cache` which works in Node.js and uses a common cache directory

src/lib/proof-system/prover-keys.ts
mitschabaude, line 1: this file provides helpers to
* encode and decode all 4 kinds of snark keys to/from bytes
* create a header which is passed to the `Cache` so that it can figure out where and if to read from cache

src/lib/proof_system.ts
```diff
@@ -578,6 +587,33 @@ async function compileProgram({
...
+    function read_(mlHeader) {
+      let header = parseHeader(proofSystemTag.name, methodIntfs, mlHeader);
+      try {
+        let bytes = cache.read(header);
+        if (bytes === undefined) return MlResult.unitError();
+        return MlResult.ok(decodeProverKey(mlHeader, bytes));
+      } catch (e: any) {
+        return MlResult.unitError();
+      }
+    },
```
mitschabaude, line 601: this code connects all the dots:

* our helpers to create cache headers and decode snark keys from bytes
* the user-defined `cache` where we can `read` and `write` from
* the `Pickles.Cache` which Pickles expects

src/lib/util/fs.ts
mitschabaude, line 1: simple wrapper around a few node-only functions, so that we can provide alternative (dummy) implementations on the web

src/lib/zkapp.ts
```diff
@@ -661,7 +662,7 @@ class SmartContract {
    * it so that proofs end up in the original finite field). These are fairly expensive operations, so **expect compiling to take at least 20 seconds**,
    * up to several minutes if your circuit is large or your hardware is not optimal for these operations.
    */
-  static async compile() {
+  static async compile({ cache = Cache.FileSystemDefault } = {}) {
```
mitschabaude, line 665: `SmartContract.compile()` now expects an optional `Cache`. same for `ZkProgram.compile()`

mitschabaude (commented):
some comments to understand what is happening here

src/lib/proof-system/cache.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (dataType === 'string') {
+      let string = readFileSync(resolve(cacheDirectory, persistentId), 'utf8');
+      return new TextEncoder().encode(string);
+    } else {
+      let buffer = readFileSync(resolve(cacheDirectory, persistentId));
+      return new Uint8Array(buffer.buffer);
+    }
+  },
+  write({ persistentId, uniqueId, dataType }, data) {
+    if (jsEnvironment !== 'node') throw Error('file system not available');
```
MartinMinkov, line 109: Same here for error message? Not sure if it's too verbose but I think giving too much info is always better than not enough

MartinMinkov (approved):
Great work! What are your thoughts on documenting the caching strategy we use? Should we have something on our docs website or/and in-depth doc comments?

I think this information is particularly valuable, and it should probably be easily accessible outside of a PR description:

```
a step prover key (step-pk) and verification key (step-vk) for every method.
a wrap prover key (wrap-pk) and verification key (wrap-vk) for the entire contract
a .header file for each cache item. this just contains a unique string which is used to determine whether we can use the cached data
```

src/examples/simple_zkapp.ts
```diff
@@ -77,7 +77,9 @@ let zkappKey = PrivateKey.random();
 let zkappAddress = zkappKey.toPublicKey();
 
 // a special account that is allowed to pull out half of the zkapp balance, once
-let privilegedKey = PrivateKey.random();
+let privilegedKey = PrivateKey.fromBase58(
+  'EKEeoESE2A41YQnSht9f7mjiKpJSeZ4jnfHXYatYi8xJdYSxWBep'
+);
```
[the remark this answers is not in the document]
MartinMinkov: Huh, why is that? :thinking:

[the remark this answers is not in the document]
mitschabaude: Constants that are used in constraints are baked into the circuit digest as well as prover and verifier keys. This example was using `PrivateKey.random()` so the circuit was different every time. Since we compare the circuit digest to decide whether or not we can use a cached key, the answer was "no" every time.

src/lib/proof-system/cache.ts
mitschabaude: this error message will never be read by a user! when `read()` or `write()` throw an error this is just treated as "caching wasn't successful", i.e. it fails silently. this makes it easier to write the caching logic (but arguably a bit harder to debug it..)

src/examples/simple_zkapp.web.ts
```diff
@@ -0,0 +1,131 @@
...
+await tx.sign([senderKey]).send();
+
+console.log('payout');
+tx = await Mina.transaction(sender, () => {
+  AccountUpdate.fundNewAccount(sender);
+  zkapp.payout(privilegedKey);
+});
+await tx.prove();
+await tx.sign([senderKey]).send();
+sender;
```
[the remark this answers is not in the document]
mitschabaude: well I've created versions of this file a couple of times already, and it's very useful for me to have a default example that works on the web (the other simple zkapp example doesn't, because it uses the file system to store performance timings)

I agree that it's not pretty to have many similar versions of the same file 🤔 maybe if I move the contract to its own file it's better?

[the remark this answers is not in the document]
MartinMinkov: Oh no, sorry, I just meant the line on 128, specifically the `sender;` statement

src/examples/simple_zkapp.ts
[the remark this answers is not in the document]
MartinMinkov: Ah, yes, pretty straightforward. That makes sense, thank you!

src/examples/simple_zkapp.web.ts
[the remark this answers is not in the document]
mitschabaude: aah ok lol

mitschabaude:
> I think this information is particularly valuable, and it should probably be easily accessible outside of a PR description:

Addressed in the SRS caching PR!

Issue #1185, opened 2023-10-19
Expose dummyBase64Proof()


maht0rz:
Would it be possible to expose `dummyBase64Proof()`, or is there a specific reason not to?

mitschabaude:
Could be put as a static method on the `Proof` class maybe

maht0rz:
Just to add, we also need `Pickles` exposed to do `Pickles.proofOfBase64(...)`, i'll make a PR!

mitschabaude:
made a PR: https://github.com/o1-labs/o1js/pull/1188

PR #1188, opened 2023-10-20
Expose dummy proofs

mitschabaude:
exposes `Proof.dummy()`

you need dummy proofs to write ZkProgram methods that cover the base case and the recursive case at the same time. if it's the case case you pass a dummy proof and don't verify it. this, empirically, can significantly reduce the proof time of a recursive chain, without negatively impacting the base case efficiency

closes https://github.com/o1-labs/o1js/issues/1185
bindings: https://github.com/o1-labs/o1js-bindings/pull/188

### DIFF
--- a/src/lib/proof_system.ts
+++ b/src/lib/proof_system.ts
@@ -143,9 +143,10 @@ class Proof<Input, Output> {
   static async dummy<Input, OutPut>(
     publicInput: Input,
     publicOutput: OutPut,
-    maxProofsVerified: 0 | 1 | 2
+    maxProofsVerified: 0 | 1 | 2,
+    domainLog2: number = 14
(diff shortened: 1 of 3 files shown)


mrmr1993 (approved):

src/lib/proof_system.ts
```diff
@@ -139,6 +139,21 @@ class Proof<Input, Output> {
     this.proof = proof; // TODO optionally convert from string?
     this.maxProofsVerified = maxProofsVerified;
   }
+
```
barriebyron, line 142: did you want to add doc comments here?

mitschabaude: good idea :D

Trivo25 (approved):

PR #1194, opened 2023-10-23
Add left shift and right shift gadgets to o1js

MartinMinkov:
# Summary
This PR adds the implementation for left shift and right shift gates to the existing Gadgets namespace. Bitwise shifting is an operation that moves the bits of a binary number to the left or right. Unlike rotation, the bits that "fall off" at the end are discarded and the vacant positions are filled with zeros.

The implementations is based on the original OCaml implementation which is specified here for [left shift](https://github.com/MinaProtocol/mina/blob/4b2e023eec39718813fbacc38661558a3193a433/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L171) and [right shift](https://github.com/MinaProtocol/mina/blob/4b2e023eec39718813fbacc38661558a3193a433/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L188)

In our shift implementations, we handle the constant case by using the added functionality in the bindings, and for the prover case, we specify the implementation in this PR.

# Tested
This PR adds unit tests, and verification key tests for the shift gates.

:link: bindings: https://github.com/o1-labs/o1js-bindings/pull/191


src/lib/gadgets/rot.ts
```diff
@@ -27,6 +24,30 @@ function rot(field: Field, bits: number, direction: 'left' | 'right' = 'left') {
...
 
+function rightShift(field: Field, bits: number) {
+  // Check that the shift bits are in range
+  checkBitsRange(bits);
+
+  if (field.isConstant()) {
+    checkMaxBits(field);
+    return new Field(Fp.rightShift(field.toBigInt(), bits));
+  }
+  const [, excess] = rotate(field, bits, 'right');
```
querolita, line 35: Why is there only one `,` whereas there's three outputs returned by rotate?

src/lib/gadgets/gadgets.ts
```diff
@@ -62,4 +62,58 @@ const Gadgets = {
...
+  /**
+   * Performs a right shift operation on the provided {@link Field} element.
+   * This is similar to the `>>` shift operation in JavaScript, where bits are moved to the right.
+   * The `rightShift` function utilizes the rotation method internally to implement this operation.
+   *
+   * **Note:** You cannot shift {@link Field} elements that exceed 64 bits.
+   * For elements that exceed 64 bits, this operation will fail.
+   *
+   * @param field {@link Field} element to shift.
+   * @param bits Amount of bits to shift the {@link Field} element to the right.
```
querolita, line 102: Specify that this amount should be between 0 and 64 (or else rotation will fail). Meaning, this is not just any shift, but shift up to 64 bits in length.

src/lib/gadgets/gadgets.ts
```diff
@@ -62,4 +62,58 @@ const Gadgets = {
...
+   *
+   * @throws Throws an error if the input value exceeds 64 bits.
+   *
+   * @example
```
querolita, line 106: Examples would be clearer if given in binary or hexadecimal format

src/examples/regression_test.json
```diff
@@ -171,6 +171,14 @@
       "rot": {
         "rows": 17,
         "digest": "916f4017a60f48d56d487c6869919b9c"
+      },
+      "leftShift": {
+        "rows": 9,
```
querolita, line 176: Okay so this is 4 + 4 + 1 nice

querolita (approved):
Looks good to me! Just a nit to clarify some things in the docs

src/lib/gadgets/gadgets.ts
```diff
@@ -62,4 +62,58 @@ const Gadgets = {
   rot(field: Field, bits: number, direction: 'left' | 'right' = 'left') {
     return rot(field, bits, direction);
   },
+
+  /**
+   * Performs a left shift operation on the provided {@link Field} element.
+   * This is akin to the `<<` shift operation in JavaScript, where bits are moved to the left.
```
Trivo25, line 68: In light of https://github.com/o1-labs/o1js-bindings/pull/191/files#r1370549886 we should probably slightly change this part

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -116,3 +117,115 @@ test(
...
+
+function testShift(
+  field: Field,
+  bits: number,
+  mode: 'left' | 'right',
+  result: Field
+) {
+  Provable.runAndCheck(() => {
+    let w = Provable.witness(Field, () => field);
+    let r = Provable.witness(Field, () => result);
```
Trivo25, line 133: nit: the result doesnt need to be a witness

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -116,3 +117,115 @@ test(
...
+    let output = Provable.if(
```
Trivo25, line 134: nit: we can just use a normal JS `if` statement here

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -116,3 +117,115 @@ test(
...
+      },
+    },
+  },
+});
+
+await LS.compile();
+await equivalentAsync({ from: [maybeUint64], to: boolean }, { runs: 3 })(
+  (x) => {
+    if (x >= 1n << 64n) throw Error('expected 64 bits');
+    return true;
```
Trivo25, line 161: Instead of just returning true here, we can also make the `ZkProgram` return the output of the shift operation and make sure that its equivalent to the out of circuit variant not just in generating a valid proof and catching errors but also the correctness of the actual output

src/lib/gadgets/gadgets.unit-test.ts
```diff
@@ -116,3 +117,115 @@ test(
...
+      },
+    },
+  },
+});
+
+await RS.compile();
+await equivalentAsync({ from: [maybeUint64], to: boolean }, { runs: 3 })(
+  (x) => {
+    if (x >= 1n << 64n) throw Error('expected 64 bits');
+    return true;
```
Trivo25, line 184: see comment above

Trivo25 (approved):
Good job!

src/lib/gadgets/gadgets.unit-test.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1194/commits/b7e427c24a0396556450836a18302814a511ee74

src/lib/gadgets/gadgets.unit-test.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1194/commits/b7e427c24a0396556450836a18302814a511ee74

src/lib/gadgets/gadgets.unit-test.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1194/commits/b7e427c24a0396556450836a18302814a511ee74

src/lib/gadgets/gadgets.unit-test.ts
MartinMinkov: Fixed in https://github.com/o1-labs/o1js/pull/1194/commits/b7e427c24a0396556450836a18302814a511ee74****

src/lib/gadgets/gadgets.ts
MartinMinkov: Addressed this in https://github.com/o1-labs/o1js/pull/1194/commits/1b1663b6351a9499d0d074bffa6b0e615d4a7aa8

src/lib/gadgets/gadgets.ts
MartinMinkov: Also addressed in https://github.com/o1-labs/o1js/pull/1194/commits/e74b623cbb9a49699e78973d3511774d1d598eec

src/lib/gadgets/gadgets.ts
MartinMinkov: Addressed in https://github.com/o1-labs/o1js/pull/1194/commits/575920dbd2a4e50564ca422de7e7da4cc9b57542, how does it sound now? :D

Trivo25: sounds good!

ymekuria (approved):

src/lib/gadgets/gadgets.ts
```diff
@@ -128,11 +194,12 @@ const Gadgets = {
    * **Note:** Both {@link Field} elements need to fit into `2^paddedLength - 1`. Otherwise, an error is thrown and no proof can be generated.
    * For example, with `length = 2` (`paddedLength = 16`), `and()` will fail for any input that is larger than `2**16`.
    *
+   * @example
    * ```typescript
    * let a = Field(3);    // ... 000011
    * let b = Field(5);    // ... 000101
    *
-   * let c = and(a, b, 2);    // ... 000001
+   * let c = Gadgets.and(a, b, 2);    // ... 000001
```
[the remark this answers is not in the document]
MartinMinkov: I'm not sure what you mean exactly here :/ Do you mean adding the `Gadgets` prefix? If that's the case, I think so! I think the rest of them do right now.

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -52,27 +70,20 @@ let uint = (length: number) => fieldWithRng(Random.biguint(length));
...
+    (x) => Fp.rot(x, 12, 'left'),
+    (x) => Gadgets.rotate(x, 12, 'left')
+  );
+  equivalent({ from: [uint(length)], to: field })(
+    (x) => Fp.leftShift(x, 12),
+    (x) => Gadgets.leftShift(x, 12)
+  );
+  equivalent({ from: [uint(length)], to: field })(
+    (x) => Fp.rightShift(x, 12),
+    (x) => Gadgets.rightShift(x, 12)
```
Trivo25, line 84: nice, very clean!

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,72 @@ const Gadgets = {
...
+   * Performs a left shift operation on the provided {@link Field} element.
+   * This operation is similar to the `<<` shift operation in JavaScript,
+   * where bits are shifted to the left, and the overflowing bits are discarded.
+   *
+   * It’s important to note that these operations are performed considering the big-endian 64-bit representation of the number,
+   * where the most significant (64th) bit is on the left end and the least significant bit is on the right end.
+   *
+   * **Important:** The gadgets assumes that its input is at most 64 bits in size.
+   *
+   * If the input exceeds 64 bits, the gadget is invalid and does not prove correct execution of the shift.
```
Trivo25, line 120: If the input exceeds 64 bit, does proving fail or does the gadget provide invalid results?

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,72 @@ const Gadgets = {
...
+   * Performs a right shift operation on the provided {@link Field} element.
+   * This is similar to the `>>` shift operation in JavaScript, where bits are moved to the right.
+   * The `rightShift` function utilizes the rotation method internally to implement this operation.
+   *
+   * * It’s important to note that these operations are performed considering the big-endian 64-bit representation of the number,
+   * where the most significant (64th) bit is on the left end and the least significant bit is on the right end.
+   *
+   * **Important:** The gadgets assumes that its input is at most 64 bits in size.
+   *
+   * If the input exceeds 64 bits, the gadget is invalid and does not prove correct execution of the shift.
```
Trivo25, line 153: see comment above, just for clarification

Trivo25 (approved):
Great job!

src/lib/gadgets/gadgets.ts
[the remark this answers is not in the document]
ymekuria: I wasn't clear in my comment. Yes I was referring to the Gadgets prefix.

src/lib/gadgets/gadgets.ts
MartinMinkov: I think the former. I just tried it as a sanity check and I get a prover error:
`Error: the proof could not be constructed: rest of division by vanishing polynomial`

mitschabaude (approved):
Looks good to me! Changelog update would be nice (and while you're at it -- maybe you can also add the one for `Gadgets.and()` which we forgot)

src/lib/gadgets/gadgets.ts
mitschabaude: But that's because you're not maliciously changing the witness!! If you'd modify the witness blocks, and could also control the inputs (and the circuit doesn't prevent them from exceeding 64 bits) you could create a proof and get the gadget to return invalid results

mitschabaude: Anyway, the updated wording also reads good to me 👍🏻

PR #1196, opened 2023-10-24
Added CODEOWNERS file structure to help with x-team reviews

joseandro:


CODEOWNERS
```diff
@@ -0,0 +1 @@
+/src/lib/gadgets @o1-labs/crypto-eng-reviewers
```
mitschabaude, line 1:
```suggestion
/src/lib/gadgets @o1-labs/crypto-eng-reviewers @mitschabaude

```

mitschabaude (changes requested):
can we add my name there? I'd like to be able to make the call whether we need crypto review, and unblock urgent changes if not

mitschabaude (approved):

PR #1198, opened 2023-10-25
Implement NOT gadget

### LINKED ISSUE #1141: Port NOT gadget to TypeScript

ymekuria:
# Description 
Closes #1141 

This PR adds the implementation for a `NOT` gate to the existing `Gadgets` namespace. A bitwise `NOT` is an operation that returns `1` in each bit position if the corresponding bit of the operand is `0`, and returns `0` if the corresponding bit of the operand is `1`.

The original OCaml implementation is [here](https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L510). 

🔗 bindings: https://github.com/o1-labs/o1js-bindings/pull/194


MartinMinkov:
Do you mind merging `main` into your branch before reviewing? There were quite a few files that moved around, so it would really help to have that laid out before trying to review and then resolving conflicts.

src/lib/gadgets/bitwise.ts
```diff
@@ -3,7 +3,44 @@ import { Field as Fp } from '../../provable/field-bigint.js';
...
+    let max = 1n << BigInt(padLength);
+    assert(
+      a.toBigInt() < max,
+      `${a.toBigInt()} does not fit into ${padLength} bits`
+    );
+    return new Field(Fp.not(a.toBigInt()));
+  }
+
+  // create a bitmask with all ones
+  let allOnesF = new Field(BigInt(2 ** length - 1));
```
Trivo25, line 32: (not so) fun fact: these all return different values

```ts
length = 64
let a = 2 ** length - 1;
console.log(a)
console.log(a.toString(2))

let b = 2n ** BigInt(length) - 1n;
console.log(b)
console.log(b.toString(2))

let c = BigInt(2 ** 64 - 1);
console.log(c)
console.log(c.toString(2))
```

```sh
18446744073709552000 
"10000000000000000000000000000000000000000000000000000000000000000" 

18446744073709551615 
"1111111111111111111111111111111111111111111111111111111111111111" 

8446744073709551616 
"10000000000000000000000000000000000000000000000000000000000000000" 

```

so I think we should go with `2n ** BigInt(length) - 1n`

src/lib/gadgets/bitwise.ts
```diff
@@ -10,7 +10,44 @@ import {
...
+  // create a bitmask with all ones
+  let allOnesF = new Field(2n ** BigInt(length) - 1n);
+
+  let allOnes = Provable.witness(Field, () => {
+    return allOnesF;
+  });
+
+  allOnesF.assertEquals(allOnes);
+
+  let notOutput = xor(a, allOnes, length);
```
Trivo25, line 47: I think we should actually implement the unchecked version `bnot_unchecked`, it saves a good amount of constraints (`bnot_unchecked` has only 1 constraint, no matter how large the bit length gets whereas `bnot_checked` increases in constraints with the length of the bit size because it generates a bunch of XOR gates). 

This is actually only a simple change - instead of calling into `xor`, instead its a simple subtraction

```ocaml
  (* Negating is equivalent to subtracting with all one word *)
  (* [2^len - 1] - input = not (input) *)
  Generic.sub (module Circuit) all_ones_var input
```

This could also be a conditional check wdyt @mitschabaude @MartinMinkov

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -22,6 +22,12 @@ let Bitwise = ZkProgram({
         return Gadgets.xor(a, b, 64);
       },
     },
+    not: {
+      privateInputs: [Field],
+      method(a: Field) {
+        return Gadgets.not(a, 64);
+      },
+    },
```
Trivo25, line 30: I think we should actually test all values up to the maximum, 255 bit. I didnt do that for XOR, so would be good if you could change that part as well!

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -79,6 +85,24 @@ await equivalentAsync(
...
+// not
+[2, 4, 8, 16, 32, 64, 128].forEach((length) => {
+  equivalent({ from: [uint(length), uint(length)], to: field })(Fp.not, (x) =>
+    Gadgets.not(x, length)
+  );
+});
+
+await equivalentAsync({ from: [maybeUint64], to: field }, { runs: 3 })(
+  (x) => {
+    if (x >= 2n ** 64n) throw Error('Does not fit into 64 bits');
```
Trivo25, line 97: see comment above

src/lib/gadgets/gadgets.ts
```diff
@@ -104,4 +104,33 @@ const Gadgets = {
   xor(a: Field, b: Field, length: number) {
     return xor(a, b, length);
   },
+
+  /**
+   * Bitwise NOT gate on {@link Field} elements. Equivalent to the [bitwise
+   * NOT `~` operator in JavaScript](https://developer.mozilla.org/en-US/docs/
```
Trivo25, line 110: we should add a small disclaimer here that it behaves slightly different than the JavaScript operator - see the bindings PR for an explanation

Trivo25 (commented):
Nice! Left some comments for discussion

src/lib/gadgets/bitwise.ts
MartinMinkov: Just to clarify for myself, why wouldn't someone want to use the version with a checked length? Is the idea that the length is checked elsewhere and then does not have to be redone in the NOT gate? If that's the case, then we are trading off API safety/simplicity with low-level power usage. This seems fine to me! I would prefer adding a conditional check/flag over adding another unchecked version.

src/lib/gadgets/gadgets.ts
```diff
@@ -104,4 +104,33 @@ const Gadgets = {
...
+   * The `length` parameter lets you define how many bits to NOT. It
+   * defaults to the size of the field in bits ({@link Fp.sizeInBits}).
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional * * *
+   *   constraints.
+   *
+   * @example
+   * ```ts
+   * let a = Field(5);    // ... 101
+   * let b = not(5,3);    // ... 010
```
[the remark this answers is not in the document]
MartinMinkov: These should probably use binary inputs just to make it very clear, like the other examples.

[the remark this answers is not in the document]
Trivo25: Oh yes, agree!

src/lib/gadgets/bitwise.ts
ymekuria: I slightly prefer adding a flag over another unchecked version. The api could be something like this.
```ts
not(a: Field, length: number, checked: boolean = true)
```
We can set the default to checked or even unchecked.

src/lib/gadgets/bitwise.ts
```diff
@@ -10,7 +10,44 @@ import {
...
+  }
+
+  // create a bitmask with all ones
+  let allOnesF = new Field(2n ** BigInt(length) - 1n);
+
+  let allOnes = Provable.witness(Field, () => {
+    return allOnesF;
+  });
+
+  allOnesF.assertEquals(allOnes);
```
mitschabaude, line 45: I was wondering why we can't just pass in `allOnesF` to `xor()`. I assume this was done because otherwise when passing the constant to the xor16 gate, snarky would add a generic gate to resolve the constant, and that would break the xor chain.

I think it's actually a major gotcha that should be handled _in `xor()`_. I.e. `xor()` should do something like this when one of its inputs is constant. (Assuming it's really a problem)

@Trivo25 you wrote xor, do you agree?

src/lib/gadgets/bitwise.ts
mitschabaude: @MartinMinkov we have many "data types with intrinsic constraints", like `Bool` or `UInt64` which always get constraints attached when introduced in a circuit. We _never_ additionally check those constraints in operations. For example, `Bool.and()` doesn't check that its inputs are booleans, `UInt64.add()` doesn't check that its inputs fit in 64 bits and so on. The rule is that you should constraint outputs, but not inputs. IMO doing the checked version by default would violate that rule.

@ymekuria I therefore think the _unchecked_ version should be the default. I'm in favor of the flag if we can find any case where the checked version is useful, i.e. better than just doing `rangeCheck64(x); not(x)`.

@querolita maybe you know why we had two versions here? Is there a case where the checked version is better?

mitschabaude (commented):
Also some comments from me!

src/lib/gadgets/bitwise.ts
Trivo25: I am not sure, to be honest

I checked that if one of the inputs is a constant that the chain doesnt break, and it seems to work

```ts
let cs = Provable.constraintSystem(() => {
  let variable = Provable.witness(Field, () => Field(5));
  let constant = Field(4);
  Gadgets.xor(variable, constant, 128);
});

console.log(cs);
```

```
  gates: [
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Xor16', wires: [Array], coeffs: [] },
    { type: 'Zero', wires: [Array], coeffs: [] },
    { type: 'Generic', wires: [Array], coeffs: [Array] }
  ],
```

Trivo25: perhaps @querolita can explain why we do this check? What are your thoughts?

mitschabaude: Right - because the constant resolving already happens _before_ the first xor16 gate, and after that it's only witnesses, no constants

Trivo25:
when you are ready, can you fix the bindings merge conflict so the CI can run through? thanks!

src/lib/gadgets/bitwise.ts
Trivo25: Do you think there is still an issue? If so, I think I am having trouble understanding :D

src/lib/gadgets/bitwise.ts
```diff
@@ -11,7 +11,45 @@ import {
...
+
+  let notChecked = xor(a, allOnes, length);
+  let notUnchecked = allOnes.sub(a);
+
+  return checked ? notChecked : notUnchecked;
```
Trivo25, line 51: we should wrap the actual logic in an if statement, rather than the return values using a ternary operation

```ts
if(checked) {
   // do checked, return checked
} else {
   // do unchecked, return unchecked 
}

```

src/examples/primitive_constraint_system.ts
```diff
@@ -79,6 +79,14 @@ const BitwiseMock = {
     Gadgets.xor(a, b, 48);
     Gadgets.xor(a, b, 64);
   },
+
+  not() {
```
Trivo25, line 83: we should add the unchecked variant as well, either as its own method or directly in the `not` method

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -19,7 +19,13 @@ let Bitwise = ZkProgram({
...
       privateInputs: [Field, Field],
       method(a: Field, b: Field) {
-        return Gadgets.xor(a, b, 64);
+        return Gadgets.xor(a, b, 255);
+      },
+    },
+    not: {
+      privateInputs: [Field],
+      method(a: Field) {
+        return Gadgets.not(a, 255);
```
Trivo25, line 28: we should add a checked/unchecked version here as well

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -50,6 +56,10 @@ let uint = (length: number) => fieldWithRng(Random.biguint(length));
     (x, y) => x & y,
     (x, y) => Gadgets.and(x, y, length)
   );
+  equivalent({ from: [uint(length), uint(length)], to: field })(
```
Trivo25, line 59: see comment above

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,49 @@ const Gadgets = {
...
+   * for a single negation. If the 'checked' parameter is set to 'false' NOT is
+   * implementad as a subtraction of the input from the all one bitmask. This
+   * implementation is returned by default if no 'checked' parameter is provided.
+   *
+   *You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#not)
+   *
+   * @example
+   * ```ts
+   * let a = Field(0b0101);
+   * let b = not(a,4,false);    // not-ing 4 bits
```
Trivo25, line 139: recommendation:

 `let b = Gadget.not(a, 4, false); // not-ing 4 bits`

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,49 @@ const Gadgets = {
...
+   *
+   * b.assertEquals(0b1010);
+   * ```
+   *
+   * @param a - The value to apply NOT to.
+   * @param length - The number of bits to be considered for the NOT operation.
+   * @param checked - Optional Boolean to determine if the checked or unchecked not implementation is used.
```
Trivo25, line 146: We should clarify what this means for the developer and how to utilize the two variants

src/lib/gadgets/gadgets.ts
```diff
@@ -117,9 +160,9 @@ const Gadgets = {
    * Where:\
    * `a + b = sum`\
    * `a ^ b = xor`\
-   * `a & b = and`
+   * `a & b = and`You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and)
```
Trivo25, line 163: space missing after `and`

src/lib/gadgets/bitwise.ts
mitschabaude: Seems like there's no issue with just passing a constant directly 👍🏻

mitschabaude (approved):

src/lib/gadgets/bitwise.ts
Trivo25: ahh amazing :D

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   * corresponding bit of the operand is `0`, and returning `0` if the
+   * corresponding bit of the operand is `1`.
+   *
+   * The `length` parameter lets you define how many bits to NOT.
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional constraints.
+   *
+   *
+   * NOT is implemented in two different ways. If the `checked` parameter is set to `true`
+   * the {@link Gadgets.xor} gadget is reusesd the with a second argument to be an
```
MartinMinkov, line 128:
```suggestion
   * the {@link Gadgets.xor} gadget is reused with a second argument to be an
```

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   * ```ts
+   * // not-ing 4 bits with the unchecked version
+   * let a = Field(0b0101);
+   * let b = gadgets.not(a,4,false);
+   *
+   * b.assertEquals(0b1010);
+   *
+   * // not-ing 4 bits with the checked version utilizing the xor gadget
+   * let a = Field(0b0101);
+   * let b = gadgets.not(a,4,true);
```
MartinMinkov, line 146:
```suggestion
   * let b = Gadgets.not(a,4,true);
```

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   *
+   * b.assertEquals(0b1010);
+   * ```
+   *
+   * @param a - The value to apply NOT to.
+   * @param length - The number of bits to be considered for the NOT operation.
+   * @param checked - Optional Boolean to determine if the checked or unchecked not implementation is used. If it is set to `true` the {@link Gadgets.xor} gadget is reusesd.
```
MartinMinkov, line 153:
```suggestion
   * @param checked - Optional boolean to determine if the checked or unchecked not implementation is used. If it is set to `true` the {@link Gadgets.xor} gadget is reused.
```

src/lib/gadgets/gadgets.ts
```diff
@@ -119,7 +170,7 @@ const Gadgets = {
    * `a ^ b = xor`\
    * `a & b = and`
    *
-   * You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and)
+   *You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and)
```
MartinMinkov, line 173:
```suggestion
   * You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and)
```

MartinMinkov:
Adding links to the mina book is a great idea! Do you mind adding the same for other gadgets in this PR here as well?

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   * implementad as a subtraction of the input from the all one bitmask. This
+   * implementation is returned by default if no `checked` parameter is provided.
+   *
+   *You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#not)
+   *
+   * @example
+   * ```ts
+   * // not-ing 4 bits with the unchecked version
+   * let a = Field(0b0101);
+   * let b = gadgets.not(a,4,false);
```
MartinMinkov, line 140:
```suggestion
   * let b = Gadgets.not(a, 4, false);
```

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
     return xor(a, b, length);
   },
+
+  /**
+   * Bitwise NOT gate on {@link Field} elements. Similar to the [bitwise
+   * NOT `~` operator in JavaScript](https://developer.mozilla.org/en-US/docs/
+   * Web/JavaScript/Reference/Operators/Bitwise_NOT).
+   *
+   * **Note:** The NOT gate only operates over the amount
+   * of bits specified by the 'length' paramenter.
```
MartinMinkov, line 116:
```suggestion
   * of bits specified by the 'length' parameter.
```

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   * The `length` parameter lets you define how many bits to NOT.
+   *
+   * **Note:** Specifying a larger `length` parameter adds additional constraints.
+   *
+   *
+   * NOT is implemented in two different ways. If the `checked` parameter is set to `true`
+   * the {@link Gadgets.xor} gadget is reusesd the with a second argument to be an
+   * all one bitmask the same length. This approach needs as many rows as an XOR would need
+   * for a single negation. If the `checked` parameter is set to `false`, NOT is
+   * implementad as a subtraction of the input from the all one bitmask. This
```
MartinMinkov, line 131:
```suggestion
   * implemented as a subtraction of the input from the all one bitmask. This
```

src/lib/gadgets/gadgets.ts
```diff
@@ -106,6 +106,57 @@ const Gadgets = {
...
+   * let a = Field(0b0101);
+   * let b = gadgets.not(a,4,true);
+   *
+   * b.assertEquals(0b1010);
+   * ```
+   *
+   * @param a - The value to apply NOT to.
+   * @param length - The number of bits to be considered for the NOT operation.
+   * @param checked - Optional Boolean to determine if the checked or unchecked not implementation is used. If it is set to `true` the {@link Gadgets.xor} gadget is reusesd.
+   * If it is set to `false`, NOT is implementad as a subtraction of the input from the all one bitmask. It is set to `false` by default if no parameter is provided.
```
MartinMinkov, line 154:
```suggestion
   * If it is set to `false`, NOT is implemented as a subtraction of the input from the all one bitmask. It is set to `false` by default if no parameter is provided.
```

src/lib/gadgets/bitwise.ts
```diff
@@ -11,7 +11,46 @@ import {
...
-export { xor, and, rotate };
+export { xor, not, and, rotate };
+
+function not(a: Field, length: number, checked: boolean = false) {
+  // check that input length is positive
+  assert(length > 0, `Input length needs to be positive values.`);
+
+  // Check that length does not exceed maximum field size in bits
+  assert(
+    length <= Field.sizeInBits(),
```
Trivo25, line 22: this needs to be `length < Field.sizeInBits()` - see https://github.com/MinaProtocol/mina/blob/develop/src/lib/crypto/kimchi_backend/gadgets/bitwise.ml#L514

This is also the reason why the unit tests fail occasionally if the inpuits are greater than `2n ** 254n`

ymekuria:
> Adding links to the mina book is a great idea! Do you mind adding the same for other gadgets in this PR here as well?

Yeah of course! Thanks for the comments @MartinMinkov

src/lib/gadgets/bitwise.ts
ymekuria: Great catch @Trivo25! I was trying to figure out why the test occasionally failed at very large inputs.

src/lib/gadgets/gadgets.ts
```diff
@@ -195,11 +250,12 @@ const Gadgets = {
    * For example, with `length = 2` (`paddedLength = 16`), `and()` will fail for any input that is larger than `2**16`.
    *
    * @example
+   * @example
```
Trivo25, line 253: double `@example`

src/lib/gadgets/gadgets.ts
```diff
@@ -195,11 +250,12 @@ const Gadgets = {
...
    * ```typescript
    * let a = Field(3);    // ... 000011
    * let b = Field(5);    // ... 000101
    *
-   * let c = Gadgets.and(a, b, 2);    // ... 000001
+   * let c = Gadgets.Gadgets.and(a, b, 2);    // ... 000001
```
Trivo25, line 258: double `Gadgets`

src/examples/regression_test.json
```diff
@@ -176,6 +176,22 @@
...
+      "leftShift": {
+        "rows": 7,
+        "digest": "66de39ad3dd5807f760341ec85a6cc41"
+      },
+      "rightShift": {
+        "rows": 7,
+        "digest": "a32264f2d4c3092f30d600fa9506385b"
+      },
       "leftShift": {
         "rows": 7,
```
Trivo25, line 196: looks like the entires are doubled for some reason, we should be good by doing a clean reset of this file i think

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -90,8 +112,8 @@ await equivalentAsync(
   { runs: 3 }
 )(
   (x, y) => {
-    if (x >= 2n ** 64n || y >= 2n ** 64n)
-      throw Error('Does not fit into 64 bits');
+    if (x >= 2n ** 255n || y >= 2n ** 255n)
```
Trivo25, line 115: should be `254n` now

PR #1199, opened 2023-10-26
Update wasm bindgen to 0.2.87

rbonichon:
See https://github.com/o1-labs/o1js-bindings/pull/193


mitschabaude:
Hey you can fix the verification key regression test by regenerating the verfiication keys, with this command:
```
npm run dump-vks
```
(assuming that a change to the vks is expected -- because of the chunking changes I assume?)

mitschabaude (approved):

dannywillems (approved):

PR #1200, opened 2023-10-26
Add ZkProgram name as required argument

mitschabaude:
Adds `name` argument to ZkProgram.

* `name` is required in the new non-experimental API, but still optional in the old `Experimental.ZkProgram`
* the name is an arbitrary string
* the name will be used for caching. when using the name in filenames, we'll need some way to remove unsupported letters. the caching PR will handle this.
* all examples updated to non-experimental API and to having a name

### DIFF
--- a/src/lib/proof_system.ts
+++ b/src/lib/proof_system.ts
@@ -35,6 +35,7 @@ export {
   SelfProof,
   JsonProof,
   ZkProgram,
+  ExperimentalZkProgram,
   verify,
   Empty,
   Undefined,
@@ -240,6 +241,7 @@ function ZkProgram<
   }
 >(
   config: StatementType & {
+    name: string;
     methods: {
       [I in keyof Types]: Method<
         InferProvableOrUndefined<Get<StatementType, 'publicInput'>>,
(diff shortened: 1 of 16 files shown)


src/lib/proof_system.ts
```diff
@@ -273,7 +275,7 @@ function ZkProgram<
   let publicInputType: ProvablePure<any> = config.publicInput! ?? Undefined;
   let publicOutputType: ProvablePure<any> = config.publicOutput! ?? Void;
 
-  let selfTag = { name: `Program${i++}` };
+  let selfTag = { name: config.name };
```
MartinMinkov, line 278: Do we want to sanitize this input in any way since this will be written to disk? Some thoughts that come to mind:
1. Safe encoding (making sure only alphanumeric characters and separators are specified)
2. String length

These are just some thoughts, it's entirely up to you if you find them appropriate :D

MartinMinkov (approved):
Just had one question, but approving nonetheless :D

src/lib/proof_system.ts
mitschabaude: Yes very appropriate! I'll figure it out in the caching PR, maybe serialize them before using as filename but leave the namse string on the `ZkProgram` in original form

barriebyron:
oh thanks to Maki for noticing! 
@garwalsh We missed corresponding doc updates for the ZkProgram API, tracking in https://app.zenhub.com/workspaces/zkapps-product-eng-6130fedb3b0fc600123d8796/issues/gh/o1-labs/docs2/711

PR #1202, opened 2023-10-27
Some doccomment fixes to bitwise gadgets

mitschabaude:
This fixes some doccomments to show the same usage of gadgets that is relevant for a user, i.e. `Gadgets.<method>()`

Also clarifies the bit length of 64 w.r.t which we rotate in the `rotate()` gadget.

### DIFF
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -21,10 +21,10 @@ const Gadgets = {
    * @example
    * ```ts
    * const x = Provable.witness(Field, () => Field(12345678n));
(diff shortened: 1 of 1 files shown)


Trivo25:
Since youre at it, would you mind changing the XOR gadget example in the doc comment to use binary representation as well? I think that would be more helpful

```ts
let a = Provable.witness(Field, () => Field(0b0101));
let b = Provable.witness(Field, () => Field(0b0011));

let c = Gadgets.xor(a, b, 4); // comparing first 4 bits
c.assertEquals(0b0110);
```

PR #1204, opened 2023-10-30
Use `/usr/bin/env bash` instead of `/bin/bash` in shell scripts

rbonichon:
The full description is, I believe, in the title :-)


rbonichon:
I assume I'll need to do similar PRs for other important branches (berkeley, main, ...), right?

dannywillems (approved):

mitschabaude:
Thanks for this change!

PR #1208, opened 2023-10-30
add txnVersion to verification key permission

Trivo25:
o1js implementation of the protocol change described by this [RFC](https://github.com/MinaProtocol/mina/blob/9577ad689a8e4d4f97e1d0fc3d26e20219f4abd1/rfcs/0051-verification-key-permissions.md)

This change enables developers to flag contracts as immutable but not prevent them from deadlocking with future protocol changes. In addition to the permission type a developer wants to specify for `setVerificationKey`, it is now possible to also specify a `txnVersion: UInt32` which will be stored on-chain as part of the zkApp accounts permission field. If the `txnVersion` matches the current protocol version, permissions of type `Proof` and `Impossible` act normal. If the `txnVersion` is older than the current protocol version, the permission falls back to `Signature` automatically, so that now broken smart contracts can be upgraded using a Signatures.

TODO: docs https://github.com/o1-labs/docs2/issues/750

companion of https://github.com/MinaProtocol/mina/pull/14407

bindings https://github.com/o1-labs/o1js-bindings/pull/195


src/examples/regression_test.json
```diff
@@ -1,139 +1,139 @@
...
       },
       "candidateRegistration": {
         "rows": 1260,
-        "digest": "abf93b4d926a8743bf622da4ee0f88b2"
+        "digest": "6e9e689b9d1bd2eedcc323c842cf8c87"
       },
       "approveRegistrations": {
-        "rows": 1146,
-        "digest": "197ce95e1895f940278034d20ab24274"
+        "rows": 1147,
```
Trivo25, line 14: I am actually a bit surprised by this change - this change shouldn't impact this as far as I understand, but maybe something on the Mina branch changed how snarky works :thinking:

Trivo25: Thats right! Didnt think about that at all

PR #1209, opened 2023-10-31
Readme and repo grooming, release 0.14

mitschabaude:
A few small things lumped together

* `README-dev.md` was outdated in two places, this updates it
* Move example code to a better place
* Bumps version and changelog

bindings: https://github.com/o1-labs/o1js-bindings/pull/196


src/examples/plain-html/index.html
mitschabaude, line 1: these files are moved here from the bindings repo. they used to be in `src/bindings/compiled/web_bindings` which is completely the wrong place for test/example files

package.json
```diff
@@ -53,19 +52,18 @@
     "build:web": "rimraf ./dist/web && node src/build/buildWeb.js",
     "build:examples": "rimraf ./dist/examples && npx tsc -p tsconfig.examples.json || exit 0",
     "build:docs": "npx typedoc",
-    "serve:web": "cp src/bindings/compiled/web_bindings/server.js src/bindings/compiled/web_bindings/index.html src/examples/simple_zkapp.js dist/web && node dist/web/server.js",
```
mitschabaude, line 56: removed a couple scripts here that, to my knowledge, aren't used

Trivo25 (approved):

PR #1215, opened 2023-11-01
Expose remaining gates

mitschabaude:
* exposes all remaining gates from OCaml
* makes the `rangeCheck64` implementation match the other gadgets, in that witnessing logic is moved to the gadget and the gate is a simple wrapper
* some other tweaks to gadget code and introduce some convenience types

bindings: https://github.com/o1-labs/o1js-bindings/pull/201

### DIFF
--- a/src/lib/gadgets/common.ts
+++ b/src/lib/gadgets/common.ts
@@ -1,15 +1,30 @@
 import { Provable } from '../provable.js';
-import { Field } from '../field.js';
+import { Field, FieldConst } from '../field.js';
+import { TupleN } from '../util/types.js';
+import { Snarky } from '../../snarky.js';
+import { MlArray } from '../ml/base.js';
 
 const MAX_BITS = 64 as const;
 
 export {
   MAX_BITS,
+  exists,
   assert,
   witnessSlices,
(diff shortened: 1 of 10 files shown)


src/lib/gadgets/common.ts
```diff
@@ -1,23 +1,43 @@
 import { Provable } from '../provable.js';
-import { Field } from '../field.js';
+import { Field, FieldConst } from '../field.js';
+import { TupleN } from '../util/types.js';
+import { Snarky } from '../../snarky.js';
+import { MlArray } from '../ml/base.js';
 
 const MAX_BITS = 64 as const;
```
Trivo25, line 7: I am just seeing this - I think we should clarify what `MAX_BITS` is for (rotation gate, I believe?)

Trivo25 (approved):

src/lib/gadgets/common.ts
mitschabaude: yeah it probably doesn't belong in the `common.ts` file, complely out of context here

mitschabaude: will address in next PR

PR #1216, opened 2023-11-01
Foreign fields 1: Multi-range-check gadgets

mitschabaude:
Add multi-range check gadgets and tests

### DIFF
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -1,7 +1,11 @@
 /**
  * Wrapper file for various gadgets, with a namespace and doccomments.
  */
-import { rangeCheck64 } from './range-check.js';
+import {
+  compactMultiRangeCheck,
+  multiRangeCheck,
+  rangeCheck64,
+} from './range-check.js';
 import { rotate, xor, and } from './bitwise.js';
 import { Field } from '../core.js';
 
@@ -138,5 +142,23 @@ const Gadgets = {
    */
   and(a: Field, b: Field, length: number) {
     return and(a, b, length);
+  },
+
+  /**
+   * Multi-range check
+   *
+   * TODO
+   */
+  multiRangeCheck(x: Field, y: Field, z: Field) {
+    multiRangeCheck(x, y, z);
+  },
+
+  /**
+   * Compact multi-range check
+   *
+   * TODO
+   */
+  compactMultiRangeCheck(xy: Field, z: Field) {
+    return compactMultiRangeCheck(xy, z);
   },
 };
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -145,9 +145,15 @@ const Gadgets = {
   },
 
   /**
-   * Multi-range check
+   * Multi-range check.
    *
-   * TODO
+   * Proves that x, y, z are all in the range [0, 2^88).
+   *
+   * This takes 4 rows, so it checks 88*3/4 = 66 bits per row. This is slightly more efficient
+   * than 64-bit range checks, which can do 64 bits in 1 row.
+   *
+   * In particular, the 3x88-bit range check supports bigints up to 264 bits, which in turn is enough
+   * to support foreign field multiplication with moduli up to 2^259.
    */
   multiRangeCheck(x: Field, y: Field, z: Field) {
     multiRangeCheck(x, y, z);
@@ -156,7 +162,15 @@ const Gadgets = {
   /**
    * Compact multi-range check
    *
-   * TODO
+   * This is a variant of {@link multiRangeCheck} where the first two variables are passed in
+   * combined form xy = x + 2^88*y.
+   *
+   * The gadget
+   * - splits up xy into x and y
+   * - proves that xy = x + 2^88*y
+   * - proves that x, y, z are all in the range [0, 2^88).
+   *
+   * The split form [x, y, z] is returned.
    */
   compactMultiRangeCheck(xy: Field, z: Field) {
     return compactMultiRangeCheck(xy, z);
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -253,7 +253,7 @@ const Gadgets = {
    *
    * @throws Throws an error if `xy` exceeds 2*88 = 176 bits, or if z exceeds 88 bits.
    */
-  compactMultiRangeCheck(limbs: [Field, Field]) {
-    return compactMultiRangeCheck(limbs);
+  compactMultiRangeCheck(xy: Field, z: Field) {
+    return compactMultiRangeCheck(xy, z);
   },
 };
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -221,9 +221,16 @@ const Gadgets = {
    *
    * In particular, the 3x88-bit range check supports bigints up to 264 bits, which in turn is enough
    * to support foreign field multiplication with moduli up to 2^259.
+   *
+   * @example
+   * ```ts
+   * Gadgets.multiRangeCheck([x, y, z]);
+   * ```
+   *
+   * @throws Throws an error if one of the input values exceeds 88 bits.
    */
-  multiRangeCheck(x: Field, y: Field, z: Field) {
-    multiRangeCheck(x, y, z);
+  multiRangeCheck(limbs: [Field, Field, Field]) {
+    multiRangeCheck(limbs);
   },
 
   /**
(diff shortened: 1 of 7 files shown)


src/lib/gadgets/range-check.ts
```diff
@@ -48,3 +48,154 @@ function rangeCheck64(x: Field) {
...
+    if (xy.toBigInt() >> twoL || z.toBigInt() >> L) {
+      throw Error(
+        `Expected fields to fit in ${twoL} and ${L} bits respectively, got ${xy}, ${z}`
+      );
+    }
+    let [x, y] = splitCompactLimb(xy.toBigInt());
+    return [new Field(x), new Field(y), z];
+  }
+
+  let [x, y] = exists(2, () => splitCompactLimb(xy.toBigInt()));
```
MartinMinkov, line 92: Just a question for me here :D 
My understanding is that `exists` brings in a witness, just like `Provable.witness` does. When should you use `exists` vs `Provable.witness`?

MartinMinkov (approved):
Looks great to me! I just had one question for myself :D

src/lib/gadgets/gadgets.ts
```diff
@@ -139,4 +143,36 @@ const Gadgets = {
...
+   *
+   * Proves that x, y, z are all in the range [0, 2^88).
+   *
+   * This takes 4 rows, so it checks 88*3/4 = 66 bits per row. This is slightly more efficient
+   * than 64-bit range checks, which can do 64 bits in 1 row.
+   *
+   * In particular, the 3x88-bit range check supports bigints up to 264 bits, which in turn is enough
+   * to support foreign field multiplication with moduli up to 2^259.
+   */
+  multiRangeCheck(x: Field, y: Field, z: Field) {
```
MartinMinkov, line 158: It would be great to match the template similar to the other gates, such as having an `@example` block, describing inputs, if it can throw, etc, just for consistency stake.

src/lib/gadgets/range-check.ts
mitschabaude: `Provable.witness` handles any provable data types. I created `exists` to have something that's quicker to use if you just want an array of field elements. It's better than `witness` for that in three ways:
* no boilerplate for specifying the type `Provable.Array(Field, 5)`
* you can return bigints from the callback instead of `Fields`
* the array length is typed! i.e. it's a type error to return or unpack too many outputs, like `[x,y,z]`in this example

My thinking was that for low-level gadgets, everything will be arrays of field elements so it makes to have something which optimizes for that

MartinMinkov: Awesome! Thank you for the explanation. I must admit, I used `Provable.witness` for the rotation gadget, not a big deal but I wish I used `exists` instead!

src/lib/gadgets/range-check.ts
```diff
@@ -48,3 +48,154 @@ function rangeCheck64(x: Field) {
...
+  yz: Field;
+}) {
+  let { x64, x76, y64, y76, z, yz } = inputs;
+
+  // create limbs for current row
+  let [z22, z24, z26, z28, z30, z32, z34, z36, z38, z50, z62, z74, z86] =
+    exists(13, () => {
+      let zz = z.toBigInt();
+      return [
+        bitSlice(zz, 22, 2),
```
jspada, line 161: Which endianness is this in?  The `curr` and `next` row seem in opposite order of the rust implementation.  Probably just endianness.  Otherwise, lgtm!

src/lib/gates.ts
```diff
@@ -19,6 +19,24 @@ function rangeCheck0(
   );
 }
 
+/**
+ * the rangeCheck1 gate is used in combination with the rangeCheck0,
+ * for doing a 3x88-bit range check
+ */
```
jspada, line 25: If this is part of the public API you probably want to comment the inputs and output of this gadget.

src/lib/gadgets/range-check.ts
```diff
@@ -48,3 +48,154 @@ function rangeCheck64(x: Field) {
...
+      bitSlice(xx, 4, 2),
+      bitSlice(xx, 6, 2),
+      bitSlice(xx, 8, 2),
+      bitSlice(xx, 10, 2),
+      bitSlice(xx, 12, 2),
+      bitSlice(xx, 14, 2),
+    ];
+  });
+
+  // 12-bit limbs
```
jspada, line 121: This seems different than the rust implementation.  In the rust implementation the crumbs are in columns 7-14.  Maybe it has to do with the endianness... in the rust implementation of the gate limbs are mapped to columns in big-endian order (the lowest columns contain the highest bits).

The reason for this is so that the hiehgest bits are covered by copy constraints and lookups, so that we can copy the highest two 12-bit limbs to zero in order to use the `RangeCheck0` gate for 64-bit range-checks.

You probably have this right here and it's just because of a different byte-order, but I just wanted to call it out explicitly to make sure.

src/lib/gadgets/range-check.unit-test.ts
```diff
@@ -1,49 +1,127 @@
...
   (x) => {
-    if (x >= 1n << 64n) throw Error('expected 64 bits');
+    assert(x < 1n << 64n);
     return true;
   },
   async (x) => {
-    let proof = await RangeCheck64.run(x);
-    return await RangeCheck64.verify(proof);
+    let proof = await RangeCheck.check64(x);
+    return await RangeCheck.verify(proof);
```
jspada, line 97: cool

src/lib/gadgets/range-check.unit-test.ts
```diff
@@ -1,49 +1,127 @@
...
+  { from: [maybeUint(2n * L), uint(L)], to: boolean },
+  { runs: 3 }
+)(
+  (xy, z) => {
+    assert(!(xy >> (2n * L)) && !(z >> L), 'compact: not out of range');
+    return true;
+  },
+  async (xy, z) => {
+    let proof = await RangeCheck.checkCompact(xy, z);
+    return await RangeCheck.verify(proof);
```
jspada, line 125: I don't see any out of range tests. might be a good idea.

src/lib/gadgets/range-check.ts
```diff
@@ -48,3 +48,154 @@ function rangeCheck64(x: Field) {
...
+}
+
+/**
+ * Compact multi-range-check - checks
+ * - xy = x + 2^88*y
+ * - x, y, z \in [0, 2^88)
+ *
+ * Returns the full limbs x, y, z
+ */
+function compactMultiRangeCheck(xy: Field, z: Field): [Field, Field, Field] {
```
jspada, line 80: lgtm.  Same order of params as rusts impl's `create_multi_compact‎`

jspada (approved):
Minor suggestions and a question about endianness.  Otherwise, lgtm!

src/lib/gates.ts
mitschabaude: The gates folder is not currently part of the public API

src/lib/gadgets/gadgets.ts
```diff
@@ -139,4 +143,36 @@ const Gadgets = {
...
+    multiRangeCheck(x, y, z);
+  },
+
+  /**
+   * Compact multi-range check
+   *
+   * This is a variant of {@link multiRangeCheck} where the first two variables are passed in
```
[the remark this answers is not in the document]
mitschabaude: Yeah - so for context, the way we expose this is on a `Gadgets` namespace which is presented as a set of low-level provable methods for advanced users (including ourselves, in higher-level APIs). So I think the compact range check fits well to include. (It gives you something you can't do in any other way, so in a sense it can't be left out)

What do you think are limitations/security concerns that should be highlighted?

src/lib/gadgets/range-check.ts
mitschabaude: so, I guess I create the witnesses in little endian order here (which seemed more natural to me), but note that I pass them to the gate in big endian order, [see here](https://github.com/o1-labs/o1js/pull/1216/files#diff-41b34b8e0533ea367eb60a3dea585c5505c8b3590f0e5459f3d39fa6f1f10b4dR134-R139)

src/lib/gadgets/range-check.unit-test.ts
mitschabaude: this is a test for both in range and out of range! the function signature spec is
```ts
{ from: [maybeUint(2n * L), uint(L)], to: boolean },
```
in our test framework, this means that the first argument `xy` will sometimes be an invalid and sometimes a valid uint of `2L` bits. the test is run on a number of random samples. since it does a proof everytime, we only do 3 samples per test run (`{ runs: 3 }`) but locally I do 20 runs or so when first creating a test like this. the distributions are pretty clever so this always tends to uncover the bugs

mitschabaude: The test checks that the two functions either both throw an error or return the same result. The first function just checks the range and throws an error if it's out of range, so we want to be equivalent to that

PR #1220, opened 2023-11-02
Foreign fields 3: Addition

### LINKED ISSUE #1152: Port FFSum_check gadget to TypeScript

### LINKED ISSUE #1180: Port FFadd and FFsub gadget to TypeScript
depends on interface to #1152

mitschabaude:
* adds a `ForeignField` gadget namespace with methods for addition, subtraction and general sums using non-native modular arithmetic.
* enhances the `constraintSystem()` test runner introduced in #1241 so that it becomes very easy to check methods of a zkprogram

closes #1152
closes #1180

depends on bindings PRs https://github.com/o1-labs/o1js-bindings/pull/203 and https://github.com/o1-labs/o1js-bindings/pull/205

### Proof that we can save range checks in `sum()`

An optimization used here is to not range-check intermediate results in addition chains of length > 2, i.e. what is exposed as `sum()`. The reasoning why this is sound is as follows:

Assume we are doing an addition chain over summands $a_0,\ldots,a_n$. Let's denote the lower input limbs by $a_i^{(01)}$ and the high limb by $a_i^{(2)}$. Each FFadd gate individually proves that

$$
a_i^{(01)} + a_{i+1}^{(01)} - o_i f^{(01)} - c_i 2^{2\ell} = r_i^{(01)} \mod{n}
$$

$$
a_i^{(2)} + a_{i+1}^{(2)} - o_i f^{(2)} + c_i = r_i^{(2)} \mod{n}
$$

where $o_i$ and $c_i$ are overflow and carry, and $n$ is the native modulus. Since addition modulo $n$ is associative, a full chain of ffadd gates proves the following equations about the _final output_ $r$:

$$
a_0^{(01)} + \ldots + a_{n}^{(01)} - (o_1 + \ldots + o_n) f^{(01)} - (c_1 + \ldots + c_n) 2^{2\ell} = r^{(01)} \mod{n}
$$

$$
a_0^{(2)} + \ldots + a_{n}^{(2)} - (o_1 + \ldots + o_n) f^{(2)} + (c_1 + \ldots + c_n) = r^{(2)} \mod{n}
$$

Since we assume the inputs $a_i$ to be range-checked, and both $o_i$ and $c_i$ are 0 or 1 or -1, and the final output $r$ is range-checked as well, the difference of left and right hand side is too small to overflow the native field:

$$
|LHS - RHS| < n
$$

This is enough to conclude that $LHS = RHS$ over the integers. Which in this case just means that we are doing correct addition or subtraction modulo the foreign field.


src/lib/testing/equivalent.ts
```diff
@@ -235,16 +244,63 @@ let bool: ProvableSpec<boolean, Bool> = {
...
+// spec combinators
+
+function array<T, S>(
+  spec: ProvableSpec<T, S>,
+  n: number
+): ProvableSpec<T[], S[]>;
+function array<T, S>(
+  spec: Spec<T, S>,
+  n: Random<number> | number
+): Spec<T[], S[]>;
```
rbonichon, line 262: Are these function declarations?
Why is the second declaration needed (it's the same type as the implementation right below)?

mitschabaude: This is how you declare if there are two different function signatures. The third one is not exposed, it's the "internal" signature used in writing the function. So if I remove the second one, consumers only see the first one

src/lib/gadgets/foreign-field.ts
mitschabaude, line 1: the code for the new gadgets! please review carefully

src/lib/gadgets/gadgets.ts
```diff
@@ -312,4 +313,118 @@ const Gadgets = {
...
+      return ForeignField.sum(xs, signs, f);
+    },
+  },
+
+  /**
+   * Helper methods to interact with 3-limb vectors of Fields.
+   *
+   * **Note:** This interface does not contain any provable methods.
+   */
+  Field3,
```
mitschabaude, line 422: I'm a bit conflicted about this. I sometimes got the feedback that it's confusing how o1js APIs are a mix of provable and non-provable methods. I'm not sure if putting them on a separate sub-namespace of `Gadgets` is enough separation. In any case, these helper methods are essential to reduce boiler-plate of interacting with all foreign field and foreign curve gadgets, and I also didn't feel good about another top-level export for them.

src/lib/gadgets/gadgets.ts
```diff
@@ -312,4 +313,118 @@ const Gadgets = {
...
 };
+
+export namespace Gadgets {
+  /**
+   * A 3-tuple of Fields, representing a 3-limb bigint.
+   */
+  export type Field3 = [Field, Field, Field];
+}
```
mitschabaude, line 430: note: this is how to get a type namespace alongside a value namespace. It's possible to use the same name if the type namespace is only types. We could've also put everything in a single `namespace`, but I prefer the syntax for writing JS objects over the clunky namespace syntax which isn't even real JS (just a TS addition to the language).

src/lib/testing/constraint-system.ts
```diff
@@ -85,6 +86,37 @@ function constraintSystem<Input extends Tuple<CsVarSpec<any>>>(
...
+ * Convenience function to run {@link constraintSystem} on the method of a {@link ZkProgram}.
+ *
+ * @example
+ * ```ts
+ * const program = ZkProgram({ methods: { myMethod: ... }, ... });
+ *
+ * constraintSystem.fromZkProgram(program, 'myMethod', contains('Rot64'));
+ * ```
+ */
+constraintSystem.fromZkProgram = function fromZkProgram<
```
mitschabaude, line 99: This is a continuation of #1241. I figured I could make writing constraint system tests for zkprogram methods even easier by reading the input types off the zkprogram.

src/lib/proof_system.ts
mitschabaude, line 1: changes in this file are in support of the new `constraintSystem.fromZkProgram()`

src/lib/proof_system.ts
```diff
@@ -427,21 +434,34 @@ function ZkProgram<
...
       digest,
+      analyzeMethods,
       publicInputType: publicInputType as ProvableOrUndefined<
         Get<StatementType, 'publicInput'>
       >,
       publicOutputType: publicOutputType as ProvableOrVoid<
         Get<StatementType, 'publicOutput'>
       >,
-      analyzeMethods,
+      privateInputTypes: Object.fromEntries(
```
mitschabaude, line 444: exposing the private input types on zkprogram seems consistent, since we already expose the public input types and the methods

src/lib/proof_system.ts
```diff
@@ -925,6 +945,7 @@ ZkProgram.Proof = function <
     static tag = () => program;
   };
 };
+ExperimentalZkProgram.Proof = ZkProgram.Proof;
```
mitschabaude, line 948: this was a bug which accidentally broke `Experimental.ZkProgram`, fixed it while I was here

src/lib/testing/equivalent.ts
mitschabaude, line 1: additions here are in support of the foreign field unit tests

src/lib/testing/random.ts
```diff
@@ -307,6 +308,7 @@ const Random = Object.assign(Random_, {
   reject,
   dice: Object.assign(dice, { ofSize: diceOfSize() }),
   field,
+  otherField: fieldWithInvalid,
```
mitschabaude, line 311: random generator for foreign fields, for tests

mitschabaude (commented):
some notes for reviewers!

src/lib/gadgets/foreign-field.ts
```diff
@@ -0,0 +1,138 @@
...
+    let overflow = 0n;
+    if (sign === 1n && r >= f) overflow = 1n;
+    if (sign === -1n && r < 0n) overflow = -1n;
+    if (f === 0n) overflow = 0n; // special case where overflow doesn't change anything
+
+    // do the add with carry
+    // note: this "just works" with negative r01
+    let r01 = collapse2(x_) + sign * collapse2(y_) - overflow * collapse2(f_);
+    let carry = r01 >> twoL;
+    r01 &= twoLMask;
```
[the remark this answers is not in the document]
querolita: Oh this is 2^176 - 1 perhaps?

src/lib/gadgets/range-check.ts
```diff
@@ -53,6 +61,7 @@ function rangeCheck64(x: Field) {
 const L = 88n;
 const twoL = 2n * L;
 const lMask = (1n << L) - 1n;
+const twoLMask = (1n << twoL) - 1n;
```
querolita, line 64: Oh it is indeed 😄

src/lib/proof_system.ts
querolita: What is this exactly for?

src/lib/testing/random.ts
```diff
@@ -843,12 +845,12 @@ function bignatWithInvalid(max: bigint): RandomWithInvalid<bigint> {
...
 
-function fieldWithInvalid(
-  F: typeof Field | typeof Scalar
-): RandomWithInvalid<bigint> {
+function fieldWithInvalid(F: FiniteField): RandomWithInvalid<bigint> {
   let randomField = Random_(F.random);
-  let specialField = oneOf(0n, 1n, F(-1));
-  let field = oneOf<bigint[]>(randomField, randomField, uint64, specialField);
+  let specialField = oneOf(0n, 1n, F.negate(1n));
+  let roughLogSize = 1 << Math.ceil(Math.log2(F.sizeInBits) - 1);
```
querolita, line 851: Why are you using this size?

querolita (approved):
Approving with minor comments

src/lib/proof_system.ts
mitschabaude: It's in support of a testing utility I added: https://github.com/o1-labs/o1js/pull/1220#discussion_r1392513182
You can ignore it, it's not related to the circuit gadgets added here :D

src/lib/gadgets/foreign-field.ts
```diff
@@ -0,0 +1,138 @@
...
+  },
+  sum,
+};
+
+/**
+ * computes x[0] + sign[0] * x[1] + ... + sign[n-1] * x[n] modulo f
+ *
+ * assumes that inputs are range checked, does range check on the result.
+ */
+function sum(x: Field3[], sign: Sign[], f: bigint) {
```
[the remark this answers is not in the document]
mitschabaude: `sum()` is exposed! it is actually the same as `sumchain` in OCaml, I renamed it because I think the "chain" aspect is more about the internal implementation and less important for users

src/lib/gadgets/foreign-field.ts
```diff
@@ -0,0 +1,138 @@
...
+    return Field3.from(mod(sum, f));
+  }
+  // provable case - create chain of ffadd rows
+  x = x.map(toVars);
+  let result = x[0];
+  for (let i = 0; i < sign.length; i++) {
+    ({ result } = singleAdd(result, x[i + 1], sign[i], f));
+  }
+  // final zero row to hold result
+  Gates.zero(...result);
```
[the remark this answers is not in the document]
mitschabaude: I completely agree, this isn't yet the best version of the sumchain gadget. Wait for the ECDSA PR, there I needed chaining into ffmul and I came up with an API which is both easy to use and flexible

src/lib/gadgets/foreign-field.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
+  let f_ = split(f);
+
+  let [r0, r1, r2, overflow, carry] = exists(5, () => {
+    let x_ = bigint3(x);
+    let y_ = bigint3(y);
+
+    // figure out if there's overflow
+    let r = collapse(x_) + sign * collapse(y_);
```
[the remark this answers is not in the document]
mitschabaude: yeah, bigints can be negative in js

src/lib/gadgets/foreign-field.ts
```diff
@@ -0,0 +1,138 @@
...
+    let overflow = 0n;
+    if (sign === 1n && r >= f) overflow = 1n;
+    if (sign === -1n && r < 0n) overflow = -1n;
+    if (f === 0n) overflow = 0n; // special case where overflow doesn't change anything
```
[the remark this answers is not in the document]
mitschabaude: yes I think it could be useful if you want to do plain bigint addition (not modulo a field)

src/lib/testing/random.ts
mitschabaude: because the `biguint` random generator I'm passing it in needs the bit size to be a power of 2 (for efficiency)

Trivo25 (approved):

PR #1223, opened 2023-11-02
Foreign fields 4: Multiplication

### LINKED ISSUE #1142: Port FFMUL gadget to TypeScript

mitschabaude:
* add 3 different gadgets that use the ffmul gate: `ForeignField.mul()`, `ForeignField.inv()`, `ForeignField.div()`
  * gadget design closely follows https://github.com/o1-labs/rfcs/blob/main/0006-ffmul-revised.md
* add foreign field unit tests with numbers not reduced modulo f, where applicable
* change the return signature of `ZkProgram.analyzeMethods()` to be nicer to read and to match `SmartContract.analyzeMethods()` (seems kinda unrelated; I did this to inspect number of rows of some ffmul methods. It's technically a breaking change but low impact)

bindings: https://github.com/o1-labs/o1js-bindings/pull/206

closes #1142


src/lib/gadgets/gadgets.ts
```diff
@@ -412,6 +407,62 @@ const Gadgets = {
...
+     * **Assumptions**: In addition to the assumption that inputs are in the range [0, 2^88), as in all foreign field gadgets,
+     * this assumes an additional bound on the inputs: `x * y < 2^264 * p`, where p is the native modulus.
+     * We usually assert this bound by proving that `x[2] < f[2] + 1`, where `x[2]` is the most significant limb of x.
+     * To do this, use an 88-bit range check on `2^88 - x[2] - (f[2] + 1)`, and same for y.
+     * The implication is that x and y are _almost_ reduced modulo f.
+     *
+     * @example
+     * ```ts
+     * // example modulus: secp256k1 prime
+     * let f = (1n << 256n) - (1n << 32n) - 0b1111010001n;
```
Trivo25, line 425: any reason for using binary notation here?

Trivo25 (approved):
this is awesome! Really appreciate all the descriptive comments

src/lib/gadgets/foreign-field.ts
```diff
@@ -23,6 +34,10 @@ const ForeignField = {
     return sum([x, y], [-1n], f);
   },
   sum,
+
+  mul: multiply,
+  inv: inverse,
```
querolita, line 39: Oh, interesting feature!

src/lib/gadgets/foreign-field.unit-test.ts
```diff
@@ -117,11 +199,29 @@ constraintSystem.fromZkProgram(
   )
 );
 
+let mulChain: GateType[] = ['ForeignFieldMul', 'Zero'];
+let mulLayout = ifNotAllConstant(
+  and(
+    contains([mulChain, mrc, mrc, mrc, mrc]),
```
querolita, line 205: which are these mrc's exactly?

I assume the first one is for compact for the remainder,
the second one for quotient range check,
the third one for quotient bound, product1_lo and hi_0,
the last one for result bound, left bound, and right bound

?

src/lib/gadgets/foreign-field.unit-test.ts
```diff
@@ -117,11 +199,29 @@ constraintSystem.fromZkProgram(
...
+    withoutGenerics(equals([...mulChain, ...repeat(4, mrc)]))
+  )
+);
+let invLayout = ifNotAllConstant(
+  and(
+    contains([mrc, mulChain, mrc, mrc, mrc]),
+    withoutGenerics(equals([...mrc, ...mulChain, ...repeat(3, mrc)]))
```
querolita, line 212: So I guess this means generics (which we have) can be anywhere else? We needed 3 to compute the result bound, left bound, and right bound iirc.

querolita (commented):
Excellent job! I would nonetheless like to clarify a question about the circuit shape before accepting, to make sure all checks are getting in. I also have a question regarding division 💯

src/lib/gadgets/foreign-field.ts
```diff
@@ -79,8 +94,8 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
     // do the add with carry
     // note: this "just works" with negative r01
     let r01 = collapse2(x_) + sign * collapse2(y_) - overflow * collapse2(f_);
-    let carry = r01 >> twoL;
-    r01 &= twoLMask;
+    let carry = r01 >> L2;
+    r01 &= l2Mask;
```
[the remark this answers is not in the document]
mitschabaude: good point, I probably should lowercase both

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+
+  // provable case
+  let { r01, r2, q, q2Bound } = multiplyNoRangeCheck(a, b, f);
+
+  // limb range checks on quotient and remainder
+  multiRangeCheck(q);
+  let r = compactMultiRangeCheck(r01, r2);
+
+  // range check on q and r bounds
+  // TODO: this uses one RC too many.. need global RC stack, or get rid of bounds checks
```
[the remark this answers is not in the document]
mitschabaude: fwiw I think we can get rid of the bounds checks entirely if we just slightly tweak the rangecheck1 gate. I'll do this after the HF.

therefore, I don't think it's worth introducing the global variable storing pending limbs. this would complicate the architecture, and the impact is pretty small: in the ec addition gadget which takes 70 rows, this would currently only save us 2/3s of a MRC gate, so only 2.66 rows. That's just a 3.8% improvement. I still have a couple of ideas for improvements that are much more impactful and don't complicate the overall architecture.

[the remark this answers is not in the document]
mitschabaude: Thanks for asking about the exact layout and giving an example @querolita. It made me realize a mistake I made -- I accidentally range-checked the carry c0, even though it is just 2 bits and is checked in the ffmul gate 😬

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+    let [f0, f1, f2] = split(f);
+    let f01 = collapse2([f0, f1]);
+    y01.equals(f01).and(y[2].equals(f2)).assertFalse();
+  }
+
+  return z;
+}
+
+/**
+ * Common logic for gadgets that expect a certain multiplication result instead of just using the remainder.
```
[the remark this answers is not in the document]
mitschabaude: No, I mean gadgets that know what the result should be before the multiplication. For example, the inverse gadget uses this. It knows that the result of the multiplication should be 1

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+  let z2Bound = weakBound(z[2], f);
+  let q2Bound = assertMul(z, y, x, f);
+
+  // range check on q and result bounds
+  multiRangeCheck([q2Bound, z2Bound, Field.from(0n)]);
+
+  if (!allowZeroOverZero) {
+    // assert that y != 0 mod f by checking that it doesn't equal 0 or f
+    // this works because we assume y[2] <= f2
+    // TODO is this the most efficient way?
```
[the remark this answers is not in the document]
mitschabaude: What do you mean with proving y has an inverse? Do you mean prove that it has an inverse in the foreign field? That would be another ff multiplication, which is much more expensive than what I'm doing here (that's the reason that my `div()` is not just `inv()` + `mul()`)

Or do you mean proving that y has an inverse in the native field? That's not possible, because y is a vector of 3 limbs. If we would sum them and prove that the result has an inverse, this would only prove that y is not 0. But what if y is f. Also, this would ruin completeness, since it would not allow y to equal the native modulus (or any multiple of it), but that should actually be allowed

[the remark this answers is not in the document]
mitschabaude: The approach here is to check that y does not equal 0, and it does not equal f, by proving that not all the individual limbs are equal to 0 or f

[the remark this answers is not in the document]
mitschabaude: it uses the `equals()` gadget for native field elements, which you can check the code for here: https://github.com/o1-labs/o1js/blob/aec1e893774714696e621a64bb666d2c1d8f57ca/src/lib/field.ts#L651-L656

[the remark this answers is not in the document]
mitschabaude: the real magic happens inside `isZero()`:
https://github.com/o1-labs/o1js/blob/aec1e893774714696e621a64bb666d2c1d8f57ca/src/lib/field.ts#L609-L629

src/lib/gadgets/foreign-field.ts
```diff
@@ -79,8 +94,8 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
     // do the add with carry
     // note: this "just works" with negative r01
     let r01 = collapse2(x_) + sign * collapse2(y_) - overflow * collapse2(f_);
-    let carry = r01 >> twoL;
-    r01 &= twoLMask;
+    let carry = r01 >> l2;
```
jspada, line 97: maybe `combine` is a better name that `collapse` ?

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+  if (x.every((x) => x.isConstant())) {
+    let xInv = modInverse(Field3.toBigint(x), f);
+    assert(xInv !== undefined, 'inverse exists');
+    return Field3.from(xInv);
+  }
+
+  // provable case
+  let xInv = exists(3, () => {
+    let xInv = modInverse(Field3.toBigint(x), f);
+    return xInv === undefined ? [0n, 0n, 0n] : split(xInv);
```
jspada, line 141: so if an inverse doesn't exist it return 0s, but does something else pick this up and fail?

jspada, line 141: maybe `toLimbs` would be a better name?

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+  });
+  multiRangeCheck(xInv);
+  let xInv2Bound = weakBound(xInv[2], f);
```
jspada, line 144: would be nice to have a comment explaining why use a weak bound check here

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+
+  let one: Field2 = [Field.from(1n), Field.from(0n)];
+  assertMul(x, xInv, one, f);
+
+  // range check on result bound
+  // TODO: this uses two RCs too many.. need global RC stack
```
jspada, line 150: would be nice to fix this.

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+    // this works because we assume y[2] <= f2
+    // TODO is this the most efficient way?
+    let y01 = y[0].add(y[1].mul(1n << l));
+    y01.equals(0n).and(y[2].equals(0n)).assertFalse();
+    let [f0, f1, f2] = split(f);
+    let f01 = collapse2([f0, f1]);
+    y01.equals(f01).and(y[2].equals(f2)).assertFalse();
+  }
+
+  return z;
```
jspada, line 196: looks good to me

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+    let r = ab - q * f;
+
+    let [q0, q1, q2] = split(q);
+    let [r0, r1, r2] = split(r);
+    let r01 = collapse2([r0, r1]);
+
+    // compute product terms
+    let p0 = a0 * b0 + q0 * f_0;
+    let p1 = a0 * b1 + a1 * b0 + q0 * f_1 + q1 * f_0;
+    let p2 = a0 * b2 + a1 * b1 + a2 * b0 + q0 * f_2 + q1 * f_1 + q2 * f_0;
```
jspada, line 247: lgtm!

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+
+    let [p10, p110, p111] = split(p1);
+    let p11 = collapse2([p110, p111]);
+
+    // carry bottom limbs
+    let c0 = (p0 + (p10 << l) - r01) >> l2;
+
+    // carry top limb
+    let c1 = (p2 - r2 + p11 + c0) >> l;
```
jspada, line 256: looks right.

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+    // compute product terms
+    let p0 = a0 * b0 + q0 * f_0;
+    let p1 = a0 * b1 + a1 * b0 + q0 * f_1 + q1 * f_0;
+    let p2 = a0 * b2 + a1 * b1 + a2 * b0 + q0 * f_2 + q1 * f_1 + q2 * f_0;
+
+    let [p10, p110, p111] = split(p1);
+    let p11 = collapse2([p110, p111]);
+
+    // carry bottom limbs
+    let c0 = (p0 + (p10 << l) - r01) >> l2;
```
jspada, line 253: this is `p01 * 2^L` in the RFC, but probably you just took a different notation here?

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+  Gates.foreignFieldMul({
+    left: a,
+    right: b,
+    remainder: [r01, r2],
+    quotient: q,
+    quotientHiBound: q2Bound,
+    product1: [p10, p110, p111],
+    carry0: c0,
+    carry1p: [c1_00, c1_12, c1_24, c1_36, c1_48, c1_60, c1_72],
+    carry1c: [c1_84, c1_86, c1_88, c1_90],
```
jspada, line 309: cool

src/lib/gadgets/foreign-field.ts
```diff
@@ -92,12 +107,230 @@ function singleAdd(x: Field3, y: Field3, sign: Sign, f: bigint) {
...
+    negForeignFieldModulus: [f_0, f_1, f_2],
+  });
+
+  // multi-range check on internal values
+  multiRangeCheck([p10, p110, q2Bound]);
+
+  // note: this function is supposed to be the most flexible interface to the ffmul gate.
+  // that's why we don't add range checks on q and r here, because there are valid use cases
+  // for not range-checking either of them -- for example, they could be wired to other
+  // variables that are already range-checked, or to constants / public inputs.
```
jspada, line 320: makes sense and this is what I did in the ocaml implementation also.

src/lib/gadgets/foreign-field.unit-test.ts
```diff
@@ -56,11 +74,57 @@ for (let F of fields) {
...
+    F.mul,
+    (x, y) => ForeignField.mul(x, y, F.modulus),
+    'mul unreduced'
+  );
+  equivalentProvable({ from: [big258], to: f })(
+    (x) => F.inverse(x) ?? throwError('no inverse'),
+    (x) => ForeignField.inv(x, F.modulus),
+    'inv unreduced'
+  );
+  // the div() gadget doesn't work with unreduced x because the backwards check (x/y)*y === x fails
```
jspada, line 114: It may be useful to document this at the interface level, i.e., what is compatible with unreduced values

src/lib/gates.ts
```diff
@@ -184,3 +187,56 @@ function foreignFieldAdd({
...
+  right: TupleN<Field, 3>;
+  remainder: TupleN<Field, 2>;
+  quotient: TupleN<Field, 3>;
+  quotientHiBound: Field;
+  product1: TupleN<Field, 3>;
+  carry0: Field;
+  carry1p: TupleN<Field, 7>;
+  carry1c: TupleN<Field, 4>;
+  foreignFieldModulus2: bigint;
+  negForeignFieldModulus: TupleN<bigint, 3>;
```
jspada, line 205: nice and neat!

jspada (approved):
Approved with minor suggestions and a question.

src/lib/gadgets/foreign-field.ts
mitschabaude: yes! we later prove that `x * xInv === 1` using the ffmul gate. This would fail if an inverse doesn't exist.

src/lib/gadgets/foreign-field.ts
mitschabaude: How about join()? That matches with split()

src/lib/gadgets/foreign-field.ts
mitschabaude: oh you're right! that's actually a typo in the RFC 😓

src/lib/gadgets/foreign-field.ts
mitschabaude: went with combine

src/lib/gadgets/foreign-field.ts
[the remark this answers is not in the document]
querolita: > and it does not equal f, by proving that not all the individual limbs

What if it was any other multiple of `f`? You don't consider this case because it is assumed to be range-checked to be at most `f`?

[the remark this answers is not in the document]
querolita: >     `y01.equals(0n).and(y[2].equals(0n)).assertFalse();`

I didn't know you could do this type of assert_neq's in o1js. How does it work under the hood though? In Kimchi, we just create the constraints to satisfy the properties, not the opposite. Meaning, if one wanted to check that a value is **not** zero in one gate, one would have to either write all permitted values (insanely complex) or provide the inverse in the witness and just prove that $1 - x*inv = 0$.

Issue #1222, opened 2023-11-02
Fix all example zkApps inside `src/examples`, ensure they can all build and execute correctly


MartinMinkov:
# Summary

Some examples fail to build when running `npm run build:test`. I'd assume all these errors are because they were assuming an older version of `o1js` and never updated their APIs. It would be great to get these to build and run again since they are a beneficial resource when everything is up to date.

barriebyron:
Contributor guidance, submit one PR for each example file

mitschabaude:
> Contributor guidance, submit one PR for each example file

Note: as there are _a lot_ of examples, and many fixes will be trivial, I don't recommend to take that too literally

ymekuria:
> It would also be great to publish the examples to some sandbox site like https://stackblitz.com and add links to the sandbox code to the Tutorials docs.

Great idea! We are exploring ways to incorporate something like this.

PR #1225, opened 2023-11-03
Some example improvements

mitschabaude:
fixes https://github.com/o1-labs/o1js/issues/1228

start improving `/examples` folder by moving more files into their own folders, cleaning up a few examples, and adding some comments and readmes. Also fixes a type error in `ZkProgram.compile()`

I'm still not really settling on a folder structure I like, and there's still quite a bit of random junk left in `/examples`, but this is a start


src/examples/internals/README.md
```diff
@@ -0,0 +1,5 @@
+# Examples: Internals
+
+This folder contains examples which highlight inner workings and less-documented behaviours of o1js.
+
+These examples might be useful for advanced users and contributors.
```
barriebyron, line 5:
```suggestion
This folder contains examples that highlight the inner workings and less-documented behaviors of o1js.

These examples might be useful for advanced users and contributors.

Run the examples with `npm run build`.
```

barriebyron: @mitschabaude It was only yesterday when I saw the description in https://github.com/o1-labs/o1js/issues/1222 that I knew we had runnable examples... we might not need the "Run the examples with..." here, but perhaps at a top-level?

mitschabaude: that's not correct :D but good idea to add an isntruction how to run examples! I'll add it to the top level examples folder

MartinMinkov (approved):

MartinMinkov (approved):

barriebyron:
soooooo goood thank you!!

PR #1229, opened 2023-11-05
RSA example

mitschabaude:
* enable 52-, 40-, 28- and 16-bit range checks with `Gadgets.rangeCheck64()`, by returning the high bit slices which can be forced to equal 0
* expose `Provable.witnessFields()`, for fixed-length tuples of field elements
* add RSA example and tests

includes https://github.com/o1-labs/o1js/pull/1599 by @Shigoto-dev19 who fixed the bigint mul circuit and added tests for RSA

### Results

RSA verification was built only from range checks and generic gates, and uses 12k Kimchi constraints.
This is a pretty reasonable number - for comparison, a comparable [circom implementation](https://github.com/zkemail/zk-email-verify/blob/1b494cdc6f7b755431b71eed9189899650e1b6f0/packages/circuits/email-verifier.circom#L68-L69) uses ~150k R1CS constrants.

```
$ ./run src/examples/crypto/rsa/run.ts --bundle
{ 'Total rows': 12401, RangeCheck0: 2488, Generic: 9913 }
```


Trivo25:
I doubt this will land soon, while the core functionality is there, this needs to be polished, tested and documented. Wdyt @mitschabaude

mitschabaude:
Is there anything here in particular that you want landed @emlautarom1?
Note that RSA is just an example, not part of the library

Trivo25 (approved):

PR #1238, opened 2023-11-13
Side-loaded keys

mrmr1993:


mitschabaude:
The CI errors are because the bindings module is missing stuff from `o1js-bindings/main`

Trivo25:
https://github.com/o1-labs/o1js/pull/1606

PR #1239, opened 2023-11-13
Make `tx.wait()` more reliable by increasing block length check to default of 20 blocks

MartinMinkov:
# Summary
This update introduces improvements to `tx.wait()`, addressing reliability issues in transaction inclusion checking within blocks. It includes a fix by checking for the last 20 blocks of a transaction inclusion instead of 1. Additionally, the number of blocks to check for `fetchLatestBlockZkappStatus` is now customizable.


mitschabaude (approved):
Nice @MartinMinkov, I really appreciate that you're looking after the zkApp DX aspects of o1js

PR #1241, opened 2023-11-13
Test DSL to detect broken gate chains

mitschabaude:
- first version of constraint system test dsl
- properly implement fieldvar combinators
- pass inputs to cs tests, and make them combinable
- more cs tests and combinators
- replace existing test with new dsl test, which passes
- usability tweaks
- tweak error output
- replace other test and expose bug where generic gate disrupts mrc chain
- print constraint system

### DIFF
--- a/src/lib/field.ts
+++ b/src/lib/field.ts
@@ -78,13 +78,22 @@ const FieldVar = {
   isConstant(x: FieldVar): x is ConstantFieldVar {
     return x[0] === FieldType.Constant;
   },
-  // TODO: handle (special) constants
   add(x: FieldVar, y: FieldVar): FieldVar {
+    if (FieldVar.isConstant(x) && x[1][1] === 0n) return y;
+    if (FieldVar.isConstant(y) && y[1][1] === 0n) return x;
+    if (FieldVar.isConstant(x) && FieldVar.isConstant(y)) {
+      return FieldVar.constant(Fp.add(x[1][1], y[1][1]));
+    }
     return [FieldType.Add, x, y];
   },
-  // TODO: handle (special) constants
-  scale(c: FieldConst, x: FieldVar): FieldVar {
-    return [FieldType.Scale, c, x];
+  scale(c: bigint | FieldConst, x: FieldVar): FieldVar {
+    let c0 = typeof c === 'bigint' ? FieldConst.fromBigint(c) : c;
+    if (c0[1] === 0n) return FieldVar.constant(0n);
+    if (c0[1] === 1n) return x;
+    if (FieldVar.isConstant(x)) {
+      return FieldVar.constant(Fp.mul(c0[1], x[1][1]));
+    }
+    return [FieldType.Scale, c0, x];
   },
   [0]: [FieldType.Constant, FieldConst[0]] satisfies ConstantFieldVar,
   [1]: [FieldType.Constant, FieldConst[1]] satisfies ConstantFieldVar,
--- a/src/lib/gadgets/range-check.unit-test.ts
+++ b/src/lib/gadgets/range-check.unit-test.ts
@@ -1,8 +1,6 @@
-import type { Gate } from '../../snarky.js';
 import { mod } from '../../bindings/crypto/finite_field.js';
 import { Field } from '../../lib/core.js';
 import { ZkProgram } from '../proof_system.js';
-import { Provable } from '../provable.js';
 import {
   Spec,
   boolean,
@@ -10,12 +8,12 @@ import {
   fieldWithRng,
 } from '../testing/equivalent.js';
 import { Random } from '../testing/property.js';
-import { assert, exists } from './common.js';
+import { assert } from './common.js';
 import { Gadgets } from './gadgets.js';
 import { L } from './range-check.js';
-import { expect } from 'expect';
 import {
   constraintSystem,
+  contains,
   equals,
   ifNotAllConstant,
   withoutGenerics,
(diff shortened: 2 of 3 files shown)


MartinMinkov (approved):
This is fantastic! Bonus points for your DSL design, seems straightforward to extend it! But you added a pretty extensive suite for our Gadgets already! :tada:

ymekuria (approved):
This is great @mitschabaude! I just had a question.

src/lib/gadgets/bitwise.ts
```diff
@@ -37,18 +38,12 @@ function not(a: Field, length: number, checked: boolean = false) {
...
-  });
-
-  allOnesF.assertEquals(allOnes);
+  let allOnes = new Field(2n ** BigInt(length) - 1n);
 
   if (checked) {
     return xor(a, allOnes, length);
   } else {
-    return allOnes.sub(a);
+    return allOnes.sub(a).seal();
```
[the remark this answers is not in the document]
mitschabaude: It makes sure that the constraint created by the subtraction is added at this point. In my tests I found that the constraint system created by the unchecked version is empty.

That's because additions and subtractions are treated as a form of lazy constraint / AST, which are only turned into an actual constraint once the result is included in some multiplication or assertion or custom gate.

`seal()` creates a new variable that is asserted equal with the old variable, which causes snarky to concretize the AST into constraints. In the case of a gadget, I thought this was the better behavior, as it's normalized and not dependent on the remaining circuit where this subtraction will end up in the CS.

src/lib/gadgets/common.ts
```diff
@@ -27,6 +35,31 @@ function exists<N extends number, C extends () => TupleN<bigint, N>>(
   return TupleN.fromArray(n, vars);
 }
 
+/**
+ * Given a Field, collapse its AST to a pure Var. See {@link FieldVar}.
+ *
+ * This is useful to prevent rogue Generic gates added in the middle of gate chains,
+ * which are caused by snarky auto-resolving constants, adds and scales to vars.
```
jspada, line 42: This is really awesome

src/lib/gadgets/common.ts
```diff
@@ -27,6 +35,31 @@ function exists<N extends number, C extends () => TupleN<bigint, N>>(
...
+ *
+ * Same as `Field.seal()` with the difference that `seal()` leaves constants as is.
+ */
+function toVar(x: Field | bigint) {
```
jspada, line 46: Maybe this is better called `ConstrainConstantHere` or something like that?

src/lib/gadgets/common.ts
```diff
@@ -27,6 +35,31 @@ function exists<N extends number, C extends () => TupleN<bigint, N>>(
...
+  if (x instanceof Field && x.value[1] === FieldType.Var) return x;
+  let xVar = existsOne(() => Field.from(x).toBigInt());
+  xVar.assertEquals(x);
+  return xVar;
+}
+
+/**
+ * Apply {@link toVar} to each element of a tuple.
+ */
+function toVars<T extends Tuple<Field | bigint>>(
```
jspada, line 57: `ConstrainConstantsHere`?

jspada (approved):
I think this code is great.  Approved with some minor suggestions and reminders.

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -183,3 +193,63 @@ await equivalentAsync({ from: [field], to: field }, { runs: 3 })(
...
+  { from: [Field] },
+  Bitwise.rawMethods.notUnchecked,
+  ifNotAllConstant(contains('Generic'))
+);
+
+constraintSystem(
+  'and',
+  { from: [Field, Field] },
+  Bitwise.rawMethods.and,
+  ifNotAllConstant(contains(xorChain(64)))
```
querolita, line 228: Are you only supporting And64 then?

src/lib/gadgets/bitwise.unit-test.ts
```diff
@@ -183,3 +193,63 @@ await equivalentAsync({ from: [field], to: field }, { runs: 3 })(
...
+);
+
+let rotChain: GateType[] = ['Rot64', 'RangeCheck0', 'RangeCheck0'];
```
querolita, line 231: Nice, but remember to have copy constraints to the RangeCheck0's to constant zero for the two most significant sublimbs to make it 64-bit instead of 88-bit.

querolita (approved):
Very useful for debugging, congrats!

src/lib/gadgets/bitwise.ts
```diff
@@ -252,6 +247,10 @@ function rot(
     }
   );
 
+  // flush zero var to prevent broken gate chain (zero is used in rangeCheck64)
+  // TODO this is an abstraction leak, but not clear to me how to improve
+  toVar(0n);
```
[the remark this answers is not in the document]
mitschabaude: @querolita I think my tests already exhibit that case, because I am creating random linear combinations of variables as inputs, something like `x + 20y + 3z + 17`, and this will generate an even or odd number of generic gates when passed to a gate, depending on the length of the linear combination. Since we test many random inputs, I'm pretty sure all cases are covered

src/lib/gadgets/bitwise.ts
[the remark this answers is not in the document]
mitschabaude: The snarky version of `seal()` could be used for that! The o1js version not because it doesn't "seal" constants. The o1js version was intended more to prevent unnecessary large numbers of constraints when the same long AST is flushed to generic gates again and again. The function added in this PR that does what you descirbe @jspada is called `toVar()`

src/lib/gadgets/bitwise.unit-test.ts
mitschabaude: no, this tests a specific example circuit -- the `and()` method on the `Bitwise` zkprogram, which happens to use 64-bit and. but we support all bit lengths

src/lib/gadgets/bitwise.ts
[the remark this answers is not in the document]
mitschabaude: > In OCaml we had a PR where whenever you add a gate it adds a "shadow row" that specifies what the next row should be with Some or None (for anything), per cell in the next row.

That sounds interesting! Do you know where that PR is?

PR #1256, opened 2023-11-20
Lightnet namespace API updates.

shimkiv:
- `listAcquiredKeyPairs()`

---

Related namespace API: https://github.com/o1-labs/o1js/pull/1167

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
-- `Lightnet` namespace API updates https://github.com/o1-labs/o1js/pull/1256
+- `Lightnet` namespace API updates with added `listAcquiredKeyPairs()` method https://github.com/o1-labs/o1js/pull/1256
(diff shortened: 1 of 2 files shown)


mitschabaude:
* What does it mean if an account is "locked"?
* Can we abstract this away from users? Is there any benefit to not always having all accounts unlocked?

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Changed
 
+- `Lightnet` namespace API updates https://github.com/o1-labs/o1js/pull/1256
```
Trivo25, line 30: nit: We can leave this out since we already talk about it under `CHANGED` - no need to double-specify what has changed

Trivo25 (approved):

CHANGELOG.md
mitschabaude: the other changelog entry modified here is from much earlier!

Trivo25: Ohh right - the github diff tricked me :D Then thats fine, we should just specify what actually changed!

PR #1259, opened 2023-11-21
add support for bitwise gadgets on UInt64 and UInt32 type

Trivo25:


src/lib/gadgets/arithmetic.ts
```diff
@@ -0,0 +1,40 @@
...
+    let nBigInt = n.toBigInt();
+    let q = nBigInt / (1n << 32n);
+    let r = nBigInt - q * (1n << 32n);
+    return [new Field(q), new Field(r)];
+  });
+  let [q, r] = qr;
+
+  // I think we can "skip" this here and do it in the caller - see rotate32
+  // rangeCheck32(q);
+  // rangeCheck32(r);
```
mitschabaude, line 28: we need those! if not both are constrained, either of them can be invalid

src/lib/gadgets/arithmetic.ts
```diff
@@ -0,0 +1,39 @@
...
+    };
+  }
+
+  let qr = Provable.witness(Provable.Array(Field, 2), () => {
+    let nBigInt = n.toBigInt();
+    let q = nBigInt / (1n << 32n);
+    let r = nBigInt - q * (1n << 32n);
+    return [new Field(q), new Field(r)];
+  });
+  let [quotient, remainder] = qr;
```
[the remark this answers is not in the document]
mitschabaude: woah wtf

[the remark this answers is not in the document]
mitschabaude: fascinating, the inferred return type seems to be fully determined by the callback and result type, and completely overrides the `Provable<T>` type. So this is inferred to be `[Field, Field]` but then `Provable.Array(Field, 2)` doesn't match that.

This can probably be changed in either of two ways -- tweaking the type signature of `witness()`, or making `Provable.Array()` infer a tuple. The second would make it harder to use in some circumstances though.

We should probably separate "tuple" and "array", you can already do the former with `provableTuple` and this works:
```ts
let [quotient, remainder] = Provable.witness(
    provableTuple([Field, Field]),
    () => { ... });
```

src/lib/gadgets/range-check.ts
```diff
@@ -12,6 +13,21 @@ export {
   twoLMask,
 };
 
+/**
+ * Asserts that x is in the range [0, 2^32)
+ */
+function rangeCheck32(x: Field) {
```
Trivo25, line 19: I am happy to refactor this as well and move all the helpers (mainly `rangeCheckHelper`) out of `Field` entirely if we think we should

mitschabaude: yes we should! btw `rangeCheckHelper` is a bad API that should've never been exposed in this way. It returns a value that fits within `16*n` bits but is otherwise unconstrained. Only by asserting it to be equal to something else, it becomes a range check. But our current docs frame it as it would do provable reduction modulo `16n`

src/lib/gadgets/range-check.ts
```diff
@@ -12,6 +13,22 @@ export {
...
+  if (x.isConstant()) {
+    if (x.toBigInt() >= 1n << 32n) {
+      throw Error(`rangeCheck32: expected field to fit in 32 bits, got ${x}`);
+    }
+    return;
+  }
+
+  // can we make this more efficient? its 3 gates :/
+  let actual = x.rangeCheckHelper(32);
```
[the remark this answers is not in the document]
mitschabaude: also with two 64 bit range checks and half a generic gate you can do it (2.5 rows):
```ts
rangeCheck64(x);
rangeCheck64(x.mul(1n << 32n));
```

src/lib/gadgets/bitwise.ts
```diff
@@ -209,16 +210,42 @@ function rotate(
...
       `rotation: expected field to be at most 64 bits, got ${field.toBigInt()}`
     );
     return new Field(Fp.rot(field.toBigInt(), bits, direction));
   }
-  const [rotated] = rot(field, bits, direction);
+  const [rotated] = rot64(field, bits, direction);
+  return rotated;
+}
+
+function rotate32(
```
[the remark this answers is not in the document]
mitschabaude: I would've guessed:
* 2.5 per `rangeCheck32` = 7.5
* 0.5 for `2^r * field === 2^32 * excess + shifted` -- maybe snarky doesn't do this optimally
* 0.5 for `rotated === shifted + excess`
= 8.5

src/lib/gadgets/range-check.ts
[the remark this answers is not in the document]
Trivo25: Oh nice! Can we take this as a full `rangeCheck32` for all the checks, replacing `rangeCheckHelper`?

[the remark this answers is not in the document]
Trivo25: It actually seems to be slightly less efficient :thinking:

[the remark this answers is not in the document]
mitschabaude: yes, we shouldn't use `rangeCheck64` probably because it adds a custom gate that we might otherwise avoid

src/lib/gadgets/bitwise.ts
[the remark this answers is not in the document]
mitschabaude: you can tell by using `printGates(gates)`. Next to each generic gate it prints two vectors of 5 coefficients, except if it's half a gate, then it only prints one vector

src/lib/gadgets/arithmetic.unit-test.ts
```diff
@@ -0,0 +1,68 @@
...
+import { Random } from '../testing/property.js';
+import { provable } from '../circuit_value.js';
+import { assert } from './common.js';
+
+const maybeField = {
+  ...field,
+  rng: Random.map(Random.oneOf(Random.field, Random.field.invalid), (x) =>
+    mod(x, Field.ORDER)
+  ),
+};
```
mitschabaude, line 20: since we do `mod(x, Field.ORDER)` on the result, it gets mapped into the valid range for field elements. so this `maybeField` ends up being the same as `field` with a slightly different random distribution.

I guess, just use `field` instead?

mitschabaude (commented):
I know still a draft, but looks great to me!

src/lib/gadgets/range-check.ts
```diff
@@ -1,10 +1,56 @@
...
+  }
+
+  let actual = rangeCheckHelper(32, x);
+  actual.assertEquals(x);
+}
+
+/**
+ * Create a new {@link Field} element from the first `length` bits of this {@link Field} element.
+ */
+function rangeCheckHelper(length: number, x: Field) {
```
Trivo25, line 35: what about calling this `Gadgets.truncate` and putting into the common gadget section?

mitschabaude: again, `truncate()` suggests that the truncation is provable. But it's just the witness that happens to be the truncated version of the input -- no constraints enforce this. The only thing that constraints enforce is that the return value of `rangeCheckHelper` fits within 16*n bits. So, it's really just a "range check helper" and not truncation gadget.

My proposal: Create a gadget `rangeCheckMultiple16(x: Field, nBits: <multiple of 16>)` which uses `rangeCheckHelper(x).assertEquals(x)` internally, but remove `rangeCheckHelper` from the public API.

Also, would be nice to implement that gadget in TS on top of the raw EC scalar gate 👀

src/lib/gadgets/arithmetic.unit-test.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+await Arithmetic.compile();
+
+const divMod32Helper = (x: bigint) => {
+  let quotient = x / (1n << 32n);
+  let remainder = x - quotient * (1n << 32n);
+  return { remainder, quotient };
+};
+const divMod32Output = record({ remainder: field, quotient: field });
```
mitschabaude, line 39: nice

src/lib/gadgets/gadgets.ts
```diff
@@ -39,6 +50,49 @@ const Gadgets = {
...
+   * ```ts
+   * const x = Provable.witness(Field, () => Field(12345678n));
+   * Gadgets.rangeCheck32(x); // successfully proves 32-bit range
+   *
+   * const xLarge = Provable.witness(Field, () => Field(12345678901234567890123456789012345678n));
+   * Gadgets.rangeCheck32(xLarge); // throws an error since input exceeds 32 bits
+   * ```
+   *
+   * **Note**: Small "negative" field element inputs are interpreted as large integers close to the field size,
+   * and don't pass the 32-bit check. If you want to prove that a value lies in the int64 range [-2^31, 2^31),
```
mitschabaude, line 74:
```suggestion
   * and don't pass the 32-bit check. If you want to prove that a value lies in the int32 range [-2^31, 2^31),
```

src/lib/gadgets/gadgets.ts
```diff
@@ -39,6 +50,49 @@ const Gadgets = {
...
+   * you could use `rangeCheck32(x.add(1n << 31n))`.
+   */
+  rangeCheck32(x: Field) {
+    return rangeCheck32(x);
+  },
+
+  /**
+   * Create a new {@link Field} element from the first `length` bits of this {@link Field} element.
```
mitschabaude, line 82: 👎🏻 for reasons discussed

src/lib/gadgets/gadgets.ts
```diff
@@ -482,6 +574,42 @@ const Gadgets = {
    * **Note:** This interface does not contain any provable methods.
    */
   Field3,
+  /**
+   * Division modulo 2^32. The operation decomposes a {@link Field} element in the range [0, 2^64] into two 32-bit limbs, `remainder` and `quotient`, using the following equation: `n = quotient * 2^32 + remainder`.
+   *
+   * **Note:** The Gadget expects the input to be in the range [0, 2^64]. If the input exceeds 64 bits, the gadget is invalid and fails to prove correct execution.
```
mitschabaude, line 580: Two comments:
1) `quotient * 2^32 + remainder` can be at most $2^{64} -1$, not $2^{64}$
2) `divMod32` actually acts as a range check on the input, so I'd tweak the wording from "the gadget becomes invalid" to "the gadget fails" for out-of-range inputs.

```suggestion
   * Division modulo 2^32. The operation decomposes a {@link Field} element in the range [0, 2^64) into two 32-bit limbs, `remainder` and `quotient`, using the following equation: `n = quotient * 2^32 + remainder`.
   *
   * **Note:** The gadget acts as a proof that the input is in the range [0, 2^64). If the input exceeds 64 bits, the gadget fails.
```

src/lib/gadgets/gadgets.ts
```diff
@@ -482,6 +574,42 @@ const Gadgets = {
...
+  divMod32,
+
+  /**
+   * Addition modulo 2^32. The operation adds two {@link Field} elements in the range [0, 2^64] and returns the result modulo 2^32.
+   *
+   * Asserts that the result is in the range [0, 2^32) using {@link Gadgets.rangeCheck32}.
+   *
+   * It uses {@link Gadgets.divMod32} internally by adding the two {@link Field} elements and then decomposing the result into `remainder` and `quotient` and returning the `remainder`.
+   *
+   * **Note:** The Gadget expects the input to be in the range [0, 2^64]. If the input exceeds 64 bits, the gadget is invalid and fails to prove correct execution.
```
mitschabaude, line 602:
```suggestion
   * **Note:** The gadget assumes both inputs to be in the range [0, 2^64). When called with non-range-checked inputs, be aware that the sum `a + b` can overflow the native field and the gadget can succeed but return an invalid result.
```

Just trying to clarify the behaviour w.r.t to input range checks.

For example: `Gadgets.addMod32(Field(-1), Field(1))` will return `0` even though
$$(p-1) + 1 \ne  0 \mod{2^{32}}$$

tests/vk-regression/vk-regression.json
```diff
@@ -168,7 +168,11 @@
   "Bitwise Primitive": {
     "digest": "Bitwise Primitive",
     "methods": {
-      "rot": {
+      "rot32": {
+        "rows": 31,
+        "digest": "3e39e3f7bfdef1c546700c0294407fc2"
+      },
```
mitschabaude, line 174: I can't wait to see how many constraints sha256 takes! if it's mostly using no custom gates we end up with a similar efficiency as those guys who originally implemented it

src/lib/int.ts
```diff
@@ -200,10 +201,185 @@ class UInt64 extends CircuitValue {
...
+   *
+   * ```
+   *
+   * @param a - The value to apply NOT to.
+   * @param checked - Optional boolean to determine if the checked or unchecked not implementation is used. If it
+   * is set to `true` the {@link Gadgets.xor} gadget is reused. If it is set to `false`, NOT is implemented
+   *  as a subtraction of the input from the all one bitmask. It is set to `false` by default if no parameter is provided.
+   *
+   */
+  not(checked = true) {
```
mitschabaude, line 276: I vote for always using `checked = false` here. it's just 1 row while the xor for 64 bits will be 4 + 1 rows IIRC

src/lib/int.ts
```diff
@@ -215,12 +391,17 @@ class UInt64 extends CircuitValue {
...
-        .rangeCheckHelper(UInt64.NUM_BITS)
-        .equals(xMinusY);
-      let yMinusXFits = yMinusX
-        .rangeCheckHelper(UInt64.NUM_BITS)
-        .equals(yMinusX);
+
+      let xMinusYFits = Gadgets.rangeCheckHelper(
+        UInt64.NUM_BITS,
+        xMinusY
+      ).equals(xMinusY);
```
mitschabaude, line 398: aah.. now I get why `rangeCheckHelper()` was exposed and not just `rangeCheck()` -- because here we are using it with `equals()` instead of `assertEquals()`

Maybe we can create two different gadgets for those cases?
```ts
rangeCheck(n: number, x: Field): void; // rangeCheckHelper + assertEquals
isDefinitelyInRange(n: number, x: Field): Bool; // rangeCheckHelper + equals
```

weird name `isDefinitelyInRange()` because a false result doesn't prove that it's not in the range. feel free to pick a better name.

src/lib/int.ts
```diff
@@ -523,25 +711,201 @@ class UInt32 extends CircuitValue {
...
+   *
+   * @param a - The value to apply NOT to.
+   * @param checked - Optional boolean to determine if the checked or unchecked not implementation is used. If it
+   * is set to `true` the {@link Gadgets.xor} gadget is reused. If it is set to `false`, NOT is implemented
+   *  as a subtraction of the input from the all one bitmask. It is set to `false` by default if no parameter is provided.
+   *
+   */
+  not(checked = true) {
+    return Gadgets.not(this.value, UInt32.NUM_BITS, checked);
+  }
```
mitschabaude, line 804: IMO remove `checked` and set to false

src/lib/gadgets/range-check.ts
mitschabaude: EDIT: see https://github.com/o1-labs/o1js/pull/1259#discussion_r1407701112

mitschabaude:
@Trivo25 while you're here -- can you add correctly typed constructors for `UInt32` and `UInt64`?

CHANGELOG.md
```diff
@@ -19,6 +19,25 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+- `Gadgets.rangeCheck32` range check for 32 bit values https://github.com/o1-labs/o1js/pull/1259
+- `Gadgets.rangeCheckHelper` for returning a new Field element of the first `length` bits of an element https://github.com/o1-labs/o1js/pull/1259
+- `Gadgets.addMod32` addition modulo 2^32 https://github.com/o1-labs/o1js/pull/1259
+- Expose new bitwise gadgets on `UInt32` and `UInt64` https://github.com/o1-labs/o1js/pull/1259
+  - bitwise XOR via `{UInt32, UInt64}.xor`
+  - bitwise NOT via `{UInt32, UInt64}.not`
+  - bitwise ROTATE via `{UInt32, UInt64}.rotate`
+  - bitwise LEFTSHIFT via `{UInt32, UInt64}.leftShift`
+  - bitwise RIGHTSHIFT via `{UInt32, UInt64}.rightShift`
+  - bitwise AND via `{UInt32, UInt64}.and`
```
mitschabaude, line 39: elsewhere we often write `()` to indicate functions, `Gadgets.rotate32()`

src/lib/int.ts
```diff
@@ -523,25 +711,201 @@ class UInt32 extends CircuitValue {
...
+   * @example
+   * ```ts
+   * const x = UInt32.from(0b001100); // 12 in binary
+   * const y = x.leftShift(2); // left shift by 2 bits
+   * y.assertEquals(0b110000); // 48 in binary
+   * ```
+   */
+  leftShift(bits: number) {
+    return Gadgets.leftShift(this.value, bits);
+  }
```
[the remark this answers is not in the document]
Trivo25: Oh thats right, it can overflow :X

src/lib/int.ts
```diff
@@ -523,25 +711,201 @@ class UInt32 extends CircuitValue {
...
+   * @example
+   * ```ts
+   * const x = UInt32.from(0b001100); // 12 in binary
+   * const y = x.rightShift(2); // left shift by 2 bits
+   * y.assertEquals(0b000011); // 48 in binary
+   * ```
+   */
+  rightShift(bits: number) {
+    return Gadgets.rightShift(this.value, bits);
+  }
```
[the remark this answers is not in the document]
Trivo25: Why is that? using the 64bit right shift shouldn't matter, since all bits are discarded anyway and we only apply the operation on the 32bits given 🤔

src/lib/gadgets/gadgets.ts
Trivo25: What about

```
Creates an _unconstrained witness_ from the first `length` bits of the {@link Field} element.
In order to constrain the new element further, you can use `assertEquals` to constrain the new witness to the original element.
```

src/lib/gadgets/gadgets.ts
```diff
@@ -39,6 +52,80 @@ const Gadgets = {
...
+   */
+  rangeCheckN(n: number, x: Field, message?: string) {
+    return rangeCheckN(n, x, message);
+  },
+
+  /**
+   * Checks whether the input value is in the range [0, 2^n). `n` must be a multiple of 16.
+   *
+   * This function proves that the provided field element can be represented with `n` bits.
+   * If the field element exceeds `n` bits, `Bool(false)` is returned and `Bool(true)` otherwise.
```
mitschabaude, line 112: this is misleading, because it only proves that the result is in range if the return value of `isInRangeN()` is true. It doesn't prove that the result is _not_ in range if the return value is false.

See my comment on where this is used: https://github.com/o1-labs/o1js/pull/1259/files#r1409932102

So maybe better to rephrase the doccomment, or frame it as a helper function for now. I can follow up with some code that actually proves the full thing, which I think would also make the UInt64 comparison cheaper (currently the comparison calls rangeCheckHelper twice, but it can be done with one call)

mitschabaude (approved):
Left shift and right shift look good to me!

My only concern now is the misleading documentation of the `isInRange()` gadget that I requested :D But I think it can be just fixed in a follow-up where we make it do what it says, so approving!

mitschabaude:
Created a follow-up issue that doesn't need to be addressed here: https://github.com/o1-labs/o1js/issues/1286

src/lib/gadgets/gadgets.ts
```diff
@@ -223,15 +373,15 @@ const Gadgets = {
    * @example
    * ```ts
    * const x = Provable.witness(Field, () => Field(0b001100)); // 12 in binary
-   * const y = Gadgets.rightShift(x, 2); // right shift by 2 bits
+   * const y = Gadgets.rightShift64(x, 2); // right shift by 2 bits
    * y.assertEquals(0b000011); // 3 in binary
    *
    * const xLarge = Provable.witness(Field, () => Field(12345678901234567890123456789012345678n));
    * rightShift(xLarge, 32); // throws an error since input exceeds 64 bits
```
barriebyron, line 380: two questions:
1. will we have a leftShift64 operation?
2. should this comment be updated to rightShift64 or the second part of the example is no longer needed?
```suggestion
   * rightShift64(xLarge, 32); // throws an error since input exceeds 64 bits
```

PR #1260, opened 2023-11-21
Fix TS build and Mina tests

mitschabaude:
- ffmul gate wrapper
- ffmul witnesses
- raw gate wrapper
- ffmul constraints
- constant case
- start testing
- another small witness test
- add proof mul test which doesn't work
- revert test to normal
- refactor  common mul logic into function
- ff inverse
- fix inverse
- division
- improve zkprogram analyzemethods return
- fix division
- test inv and div
- minor
- add cs tests for mul
- fix: add missing mrc
- introduce assertmul and use it for inv and div
- add ff tests with unreduced inputs
- a bit of clean up
- remove unused logic
- add to gadgets namespace, mul doccomment
- fix unit tests and code tweak
- document inv and div
- bindings
- minor
- minor tweaks + changelog
- fix: don't range-check 2-bit carry
- consistent lower-casing of limb size constant
- remove remainder bounds check in mul() gadget
- fixup unit test
- address feedback
- fix ts build
- fix mina tests
- bindings

### DIFF
--- a/src/lib/gadgets/bitwise.ts
+++ b/src/lib/gadgets/bitwise.ts
@@ -212,7 +212,7 @@ function rotate(
       field.toBigInt() < 2n ** BigInt(MAX_BITS),
       `rotation: expected field to be at most 64 bits, got ${field.toBigInt()}`
     );
-    return new Field(Fp.rot(field.toBigInt(), bits, direction));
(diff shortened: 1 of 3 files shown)


MartinMinkov (approved):
There's some build errors but I'm pre approving!

PR #1262, opened 2023-11-21
Foreign fields 5: Optimized multiplication of sums

mitschabaude:
- start ec add
- much more efficient ecadd
- chain add into mul
- adapt to ffmul updates
- replace custom logic in ec add with easy to use assertRank1 method
- chore(package.json): bump version from 0.14.1 to 0.14.2 for new release
- docs(CHANGELOG.md): update unreleased section link and add version 0.14.2 section to reflect recent changes in the project
- delete ec gadget from this branch
- adapt to main
- replace mul input rcs with generic gates
- enable strict typing of variable fields
- optimized assertOneOf gadget
- make new assertRank1 more efficient and enable it
- expose assertMul and Sum
- add assertion to prevent invalid multiplication
- document assertMul


querolita:
Before giving an actual code review, I will start by writing comments on the above explanation. 

>  the ffadd gate constrains its result a + b in compact limb form [r01, r2]

This is true for FFMul, where the remainder is indeed given in compact form in the first 2 columns of the next row. But in the FFAdd gate, the result was given in 3-limb form, in the first 3 columns of the next row so it could chain easily with FFAdd and FFMul. Thus, what exactly are you referring to with the above sentence?

> the naive version also suffers because it computes the left and right side separately, and so applies range-checks to the same result twice. 

You mean one would naively range check `x` and `y` twice? Meaning, when computing $x^2$, once for "left" $x$ and again for "right" $x$? If this is what you meant then yes, this should be avoided and only range check it once to avoid duplicates.

src/lib/gadgets/gadgets.ts
```diff
@@ -474,6 +474,45 @@ const Gadgets = {
...
+     *
+     * Note: This is much more efficient than using {@link ForeignField.add} and {@link ForeignField.sub} separately to
+     * compute the multiplication inputs and outputs, and then using {@link ForeignField.mul} to constrain the result.
+     *
+     * The sums passed into this gadgets are "lazy sums" created with {@link ForeignField.Sum}.
+     * You can also pass in plain {@link Field3} elements.
+     *
+     * **Assumptions**: The assumptions on the _summands_ are analogous to the assumptions described in {@link ForeignField.mul}:
+     * - each summand's limbs are in the range [0, 2^88)
+     * - summands that are part of a multiplication input satisfy `x[2] <= f[2]`
```
querolita, line 489: Is this an assumption or constrained somewhere?

src/lib/gadgets/gadgets.ts
```diff
@@ -474,6 +474,45 @@ const Gadgets = {
...
+     * @throws if the provided multiplication result is not correct modulo f.
+     *
+     * @example
+     * ```ts
+     * // we assume that x, y, z, a, b, c are range-checked, analogous to `ForeignField.mul()`
+     * let xMinusY = ForeignField.Sum(x).sub(y);
+     * let aPlusBPlusC = ForeignField.Sum(a).add(b).add(c);
+     *
+     * // assert that (x - y)*z = a + b + c mod f
+     * ForeignField.assertMul(xMinusY, z, aPlusBPlusC, f);
```
querolita, line 503: Maybe it would be helpful if you show what are the concrete gates that are created when calling `assertMul`. Well, maybe this is commented somewhere else, if so, I will find it while reviewing the other files of the PR.

src/lib/gadgets/basic.ts
```diff
@@ -0,0 +1,79 @@
...
+
+// TODO: create constant versions of these and expose on Gadgets
+
+/**
+ * Assert that a value equals one of a finite list of constants:
+ * `(x - c1)*(x - c2)*...*(x - cn) === 0`
+ *
+ * TODO: what prevents us from getting the same efficiency with snarky DSL code?
+ */
+function assertOneOf(x: Field, allowed: [bigint, bigint, ...bigint[]]) {
```
querolita, line 19: What does this notation `... bigint[]` mean?

querolita: Is it like a tail of a list with potentially more bigints?

src/lib/gadgets/basic.ts
```diff
@@ -0,0 +1,79 @@
...
+  let xv = toVar(x);
+  let [c1, c2, ...c] = allowed;
+  let n = c.length;
+  if (n === 0) {
+    // (x - c1)*(x - c2) === 0
+    assertBilinear(xv, xv, [1n, -(c1 + c2), 0n, c1 * c2]);
+    return;
+  }
+  // z = (x - c1)*(x - c2)
+  let z = bilinear(xv, xv, [1n, -(c1 + c2), 0n, c1 * c2]);
```
querolita, line 29: Is there any reason why you used `-(c1+c2)` instead of `-c1` and `-c2` where it is a `0n` now? Meaning, efficiency-wise. Just curious, also wonder if it changes anyhow the way the term `-(c1+c2)` is encoded into the circuit.

src/lib/gadgets/basic.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  for (let i = 0; i < n; i++) {
```
querolita, line 31: Because this only uses generics, I guess the maximum length of the list doesn't really matter. Meaning, you don't need to set a bound to X terms so that the underneath constraints are not higher than degree X. In other words, this is just concatenating generics and copy constraints between outputs and inputs among them, right? Whether or not the check becomes too costly in terms of rows, just depends on the zkapp developer criterion, no?

querolita (commented):
Leaving some comments for now regarding collateral changes introduced in the PR, which are not just the FFMul optimization. Leaving the rest of it (the most remarkable part) for later (but not too much).

mitschabaude:
> You mean one would naively range check x and y twice? Meaning, when computing , once for "left" and again for "right" ?

I mean that you'd check `x^2 - y^2` when calling `sub()`, and you'd also range-check `(x - y)(x + y)` when calling `mul()`, and then you assert they are equal. But if you already know they should be equal you could avoid one of the range checks. It's a minor point, of course we could add a flag to `sub()` or `mul()` that tells it to skip the range check. The nice thing about this API here is that it doesn't need any such flags

src/lib/gadgets/basic.ts
mitschabaude: yes! I made the type require at least two inputs so that I don't have to implement the case of 0 or 1 inputs

src/lib/circuit_value.ts
```diff
@@ -454,6 +457,93 @@ function Struct<
...
+and Provable.asProver() blocks, which execute outside the proof.
+`);
+    assert(this.option.isSome, 'Empty `Unconstrained`'); // never triggered
+    return this.option.value;
+  }
+
+  /**
+   * Modify the unconstrained value.
+   */
+  set(value: T) {
```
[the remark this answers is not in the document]
mitschabaude: yes! that's not an issue actually

src/lib/gadgets/gadgets.ts
mitschabaude: assumption on the inputs

mitschabaude: (needs to be constrained by the caller)

mitschabaude:
This is the full constraint system for the `assertMul()` example I use in the test. It does `(x - y)(x + y) === x2 + y2`, where `x2` and `y2` are just witnessed (not constrained)
```
0    Generic         [1 0 0 1 0] [-1 0 -1 1 0]      0->2, 1->3, 2->5, 3->4, 4->(1,3), 5->0
1    Generic         [1 1 -1 0 0] [2894.. 2894.. -1 0 0] 0->(2,3), 1->5, 2->(6,4), 3->(0,1), 4->(2,6), 5->1
2    ForeignFieldAdd [3094.. 3094.. 1208.. 1]       0->(6,3), 1->(9,1), 2->(9,2), 3->(8,3), 4->(9,4), 5->(9,5), 6->(1,4)
3    Zero                                           0->(6,5), 1->(10,4), 2->(10,5)
4    ForeignFieldAdd [3094.. 3094.. 1208.. -1]      
5    Zero                                           0->(20,0), 1->(20,1), 2->(11,1)
6    Generic         [-1 0 -1 1 0] [1 1 -1 0 0]     0->1, 1->(7,0), 2->(7,3), 3->(8,0), 4->(1,2), 5->(10,3)
7    Generic         [2894.. 2894.. -1 0 0] [1 0 0 1 0] 0->4, 1->(9,6), 2->(8,4), 3->5, 4->(6,0), 5->(6,2)
8    Generic         [1 1 -1 0 0] [-1 1 -1 0 0]     0->(9,0), 1->5, 2->(10,0), 3->(9,3), 4->(7,2), 5->1
9    ForeignFieldAdd [3094.. 3094.. 1208.. -1]      0->(2,0), 1->(2,1), 2->(2,2), 3->(1,0), 4->(2,4), 5->(2,5), 6->(7,1)
10   ForeignFieldMul [1208.. 4294.. 0 3082..]       0->(8,2), 3->(3,0), 4->(3,1), 5->(3,2), 6->(12,0)
11   Zero                                           0->(20,2), 1->(5,2), 2->(16,0), 3->(17,0), 4->(18,0), 5->(14,0), 6->(13,0)
12   RangeCheck0     [0]                            0->(10,6), 1->(15,3), 2->(15,4)
13   RangeCheck0     [0]                            0->(11,6), 1->(15,5), 2->(15,6)
14   RangeCheck1                                    0->(11,5), 1->(18,1)
15   Zero                                           3->(12,1), 4->(12,2), 5->(13,1), 6->(13,2)
16   RangeCheck0     [0]                            0->(11,2), 1->(19,3), 2->(19,4)
17   RangeCheck0     [0]                            0->(11,3), 1->(19,5), 2->(19,6)
18   RangeCheck1                                    0->(11,4), 1->(20,3)
19   Zero                                           3->(16,1), 4->(16,2), 5->(17,1), 6->(17,2)
20   Generic         [1 3094.. -1 0 0] [1 0 0 0 0]  0->(5,0), 1->(5,1), 2->(11,0), 3->(14,1)
```

src/lib/gadgets/gadgets.ts
mitschabaude: Ok, I posted an example constraint system here now: https://github.com/o1-labs/o1js/pull/1262#issuecomment-1823345067

src/lib/gadgets/basic.ts
mitschabaude: No reason! I could've used `-c1` and `-c2` separately as well. Yeah if you look at the function below, those constants directly become generic gate coefficients. Don't think it makes a difference efficiency-wise (but if anything, more 0s is probably better)

src/lib/gadgets/basic.ts
mitschabaude: yes exactly, just concatenates generic gates

Trivo25 (approved):

src/lib/gadgets/foreign-field.ts
```diff
@@ -431,3 +456,216 @@ function combine2([x0, x1]: bigint3 | [bigint, bigint]) {
...
+
+        // add with carry, only on the lowest limb
+        x0 = x0 + sign * xi[0] - overflow * f0;
+        let carry = x0 >> l;
+        return [carry, overflow];
+      });
+      overflows.push(overflow);
+
+      // constrain carry
+      assertOneOf(carry, [0n, 1n, -1n]);
```
jspada, line 632: don't you need to constrain `overflow` also?

src/lib/gadgets/foreign-field.ts
```diff
@@ -431,3 +456,216 @@ function combine2([x0, x1]: bigint3 | [bigint, bigint]) {
...
+          .sub(overflow.mul(f0))
+          .sub(carry.mul(1n << l))
+      );
+      x0s.push(x0);
+    }
+
+    // ffadd chain
+    let x = xs[0];
+    for (let i = 0; i < n; i++) {
+      let { result, overflow } = singleAdd(x, xs[i + 1], signs[i], f);
```
jspada, line 647: ah, maybe this is constraining the overflows as `0 | 1`?

src/lib/gadgets/foreign-field.ts
```diff
@@ -431,3 +456,216 @@ function combine2([x0, x1]: bigint3 | [bigint, bigint]) {
...
+      // wire low limb and overflow to previous values
+      result[0].assertEquals(x0s[i]);
```
jspada, line 649: looks good.

src/lib/gadgets/foreign-field.ts
```diff
@@ -431,3 +456,216 @@ function combine2([x0, x1]: bigint3 | [bigint, bigint]) {
...
+  // we assume that all summands si are bounded with si[2] <= f[2] checks, which implies si < 2^k where k := ceil(log(f))
+  // our assertion below gives us
+  // |x|*|y| + q*f + |r| < (x.length * y.length) 2^2k + 2^2k + 2^2k < 3 * 2^(2*258) < 2^264 * (native modulus)
+  assert(
+    BigInt(Math.ceil(Math.sqrt(x.length * y.length))) * f < 1n << 258n,
+    `Foreign modulus is too large for multiplication of sums of lengths ${x.length} and ${y.length}`
+  );
+
+  // finish the y and xy sums with a zero gate
+  let y0 = y.finishForMulInput(f);
```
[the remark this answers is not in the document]
mitschabaude: but `finish()` is already reserved for when you don't use it as multiplication input!

src/lib/gadgets/foreign-field.ts
mitschabaude: exactly! because this overflow is wired to the one pushed to the array earlier

src/lib/field.ts
```diff
@@ -11,10 +11,12 @@ export { Field };
 
 // internal API
 export {
-  ConstantField,
   FieldType,
   FieldVar,
   FieldConst,
+  ConstantField,
```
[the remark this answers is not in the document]
mitschabaude: heh, it gets a bit confusing with the similar names.

* `Field` - o1js main circuit type. It is a class with one property, `value: FieldVar`
* `FieldVar` - the JSOO representation of OCaml's main circuit type, which in different places is called `Field.t`, `Circuit.Field.t`, or `field Cvar.t`
* `FieldConst` - the JSOO representation of the _constant_ field elements, i.e. the `field` in `field Cvar.t`. In some places called `Field.Constant.t`

Now, `FieldVar` / `Cvar.t` is an enum with four possible variants: `Constant`, `Var`, `Add`, `Scale`. We use these as modifiers in front of `Field` / `FieldVar` to name the narrower type that only can be this variant.

For example:
* `ConstantFieldVar` - a `FieldVar` which is of the `Constant` variant
* `VarFieldVar` - a `FieldVar` which is of the `Var` variant
* `ConstantField` - a `Field` whose `value` is of the `Constant` variant
* `VarField` - a `Field` whose `value` is of the `Var` variant

src/lib/gadgets/foreign-field.ts
mitschabaude: no, see https://github.com/o1-labs/o1js/pull/1262#discussion_r1415877018

src/lib/gadgets/basic.ts
```diff
@@ -0,0 +1,79 @@
...
+function assertBilinear(
+  x: VarField,
+  y: VarField,
+  [a, b, c, d]: TupleN<bigint, 4>,
+  z?: VarField
+) {
+  // b*x + c*y - z? + a*x*y + d === 0
+  Gates.generic(
+    { left: b, right: c, out: z === undefined ? 0n : -1n, mul: a, const: d },
+    { left: x, right: y, out: z === undefined ? x : z }
```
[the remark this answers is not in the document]
mitschabaude: yeah.. but it is not of the accepted type. I don't currently have a way here to specifiy "empty" cells, so I need to fill it with a witness and give it a 0 coefficient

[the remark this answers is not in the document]
mitschabaude: well I could create a witness here which isn't wired to anything else. but it would make the code more messy

src/lib/gadgets/basic.ts
```diff
@@ -0,0 +1,79 @@
...
+import { existsOne, toVar } from './common.js';
+import { Gates } from '../gates.js';
+import { TupleN } from '../util/types.js';
+
+export { assertOneOf };
+
+// TODO: create constant versions of these and expose on Gadgets
+
+/**
+ * Assert that a value equals one of a finite list of constants:
```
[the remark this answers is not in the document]
mitschabaude: lookups add overhead to the kimchi and pickles provers, so I think whether it's worth it depends a lot on how often we use this in the circuit overall and whether we use lookups for other things as well

src/lib/gadgets/basic.ts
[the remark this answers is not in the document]
mitschabaude: I ended up creating an `emptyCell()` helper and using it here. It seems cleaner than to wire some unrelated other witnesses to spots that are disabled with a 0 coefficient

src/lib/gadgets/foreign-field.unit-test.ts
```diff
@@ -251,8 +263,76 @@ await equivalentAsync({ from: [f, f], to: f }, { runs: 3 })(
...
+
 // helper
 
+function containsNTimes(n: number, pattern: readonly GateType[]) {
+  return and(
+    contains(repeat(n, pattern)),
+    not(contains(repeat(n + 1, pattern)))
+  );
+}
+
```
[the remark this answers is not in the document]
mitschabaude: done in https://github.com/o1-labs/o1js/pull/1262/commits/74ff146b5a10da8846fc335d2e7162c7b62d9ce2

src/lib/gadgets/foreign-field.ts
```diff
@@ -431,3 +456,216 @@ function combine2([x0, x1]: bigint3 | [bigint, bigint]) {
...
+        let xi = toBigint3(xs[i + 1]);
+        let sign = signs[i];
+
+        // figure out if there's overflow
+        x += sign * combine(xi);
+        let overflow = 0n;
+        if (sign === 1n && x >= f) overflow = 1n;
+        if (sign === -1n && x < 0n) overflow = -1n;
+        if (f === 0n) overflow = 0n;
+        xRef.set(x - overflow * f);
```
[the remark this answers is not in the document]
jspada: can't remember now

PR #1271, opened 2023-11-27
Encode how to map provables to normal values

mitschabaude:
- add toValue impls in most places
- fixup events, scalar, group
- get compilation working
- bindings
- fix test compilation
- a non-trivial value type

### DIFF
--- a/src/lib/account_update.ts
+++ b/src/lib/account_update.ts
@@ -1285,6 +1285,11 @@ class AccountUpdate implements Types.AccountUpdate {
       other
     );
   }
+  static toValue = Types.AccountUpdate.toValue;
+  static fromValue(value: TypesBigint.AccountUpdate): AccountUpdate {
+    let accountUpdate = Types.AccountUpdate.fromValue(value);
+    return new AccountUpdate(accountUpdate.body, accountUpdate.authorization);
+  }
 
   static witness<T>(
     type: FlexibleProvable<T>,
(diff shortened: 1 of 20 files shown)


src/lib/provable/core/provable-context.ts
```diff
@@ -71,6 +71,7 @@ function inCompileMode() {
 
 function asProver(f: () => void) {
   if (inCheckedComputation()) {
+    // TODO make this start a "witness block" context
```
MartinMinkov, line 74: Can you make an issue and reference it to the comment to tackle later? (if you don't mind, of course) 

I'd like to see more issues in TODO comments like this to make it easier to pick up :D

src/lib/provable/crypto/poseidon.ts
```diff
@@ -221,6 +222,19 @@ const TokenSymbolPure: ProvableExtended<
   check({ field }: TokenSymbol) {
     rangeCheckN(48, field);
   },
+  toValue({ symbol }) {
+    return symbol;
+  },
+  fromValue(symbol: string | TokenSymbol) {
+    if (typeof symbol !== 'string') return symbol;
+    let bytesLength = new TextEncoder().encode(symbol).length;
+    if (bytesLength > 6)
```
MartinMinkov, line 231: If you think it's a better idea, can you put `6`  into https://github.com/o1-labs/o1js/blob/main/src/lib/mina/constants.ts? It's a minor change, but it could really help explain later why it's this value.

MartinMinkov (approved):

mitschabaude:
TODO: forgot to add changelog for this

PR #1278, opened 2023-11-29
Add automated version bump workflow

MartinMinkov:
This PR introduces a new GitHub Actions workflow that automatically bumps the project's patch version bi-weekly on Fridays. 

The workflow also creates a new branch and a PR to `main` for the new version. This automation will help in maintaining a regular release cycle and reduce manual effort.

**Note:** This PR only bumps the patch version by default and does not make any edits to the CHANGELOG. Those edits can be made in this Github Action, but was left out since I wasn't sure if we wanted to include that or just let the human reviewer make those changes manually. Should this be changed?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,66 @@
+# Purpose:
+# Automatically bumps the project's patch version bi-weekly on Fridays.
+#
+# Details:
+# - Triggered at 00:00 UTC every Friday; runs on even weeks of the year.
+# - Sets up the environment by checking out the repo and setting up Node.js.
+# - Bumps patch version using `npm version patch`, then creates a new branch 'release/x.x.x'.
+# - Pushes changes and creates a PR to `main` using GitHub CLI.
+# - Can also be triggered manually via `workflow_dispatch`.
+name: Version Bump
+
+on:
+  workflow_dispatch: # Allow to manually trigger the workflow
+  schedule:
+    - cron: "0 0 * * 5" # At 00:00 UTC every Friday
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+    - cron: "0 0 * * 5" # At 00:00 UTC every Friday
+
+jobs:
+  version-bump:
+    runs-on: ubuntu-latest
+
+    steps:
+      # Since cronjob syntax doesn't support bi-weekly schedule, we need to check if it's an even week or not
+      - name: Check if it's an even week
+        id: check-week
```
shimkiv, line 24: Do we need `id`s? I don't see this particular step results being referenced elsewhere.

.github/workflows/release.yml
```diff
@@ -0,0 +1,66 @@
...
+        with:
+          node-version: "18"
+
+      - name: Configure Git
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+
+      - name: Bump patch version
+        id: version-bump
```
shimkiv, line 47: Same as above + this id duplicates Job name.

shimkiv (approved):
Great stuff! Left couple minor comments.

.github/workflows/release.yml
MartinMinkov: fixed!

.github/workflows/release.yml
MartinMinkov: fixed!

mitschabaude (approved):
Awesome!

Issue #1146, opened 2023-10-03
Port Keccak (SHA3) gadget to TypeScript


nicc:
partly depends on interfaces to #1138 #1140 #1141 #1139 #1148

Trivo25:
cc @jackryanservia to pick this up. if not, let me know

barriebyron:
Docs need to land with these code updates for Keccak 
Docs issue https://app.zenhub.com/workspaces/zkapps-product-eng-6130fedb3b0fc600123d8796/issues/gh/o1-labs/docs2/741

PR #1281, opened 2023-12-01
Keccak block transformation

jackryanservia:
Here is implementation of block transformation for #1146.

You can check things out by running the test!

And here are the constraint systems for each step!
[Theta.txt](https://github.com/o1-labs/o1js/files/13523461/Theta.txt)
[PiRho.txt](https://github.com/o1-labs/o1js/files/13523462/PiRho.txt)
[Iota.txt](https://github.com/o1-labs/o1js/files/13523463/Iota.txt)
[Chi.txt](https://github.com/o1-labs/o1js/files/13523464/Chi.txt)

I think everything makes sense! But maybe I'm missing something :)

### DIFF
--- a/src/lib/keccak.ts
+++ b/src/lib/keccak.ts
@@ -1,0 +1,217 @@
+import { Field } from './field.js';
+import { Gadgets } from './gadgets/gadgets.js';
+
+// KECCAK CONSTANTS
+
+// Length of the square matrix side of Keccak states
+const KECCAK_DIM = 5;
+
+// Value `l` in Keccak, ranges from 0 to 6 and determines the lane width
+const KECCAK_ELL = 6;
+
+// Width of a lane of the state, meaning the length of each word in bits (64)
+const KECCAK_WORD = 2 ** KECCAK_ELL;
+
+// Number of bytes that fit in a word (8)
+const BYTES_PER_WORD = KECCAK_WORD / 8;
+
+// Length of the state in bits, meaning the 5x5 matrix of words in bits (1600)
+const KECCAK_STATE_LENGTH = KECCAK_DIM ** 2 * KECCAK_WORD;
+
+// Number of rounds of the Keccak permutation function depending on the value `l` (24)
+const KECCAK_ROUNDS = 12 + 2 * KECCAK_ELL;
+
+// Creates the 5x5 table of rotation offset for Keccak modulo 64
+//  | x \ y |  0 |  1 |  2 |  3 |  4 |
+//  | ----- | -- | -- | -- | -- | -- |
+//  | 0     |  0 | 36 |  3 | 41 | 18 |
+//  | 1     |  1 | 44 | 10 | 45 |  2 |
+//  | 2     | 62 |  6 | 43 | 15 | 61 |
+//  | 3     | 28 | 55 | 25 | 21 | 56 |
+//  | 4     | 27 | 20 | 39 |  8 | 14 |
+const ROT_TABLE = [
+  [0, 36, 3, 41, 18],
+  [1, 44, 10, 45, 2],
+  [62, 6, 43, 15, 61],
+  [28, 55, 25, 21, 56],
+  [27, 20, 39, 8, 14],
+];
+
+// Round constants for Keccak
+// From https://keccak.team/files/Keccak-reference-3.0.pdf
+const ROUND_CONSTANTS = [
+  0x0000000000000001n,
+  0x0000000000008082n,
+  0x800000000000808an,
+  0x8000000080008000n,
+  0x000000000000808bn,
+  0x0000000080000001n,
+  0x8000000080008081n,
+  0x8000000000008009n,
+  0x000000000000008an,
+  0x0000000000000088n,
+  0x0000000080008009n,
+  0x000000008000000an,
+  0x000000008000808bn,
+  0x800000000000008bn,
+  0x8000000000008089n,
+  0x8000000000008003n,
+  0x8000000000008002n,
+  0x8000000000000080n,
+  0x000000000000800an,
+  0x800000008000000an,
+  0x8000000080008081n,
+  0x8000000000008080n,
+  0x0000000080000001n,
+  0x8000000080008008n,
+].map(Field.from);
+
+// Return a keccak state where all lanes are equal to 0
+const getKeccakStateZeros = (): Field[][] =>
+  Array.from(Array(KECCAK_DIM), (_) => Array(KECCAK_DIM).fill(Field.from(0)));
+
+// KECCAK HASH FUNCTION
+
+// ROUND TRANSFORMATION
+
+// First algrithm in the compression step of Keccak for 64-bit words.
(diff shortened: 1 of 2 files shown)


dannywillems:
Requesting @querolita review as she worked on Keccak.
On my side, I would like to see more unit tests. In particular, I would like to see test vectors from the Keccak team. Maybe there are some on their [website](https://keccak.team).

dannywillems (dismissed):
Test vectors.

src/lib/keccak.ts
```diff
@@ -0,0 +1,217 @@
...
+// E[x,y] = A[x,y] xor D[x]
+// In the Keccak reference, it corresponds to the `theta` algorithm.
+// We use the first index of the state array as the x coordinate and the second index as the y coordinate.
+const theta = (state: Field[][]): Field[][] => {
+  const stateA = state;
+
+  // XOR the elements of each row together
+  // for all x in {0..4}: C[x] = A[x,0] xor A[x,1] xor A[x,2] xor A[x,3] xor A[x,4]
+  const stateC = stateA.map((row) =>
+    row.reduce((acc, next) => Gadgets.xor(acc, next, KECCAK_WORD))
```
querolita, line 89: What's the initial value of `acc`? All zeros I suppose?

src/lib/keccak.ts
```diff
@@ -0,0 +1,217 @@
...
+// end for
+// We use the first index of the state array as the x coordinate and the second index as the y coordinate.
+function piRho(state: Field[][]): Field[][] {
+  const stateE = state;
+  const stateB: Field[][] = getKeccakStateZeros();
+
+  // for all x in {0..4} and y in {0..4}: B[y,2x+3y] = ROT(E[x,y], r[x,y])
+  for (let x = 0; x < KECCAK_DIM; x++) {
+    for (let y = 0; y < KECCAK_DIM; y++) {
+      // No need to use mod since this is always positive
```
querolita, line 140: This comments refers to an old function called mod which dealt with negative remainders, making them positive (mod) by adding the modulus. But this was removed, and the original comment referred to not needing to use mod and use just Ocaml's `%` as it was always positive. 

What I am saying is that this comment should be removed. It adds no meaning in this context.

src/lib/keccak.ts
```diff
@@ -0,0 +1,217 @@
...
+  const stateF = getKeccakStateZeros();
+
+  // for all x in {0..4} and y in {0..4}: F[x,y] = B[x,y] xor ((not B[x+1,y]) and B[x+2,y])
+  for (let x = 0; x < KECCAK_DIM; x++) {
+    for (let y = 0; y < KECCAK_DIM; y++) {
+      stateF[x][y] = Gadgets.xor(
+        stateB[x][y],
+        Gadgets.and(
+          // We can use unchecked NOT because the length of the input is constrained to be 64 bits thanks to the fact that it is the output of a previous Xor64
+          Gadgets.not(stateB[(x + 1) % 5][y], KECCAK_WORD, false),
```
querolita, line 171: use KECCAK_DIM instead of 5 here

src/lib/keccak.ts
```diff
@@ -0,0 +1,217 @@
...
+          stateB[(x + 2) % 5][y],
```
querolita, line 172: same here

src/lib/keccak.unit-test.ts
```diff
@@ -0,0 +1,105 @@
...
+  ROUND_CONSTANTS,
+  theta,
+  piRho,
+  chi,
+  iota,
+  round,
+  blockTransformation,
+} from './keccak.js';
+
+const KECCAK_TEST_STATE = [
```
querolita, line 15: Because this is still just a portion of the hash function as it does not contain padding, sponges, etc, this test is not sufficient for the final implementation. It will be clearer in the future if you can include full hash tests which I guess is already planned. In any case, this looks ok for now.

src/lib/keccak.unit-test.ts
```diff
@@ -0,0 +1,105 @@
...
+  [0, 0, 0, 0, 0],
+  [0, 0, 1, 0, 0],
+  [1, 0, 0, 0, 0],
+  [0, 0, 0, 1, 0],
+  [0, 1, 0, 0, 0],
+].map((row) => row.map((elem) => Field.from(elem)));
+
+let KeccakBlockTransformation = ZkProgram({
+  name: 'KeccakBlockTransformation',
+  publicInput: Provable.Array(Provable.Array(Field, 5), 5),
```
querolita, line 25: Use KECCAK_DIM?

querolita (approved):

querolita:
>  would cascade out quickly such that its correctness could be verified by testing the complete hash function

I agree with Jack on this. Even if I know this test is not sufficient to know that the full hash function will work fine, it is ok for now. In any case, if any of these steps were flaky, the error will cascade out in the final tests.

src/lib/keccak.ts
```diff
@@ -0,0 +1,217 @@
...
+  0x000000000000800an,
+  0x800000008000000an,
+  0x8000000080008081n,
+  0x8000000000008080n,
+  0x0000000080000001n,
+  0x8000000080008008n,
+].map(Field.from);
+
+// Return a keccak state where all lanes are equal to 0
+const getKeccakStateZeros = (): Field[][] =>
```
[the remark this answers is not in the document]
jackryanservia: It gives a square array of `Field` but does not specify the length of each dimension. It is possible to do so, but it's a bit clunky, and I'm not sure if the TypeScript compiler will generate JavaScript that actually checks this at run time. I'm also fairly new to this! :)

src/lib/keccak.ts
jackryanservia: Ah! It's just the first index of the array`row[0]`.

src/lib/keccak.ts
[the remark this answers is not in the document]
mitschabaude: Typescript never affects the runtime in any way

src/lib/keccak.ts
```diff
@@ -0,0 +1,215 @@
...
+  const stateE = stateA.map((row, index) =>
+    row.map((elem) => Gadgets.xor(elem, stateD[index], KECCAK_WORD))
+  );
+
+  return stateE;
+};
+
+// Second and third steps in the compression step of Keccak for 64-bit words.
+// B[y,2x+3y] = ROT(E[x,y], r[x,y])
+// which is equivalent to the `rho` algorithm followed by the `pi` algorithm in the Keccak reference as follows:
```
mitschabaude, line 111: kinda hard to see this equivalence 0.O

mitschabaude (approved):
looks great to me. I didn't compare the algorithms with a reference, but checked everything for common problems of provable code like insufficient range checks. conclusion: the `permutation` function is sound and given a state of 64-bit inputs it will again return such a state

src/lib/keccak.ts
mitschabaude: fun fact - we have a function called `mod` which handles negative inputs in our JS code base as well. I understood the comment that way :)

PR #1283, opened 2023-12-01
Twisted edwards curve operations

mitschabaude:
#985 + #1259 + #1262

everything that should be needed to build powerful new crypto primitives!


querolita (approved):
After checking the operations proposed here, I am accepting this PR as a basis for EDDSA, but things will likely change in upcoming PRs and will look more like this ECDSA interface.

PR #1284, opened 2023-12-01
Expose lookup constraint and implement three12Bit gadget for Keccak

jackryanservia:
Implementation of three12Bit for https://github.com/o1-labs/o1js/issues/1146.

### DIFF
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -9,6 +9,7 @@ import {
 import { not, rotate, xor, and, leftShift, rightShift } from './bitwise.js';
 import { Field } from '../core.js';
 import { ForeignField, Field3 } from './foreign-field.js';
+import { three12Bit } from './lookup.js';
 
 export { Gadgets };
 
@@ -312,6 +313,31 @@ const Gadgets = {
    */
   compactMultiRangeCheck(xy: Field, z: Field) {
     return compactMultiRangeCheck(xy, z);
+  },
+
+  /**
+   * Checks that three {@link Field} elements are less than 2^12 (using only one row) by verifying that they exist in the [RANGE_CHECK_TABLE](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/tables/mod.rs).
+   *
+   * It's possible to use this to check that values are less than 12 bits by scaling the value by 2^(12 - bits) before passing it in.
+   *
+   * You can find more details about lookups in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=lookup%20gate#lookup)
+   *
+   * @param v0 - The first {@link Field} element to be checked.
+   * @param v1 - The second {@link Field} element to be checked.
+   * @param v2 - The third {@link Field} element to be checked.
+   *
+   * @throws Throws an error if one of the input values exceeds 2^12.
+   *
+   * @example
+   * ```typescript
+   * let a = Field(4000);
+   * let b = a.mul(2 ** 4); // scale `a` so we can check if it's less than 8 bits
+   * three12Bit(a, a, a); // works
+   * three12Bit(a, a, b); // throws an error since b is greater than 12 bits (and `a` is greater than 8 bits)
+   * ```
+   */
+  three12Bit(v0: Field, v1: Field, v2: Field) {
+    return three12Bit(v0, v1, v2);
   },
 
   /**
--- a/src/lib/gadgets/gadgets.ts
+++ b/src/lib/gadgets/gadgets.ts
@@ -318,7 +318,7 @@ const Gadgets = {
   /**
    * Checks that three {@link Field} elements are less than 2^12 (using only one row) by verifying that they exist in the [RANGE_CHECK_TABLE](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/tables/mod.rs).
    *
-   * It's possible to use this to check that values are less than 12 bits by scaling the value by 2^(12 - bits) before passing it in.
+   * It's possible to use this to check that a value is less than x bits by scaling the value by 2^(12 - x) before passing it in (with x < 12).
    *
    * You can find more details about lookups in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=lookup%20gate#lookup)
    *
--- a/src/lib/gadgets/lookup.ts
+++ b/src/lib/gadgets/lookup.ts
@@ -1,0 +1,17 @@
+import { Field } from '../field.js';
+import { Gates } from '../gates.js';
+
+export { three12Bit };
+
+function three12Bit(v0: Field, v1: Field, v2: Field) {
+  // Checks that all three input values exist in the RANGE_CHECK_TABLE (tableId: 1) and are equal to 0.
+  Gates.lookup(
+    Field.from(1),
+    v0,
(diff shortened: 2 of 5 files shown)


src/lib/gadgets/lookup.ts
jackryanservia, line 1: Does it make sense for me to add another function `threeHoweverManyBits(bits0, value0, ...)` that also handles the scaling here, or does that just clutter things up?

src/lib/gadgets/lookup.unit-test.ts
```diff
@@ -0,0 +1,57 @@
...
+      privateInputs: [Field, Field, Field],
+      method(v0: Field, v1: Field, v2: Field) {
+        // Dummy range check to make sure the lookup table is initialized
+        // It should never fail because 64 > 12
+        Gadgets.rangeCheck64(v0);
+        Gadgets.three12Bit(v0, v1, v2);
+      },
+    },
+  },
+});
```
jackryanservia, line 35: I wasn't able to get the `ZkProgram` to compile unless I included `rangeCheck64`. Without it the process would just hang and not use any CPU time. I spent an hour or so trying to figure out why before I added `rangeCheck64`. Is this because of the [dummy gate thing](https://github.com/MinaProtocol/mina/blob/1d35680ad7aa654b57ba40562c6025d04ed89019/src/lib/crypto/kimchi_backend/gadgets/lookup.ml#L6)?

src/lib/gadgets/lookup.unit-test.ts
```diff
@@ -0,0 +1,57 @@
...
+import { assert } from './common.js';
+import { Gadgets } from './gadgets.js';
+import { constraintSystem, contains } from '../testing/constraint-system.js';
+
+let maybeUint = (n: number | bigint): Spec<bigint, Field> => {
+  let uint = Random.bignat((1n << BigInt(n)) - 1n);
+  return fieldWithRng(
+    Random.map(Random.oneOf(uint, uint.invalid), (x) => mod(x, Field.ORDER))
+  );
+};
```
jackryanservia, line 20: I originally misunderstood this. It just returns a value that's sometimes less than `n` bits and sometimes more, correct? Its not doing anything else?

src/lib/gates.ts
```diff
@@ -149,6 +151,41 @@ function generic(
...
+ * @param tableId the [id](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/tables/mod.rs) of the lookup table.
+ * @param index0 the index of the first value to lookup.
+ * @param value0 the first value to lookup.
+ * @param index1 the index of the second value to lookup.
+ * @param value1 the second value to lookup.
+ * @param index2 the index of the third value to lookup.
+ * @param value2 the third value to lookup.
+ *
+ */
+function lookup(
```
querolita, line 168: Maybe don't expose this to the very end user? It is a bit low level.

querolita (approved):
Looks good to me, but I wonder if this is a functionality that should be provided to zkapp developers or if it should remain as an internal feature inside o1js.

src/lib/gadgets/lookup.unit-test.ts
querolita: Oh I see, so it was not fixed eventually? I would have sweared that it was addressed at some point in the summer... Well, yes, a few months ago, it wouldn't work unless your circuit had a previous gate which was already using the range check table, even if it is just a dummy zero value. But you needed to include that row... Seems like it is still the case 😞

src/lib/gadgets/lookup.unit-test.ts
mitschabaude: Yes

src/lib/gadgets/lookup.ts
```diff
@@ -0,0 +1,17 @@
+import { Field } from '../field.js';
+import { Gates } from '../gates.js';
+
+export { three12Bit };
+
+function three12Bit(v0: Field, v1: Field, v2: Field) {
+  // Checks that all three input values exist in the RANGE_CHECK_TABLE (tableId: 1) and are equal to 0.
```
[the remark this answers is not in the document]
jackryanservia: Ahhh, yes, this isn't worded very well. I was trying to express that the hard-coded zeros are the values in the table at the index we are looking up, but I don't think it needs to be explained. I will just abstract over this as you suggested.

src/lib/gadgets/gadgets.ts
```diff
@@ -314,6 +315,31 @@ const Gadgets = {
     return compactMultiRangeCheck(xy, z);
   },
 
+  /**
+   * Checks that three {@link Field} elements are less than 2^12 (using only one row) by verifying that they exist in the [RANGE_CHECK_TABLE](https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/circuits/lookup/tables/mod.rs).
+   *
+   * It's possible to use this to check that a value is less than x bits by scaling the value by 2^(12 - x) before passing it in (with x < 12).
```
mitschabaude, line 321: as Anais said below, this has to make clearer that for checking fewer bits you have to check _both_ $z$ and $z 2^{12-x}$. 

also, nit: `x` is not the best variable to use for an exponent here because we use it everywhere for field elements (usually in mathematics you try to pick different symbols for different kinds of value). exponent could be $i$ or $n$ or $e$ for example

src/lib/gadgets/lookup.ts
mitschabaude: just note that you need to use two lookup slots per input for a different number of checked bits. assuming three inputs at a time is still a good API in that case because 3*2 exactly fit in 2 lookup gates!

src/lib/gadgets/gadgets.ts
```diff
@@ -314,6 +315,31 @@ const Gadgets = {
...
+   *
+   * @example
+   * ```typescript
+   * let a = Field(4000);
+   * let b = a.mul(2 ** 4); // scale `a` so we can check if it's less than 8 bits
+   * three12Bit(a, a, a); // works
+   * three12Bit(a, a, b); // throws an error since b is greater than 12 bits (and `a` is greater than 8 bits)
+   * ```
+   */
+  three12Bit(v0: Field, v1: Field, v2: Field) {
```
mitschabaude, line 339: can we name this function more similarly to other range check gadgets? we have `rangeCheck64` and `rangeCheck32`, as well as `multiRangeCheck` for the 3x88 bit check. this one could be `rangeCheck12x3` maybe? at least it should say "range check"

src/lib/gadgets/lookup.unit-test.ts
mitschabaude: nice @jackryanservia - I guess you found out why https://github.com/o1-labs/o1js/pull/1253 wasn't working, so we can land that PR as well

src/lib/gates.ts
mitschabaude: gates are not currently exported from o1js! but I think they will eventually, having both low- and high-level APIs makes sense IMO

mitschabaude (commented):
Good stuff!

Feedback:
- can you add a changelog item about the gadget?
- maybe we can come up with a better name than `three12bit`?
- the `three12bit` doccomment must not trick users into only range-checking $x2^{12-n}$, because we're in a finite field and all values, including $2^{12-n}$, have inverses, so you can trivially pick an $x$ which violates the supposed range check

mitschabaude:
oh - the lookup unit test fails, during proving the zkprogram. I wonder why it worked on your machine 🤔

PR #1285, opened 2023-12-01
SHA256

Trivo25:
Implementation of static sized SHA2-256 https://csrc.nist.gov/pubs/fips/180-4/upd1/final


src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+  // now pad the data to reach the format expected by sha256
+  // pad 1 bit, followed by k zero bits where k is the smallest non-negative solution to
+  // l + 1 + k = 448 mod 512
+  // then append a 64bit block containing the length of the original message in bits
+
+  let l = zeroPadded.length * 8; // length in bits
+  let k = (448 - (l + 1)) % 512;
+
+  // pad message exceeds 512bit and needs a new block
+  if (k < 0) k += 512;
```
mitschabaude, line 25: logic checks out for me 👍🏻  but would be less confusing if you used `mod()` from `finite_field.ts`, which handles the negative case:
```suggestion
  let k = mod(448 - (l + 1), 512);
```

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  let paddingBits = (
+    '1' + // append 1 bit
+    '0'.repeat(k) + // append k zero bits
+    '0'.repeat(64 - l.toString(2).length) + // append 64bit containing the length of the original message
+    l.toString(2)
```
mitschabaude, line 31: pretty elegant!

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+  ).match(/.{1,8}/g)!; // this should always be devisable by 8
```
mitschabaude, line 32: woah what kind of black magic is this 🥷
we also have a `chunkString()` helper but please leave this in for its educational value :D

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  // map the padding bit string to UInt8 elements
+  let padding = paddingBits.map((x) => UInt8.from(BigInt('0b' + x)));
```
mitschabaude, line 35: oh - it's unexpected to me that you go from bits to bytes in big-endian fashion here. like, here `1000 0000` becomes 128 where I would expect it to become 1

I've been comparing your code with [wikipedia](https://en.wikipedia.org/wiki/SHA-2), but this detail is missing from there. it just says "copy bits into 32-bit words"

could you find a source for how this is supposed to work?

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+    chunks.push(concatToUInt32(paddedMessage.slice(i, i + 4)));
+  }
+
+  // split message into 16 element sized message blocks
+  // SHA256 expects n-blocks of 512bit each, 16*32bit = 512bit
+  let messageBlocks: UInt32[][] = [];
+  for (let i = 0; i < chunks.length; i += 16) {
+    messageBlocks.push(chunks.slice(i, i + 16));
+  }
+  return messageBlocks;
```
mitschabaude, line 54: this could use the `chunk()` helper!
```ts
let messageBlocks = chunk(chunks, 16);
```

src/lib/gadgets/sha256.ts
mitschabaude: huh but I see now that `concatToUInt32` is big-endian as well -- later elements become smaller parts of the uint32. so at least this matches.

aaah, I actually found this on wikipedia now:

> Note 4: Big-endian convention is used when expressing the constants in this pseudocode,
>     **and when parsing message block data from bytes to words**, for example,
>     the first word of the input message "abc" after padding is 0x61626380

all good

src/lib/gadgets/sha256.unit-test.ts
```diff
@@ -0,0 +1,76 @@
...
+import { bytes } from './test-utils.js';
+import { equivalent, equivalentAsync } from '../testing/equivalent.js';
+import { Random, sample } from '../testing/random.js';
+import { expect } from 'expect';
+
+sample(Random.nat(400), 10).forEach((preimageLength) => {
+  let inputBytes = bytes(preimageLength);
+  let outputBytes = bytes(256 / 8);
+
+  equivalent({ from: [inputBytes], to: outputBytes, verbose: true })(
```
mitschabaude, line 14: maybe use `equivalentProvable` here?

src/lib/gadgets/sha256.unit-test.ts
```diff
@@ -0,0 +1,76 @@
...
+  { runs: RUNS }
+)(nobleSha256, async (x) => {
+  const proof = await Sha256Program.sha256(x);
+  await Sha256Program.verify(proof);
+  return proof.publicOutput;
+});
+
+for (let { preimage, hash } of testVectors()) {
+  class BytesN extends Bytes(preimage.length) {}
+  let actual = Gadgets.SHA256.hash(BytesN.fromString(preimage));
```
mitschabaude, line 52: isn't this the same as
```ts
let actual = Gadgets.SHA256.hash(preimage);
```

src/lib/gadgets/sha256.unit-test.ts
```diff
@@ -0,0 +1,76 @@
...
+  expect(actual.toHex()).toEqual(hash);
+}
+
+function testVectors() {
```
mitschabaude, line 56: nice! where are they from?

mitschabaude (commented):
fantastic work!

src/lib/gadgets/gadgets.ts
```diff
@@ -796,6 +797,27 @@ const Gadgets = {
...
+  /**
+   * Implementation of the [SHA256 hash function.](https://en.wikipedia.org/wiki/SHA-2) Hash function with 256bit output.
+   *
+   * Applies the SHA2-256 hash function to a list of big-endian byte-sized elements.
+   *
+   * The function accepts {@link Bytes} as the input message, which is a type that represents a static-length list of byte-sized field elements (range-checked using {@link Gadgets.rangeCheck8}).
+   * Alternatively, you can pass plain `number[]`, `bigint[]` or `Uint8Array` to perform a hash outside provable code.
+   *
+   * Produces an output of {@link Bytes} that conforms to the chosen bit length.
+   * Both input and output bytes are big-endian.
```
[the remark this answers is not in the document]
Trivo25: Actually -- maybe it does matter, i think the output endianness should be mentioned. Over the holidays I noticed that there can be a lot of confusion about the endianness representation

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+      let h = H[7];
+
+      // main loop
+      for (let t = 0; t <= 63; t++) {
+        // T1 is unreduced and not proven to be 32bit, we will do this later to save constraints
+        const unreducedT1 = h.value
+          .add(SigmaOne(e).value)
+          .add(Ch(e, f, g).value)
+          .add(K[t].value)
+          .add(W[t].value);
```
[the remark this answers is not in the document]
mitschabaude: tried it myself in https://github.com/o1-labs/o1js/pull/1285/commits/6b8d48a7af93fc74c2bea224fb73cc1002e9c885, and it works but saves a bit less than expected

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,306 @@
...
+  let x01 = x0.add(x1.mul(1 << d0)).seal();
+  let xRotR1 = x23.add(x01.mul(1 << (d2 + d3))).seal();
+  // ^ proves that 2^(32-d1)*x1 < xRotR1 => x1 < 2^d1 if we check xRotR1 < 2^32 later
+
+  // rotr(x, r2) = x3 + x0*2^d3 + x1*2^(d3+d0) + x2*2^(d3+d0+d1)
+  let x012 = x01.add(x2.mul(1 << (d0 + d1))).seal();
+  let xRotR2 = x3.add(x012.mul(1 << d3)).seal();
+  // ^ proves that 2^(32-d2)*x2 < xRotR2 => x2 < 2^d2 if we check xRotR2 < 2^32 later
+
+  return UInt32.from(xRotR0).xor(new UInt32(xRotR1)).xor(new UInt32(xRotR2));
```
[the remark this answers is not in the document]
mitschabaude: added in https://github.com/o1-labs/o1js/pull/1285/commits/d71872dee0e387a9e7c886d6ab59d1d4f78ed878

mitschabaude (approved):
All my comments are addressed now!

src/lib/gadgets/bit-slices.ts
```diff
@@ -0,0 +1,156 @@
...
+import { l } from './range-check.js';
+
+export { bytesToWord, wordToBytes, wordsToBytes, bytesToWords, sliceField3 };
+
+// conversion between bytes and multi-byte words
+
+/**
+ * Convert an array of UInt8 to a Field element. Expects little endian representation.
+ */
+function bytesToWord(wordBytes: UInt8[]): Field {
```
rbonichon, line 21: Suggestion: add `le` in the function name, especially since it's exported, to explicit the little endianness part, like `wordOfLeBytes` (it does not look good with Camel case, though)

src/lib/gadgets/bit-slices.ts
```diff
@@ -0,0 +1,156 @@
...
+    const shift = 1n << BigInt(8 * idx);
+    return acc.add(byte.value.mul(shift));
+  }, Field.from(0));
+}
+
+/**
+ * Convert a Field element to an array of UInt8. Expects little endian representation.
+ * @param bytesPerWord number of bytes per word
+ */
+function wordToBytes(word: Field, bytesPerWord = 8): UInt8[] {
```
rbonichon, line 32: `le` too

src/lib/gadgets/bit-slices.ts
```diff
@@ -0,0 +1,156 @@
...
+  bytesToWord(bytes).assertEquals(word);
+
+  return bytes;
+}
+
+/**
+ * Convert an array of Field elements to an array of UInt8. Expects little endian representation.
+ * @param bytesPerWord number of bytes per word
+ */
+function wordsToBytes(words: Field[], bytesPerWord = 8): UInt8[] {
```
rbonichon, line 50: same here

src/lib/gadgets/bit-slices.ts
```diff
@@ -0,0 +1,156 @@
...
+  return words.flatMap((w) => wordToBytes(w, bytesPerWord));
+}
+/**
+ * Convert an array of UInt8 to an array of Field elements. Expects little endian representation.
+ * @param bytesPerWord number of bytes per word
+ */
+function bytesToWords(bytes: UInt8[], bytesPerWord = 8): Field[] {
```
rbonichon, line 57: and here

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  let l = message.length * 8; // length in bits
+  let k = Number(mod(448n - (BigInt(l) + 1n), 512n));
+
+  let paddingBits = (
+    '1' + // append 1 bit
+    '0'.repeat(k) + // append k zero bits
+    '0'.repeat(64 - l.toString(2).length) + // append 64bit containing the length of the original message
+    l.toString(2)
+  ).match(/.{1,8}/g)!; // this should always be devisable by 8
```
rbonichon, line 56: Do you mean `divisible`?

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,278 @@
...
+
+  // map the padding bit string to UInt8 elements
+  let padding = paddingBits.map((x) => UInt8.from(BigInt('0b' + x)));
+
+  // concatenate the padding with the original padded data
```
rbonichon, line 61: the data is unpadded at this point, right? Or  actually padded?

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,278 @@
...
+  let paddedMessage = message.bytes.concat(padding);
+
+  // split the message into 32bit chunks
+  let chunks: UInt32[] = [];
+
+  for (let i = 0; i < paddedMessage.length; i += 4) {
+    // chunk 4 bytes into one UInt32, as expected by SHA256
+    // bytesToWord expects little endian, so we reverse the bytes
```
rbonichon, line 69: huh? Why not have `bytesToWord` be big-endian then?

src/lib/gadgets/sha256.ts
```diff
@@ -0,0 +1,278 @@
...
+      H[3] = H[3].addMod32(d);
+      H[4] = H[4].addMod32(e);
+      H[5] = H[5].addMod32(f);
+      H[6] = H[6].addMod32(g);
+      H[7] = H[7].addMod32(h);
+    }
+
+    // the working variables H[i] are 32bit, however we want to decompose them into bytes to be more compatible
+    // wordToBytes expects little endian, so we reverse the bytes
+    return Bytes.from(H.map((x) => wordToBytes(x.value, 4).reverse()).flat());
```
rbonichon, line 154: Well yeah. Looks like bytes/word conversion should be big-endian.

rbonichon (commented):
Need to re-read the sha256 implementation. I'm assuming it's correct or tests would break. But I'll need to verify. 

It does seem strange to use little-endian conversions and always have to reverse them. What would be the reason?

rbonichon:
> Also, a general comment from me is that little-endian bytes layout is by far the more common overall, and also more logical IMO, or in other words, SHA256 is weird here.

It is the most common these days in hardware architecture but most network protocols are big-endian (very recent example being the transfer protocol in the Optimism Cannon VM). Human text or numbers are big-endian too. All this is just a matter of convention. 

> So, I don't think we need to add a "le" suffix to methods that use little-endian convention (we have many places where that would need to be added..). I would treat little-endian as the general default and instead use a "be" suffix for methods that use big-endian convention

My suggestion to add endianness in the name was more akin to following the [principle of least astonishment](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Rust or OCaml do that

rbonichon (approved):
The implementation looks correct.

src/lib/gadgets/sha256.ts
Trivo25: yes, the original data is unpadded but will then be concantenated with the padding

PR #1291, opened 2023-12-04
Implement Keccak sponge

jackryanservia:
Implementation of the Keccak sponge and, preNist, ethereum, and SHA-3 hash functions.

This PR currently is using a workaround in `checkBytes` because lookup arguments are not working in pickles, but that will be updated as soon as a fix is available.

### DIFF
--- a/src/lib/keccak.ts
+++ b/src/lib/keccak.ts
@@ -190,49 +190,26 @@ function keccakStateXor(a: Field[][], b: Field[][]): Field[][] {
 
 // Computes the number of required extra bytes to pad a message of length bytes
 function bytesToPad(rate: number, length: number): number {
-  return Math.floor(rate / 8) - (length % Math.floor(rate / 8));
+  return rate - (length % rate);
 }
 
 // Pads a message M as:
 // M || pad[x](|M|)
-// Padding rule 0x06 ..0*..1.
-// The padded message vector will start with the message vector
-// followed by the 0*1 rule to fulfill a length that is a multiple of rate (in bytes)
-// (This means a 0110 sequence, followed with as many 0s as needed, and a final 1 bit)
-function padNist(message: Field[], rate: number): Field[] {
+// The padded message will begin with the message and end with the padding rule (below) to fulfill a length that is a multiple of rate (in bytes).
+// If nist is true, then the padding rule is 0x06 ..0*..1.
+// If nist is false, then the padding rule is 10*1.
+function pad(message: Field[], rate: number, nist: boolean): Field[] {
   // Find out desired length of the padding in bytes
   // If message is already rate bits, need to pad full rate again
   const extraBytes = bytesToPad(rate, message.length);
 
   // 0x06 0x00 ... 0x00 0x80 or 0x86
-  const last = Field.from(BigInt(2) ** BigInt(7));
+  const first = nist ? 0x06n : 0x01n;
+  const last = 0x80n;
 
   // Create the padding vector
   const pad = Array(extraBytes).fill(Field.from(0));
-  pad[0] = Field.from(6);
-  pad[extraBytes - 1] = pad[extraBytes - 1].add(last);
-
-  // Return the padded message
-  return [...message, ...pad];
-}
-
-// Pads a message M as:
-// M || pad[x](|M|)
-// Padding rule 10*1.
-// The padded message vector will start with the message vector
-// followed by the 10*1 rule to fulfill a length that is a multiple of rate (in bytes)
-// (This means a 1 bit, followed with as many 0s as needed, and a final 1 bit)
-function pad101(message: Field[], rate: number): Field[] {
-  // Find out desired length of the padding in bytes
-  // If message is already rate bits, need to pad full rate again
-  const extraBytes = bytesToPad(rate, message.length);
-
-  // 0x01 0x00 ... 0x00 0x80 or 0x81
-  const last = Field.from(BigInt(2) ** BigInt(7));
-
-  // Create the padding vector
-  const pad = Array(extraBytes).fill(Field.from(0));
-  pad[0] = Field.from(1);
+  pad[0] = Field.from(first);
   pad[extraBytes - 1] = pad[extraBytes - 1].add(last);
 
   // Return the padded message
@@ -504,8 +481,7 @@ function hash(
 
   const rate = KECCAK_STATE_LENGTH - capacity;
 
-  const padded =
-    nistVersion === true ? padNist(message, rate) : pad101(message, rate);
+  const padded = pad(message, rate, nistVersion);
 
   const hash = sponge(padded, length, capacity, rate);
 
--- a/src/lib/keccak.ts
+++ b/src/lib/keccak.ts
@@ -9,7 +9,7 @@ export { Keccak };
 const Keccak = {
   /** TODO */
   nistSha3(
-    len: number,
+    len: 224 | 256 | 384 | 512,
     message: Field[],
     inpEndian: 'Big' | 'Little' = 'Big',
     outEndian: 'Big' | 'Little' = 'Big',
@@ -28,7 +28,7 @@ const Keccak = {
   },
   /** TODO */
   preNist(
-    len: number,
+    len: 224 | 256 | 384 | 512,
     message: Field[],
     inpEndian: 'Big' | 'Little' = 'Big',
     outEndian: 'Big' | 'Little' = 'Big',
@@ -547,24 +547,13 @@ function hash(
 // Gadget for NIST SHA-3 function for output lengths 224/256/384/512.
 // Input and output endianness can be specified. Default is big endian.
 function nistSha3(
-  len: number,
+  len: 224 | 256 | 384 | 512,
   message: Field[],
   inpEndian: 'Big' | 'Little' = 'Big',
   outEndian: 'Big' | 'Little' = 'Big',
   byteChecks: boolean = false
 ): Field[] {
-  switch (len) {
-    case 224:
-      return hash(inpEndian, outEndian, byteChecks, message, 224, 448, true);
-    case 256:
-      return hash(inpEndian, outEndian, byteChecks, message, 256, 512, true);
-    case 384:
-      return hash(inpEndian, outEndian, byteChecks, message, 384, 768, true);
-    case 512:
-      return hash(inpEndian, outEndian, byteChecks, message, 512, 1024, true);
-    default:
-      throw new Error('Invalid length');
-  }
+  return hash(inpEndian, outEndian, byteChecks, message, len, 2 * len, true);
 }
 
 // Gadget for Keccak hash function for the parameters used in Ethereum.
@@ -582,22 +571,11 @@ function ethereum(
 // Input and output endianness can be specified. Default is big endian.
 // Note that when calling with output length 256 this is equivalent to the ethereum function
 function preNist(
-  len: number,
+  len: 224 | 256 | 384 | 512,
   message: Field[],
   inpEndian: 'Big' | 'Little' = 'Big',
   outEndian: 'Big' | 'Little' = 'Big',
   byteChecks: boolean = false
 ): Field[] {
-  switch (len) {
-    case 224:
-      return hash(inpEndian, outEndian, byteChecks, message, 224, 448, false);
-    case 256:
-      return ethereum(message, inpEndian, outEndian, byteChecks);
-    case 384:
-      return hash(inpEndian, outEndian, byteChecks, message, 384, 768, false);
-    case 512:
-      return hash(inpEndian, outEndian, byteChecks, message, 512, 1024, false);
-    default:
-      throw new Error('Invalid length');
-  }
+  return hash(inpEndian, outEndian, byteChecks, message, len, 2 * len, false);
 }
--- a/src/lib/keccak.ts
+++ b/src/lib/keccak.ts
@@ -1,7 +1,7 @@
 import { Field } from './field.js';
 import { Gadgets } from './gadgets/gadgets.js';
 import { assert } from './errors.js';
-import { existsOne, exists } from './gadgets/common.js';
+import { existsOne } from './gadgets/common.js';
 import { rangeCheck8 } from './gadgets/range-check.js';
 
 export { Keccak };
@@ -11,30 +11,21 @@ const Keccak = {
   nistSha3(
     len: 224 | 256 | 384 | 512,
     message: Field[],
-    inpEndian: 'Big' | 'Little' = 'Big',
-    outEndian: 'Big' | 'Little' = 'Big',
     byteChecks: boolean = false
   ): Field[] {
-    return nistSha3(len, message, inpEndian, outEndian, byteChecks);
+    return nistSha3(len, message, byteChecks);
   },
   /** TODO */
-  ethereum(
-    message: Field[],
-    inpEndian: 'Big' | 'Little' = 'Big',
-    outEndian: 'Big' | 'Little' = 'Big',
-    byteChecks: boolean = false
-  ): Field[] {
-    return ethereum(message, inpEndian, outEndian, byteChecks);
+  ethereum(message: Field[], byteChecks: boolean = false): Field[] {
+    return ethereum(message, byteChecks);
   },
   /** TODO */
   preNist(
     len: 224 | 256 | 384 | 512,
     message: Field[],
-    inpEndian: 'Big' | 'Little' = 'Big',
-    outEndian: 'Big' | 'Little' = 'Big',
     byteChecks: boolean = false
   ): Field[] {
-    return preNist(len, message, inpEndian, outEndian, byteChecks);
+    return preNist(len, message, byteChecks);
   },
 };
 
@@ -503,8 +494,6 @@ function checkBytes(inputs: Field[]): void {
 // - the 10*1 pad will take place after the message, until reaching the bit length rate.
 // - then, {0} pad will take place to finish the 1600 bits of the state.
 function hash(
-  inpEndian: 'Big' | 'Little' = 'Big',
-  outEndian: 'Big' | 'Little' = 'Big',
   byteChecks: boolean = false,
   message: Field[] = [],
   length: number,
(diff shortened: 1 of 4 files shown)


src/lib/keccak.unit-test.ts
```diff
@@ -1,106 +1,136 @@
...
+  (x) => {
+    let thing = x.map(Number);
+    let result = sha3_256(new Uint8Array(thing));
+    return Array.from(result).map(BigInt);
+  },
+  async (x) => {
+    let proof = await Keccak256.nistSha3(x);
+    return proof.publicOutput;
+  }
 );
```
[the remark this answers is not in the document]
mitschabaude: looks like a useful test to me!

src/lib/keccak.unit-test.ts
```diff
@@ -1,106 +1,136 @@
 import { Field } from './field.js';
```
Trivo25, line 1: now that we have a full end to end functional Keccak, could you also add that (and some of the other variants, where you see fit) to the verification key regression tests?

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
   0x8000000000008080n,
   0x0000000080000001n,
   0x8000000080008008n,
-].map(Field.from);
+];
+
+// AUXILARY FUNCTIONS
+
+// Auxiliary function to check composition of 8 bytes into a 64-bit word
+function checkBytesToWord(word: Field, wordBytes: Field[]): void {
```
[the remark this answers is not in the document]
querolita: I guess this is to behave as in `getKeccakStateOfBytes` for each of the words?

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+  word.assertEquals(composition);
+}
+
+// KECCAK STATE FUNCTIONS
 
 // Return a keccak state where all lanes are equal to 0
 const getKeccakStateZeros = (): Field[][] =>
   Array.from(Array(KECCAK_DIM), (_) => Array(KECCAK_DIM).fill(Field.from(0)));
 
+// Converts a list of bytes to a matrix of Field elements
```
querolita, line 125: This one does not create constraints for the byte length? Asking because `keccakStateToBytes` does. I guess this is because `getKeccakStateOfBytes` is only used at the beginning of Keccak where inputs are already constraints, but `keccakStateToBytes` is called at the end where you might want to check fresh bytes of the hash output?

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+        const z = idx % BYTES_PER_WORD;
+        const x = Math.floor(idx / BYTES_PER_WORD) % KECCAK_DIM;
+        const y = Math.floor(idx / BYTES_PER_WORD / KECCAK_DIM);
+        // [7 6 5 4 3 2 1 0] [x=0,y=1] [x=0,y=2] [x=0,y=3] [x=0,y=4]
+        //        [x=1,y=0] [x=1,y=1] [x=1,y=2] [x=1,y=3] [x=1,y=4]
+        //        [x=2,y=0] [x=2,y=1] [x=2,y=2] [x=2,y=3] [x=2,y=4]
+        //        [x=3,y=0] [x=3,y=1] [x=3,y=2] [x=3,y=3] [x=3,y=4]
+        //        [x=4,y=0] [x=4,y=1] [x=4,y=0] [x=4,y=3] [x=4,y=4]
+        const word = state[x][y].toBigInt();
+        const byte = (word >> BigInt(8 * z)) & BigInt('0xff');
```
querolita, line 165: looks good to me

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+// AUXILARY FUNCTIONS
+
+// Auxiliary function to check composition of 8 bytes into a 64-bit word
+function checkBytesToWord(word: Field, wordBytes: Field[]): void {
+  let composition = wordBytes.reduce((acc, x, i) => {
+    const shift = Field.from(2n ** BigInt(8 * i));
+    return acc.add(x.mul(shift));
+  }, Field.from(0));
+
+  word.assertEquals(composition);
```
querolita, line 116: Also, is this just a normal assertion or it actually produces constraints? If so, I would write this clearly upfront in the docs of the function.

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+        // Field element containing value 2^(8*z)
+        const shift = Field.from(2n ** BigInt(8 * z));
+        state[x][y] = state[x][y].add(shift.mul(wordBytes[z]));
+      }
+    }
+  }
+  return state;
+}
+
+// Converts a state of Fields to a list of bytes as Fields and creates constraints for it
```
querolita, line 148: Okay after reading I can see that the constraints here being referred to correspond to the decomposition of each word, but not the individual bytes composing each word. Wouldn't this need further calls to rangeCheck8 to constrain this is indeed the case? Or is this handled somewhere else in the code?

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+    a.length === KECCAK_DIM && a[0].length === KECCAK_DIM,
+    'Invalid a dimensions'
+  );
+  assert(
+    b.length === KECCAK_DIM && b[0].length === KECCAK_DIM,
+    'Invalid b dimensions'
+  );
+
+  // Calls Gadgets.xor on each pair (x,y) of the states input1 and input2 and outputs the output Fields as a new matrix
+  return a.map((row, rowIndex) =>
```
querolita, line 195: Isn't there a zip function for better readability of these iterators?

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+    row.map((element, columnIndex) =>
+      Gadgets.xor(element, b[rowIndex][columnIndex], 64)
+    )
+  );
+}
+
 // KECCAK HASH FUNCTION
 
+// Computes the number of required extra bytes to pad a message of length bytes
+function bytesToPad(rate: number, length: number): number {
```
querolita, line 205: maybe name the `rate` parameter something like `rate_in_bits`, or directly pass it as bytes. We don't give support for lengths that are not a multiple of 8, so avoiding bitlengths as much as possible is less error prone.

querolita: Same with length, be more explicit with the unit

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+  return Math.floor(rate / 8) - (length % Math.floor(rate / 8));
```
querolita, line 206: like, if you prefer not to include the unit in the names, then just take them as bytes and do `rate - ( length % rate)`

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+}
+
+// Pads a message M as:
+// M || pad[x](|M|)
+// Padding rule 0x06 ..0*..1.
+// The padded message vector will start with the message vector
+// followed by the 0*1 rule to fulfill a length that is a multiple of rate (in bytes)
```
querolita, line 213: Maybe remove those lines

// The padded message vector will start with the message vector
// followed by the 0*1 rule to fulfill a length that is a multiple of rate (in bytes)

and just include the following

// (This means a 0110 sequence, followed with as many 0s as needed, and a final 1 bit)

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+  return [...message, ...pad];
+}
+
+// Pads a message M as:
+// M || pad[x](|M|)
+// Padding rule 10*1.
+// The padded message vector will start with the message vector
+// followed by the 10*1 rule to fulfill a length that is a multiple of rate (in bytes)
+// (This means a 1 bit, followed with as many 0s as needed, and a final 1 bit)
+function pad101(message: Field[], rate: number): Field[] {
```
querolita, line 238: I wonder if it makes sense to create a function pad that receives either 0x01 or 0x06 as input (maybe an enum with either PreNist or Nist and a match mapping to each of those bytes) and then use a generic code inside, because the inside code is the same (to avoid duplications).

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
   );
 }
 
-// TESTING
+// KECCAK SPONGE
 
-const blockTransformation = (state: Field[][]): Field[][] =>
-  permutation(state, ROUND_CONSTANTS);
+// Absorb padded message into a keccak state with given rate and capacity
```
querolita, line 394: Again, I would prefer to only allow multiples of 8 here. So either assert it, or just use bytes by default. This is because we don't give support for non-byte inputs nor parameters.

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
-  ROUND_CONSTANTS,
-  theta,
-  piRho,
-  chi,
-  iota,
-  round,
-  blockTransformation,
-};
+  // (capacity / 8) zero bytes
+  const zeros = Array(capacity / 8).fill(Field.from(0));
```
querolita, line 404: I wonder if this is constrained behind the scenes.

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+  // Find out desired length of the padding in bytes
+  // If message is already rate bits, need to pad full rate again
+  const extraBytes = bytesToPad(rate, message.length);
+
+  // 0x01 0x00 ... 0x00 0x80 or 0x81
+  const last = Field.from(BigInt(2) ** BigInt(7));
+
+  // Create the padding vector
+  const pad = Array(extraBytes).fill(Field.from(0));
+  pad[0] = Field.from(1);
```
querolita, line 248: Is any of this constrained anywhere? Meaning, the fact that the padded bytes correspond to 0x01, then a lot of zeros, then a 0x80. It should, but I don't know enough about the inner workings of o1js to understand if it is doing so or not.

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  const outputLength = squeezes * bytesPerSqueeze;
+  // array with sufficient space to store the output
+  const outputArray = Array(outputLength).fill(Field.from(0));
+  // first state to be squeezed
+  const bytestring = keccakStateToBytes(state);
+  const outputBytes = bytestring.slice(0, bytesPerSqueeze);
+  copy(outputBytes, outputArray, 0, bytesPerSqueeze);
+
+  // for the rest of squeezes
+  for (let i = 1; i < squeezes; i++) {
```
querolita, line 462: With the usual configurations, this never gets executed though.

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+    // apply the permutation function to the state
+    newState = permutation(newState, rc);
+    // append the output of the permutation function to the output
+    const bytestringI = keccakStateToBytes(state);
+    const outputBytesI = bytestringI.slice(0, bytesPerSqueeze);
+    copy(outputBytesI, outputArray, bytesPerSqueeze * i, bytesPerSqueeze);
+  }
+
+  // Obtain the hash selecting the first bitlength/8 bytes of the output array
+  const hashed = outputArray.slice(0, length / 8);
```
querolita, line 472: Are these bytes constrained to be 8-bits somewhere?

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  return hashed;
+}
+
+// Keccak sponge function for 1600 bits of state width
+// Need to split the message into blocks of 1088 bits.
```
querolita, line 477: of RATE bits, 1088 is only for the ethereum configuration

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  capacity: number,
+  rate: number
+): Field[] {
+  // check that the padded message is a multiple of rate
+  if ((paddedMessage.length * 8) % rate !== 0) {
+    throw new Error('Invalid padded message length');
+  }
+
+  // load round constants into Fields
+  let rc = exists(24, () => TupleN.fromArray(24, ROUND_CONSTANTS));
```
querolita, line 490: I know this is how it was done in OCaml, but I wonder if it would make sense to defer the creation of `rc` until `permutation` function. It looks a bit unnecessary to be passing it to both absorb and squeeze at the first level.

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  const state = absorb(paddedMessage, capacity, rate, rc);
+
+  // squeeze
+  const hashed = squeeze(state, length, rate, rc);
+
+  return hashed;
+}
+
+// TODO(jackryanservia): Use lookup argument once issue is resolved
+// Checks in the circuit that a list of Fields are at most 8 bits each
```
querolita, line 502: Oh, okay so the code indeed still needs some calls to checkBytes here and there

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  );
+  assert(length > 0, 'length must be positive');
+  assert(length % 8 === 0, 'length must be a multiple of 8');
+
+  // Input endianness conversion
+  let messageFormatted = inpEndian === 'Big' ? message : message.reverse();
+
+  // Check each Field input is 8 bits at most if it was not done before at creation time
+  if (byteChecks) {
+    checkBytes(messageFormatted);
```
querolita, line 534: nice

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  if (nistVersion) {
+    padded = padNist(messageFormatted, rate);
+  } else {
+    padded = pad101(messageFormatted, rate);
+  }
+
+  const hash = sponge(padded, length, capacity, rate);
+
+  // Always check each Field output is 8 bits at most because they are created here
+  checkBytes(hash);
```
querolita, line 549: indeed

src/lib/keccak.unit-test.ts
```diff
@@ -1,106 +1,136 @@
...
+//     let proof = await Keccak512.nistSha3(x);
+//     return proof.publicOutput;
+//   }
+// );
+
+constraintSystem.fromZkProgram(
+  Keccak256,
+  'ethereum',
+  contains([['Generic'], ['Xor16'], ['Zero'], ['Rot64'], ['RangeCheck0']])
+);
```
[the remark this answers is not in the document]
querolita: Mmmmm I guess the correct shape of each component was checked before. I have the feeling that this `contains` adds no extra information to Keccak256 other than the fact that all of those gates should be in place. Like, it doesn't say anything about the actual shape. 

But I guess it's fine to leave it as a sanity check: in case someone deleted a part of the code preventing some of those gates from being created, or something.

src/lib/keccak.unit-test.ts
```diff
@@ -1,106 +1,136 @@
...
+await equivalentAsync(
+  {
+    from: [array(uint(8), PREIMAGE_LENGTH)],
+    to: array(uint(8), 32),
+  },
+  { runs: RUNS }
+)(
+  (x) => {
+    let thing = x.map(Number);
+    let result = sha3_256(new Uint8Array(thing));
```
querolita, line 65: Oh so this is using another implementation of the Keccaks to compare against what the gadget provides? Nice

I wonder if it could make sense to create tests using some of the test vectors proposed by the Keccak Team. Maybe not to be run inside CI all the time, but at least having run them once before shipping it for production.

querolita (commented):
Looks good to me. I left a few comments to ask for some clarifications regarding the exact constraints that are being added behind the scenes, but other than this it all looks good to me. Will be glad to approve as soon as we solve those concerns.

src/lib/keccak.ts
jackryanservia: JavaScript doesn't inculde a `zip` function. To my knowledge this is the most readable way to do this.

src/lib/keccak.ts
jackryanservia: I added a comment which explains that this function creates constraints (`assertEquals` creates a constraint).

src/lib/keccak.ts
```diff
@@ -209,18 +389,233 @@ function permutation(state: Field[][], rc: Field[]): Field[][] {
...
+  assert(capacity > 0, 'capacity must be positive');
+  assert(
+    capacity < KECCAK_STATE_LENGTH,
+    'capacity must be less than KECCAK_STATE_LENGTH'
+  );
+  assert(length > 0, 'length must be positive');
+  assert(length % 8 === 0, 'length must be a multiple of 8');
+
+  // Input endianness conversion
+  let messageFormatted = inpEndian === 'Big' ? message : message.reverse();
```
mitschabaude, line 530: `message.reverse()` mutates the input, which no function should ever do unexpectedly. reverse a copy instead:
```suggestion
  let messageFormatted = inpEndian === 'Big' ? message : [...message].reverse();
```

mitschabaude (changes requested):
Still not done reviewing but sending it now because I found a vulnerability.

Besides that, I wanted to call out: This keccak implementation handles a fixed-length input. Like, for example, the padding and number of sponge absorb calls is hard-coded based on the message length.

I assume many applications will want to hash variable-length messages. This needs non-trivial amount of extra work, so we might want to schedule that as a follow-up task. (cc @nicc)

src/lib/keccak.ts
mitschabaude: Where exactly? I've looked carefully, and I think all words that are split up by `keccakStateToBytes` and where the bytes _are actually used_, are constrained in the final `checkBytes` call on the output in the `hash()` function

But I fully agree that not having creation of bytes and checks on them co-located is confusing and error-prone.

src/lib/keccak.ts
mitschabaude: in `hash()`. but should be done in this function already

mitschabaude (changes requested):
I'm done with the review, and this last round is mostly about efficiency.

There's a higher-level design question which also affects efficiency: Do all use cases need to pass the inputs as bytes? Because, if someone wanted to pass in data which already comes in 64-bit chunks, that would obviously be much more efficient, we could save all the bytes-to-words conversion and externally range checking the input bytes.

The same observation applies to the output, and there's _definitely_ a use case which is more efficient when returning the output as 64-bit chunks: ECDSA.


For ECDSA, the output of this needs to become a 256-bit field element which is then represented as 3 88-bit limbs. So, if we return 4 64-bit words directly, we can convert them into limbs as follows:
```
input: words w0, w1, w2, w3
output: limbs l0, l1, l2

l0 = w0 | slice(w1, 0, 24)
l1 = slice(w1, 24, 64) | slice(w2, 0, 48)
l2 = slice(w2, 48, 16) | w3
```

so we can even use two output words without cutting them apart entirely, and need to cut the middle two words into larger pieces which is much cheaper than into bytes.

therefore, my ask is to definitely make some version of this which returns full 64-bit words. for consistency, then, taking 64-bit words as input also makes sense.

maybe the following would be a useful collection of interfaces (just a suggestion):
* a low-level Keccak interface which takes and returns 64-bit fields
* a high-level ECDSA interface which takes a byte string, does both Keccak and the EC part, and handles the word-to-limb conversion internally
* helper functions which split an array of 64-bit fields into 8-bit fields and vice versa

this is just a suggestion - pick something you like with the constraint that it doesn't force us to go to intermediate bytes in ECDSA

src/lib/keccak.ts
```diff
@@ -203,24 +380,225 @@ function round(state: Field[][], rc: Field): Field[][] {
...
+    // pad the block with 0s to up to 1600 bits
+    const paddedBlock = block.concat(zeros);
+    // padded with zeros each block until they are 1600 bit long
+    assert(
+      paddedBlock.length * 8 === KECCAK_STATE_LENGTH,
+      `improper Keccak block length (should be ${KECCAK_STATE_LENGTH})`
+    );
+    const blockState = getKeccakStateOfBytes(paddedBlock);
+    // xor the state with the padded block
+    const stateXor = keccakStateXor(state, blockState);
```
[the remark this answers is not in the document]
querolita: > XORing the 8 zero words with the current state will create a bunch of XOR gates, which are all no-ops, so they could just be left out.

Yes, this could be an optimization, leaving out those XORs and avoid extra rows that do nothing to the inputs.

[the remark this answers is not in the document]
querolita: > by making Gadgets.xor() do nothing
if this is possible, then it sounds like a good idea

src/lib/keccak.ts
```diff
@@ -203,24 +380,225 @@ function round(state: Field[][], rc: Field): Field[][] {
...
+  const bytesPerSqueeze = rate / 8;
+  // number of squeezes
+  const squeezes = Math.floor(length / rate) + 1;
+  // multiple of rate that is larger than output_length, in bytes
+  const outputLength = squeezes * bytesPerSqueeze;
+  // array with sufficient space to store the output
+  const outputArray = Array(outputLength).fill(Field.from(0));
+  // first state to be squeezed
+  const bytestring = keccakStateToBytes(state);
+  const outputBytes = bytestring.slice(0, bytesPerSqueeze);
```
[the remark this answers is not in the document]
querolita: > better to only convert the first length / 64 words of the state to bytes.

indeed, the new zkvm keccak only converts the needed 4 words at the end

src/lib/keccak.ts
querolita: >  in the final checkBytes call on the output in the hash() function

yeah, I saw that when I reached to the end of the review

src/lib/keccak.ts
```diff
@@ -64,14 +102,156 @@ const ROUND_CONSTANTS = [
...
+  );
+
+  // Check all words are composed correctly from bytes
+  for (let y = 0; y < KECCAK_DIM; y++) {
+    for (let x = 0; x < KECCAK_DIM; x++) {
+      const idx = BYTES_PER_WORD * (KECCAK_DIM * y + x);
+      // Create an array containing the 8 bytes starting on idx that correspond to the word in [x,y]
+      const word_bytes = bytestring.slice(idx, idx + BYTES_PER_WORD);
+      // Assert correct decomposition of bytes from state
+      checkBytesToWord(state[x][y], word_bytes);
```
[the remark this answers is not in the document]
nicc: Thanks @mitschabaude. Agreed that we should follow up with this. Have created a new issue [here](https://github.com/o1-labs/o1js/issues/1301).

mitschabaude (approved):
I think this is good to merge, it's good as it stands. Constraint optimizations and addition of a 64-bit API can be done in another PR.

Issue #661, opened 2022-12-21
Expose `reduce` to ZkProgram, use for rolling up Merkle tree updates


mitschabaude:

Trivo25:
# Proof Requests and Recursive Reducer

## Context

Actions and reducers influence the circuit size of a smart contract significantly. The reducer is designed to process a fixed-size list of actions, with the default capacity set to 32 actions. This size can be user-defined. For each action in the list, a user-defined callback function is applied individually. This design imposes certain limitations, as actions must be processed in small batches, and the reducer's current architecture does not allow for runtime adjustments to accommodate a varying number of outstanding actions. Such a limitation could lead to potential issues for users. For instance, if the number of actions exceeds the pre-set limit, it could result in an "overflow" scenario. This overflow can trigger a deadlock in the reducer, rendering it unable to process any actions. This deadlock occurs because the precondition on the action state might not be met, preventing further processing and potentially causing operational disruptions in the smart contract's functionality.

## Proposed solution

Leveraging recursion allows the development of a reducer capable of handling lists of actions of any length, circumventing the constraints imposed by circuit size. Through the use of ZkProgram, actions can be grouped into distinct proofs. These individual proofs can then be recursively verified, enabling a more efficient and seamless reducer function. This approach ensures that the reducer can manage an extensive number of outstanding actions efficiently, without the limitations associated. By implementing this recursive strategy, we significantly enhance the reducers flexibility and scalability,.

## Hurdles 

To streamline the process, o1js is designed to automatically create a proof for the recursive reduce function without the user needing to manage it themselves. It does this by requesting proofs in advance, before the smart contract gets proven. These proofs are then generated in advance and provided to the smart contract when needed. However, o1js can't handle proofs in an "on-the-fly" manner without asynchronous circuits and witness blocks, this means all proofs must be arranged beforehand. This setup ensures the user doesn't have to deal with the complexities of proof generation, making the whole process smoother and more user friendly.

## Workaround and proposed API

Requesting proofs is beneficial not just for internal operations but also significantly enhances the developer experience when it comes to integrating ZkProgram into smart contracts.

#### Proof requesting

```ts
// ..
class RequestProof extends SmartContract {
  @method doSmth() {
    let [proof, publicOutput] = this.requestProof(MyProgram, 'baseCase', privateInputs, publicInputs);
    proof.verify();
    // ..
  }
}

```

[...]

PR #1300, opened 2023-12-06
Reducer map example

Trivo25:
This contract emulates a "mapping" data structure, which is a key-value store, similar to a dictionary or hash table or `new Map<K, V>()` in JavaScript.
In this example, the keys are public keys, and the values are arbitrary field elements.
This utilizes the `Reducer` as an append online list of actions, which are then looked at to find the value corresponding to a specific key.

### DIFF
--- a/src/examples/zkapps/reducer/map.ts
+++ b/src/examples/zkapps/reducer/map.ts
@@ -98,14 +98,12 @@ console.log(`method size for a "mapping" contract with ${k} entries`);
 console.log('get rows:', cs['get'].rows);
 console.log('set rows:', cs['set'].rows);
 
-// a test account that pays all the fees, and puts additional funds into the zkapp
+// a test account that pays all the fees
 let feePayerKey = Local.testAccounts[0].privateKey;
 let feePayer = Local.testAccounts[0].publicKey;
 
 // the zkapp account
-let zkappKey = PrivateKey.fromBase58(
-  'EKEQc95PPQZnMY9d9p1vq1MWLeDJKtvKj4V75UDG3rjnf32BerWD'
-);
+let zkappKey = PrivateKey.random();
 let zkappAddress = zkappKey.toPublicKey();
 let zkapp = new StorageContract(zkappAddress);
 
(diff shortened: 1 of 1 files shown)


emlautarom1:
This looks extremely promising. I have a couple of questions tough:
- Could you comment on the possible issues this approach has?
- How big can the `k` constant be?
- Are all actions stored in archive nodes? Or do these nodes eventually "prune" old actions?

Trivo25:
> Could you comment on the possible issues this approach has?

This is a relatively naive implementation that demonstrates how Actions/Reducers can be used as on-chain data structures. Without more work, the amount of elements `k` is limited to a relatively small number (mainly limited by the circuit size, can be extended by using [recursive proofs](https://github.com/o1-labs/o1js/issues/661)). It's also not trivial to update already set key/value pairs. But for a set-once mapping structure, this should be enough.

> How big can the k constant be?

For this particular example, around 1024 entries. 

> Are all actions stored in archive nodes? Or do these nodes eventually "prune" old actions?

Yes they are stored in archive nodes. I am not aware of archives pruning old state.

src/examples/zkapps/reducer/map.ts
```diff
@@ -0,0 +1,200 @@
...
+    let { state: optionValue } = this.reducer.reduce(
+      pendingActions,
+      Option,
+      (
+        _state: Option,
+        _action: {
+          key: Field;
+          value: Field;
+        }
+      ) => {
```
mitschabaude, line 73: nits:
* why use `_` for variables that are used
* TS shouldn't need type annotations here because `this.reducer.reduce` should infer the callback params from the state and action types

mitschabaude (approved):

dfstio:
As I understand, this example uses storage in the archive node that is not part of the proved state and does not use at all on-chain 8 Fields storage for keeping the Merkle Map root.

If we are speaking about possible designs for updating big Merkle Trees on-chain, one of the possible designs that would allow proving inclusion to the Merkle Tree is using the Sparse Merkle Trees that can calculate new root on-chain, then be fully reconstructed off-chain to get the witness. 

An example of such an approach is the Ethereum deposit contract
https://github.com/ethereum/consensus-specs/blob/34fc0a5d09fae6649e0c6ac7a0cb09ff5a999957/solidity_deposit_contract/deposit_contract.sol#L80

I was thinking about combining this approach with a Token contract, where we can mint tokens to several addresses and use several 8-field contract states to get a total of 32 or 64 Fields to store Sparse Merkle Tree branches.

This way, we can update on-chain Sparse Merkle Tree branches and root and use on-chain witness verification based on the proved state.

PR #1302, opened 2023-12-11
Fix smart contracts not verifying proofs

### LINKED ISSUE #1166: Recursive verification does not fail on proof from a foreign ZkProgram

### LINKED ISSUE #1231: Smart Contract accepts any proof as long as types match
[According to the docs](https://docs.minaprotocol.com/zkapps/o1js/recursion#recursively-verifying-a-linear-recursive-program-in-a-zkapp), we can use proofs created in the browser as parameters for a method in a Smart Contract which will get verified by the contract as follows:

```typescript
@method foo(proof: Proof) {
  proof.verify().assertTrue();
  /* ... the rest of our method
   * For example using the total value as the fee for some other transaction. */
}
```

First of all, the previous code does not compile due to:
- `Proof` being a generic type (`Proof<I, O>`)
- Not being able to use `Proof` directly (we need to subclass it).
- `assertTrue()` is not available due to `verify()` returning `void`.

I followed the code in the examples repository (https://github.com/o1-labs/docs2/blob/main/examples/zkapps/09-recursion/src/rollup.ts) and I ended up with the following example:

```typescript
import { AccountUpdate, Bool, Experimental, Mina, PrivateKey, PublicKey, SmartContract, State, UInt64, method, state } from 'o1js';

export const RealProof = Experimental.ZkProgram({
  methods: {
    make: {
      privateInputs: [UInt64],

      method(value: UInt64) {
        let expected = UInt64.from(34);
        value.assertEquals(expected);
      },
    },
  },
});

export const FakeProof = Experimental.ZkProgram({
  methods: {
    make: {
      privateInputs: [UInt64],

      method(value: UInt64) {
        Bool(true).assertTrue();
      },
    },
  },
});

class BrokenProof extends Experimental.ZkProgram.Proof(RealProof) { }
class Broken extends SmartContract {
  @state(Bool) isValid = State<Bool>();

  init() {
    super.init();
    this.isValid.set(Bool(false));
  }

  @method setValid(proof: BrokenProof) {
    proof.verify();
    this.isValid.set(Bool(true));
  }
}

describe('Broken', () => {
  let deployerAccount: PublicKey;
  let deployerKey: PrivateKey;
  let senderAccount: PublicKey;
  let senderKey: PrivateKey;
  let zkAppAddress: PublicKey;
  let zkAppPrivateKey: PrivateKey;
  let zkApp: Broken;

  beforeAll(async () => {
    await RealProof.compile();
    await FakeProof.compile();
    await Broken.compile();
  });

  beforeEach(() => {
    const Local = Mina.LocalBlockchain({ proofsEnabled: true });
    ({ privateKey: deployerKey, publicKey: deployerAccount } = Local.testAccounts[0]);
    ({ privateKey: senderKey, publicKey: senderAccount } = Local.testAccounts[1]);
    zkAppPrivateKey = PrivateKey.random();
    zkAppAddress = zkAppPrivateKey.toPublicKey();
    zkApp = new Broken(zkAppAddress);
  });

  async function localDeploy() {
    const txn = await Mina.transaction(deployerAccount, () => {
      AccountUpdate.fundNewAccount(deployerAccount);
      zkApp.deploy();
    });
    await txn.prove();
    await txn.sign([deployerKey, zkAppPrivateKey]).send();
  }

  it('accepts a fake proof', async () => {
    await localDeploy();

    const value = UInt64.from(99999);
    const fakeProof = await FakeProof.make(value);

    const txn = await Mina.transaction(senderAccount, () => {
      zkApp.setValid(fakeProof);
    });
    await txn.prove();
    await txn.sign([senderKey]).send();

    const isValid = zkApp.isValid.get();
    expect(isValid).toEqual(Bool(true));
  });
});
```

Note how we're able to use `FakeProof` despite expecting a `RealProof`. Either the code above is incorrect (should be written differently), or this is a bug.

mitschabaude:
Fixes #1166
Fixes #1231

**Description of the fix**

Requiring a proof to verify is based on mutation of the proof value: `verify()` sets the `shouldVerify` field to `Bool(true)`. On the other hand, the smart contract wrapper clones all input arguments to the prover to avoid modifying them (which can easily lead to bugs because we run the method multiple times). The cloning made the contract unable to mutate the original proof.

The short-term fix was to not clone proofs, similar to how we don't clone primitive provable types. It would be better if a future refactor made proof verification not dependent on mutating a method argument. Instead, we could introduce a global context which holds the proofs to potentially verify inside a circuit, and set properties on those. Also, it would be nice to have an error when we forget to verify a proof (and an explicit method to not verify them).

**Changes**
* adds a regression test to CI
* add a unit test checking the result of `sortMethodArguments` and the outputs of `picklesRuleFromFunction` (the latter would have caught this bug)


src/tests/fake-proof.ts
```diff
@@ -0,0 +1,97 @@
...
+const realProof = await RealProgram.make(UInt64.from(34));
+
+// zkprogram accepts proof
+const brokenProof = await RecursiveProgram.verifyReal(realProof);
+assert(
+  await verify(brokenProof, programVk.data),
+  'recursive program accepts real proof'
+);
+
+// contract rejects proof
```
nicc, line 89:
```suggestion
// contract accepts proof
```

Trivo25 (approved):

Issue #61, opened 2022-01-12
Add secp256k1 signature verification support


jasongitmail:

Comdex:
I'm interested in this feature. Also, it would be nice to add RSASSA-PKCS1-v1_5 signature verification and SHA256 hash support.
In this way, we can take advantage of the existing Internet infrastructure, such as the SubtleCrypto interface, which is widely supported by browsers, to manage the keys that can be used to control snapp. SubtleCrypto supports non-exportable keys to prevent the risk of disclosure of plaintext private keys.

mitschabaude:
closed by #1307

PR #1307, opened 2023-12-12
Integrate ECDSA + Keccak

mitschabaude:
### Changes

* Refactor Keccak
  * Note: This started as a big refactor with the goal of returning 64-bit words. Then I realized that this is not compatible with ECDSA and we really need bytes. This path might explain the large amount of changes introduced here. I think most of them are valuable
  * A beneficial side-effect is that we now use the minimal amount of constraints for to/from bytes conversion
  * A bad side-effect is that the 224 bit variant of SHA3 is no longer supported since the core `sponge()` function was rewritten with the assumption that `length` and `capacity` are multiples of 64 bits 😬 With a bit of extra work it should be possible to bring back 224 and keep constraints optimal. Honestly, I doubt 224 will be needed.
* Add simple conversion circuit which prepares Keccak output for ECDSA
* Change ECDSA `sign()` and `verify()` to take message bytes and do the Keccak hash
* Update ECDSA doccomments
* Change example to contain ECDSA and Keccak circuits in combination and individually
* Expose Keccak as an export
* Save some constraints by not xoring with zero constants

### TODOs not covered by this PR

* Doc comments on Keccak methods
* E2E test with a real signature
* Introducing `UInt8` as a safe input to Keccak / ECDSA as in https://github.com/o1-labs/o1js/pull/999
  * assigned to @mitschabaude 
* docs, docs, docs

### Results

image


mitschabaude:
closes #61

Trivo25 (approved):

src/lib/keccak.ts
```diff
@@ -1,31 +1,23 @@
 import { Field } from './field.js';
```
Trivo25, line 1: Actually, does this fit into `src/lib/gadgets` better than `src/lib`? Similar to ECDSA

jackryanservia: It probably does! IIRC, I put it there before ECDSA was done, but it should probably live wherever that lives!

mitschabaude: Agreed, I can move it in the later PR!

PR #1309, opened 2023-12-12
Add a little hint to README-dev.md

TtheBC01:
I always forget to pull external git modules when I work directly with o1js. Proposing to add a tiny section to README-dev.md to save others a little time when they first get started.


mitschabaude (approved):
Thanks! Very important to have the full dev workflow covered there

PR #1315, opened 2023-12-13
Improve keccak witness gen performance

mitschabaude:
Keccak witness generation is very slow.

This PR improves time for hashing 32 bytes by about **50%** from ~18s to ~9s.

I identified the bottleneck that causes most of the remaining time and made an issue for it here: https://github.com/o1-labs/o1js/issues/1316


run-debug
```diff
@@ -0,0 +1 @@
+node --inspect-brk --enable-source-maps src/build/run.js $@
```
Trivo25, line 1: I was about to add that as well!! :D

Trivo25 (approved):

PR #1338, opened 2023-12-19
Faster to_bignum_bigint

mitschabaude:
- Bytes.random
- keccak witness gen benchmark
- bindings
- unit-test keccak in provable version


mitschabaude:
closing because of https://github.com/MinaProtocol/mina/pull/14742#issuecomment-1862660717

PR #1345, opened 2023-12-21
Faster fields to bytes conversion

### LINKED ISSUE #1212: Performance: Faster Field conversion

mitschabaude:
Closes #1212

This speeds up conversion between three different representations of field elements:
* bigints (used in TS)
* byte arrays (used in Rust)
* hex string (used in the JSON returned from `constraintSystem()`)

The techniques used are:
* operate on 64 bits at a time instead of individual bytes, which is faster
* avoid allocating many throwaway byte arrays that are only used as an intermediate step in conversions

Most of the changes are in the bindings PR:
https://github.com/o1-labs/o1js-bindings/pull/228

## Results

* contribution of conversions to ecdsa cached compile time: down from 9.5% to <1%
* contribution of conversions to ecdsa prover time: down from 2.5% to <0.5%
* `gatesFromJson()` time during compile: down from >300ms to 80ms


src/examples/crypto/ecdsa/ecdsa.ts
```diff
@@ -27,14 +27,6 @@ const keccakAndEcdsa = ZkProgram({
         return signature.verify(message, publicKey);
       },
     },
-
-    sha3: {
-      privateInputs: [],
-      method(message: Bytes32) {
```
mitschabaude, line 33: I felt this was clobbering up the ecdsa test, and is covered in another circuit regression test already

Trivo25 (approved):

PR #1346, opened 2023-12-21
Fix hash example

mitschabaude:
Bugs in the example were reported on discord. 
* improper use of `Bytes`
* `Hash.hash()` was throwing with `this.initialState is not a function`

Also fixes the type returned by `Poseidon.update()`.


ymekuria (approved):

MartinMinkov (approved):
Nice! :t-rex:

PR #1350, opened 2024-01-02
Rename all references of `snarkyjs` to `o1js` internally

MartinMinkov:
:link: bindings: https://github.com/o1-labs/o1js-bindings/pull/231


mitschabaude (changes requested):
I think snarky.js shouldn't be changed to o1js.js, since the name refers to snarky-ml

MartinMinkov:
> I think snarky.js shouldn't be changed to o1js.js, since the name refers to snarky-ml

Fair enough! I'll revert those changes.

mitschabaude:
> > I think snarky.js shouldn't be changed to o1js.js, since the name refers to snarky-ml
> 
> Fair enough! I'll revert those changes.
> 
> EDIT: Should `getO1js` be reverted back to `getSnarky` as well?

Yes probably!

mitschabaude (approved):

PR #1363, opened 2024-01-11
Error feedback on compile-prover discrepancies

mitschabaude:
- start debugging dex example
- start adding types for low level snarky intf
- iterate on types
- iterate on api and fix type error
- runner that can compare circuit constraints against a previous run
- expose more flexible cs processing
- revert returning result
- don't return result from constraint system
- use runCircuit in analyzeMethods
- move zkapp proving logic to separate file
- minor
- WIP: compare constraints in zkapp proof to the ones created in analyzeMethods
- remove debugging, fix basic constraint data
- add memoized witnesses to avoid failing constraints
- hook prover consistency check into prover when detecting a relevant problem
- show traces to constraints in both versions
- seems more correct but I'm not sure
- helper to print account update layout
- fix the damn bug
- [dex] remove obsolete scare comments
- minor example tweak
- submodules

### DIFF
--- a/src/lib/mina/zkapp-proof.ts
+++ b/src/lib/mina/zkapp-proof.ts
@@ -124,7 +124,19 @@ async function createZkappProof(
       try {
         return await prover(publicInputFields, MlArray.to(previousProofs));
       } catch (err) {
-        console.error(`Error when proving ${ZkappClass.name}.${methodName}()`);
+        console.error(
+          `\n\nError when proving ${ZkappClass.name}.${methodName}()`
+        );
+        if (
+          err instanceof Error &&
+          err.message.includes('FieldVector.get(): Index out of bounds')
+        ) {
+          runAsIfProver(transaction, index);
+          console.error(
+            'This is likely due to a mismatch between the circuit at compile and proving time.\n'
+          );
+        }
+
         throw err;
       } finally {
         memoizationContext.leave(id);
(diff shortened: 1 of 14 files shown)


src/lib/provable-context-debug.ts
mitschabaude, line 1: This is the new circuit runner to debug constraints!

src/lib/mina/zkapp-proof.ts
mitschabaude, line 1: The code in this new file consists of:
* Code that handles zkApp proof creation, which was moved here from `account_update.ts`
* New method `debugInconsistentConstraints()`, which is triggered after encountering certain errors in the zkApp prover. It then runs the new `runCircuit()` twice, once in compile and once in prover mode, to identify any mismatch in the generated constraints. If a mismatch is found, we throw a very helpful error that points to it. Otherwise, we throw the original prover error.

src/snarky.d.ts
```diff
@@ -540,8 +547,59 @@ declare const Snarky: {
...
+    };
+  };
 };
 
+type MlRef<T> = [_: 0, contents: T];
+
+type SnarkyVector = [0, [unknown, number, FieldVector]];
+type ConstraintSystem = unknown;
+
+type SnarkyState = [
```
mitschabaude, line 587: type description of snarky's internal state

PR #1364, opened 2024-01-11
Fix in-circuit hashing of complex account update layouts

### LINKED ISSUE #706: Support advanced account updates (prover bug reloaded)
The DEX prover fails when witnessing a proper account update layout. The error that's thrown is the same as the one we resolved here: https://github.com/o1-labs/snarkyjs/issues/498

A peculiar thing about this error is that it's thrown from a completely unexpected place, which gives no info at all about what to debug.

Last time when we resolved this error in snarkyjs, it was caused by a compile vs prover discrepancy in a part of snarkyjs that's complex and not tested well. I assume the same is the case now.

The same type of error was also reported by two community members (@maht0rz and @rpanic), and afaik was resolved by fixing something in their code.

There are two tasks here:

**Task 1**: resolve the concrete error occurring in the DEX, by finding the concrete compile vs prove discrepancy and adding thorough tests for the involved snarkyjs internals. This is product-eng work.

**Task 2**: Investigate why this (likely) compile/prove discrepancy is ending up as an error in a seemingly unrelated part of Pickles. Add invariant assertions to make sure snarky or Pickles catch this error earlier and are able to tell the users that something is wrong with their circuit. Make snarky robust enough so that problems in one circuit cannot leak 'outside' of it and mess up a different part of the proving process. This is crypto work, and captured separately in the following issue: https://github.com/o1-labs/snarkyjs/issues/629

How to reproduce:

- check out `snarkyjs/main`, run `npm i` and `npm run build`
- Comment out the line of code after this comment, and instead uncomment the block of code directly below it (see all-caps comments) https://github.com/o1-labs/snarkyjs/blob/52ec32befaf7d4364ca8e71ebb4c3b73ed182afd/src/examples/zkapps/dex/dex.ts#L434
- run the following:
```
./run src/examples/zkapps/dex/happy-path-with-proofs.ts --bundle
```

mitschabaude:
Fixes #706

This is the minimal PR that fixes the bug.

_Finding_ the bug was possible thanks to new infrastructure for catching compile/prove discrepancies, which now lives in a follow-up PR: https://github.com/o1-labs/o1js/pull/1363 (still needs clean-up work, and relies on changes to snarky that need broader alignment)

Note: This doesn't add unit tests. Should it?
It doesn't seem obvious that complex account update layouts can be tested in isolation, since it was the global mutable structure of transactions, and unwanted cross-interactions between proofs at multiple layers of the callstack, that broke them. It seems that integration tests are needed to exhibit non-trivial failure cases, and this PR already fixes such an integration test as its main result


src/lib/account_update.ts
```diff
@@ -1321,6 +1347,7 @@ class AccountUpdate implements Types.AccountUpdate {
       accountUpdate.body.mayUseToken.inheritFromParent.assertFalse();
       return;
     }
+    accountUpdate.children.callsType = { type: 'None' };
```
mitschabaude, line 1350: this fixed the bug. literally one line 😓

src/lib/account_update.ts
```diff
@@ -1096,11 +1096,37 @@ class AccountUpdate implements Types.AccountUpdate {
     return { accountUpdate, calls };
   }
 
+  toPrettyLayout() {
```
mitschabaude, line 1099: this was useful for debugging, leaving it in here

nicc:
@mitschabaude, since the DEX app is run on CI now, we would pick it up if this regressed / the bug was reintroduced right? 

Intuitively, account updates seem like the kind of thing that shoud have good unit test coverage (and be easy to test) so I'm inclined to say it should have a unit test. If that's a lot of work then we could dig into why it's hard and see if that should be fixed?

mitschabaude:
> @mitschabaude, since the DEX app is run on CI now, we would pick it up if this regressed / the bug was reintroduced right?

Yes, we would pick up if this regressed. DEX was already running in CI but in an insecure version that worked around the bug, which is fixed here

We do have unit tests for account updates, but what broke here still seems hard to unit-test. I could devise a test that would catch a regression, but it would be quite artificial, setting up an interaction between multiple calls to `approve()` in different proofs which mirrors what happens in the DEX example. Not what feels like a good unit test.

Ultimately, I think the culprit for this bug is the overall structure of the code which relies heavily on mutating a shared transaction object, including mutating the individual account updates, from random places in the code. The global structure of the code is bad, and I think it needs some design and refactoring work to become better. The broad idea for me would be to make the global list of account updates largely immutable except in a few well-defined places -- such that the effects of the code fixed here become more local, after which unit tests become meaningful.

MartinMinkov (approved):

PR #1367, opened 2024-01-11
Use actual network genesis constants (genesisTimestamp, slotTime, accountCreationFee).

### LINKED ISSUE #1366: genesisTimestamp is hardcoded to a wrong date
```typescript
  // copied from mina/genesis_ledgers/berkeley.json
  // TODO fetch from graphql instead of hardcoding
  const genesisTimestampString = '2023-02-23T20:00:01Z';
  const genesisTimestamp = UInt64.from(
    Date.parse(genesisTimestampString.slice(0, -1) + '+00:00')
  );
```
https://github.com/o1-labs/o1js/blob/main/src/lib/mina.ts#L742

shimkiv:
Closes #1366


shimkiv:
It probably won't work since `getCachedGenesisConstants` won't wait for promise to resolve...

src/examples/fetch_live.ts
```diff
@@ -0,0 +1,73 @@
...
+  archive: useCustomLocalNetwork
+    ? 'http://localhost:8282'
+    : 'https://api.minascan.io/archive/berkeley/v1/graphql',
+  lightnetAccountManager: 'http://localhost:8181',
+});
+Mina.setActiveInstance(network);
+const transactionFee = 100_000_000;
+
+// Fee payer setup
+console.log('');
```
Trivo25, line 19: empty console log

src/examples/fetch_live.ts
```diff
@@ -0,0 +1,73 @@
+import { expect } from 'expect';
```
Trivo25, line 1: i think this file could use some spacing and line breaks

src/lib/mina.ts
```diff
@@ -369,6 +372,11 @@ interface Mina {
 }
 
 const defaultAccountCreationFee = 1_000_000_000;
+const defaultNetworkConstants = {
+  genesisTimestamp: UInt64.from(0),
+  slotTime: UInt64.from(3 * 60 * 1000),
+  accountCreationFee: UInt64.from(defaultAccountCreationFee),
```
Trivo25, line 378: do we need the account creation fee twice or can we simplify to only use one source?

Trivo25 (approved):
lgtm just a couple of nits!

src/lib/mina.ts
shimkiv: Maybe then we should:
1. Remove [accountCreationFee API](https://github.com/o1-labs/o1js/pull/1367/files#diff-1f92503aa572cd2ea7bc843e4e77809f3aef81f49bb25995a80d032127482ffcR749);
2. And allow return [default network constants](https://github.com/o1-labs/o1js/pull/1367/files#diff-1f92503aa572cd2ea7bc843e4e77809f3aef81f49bb25995a80d032127482ffcR375) instead of [throwing](https://github.com/o1-labs/o1js/pull/1367/files#diff-1f92503aa572cd2ea7bc843e4e77809f3aef81f49bb25995a80d032127482ffcR768);
3. Update all API usage (obviously).

WDYT?

src/examples/fetch_live.ts
shimkiv: Will reformat it, thanks.

src/lib/mina.ts
mitschabaude: sounds good, but please deprecate the old API, not remove

Trivo25 (approved):
I think we should add the new API changes to the changelog, otherwise its good to go!

shimkiv:
> I think we should add the new API changes to the changelog, otherwise its good to go!

Added one, thank you.

shimkiv:
@Trivo25 @mitschabaude if all is good now then I'll merge later today.

PR #1372, opened 2024-01-16
minor issue related to list numbering.

avocadodefi:
Your text is very well written and mostly free from typos. However, I noticed a minor issue related to list numbering. In the "Contributing to o1js" section, the list numbering restarts at 1 after the first item. Here's the corrected version:


mitschabaude:
This is actually how markdown lists work. The code has 1s, but the text is automatically rendered with the correct numbers (makes it easier to maintain, you can add or shuffle items without caring about the numbering)

PR #1373, opened 2024-01-16
Faster TS field inverse

mitschabaude:
- adapt tictoc and examples build
- add assert() in more basic module
- submodules


Trivo25 (approved):

Trivo25 (commented):
This could also be a great addition to the changelog :eyes:

PR #1377, opened 2024-01-18
Hashed provable type

mitschabaude:
bindings: https://github.com/o1-labs/o1js-bindings/pull/238

This follows up on https://github.com/o1-labs/o1js/pull/1376 with a variant of the same idea: Instead of representing a value by its packed field elements, we now represent it by its hash.

The benefit is that a hash is (potentially) even smaller than packed field elements -- it's a single field element. So, this can achieve a better improvement to constraint count as `Packed`, in the same situations.

The downside is that the overhead of hashing is much higher than that of packing, so there might be scenarios where either technique is better.

We use `Hashed` instead of `Packed` in ECDSA, which achieves a _tiny_ improvement in combination with changing one of the default parameters for point caching (it's slightly worse with the original parameters, which were the optimal ones before, so we see that there's a trade-off. Also, packing seems still better for constant points, but I don't think it's worth adding the complexity of using different representation depending on whether the point is constant)

This also directly exposes the low-level tool underneath - `Poseidon.hashPacked<T>()`, to improve the `Poseidon.hash(value.toFields())` pattern found in many code bases


src/lib/gadgets/elliptic-curve.ts
```diff
@@ -416,10 +416,10 @@ function multiScalarMul(
 
   // pack points to make array access more efficient
```
Trivo25, line 417: it's a hashed point now! 
`// hashed points to make array access more efficient`

Trivo25 (approved):

PR #1378, opened 2024-01-19
Fix poseidon hash bottleneck in Mina submodule

MartinMinkov:
mina: https://github.com/MinaProtocol/mina/pull/14921
bindings: https://github.com/o1-labs/o1js-bindings/pull/237

# Summary
When performing a Poseidon hash, the Rust code responsible for performing the update has a bottleneck, primarily due to re-computing its internal parameters every time the function is invoked, as seen in the figure below.
image
It can be seen that `mina_poseidon::pasta::fp_kimchi::params` takes nearly 50ms, where the total running time of the hash stack trace is 53ms. By using the [static version of the params function](https://github.com/o1-labs/proof-systems/blob/ea3f1c9e28f1ea3a066684149c7f924634143125/poseidon/src/pasta/fp_kimchi.rs#L832), we only generate the params once and then reload them every time it's used.

The results of the change can be seen in the figure below.
image
The profile shows that the params have not been generated this time, making the total running time 4ms. 
This means we have a **13x speedup on Poseidon hashing** :tada:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
-[submodule "src/snarkyjs-bindings"]
+[submodule "src/o1js-bindings"]
 	path = src/bindings
-	url = https://github.com/o1-labs/snarkyjs-bindings.git
+	url = https://github.com/o1-labs/o1js-bindings.git
 [submodule "src/mina"]
 	path = src/mina
 	url = https://github.com/MinaProtocol/mina.git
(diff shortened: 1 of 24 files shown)


emberian (approved):
wow!

mitschabaude (approved):
Amazing!

Trivo25 (approved):
This improvement is worthy of a changelog entry imo!

PR #1387, opened 2024-01-26
(WIP) Add support for mainnet.

shimkiv:
Accompanying PR to the https://github.com/o1-labs/zkapp-cli/pull/564


mitschabaude:
some initial feedback:
* I don't like making the network id (yet another) global variable. It should just be a property of the Mina instance
* It was a deliberate choice to not have different network ids in `signFields()` and the in-circuit signature verification. The concept of a network id is domain-specific to transaction signing and has no place in a general-purpose signing API. So let's revert using it there.

src/lib/mina.ts
```diff
@@ -685,61 +687,64 @@ function LocalBlockchain({
...
 /**
  * Represents the Mina blockchain running on a real network
  */
 function Network(graphqlEndpoint: string): Mina;
-function Network(endpoints: {
-  mina: string | string[];
-  archive?: string | string[];
-  lightnetAccountManager?: string;
-}): Mina;
+function Network(endpoints: MinaNetworkEndpoints): Mina;
```
mitschabaude, line 700: please revert this change - it would lead to worse intellisense

also, it makes more sense to me to add `networkId` as a property of the same object than to make it a nested object

shimkiv:
> some initial feedback:

Thank you!

> * I don't like making the network id (yet another) global variable. It should just be a property of the Mina instance

I thought about it as well but then the networkId is required in different places and to not evaluate complete mina.ts while importing and limiting access to whatever getters I thought to make it with global one. Will think more on it then. 

> * It was a deliberate choice to not have different network ids in `signFields()` and the in-circuit signature verification. The concept of a network id is domain-specific to transaction signing and has no place in a general-purpose signing API. So let's revert using it there.

By reverting it you mean setting it back to `'testnet'` value or set it to `'mainnet'` to deliver these changes after HF?

src/lib/mina.ts
shimkiv: > please revert this change - it would lead to worse intellisense

Will do, thanks.

> also, it makes more sense to me to add `networkId` as a property of the same object than to make it a nested object

I also was thinking about it but since there might be more options in future that probably won't be related to endpoints I decided to extract it into nested object. Will refactor then.

shimkiv:
@mitschabaude can you also please help me in understanding unit tests failures.
The pipeline for [88025a4](https://github.com/o1-labs/o1js/actions/runs/7677641684/job/20926629251) failed with `invalid signature`, then next commit [dfbea22](https://github.com/o1-labs/o1js/actions/runs/7678945839/job/20930135754#step:4:263) failed by another issue, this time mina-signer, not o1js, but there were no changes that could have fix initial failure.
Then I committed fix [8a6e3c3](https://github.com/o1-labs/o1js/pull/1387/commits/8a6e3c3090c8aba47bba2cbed053be5ae9b7a440) and all is green.
Do you have any idea why `rev:88025a4` could have fail that time?

mitschabaude:
> Do you have any idea why `rev:88025a4` could have fail that time?

I think it was an edge case test failure. I have copied the failing test case and will investigate it

mitschabaude:
> By reverting it you mean setting it back to `'testnet'` value or set it to `'mainnet'` to deliver these changes after HF?

Set it back! "testnet" / "mainnet" is meaningless for these signatures, because they have nothing to do with networks

src/mina-signer/MinaSigner.ts
```diff
@@ -124,7 +124,7 @@ class Client {
    */
   signFields(fields: bigint[], privateKey: Json.PrivateKey): Signed<bigint[]> {
     let privateKey_ = PrivateKey.fromBase58(privateKey);
-    let signature = sign({ fields }, privateKey_, 'testnet');
+    let signature = sign({ fields }, privateKey_, 'mainnet');
```
mitschabaude, line 127: revert! again, `signFields()` is unrelated to any network and back when this was implemented and internally audited, we agreed on sticking with the `'testnet'` prefix.

just to reiterate: this doesn't mean this can't be used on mainnet. signFields just isn't tied to the network in any way

src/mina-signer/MinaSigner.ts
```diff
@@ -144,7 +144,7 @@ class Client {
       Signature.fromBase58(signature),
       { fields: data },
       PublicKey.fromBase58(publicKey),
-      'testnet'
+      'mainnet'
```
mitschabaude, line 147: revert

src/lib/signature.ts
```diff
@@ -235,21 +236,23 @@ class Signature extends CircuitValue {
    * Signs a message using a {@link PrivateKey}.
    * @returns a {@link Signature}
    */
-  static create(privKey: PrivateKey, msg: Field[]): Signature {
+  static create(privKey: PrivateKey, msg: Field[], networkId: NetworkId = 'testnet'): Signature {
```
mitschabaude, line 239: revert - `Signature.create()` is the o1js version of `signFields`, and is unrelated to the Mina network

src/lib/signature.ts
```diff
@@ -235,21 +236,23 @@ class Signature extends CircuitValue {
...
     const publicKey = PublicKey.fromPrivateKey(privKey).toGroup();
     const d = privKey.s;
     const kPrime = Scalar.fromBigInt(
       deriveNonce(
         { fields: msg.map((f) => f.toBigInt()) },
         { x: publicKey.x.toBigInt(), y: publicKey.y.toBigInt() },
         BigInt(d.toJSON()),
-        'testnet'
+        networkId
```
mitschabaude, line 247: revert

src/lib/signature.ts
```diff
@@ -235,21 +236,23 @@ class Signature extends CircuitValue {
...
       )
     );
     let { x: r, y: ry } = Group.generator.scale(kPrime);
     const k = ry.toBits()[0].toBoolean() ? kPrime.neg() : kPrime;
     let h = hashWithPrefix(
-      prefixes.signatureTestnet,
+      networkId === 'mainnet'
```
mitschabaude, line 253: revert

src/lib/signature.ts
```diff
@@ -263,10 +266,12 @@ class Signature extends CircuitValue {
    * Verifies the {@link Signature} using a message and the corresponding {@link PublicKey}.
    * @returns a {@link Bool}
    */
-  verify(publicKey: PublicKey, msg: Field[]): Bool {
+  verify(publicKey: PublicKey, msg: Field[], networkId: NetworkId = 'testnet'): Bool {
```
mitschabaude, line 269: revert

src/lib/signature.ts
```diff
@@ -263,10 +266,12 @@ class Signature extends CircuitValue {
...
     const point = publicKey.toGroup();
     let h = hashWithPrefix(
-      prefixes.signatureTestnet,
+      networkId === 'mainnet'
```
mitschabaude, line 272: revert

src/mina-signer/MinaSigner.ts
```diff
@@ -493,6 +493,15 @@ class Client {
     let sk = PrivateKey.fromBase58(privateKeyBase58);
     return createNullifier(message, sk);
   }
+
+  /**
+   * Returns the network ID.
+   * 
+   * @returns {NetworkId} The network ID.
+   */
+  getNetworkId(): NetworkId {
```
mitschabaude, line 502: nit: this doesn't seem like a useful method to me since you can do `this.network`?

if you added this to use the "network id" terminology instead of "network", I suggest to make this a getter, so we can change its behaviour back to a plain property access later:

```
get networkId(): NetworkId
```

src/mina-signer/src/signature.ts
```diff
@@ -35,15 +36,14 @@ export {
...
-  NetworkId,
   signLegacy,
   verifyLegacy,
   deriveNonce,
 };
 
 const networkIdMainnet = 0x01n;
 const networkIdTestnet = 0x00n;
-type NetworkId = 'mainnet' | 'testnet';
+
```
mitschabaude, line 46: good refactor 👍🏻

src/mina-signer/tests/verify-in-snark.unit-test.ts
mitschabaude, line 1: this whole test is not concerned with signatures tied to a network, and shouldn't end up being changed

src/lib/mina.ts
```diff
@@ -389,12 +392,23 @@ const defaultNetworkConstants: NetworkConstants = {
...
   const ledger = Ledger.create();
   let networkState = defaultNetworkState();
+  let minaNetworkId: NetworkId = 'testnet';
+
+  if (networkId) {
+    if (networkId !== 'mainnet' && networkId !== 'testnet') {
+      throw Error(
+        "Invalid network id specified. Please use 'mainnet' or 'testnet'"
+      );
+    }
```
mitschabaude, line 409: remove this check - we consistently rely on types and the type forbids passing anything else than `"mainnet"` or `"testnet"`

src/lib/mina.ts
```diff
@@ -689,57 +705,68 @@ LocalBlockchain satisfies (...args: any) => Mina;
...
     Fetch.setGraphqlEndpoint(minaGraphqlEndpoint);
-  } else if (input && typeof input === 'object') {
-    if (!input.mina)
+  } else if (options && typeof options === 'object') {
+    if (options.networkId) {
+      if (options.networkId !== 'mainnet' && options.networkId !== 'testnet') {
+        throw Error(
+          "Invalid network id specified. Please use 'mainnet' or 'testnet'"
+        );
+      }
```
mitschabaude, line 738: remove this check

src/lib/mina.ts
```diff
@@ -1013,6 +1041,7 @@ function BerkeleyQANet(graphqlEndpoint: string) {
 }
 
 let activeInstance: Mina = {
+  getNetworkId: () => 'testnet',
```
mitschabaude, line 1044: this is good 👍🏻

mitschabaude (commented):
a few more comments, but looks mostly good to me!

src/mina-signer/MinaSigner.ts
shimkiv: Yeah thank you I got it from the previous review, I just thought since we're aiming to the mainnet and there is not differences then probably better to use 'mainnet'. I didn't know there was an agreement.

src/lib/mina.ts
shimkiv: @mitschabaude 
The idea behind this check is to fail as soon as network is configured. Without this checks in-browser vanilla JS will allow configure network with any value:

```javascript
const remoteBlockchain = Mina.Network({
  networkId: 'whatever',
  mina: 'http://localhost:8080/graphql',
})
Mina.setActiveInstance(remoteBlockchain)
```

And will only throw later when `networkId` will actually be used.

Do you still want these checks to be removed?

src/lib/mina.ts
shimkiv: https://github.com/o1-labs/o1js/pull/1387#discussion_r1469463590

src/lib/mina.ts
mitschabaude: yes, I'd remove them. the reasoning is consistency -- there are a million places where we don't validate input types and that break in arbitrary/unhelpful ways when passed an input of the wrong type.

if we validate inputs here, why not anywhere else?

the decision to not validate inputs in general comes with the expectation that o1js is used from typescript. Passing 'whatever' as in your example would result in an error at _compile time_, i.e. before your check would ever be run. from that perspective, the check is dead code.

if you think we shouldn't rely on typescript and do detailed input validation despite type restrictions, then this should be brought forth as part of a general discussion that applies to the entire code base; not added in an ad-hoc way like here which is inconsistent with the rest of the code base.

shimkiv: Removed, internal discussion started. Thank you.

mitschabaude (approved):

PR #1394, opened 2024-01-30
General-purpose token contract + debugging

mitschabaude:
- start debugging dex example
- start adding types for low level snarky intf
- iterate on types
- iterate on api and fix type error
- runner that can compare circuit constraints against a previous run
- expose more flexible cs processing
- revert returning result
- don't return result from constraint system
- use runCircuit in analyzeMethods
- move zkapp proving logic to separate file
- minor
- WIP: compare constraints in zkapp proof to the ones created in analyzeMethods
- remove debugging, fix basic constraint data
- add memoized witnesses to avoid failing constraints
- hook prover consistency check into prover when detecting a relevant problem
- show traces to constraints in both versions
- seems more correct but I'm not sure
- helper to print account update layout
- fix the damn bug
- [dex] remove obsolete scare comments
- minor example tweak
- submodules
- changelog
- reduce diff to main where changes weren't necessary
- add docs and changelog
- more cases in which to check inconsistent constraints
- bindings
- export token contract
- more token contract methods
- fix token contract child adopting
- unlink account updates to prevent them from being in caller's public input in compile()
- (wip) use base token contract for dex
- wip debugging / start fixing
- add data structure for in-circuit call forest construction which only persists through a contract call
- fixed Hashed default hash
- fix edge case in ecdsa unit test


mitschabaude:
not needed anymore

PR #1398, opened 2024-02-01
Merkle list and token account update iterator

mitschabaude:
This PR introduces several provable types to represent _dynamic_-sized data structures (on which we perform a fixed number of operations)
* `MerkleList<T>` represents a list which supports `push()` and `pop()` (if you pop an empty list, it returns a dummy)
* `MerkleListIterator<T>` exposes the functionality of popping through the elements of a Merkle list, but without removing them from the list. Main method: `next()`, which returns a dummy if we're past the end of the list
* `AccountUpdateForest` is built up from Merkle lists recursively to represent a forest of account updates -- i.e., the children of another account update
  * uses the Merkle list hash that is compatible with how the protocol hashes trees of account updates
  * so, as we iterate through an `AccountUpdateForest`, we will be given nodes (account updates) along with _their_ child account updates, which are a Merkle list themselves
* `TokenAccountUpdateIterator` pulls it all together: It iterates _deeply_ through an `AccountUpdateForest`, stepping down in the tree after every new account update with children is encountered, and stepping up once a layer is done. Since this will be used by token contracts, it supports skipping "inaccessible subtrees", i.e. it skips the children of an update that already doesn't inherit token permissions itself. Assuming that no subtree is skipped, the iteration will return every update in the tree exactly once, in depth-first order. After the full tree has been processed, it returns dummies.
  * the logic is heavily indebted to @mrmr1993's [token contract example](https://github.com/MinaProtocol/mina/blob/37d78d962ec070b1336401f722c0058f8dbe3609/src/app/zkapps_examples/tokens/zkapps_tokens.ml).
  * there is an exhaustive new unit test for `TokenAccountUpdateIterator`. It also tests flat iteration through a Merkle list


src/lib/hash.ts
```diff
@@ -66,6 +66,17 @@ const Poseidon = {
     return MlFieldArray.from(newState) as [Field, Field, Field];
   },
 
+  hashWithPrefix(prefix: string, input: Field[]) {
```
mitschabaude, line 69: I'm exporting this since I wanted to use this in dogfooding already as well

src/lib/mina.ts
```diff
@@ -429,8 +429,8 @@ function LocalBlockchain({
     getNetworkId: () => minaNetworkId,
     proofsEnabled,
     /**
-   * @deprecated use {@link Mina.getNetworkConstants}
-   */
+     * @deprecated use {@link Mina.getNetworkConstants}
+     */
```
mitschabaude, line 433: @shimkiv you should set up `prettier` :P

shimkiv: Done.

src/lib/provable-types/merkle-list.ts
```diff
@@ -0,0 +1,495 @@
...
+ *
+ * // now use it
+ * let list = MyList.empty();
+ *
+ * list.push(new MyType(...));
+ *
+ * let element = list.pop();
+ * ```
+ *
+ * Internal detail: `push()` adds elements to the _start_ of the internal array and `pop()` removes them from the start.
```
Trivo25, line 67: what about naming the methods  `unshift` and `shift` for `push` and `pop` respectively so that the method the overall `MerkleList` structure follows the native JavaScript array more closely?

mitschabaude: Hm.. you have a point but what gives me pause is that the names `shift()` / `unshift()` are probably less well-known and recognizable for developers not intimately familiar with the JS stdlib

Trivo25 (approved):
lgtm

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/834a44002...HEAD)
 
+### Added
+
+- `MerkleList<T>` to enable provable operations on a dynamically-sized list https://github.com/o1-labs/o1js/pull/1398
```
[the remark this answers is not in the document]
mitschabaude: Good idea! Maybe a recursive zkprogram that sums up the elements of a list

PR #1400, opened 2024-02-01
Fix import cycles in account_update.ts

mitschabaude:
On `main`, there was a knot of import cycles between the different files involved in transaction construction:
* `account_update.ts` depended on `zkapp.ts`, `mina.ts` and `precondition.ts`, all of which in turn depended on `account_update.ts`
* `zkapp.ts` and `mina.ts` depended on each other
* `precondition.ts` depended on `mina.ts` which depended on `precondition.ts` indirectly via `account_update.ts`

This PR resolves all of the above import cycles and achieves a clean import graph:
```
zkapp.ts <-- mina.ts <-- account_update.ts <-- precondition.ts
```
(where the arrow means "direction of import")

This is done by
* Moving the "current Mina instance" from `mina.ts` to a lower-level module
* Moving the global transaction context to a lower-level module
* Introducing a base class for `SmartContract` which lives in a lower level module

### DIFF
--- a/src/lib/account_update.ts
+++ b/src/lib/account_update.ts
@@ -1783,7 +1783,7 @@ function addMissingSignatures(
     let signature = signFieldElement(
       fullCommitment,
       privateKey.toBigInt(),
-      Mina.getNetworkId()
+      activeInstance.getNetworkId()
     );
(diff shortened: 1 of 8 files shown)


MartinMinkov (approved):
:tada:

Issue #1404, opened 2024-02-02
[Rosetta Compatible Signature RFC] Expose a signatureToHex in rosetta.js


robrobbins:
This is needed in order to submit a signed <payment, etc..> to the `construction/submit` rosetta endpoint.

My project expects to be able to utilize mina-signer's `client.signTransaction` method to produce a result compatible with the above rosetta endpoint.

I'm going to open a PR now with what I believe to be a solution, so i'll assign this to myself, if i can - if not i'll simply reference this issue in the PR

robrobbins:
closed as I will open a new PR once #1414 is merged to `main`.

PR #1406, opened 2024-02-02
expose signature to hex via rosetta.ts

robrobbins:
This is a draft of the possible solution for the issue/RFC #1404 

in draft as tests are still being written

### DIFF
--- a/src/mina-signer/src/rosetta.ts
+++ b/src/mina-signer/src/rosetta.ts
@@ -4,7 +4,7 @@ import { Field } from '../../provable/field-bigint.js';
 import { Memo } from './memo.js';
 import { Signature } from './signature.js';
 
-export { publicKeyToHex, rosettaTransactionToSignedCommand };
+export { publicKeyToHex, signatureToHex, rosettaTransactionToSignedCommand };
 
 function publicKeyToHex(publicKey: PublicKey) {
   return fieldToHex(Field, publicKey.x, !!publicKey.isOdd);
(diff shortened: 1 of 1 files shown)


src/mina-signer/src/rosetta.ts
```diff
@@ -20,6 +20,14 @@ function signatureFromHex(signatureHex: string): Signature {
   };
 }
```
robrobbins, line 22: my particular use case is that i will have a payload from `client.signTransaction`. this it's expected that a user would call `Signature.fromJSON(signature)` first and pass that to the method below

src/mina-signer/src/rosetta.ts
```diff
@@ -20,6 +20,14 @@ function signatureFromHex(signatureHex: string): Signature {
...
 
+function signatureToHex(signature: Signature): string {
+  // TODO possibly check isOdd like public key to hex does?
```
mitschabaude, line 24: not necessary here - "isOdd" is part of a compressed public key, but not part of a signature

mitschabaude (commented):
I noticed that `fieldToHex` seems to have two major bugs (my bad):
- It uses AND instead of OR when putting in the padding bit
- It doesn't pad a zero when turning a byte to hex

Can you fix this while you're here?

```diff
function fieldToHex<T extends Field | Scalar>(
  binable: Binable<T>,
  x: T,
  paddingBit: boolean = false
) {
  let bytes = binable.toBytes(x);
  // set highest bit (which is empty)
-  bytes[bytes.length - 1] &= Number(paddingBit) << 7;
+  bytes[bytes.length - 1] |= Number(paddingBit) << 7;
  // map each byte to a hex string of length 2
  return bytes
-    .map((byte) => byte.toString(16).split('').reverse().join(''))
+    .map((byte) => byte.toString(16).padStart(2, '0').split('').reverse().join(''))
    .join('');
}
```

robrobbins:
going to close this and re-open with a fresh branch after #1414 is merged

PR #1412, opened 2024-02-05
Fix: Keep track of changing permissions during transaction validation

mitschabaude:
This fixes a bug where a transaction that updated permissions on an existing account, and then tried to perform an action affected by the updated permissions, was incorrectly rejected by the TS pre-validation logic, because that logic didn't keep track of the changed permissions.

As a fix, we introduced a lightweight internal "ledger" on which individual account updates are applied, updating permissions.
This can be extended in the future to keep track of more account state on the TS side.

Note: There are no tests currently fixed by this. However, a unit test that would fail without this change is coming soon as part of https://github.com/o1-labs/o1js/pull/1384 (`token-contract.unit-test.ts`)

### DIFF
--- a/src/lib/mina.ts
+++ b/src/lib/mina.ts
@@ -44,6 +44,7 @@ import {
   type NetworkConstants,
 } from './mina/mina-instance.js';
 import { SimpleLedger } from './mina/transaction-logic/ledger.js';
+import { assert } from './gadgets/common.js';
 
 export {
   createTransaction,
@@ -425,6 +426,7 @@ function LocalBlockchain({
           }
         }
 
+        // TODO: verify account update even if the account doesn't exist yet, using a default initial account
         if (account !== undefined) {
           await verifyAccountUpdate(
             account,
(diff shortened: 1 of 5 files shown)


src/lib/mina.ts
```diff
@@ -1239,15 +1254,20 @@ async function verifyAccountUpdate(
...
+
       isValidProof = await verify(proof, verificationKey);
       if (!isValidProof) {
         throw Error(
           `Invalid proof for account update\n${JSON.stringify(update)}`
         );
       }
     } catch (error) {
-      errorTrace += '\n\n' + (error as Error).message;
+      errorTrace += '\n\n' + (error as Error).stack;
```
mitschabaude, line 1270: much better to also include the stack trace here - that way you can see where the code failed

src/lib/mina.ts
```diff
@@ -1239,15 +1254,20 @@ async function verifyAccountUpdate(
         publicOutput: [],
       };
 
-      let verificationKey = account.zkapp?.verificationKey?.data!;
+      let verificationKey = account.zkapp?.verificationKey?.data;
+      assert(
+        verificationKey !== undefined,
+        'Account does not have a verification key'
+      );
```
mitschabaude, line 1261: ran into this being `undefined` in a test and failed down the line - better to have an explicit error about it right away

Trivo25 (approved):
nice, looking forward to extending this!

PR #1414, opened 2024-02-06
Cleanups for mina-signer

robrobbins:
1:

*`.test.ts` files in ..`/mina-signer/tests/` would not run with their paths to the `dist` directory as-is. I noticed this when running `npm run test`. Diving in we can see the bash script just sets the NODE_OPTIONS var and executes `npx just past/to/foo.test.ts` while the `/lib/` tests all run and pass none of the tests in `/mina-signer` will compile.

you can quickly test this with ` NODE_OPTIONS=--experimental-vm-modules npx jest ./src/mina-signer/tests/client.test.ts`. It will fail in main. this fixes the relative path to correctly find the directory

there are many ways you could find that path, but this solution stays the closest to what was already in place.

this also will clear the way for my work on #1406 as i will need to write and run a few tests there

2:

this fixes some issues with `fieldToHex` that were pointed out by @mitschabaude in #1406. I placed them here as it fit with the _fixup_ theme.

### DIFF
--- a/src/mina-signer/tests/payment.test.ts
+++ b/src/mina-signer/tests/payment.test.ts
@@ -1,5 +1,5 @@
-import Client from '../dist/node/mina-signer/MinaSigner.js';
-import type { Keypair } from '../dist/node/mina-signer/src/TSTypes.js';
+import Client from '../../../dist/node/mina-signer/MinaSigner.js';
+import type { Keypair } from '../../../dist/node/mina-signer/src/TSTypes.js';
 
 describe('Payment', () => {
   describe('Mainnet network', () => {
--- a/src/mina-signer/src/rosetta.ts
+++ b/src/mina-signer/src/rosetta.ts
@@ -27,12 +27,13 @@ function fieldToHex<T extends Field | Scalar>(
 ) {
   let bytes = binable.toBytes(x);
   // set highest bit (which is empty)
-  bytes[bytes.length - 1] &= Number(paddingBit) << 7;
-  // map each byte to a hex string of length 2
+  bytes[bytes.length - 1] |= Number(paddingBit) << 7;
+  // map each byte to a 0-padded hex string of length 2
   return bytes
-    .map((byte) => byte.toString(16).split('').reverse().join(''))
+    .map((byte) => byte.toString(16).padStart(2, '0').split('').reverse().join(''))
     .join('');
 }
+
 function fieldFromHex<T extends Field | Scalar>(
   binable: Binable<T>,
   hex: string
--- a/src/mina-signer/tests/client.test.ts
+++ b/src/mina-signer/tests/client.test.ts
@@ -1,4 +1,4 @@
-import Client from '../dist/node/mina-signer/MinaSigner.js';
+import Client from '../../../dist/node/mina-signer/MinaSigner.js';
 
 describe('Client Class Initialization', () => {
   let client;
--- a/src/mina-signer/tests/keypair.test.ts
+++ b/src/mina-signer/tests/keypair.test.ts
@@ -1,4 +1,4 @@
-import Client from '../dist/node/mina-signer/MinaSigner.js';
+import Client from '../../../dist/node/mina-signer/MinaSigner.js';
 
 describe('Keypair', () => {
   let client: Client;
--- a/src/mina-signer/tests/message.test.ts
+++ b/src/mina-signer/tests/message.test.ts
@@ -1,5 +1,5 @@
-import Client from '../dist/node/mina-signer/MinaSigner.js';
-import type { PrivateKey } from '../dist/node/mina-signer/src/TSTypes.js';
+import Client from '../../../dist/node/mina-signer/MinaSigner.js';
+import type { PrivateKey } from '../../../dist/node/mina-signer/src/TSTypes.js';
 
 describe('Message', () => {
   describe('Mainnet network', () => {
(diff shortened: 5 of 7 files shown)


mitschabaude:
@robrobbins, re 1, we run these tests in CI so they work. You have to do `npm run build` in the mina-signer directory first. I think it must be that way, because mina-signer is _published_ from the `mina-signer/dist` directory. So we _have_ to test if it actually works from that directory, not when embedded within the larger o1js code base.

In conclusion: I don't want your changes for 1)

2) lgtm :)

mitschabaude (changes requested):
see above

robrobbins:
> I don't want your changes for 1)

Not the phrase I'd use for reviewing the work of a professional developer who is sacrificing their valuable time in an attempt to help your project. 

However, while we are talking about *want* here, let's look at what I wanted: a way to run tests while developing. I did run `npm build` that much is in the dev readme. What is not there is what you pointed out in the ci file. You have to `cd` int `mina-signer` and run `rpm build` again.

So now it goes like this:
1) revert path changes. 
2) add a sub section [here](https://github.com/o1-labs/o1js/blob/main/README-dev.md#running-tests) specific to mina-signer WRT the need to cd..build
3) leave rosetta.ts changes as is

mitschabaude:
> Not the phrase I'd use for reviewing the work of a professional developer who is sacrificing their valuable time in an attempt to help your project.

Sorry for the rude tone there - it definitely wasn't my intention to come off like this, your effort is highly appreciated!

> add a sub section [here](https://github.com/o1-labs/o1js/blob/main/README-dev.md#running-tests) specific to mina-signer WRT the need to cd..build

Good point, this should be documented

> So now it goes like this:

Sounds like a plan!

> add a sub section [here](https://github.com/o1-labs/o1js/blob/main/README-dev.md#running-tests) specific to mina-signer WRT the need to cd..build****

README-dev.md
```diff
@@ -141,8 +141,23 @@ npm run test
 npm run test:unit
 ```
 
+#### Running Mina-Signer Tests
```
robrobbins, line 144: So there are a few ways you could go about writing this into the readme-dev. Essentially it's the fact that

`npm run build` 

is not sufficient, and makes rise 140 above incorrect. `npm run test` will fail once it gets through the `/lib` tests and arrives at the `/mina-signer` tests. as @mitschabaude pointed out, the `ci` moves into `/mina-signer` and runs build again before running `npm run test`. 

perhaps this change should be behind a header for running mina-signer tests and should simply be at the topmost reference to building. it's not as svelte however

 ```
npm run build
cd src/mina-signer
npm run build
cd ../..
```

so i stuck it here. 🤷‍♂️

File: README-dev.md [hunk 2]
```diff
@@ -141,8 +141,23 @@ npm run test
...
+
+```sh
+cd src/mina-signer
+npm run build
+cd ../..
+```
+
 This runs all the unit tests and provides you with a summary of the test results.
 
+Note that you can run individual jest tests via the command:
```
robrobbins, line 155: optional comment here but I can't imagine outside devs don't ask this. could save some time. would have saved me some digging. up to you however.

src/mina-signer/tests/payment.test.ts
```diff
@@ -1,5 +1,5 @@
 import Client from '../dist/node/mina-signer/MinaSigner.js';
-import type { Keypair } from '../dist/node/mina-signer/src/TSTypes.js';
+import type { Keypair } from '../../../dist/node/mina-signer/src/TSTypes.js';
```
robrobbins, line 2: dammit if forgot these. lemme fix, squash in a fixup pr. 5 mins

robrobbins: fixed

robrobbins:
OK
* paths are reverted
* notes added to readme-dev. added comments on thoughts there. check what you think. I don't have any real opinions, just needs to be _somewhere_

README-dev.md
mitschabaude: That works well!

README-dev.md
```diff
@@ -141,8 +141,23 @@ npm run test
 npm run test:unit
 ```
 
+#### Running Mina-Signer Tests
+`npm run build` is not recursive, thus in order for the Mina Signer tests to run you must execute the following from the root directory:
```
mitschabaude, line 145:
```suggestion
In order for the mina-signer tests to run you must also build from inside its subdirectory:
```

Removing the header since the other kinds of tests below don't have one either

README-dev.md
mitschabaude: Useful thanks!

README-dev.md
```diff
@@ -141,8 +141,23 @@ npm run test
...
+npm run build
+cd ../..
+```
+
 This runs all the unit tests and provides you with a summary of the test results.
 
+Note that you can run individual jest tests via the command:
+
+```sh
+NODE_OPTIONS=--experimental-vm-modules npx jest <path/to/test.ts>
```
mitschabaude, line 158:
```suggestion
./jest <path/to/test.ts>
```
we actually have a little script to make this not so hard to remember

mitschabaude (approved):
Approved with some tweaks! Thanks a lot @robrobbins

robrobbins:
> Approved with some tweaks! Thanks a lot @robrobbins

:shipit: 

I have the rosetta signature-to-hex ready to go after this is in and i rebase it

PR #1420, opened 2024-02-07
implement and test signatureToHex for rosetta.ts

robrobbins:
PR for the RFC #1404


src/mina-signer/tests/rosetta.test.ts
```diff
@@ -23,6 +25,25 @@ describe('Rosetta', () => {
...
+    // TODO we could go further and use validate method on the return?
+    // or leave as is with the assumption that the above hex is known to be valid
+  });
+
+  it('generates a valid hex string from a signature', () => {
+    let signature = signatureFromHex(rosettaTnxMockSignature);
+    let hex = signatureToHex(signature);
+
+    // fails
+    expect(hex).toEqual(rosettaTnxMockSignature);
```
robrobbins, line 44: as stated this fails with a single char being different than what is expected (see test output). @mitschabaude interesting... that or my expectation is incorrect 🤷‍♂️

mitschabaude: Interesting, that test should definitely pass

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -23,6 +25,25 @@ describe('Rosetta', () => {
...
 
+  it('generates a valid signature from a hex string', () => {
+    let { r, s } = signatureFromHex(rosettaTnxMockSignature);
+    expect(r).not.toBeNaN();
+    expect(r).not.toBeFalsy();
+    expect(s).not.toBeNaN();
+    expect(s).not.toBeFalsy();
+
+    // TODO we could go further and use validate method on the return?
+    // or leave as is with the assumption that the above hex is known to be valid
```
mitschabaude, line 36: yes - ideally, we should verify the signature against the payment, using `signedRosettaTransactionToSignedCommand` and `verifyPaymentSignature` (or however it's called). that's a big hole we left in the tests when we created this

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -1,11 +1,13 @@
...
+import { signatureFromHex, signatureToHex } from '../dist/node/mina-signer/src/rosetta.js';
 
 describe('Rosetta', () => {
   let client: Client;
 
+  const rosettaTnxMockSignature = '389ac7d4077f3d485c1494782870979faa222cd906b25b2687333a92f41e40b925adb08705eddf2a7098e5ac9938498e8a0ce7c70b25ea392f4846b854086d43';
   const signedRosettaTnxMock = `
   {
-    "signature": "389ac7d4077f3d485c1494782870979faa222cd906b25b2687333a92f41e40b925adb08705eddf2a7098e5ac9938498e8a0ce7c70b25ea392f4846b854086d43",
+    "signature": "${rosettaTnxMockSignature}",
```
mitschabaude, line 10: bug
```suggestion
    "signature": `${rosettaTnxMockSignature}`,
```

src/mina-signer/tests/rosetta.test.ts
mitschabaude: even if the hex above is valid -- only by verifying the signature can we really be sure that we parse it correctly

src/mina-signer/tests/rosetta.test.ts
robrobbins: how so? the entire blob is in backticks. here is the output of `console.log(rosettaTnxMockSignature)` left as is:

```
      {
        "signature": "389ac7d4077f3d485c1494782870979faa222cd906b25b2687333a92f41e40b925adb08705eddf2a7098e5ac9938498e8a0ce7c70b25ea392f4846b854086d43",
        "payment": {
          "to": "B62qnzbXmRNo9q32n4SNu2mpB8e7FYYLH8NmaX6oFCBYjjQ8SbD7uzV",
          "from": "B62qnzbXmRNo9q32n4SNu2mpB8e7FYYLH8NmaX6oFCBYjjQ8SbD7uzV",
          "fee": "10000000",
          "token": "1",
          "nonce": "0",
          "memo": null,
          "amount": "1000000000",
          "valid_until": "4294967295"
        },
        "stake_delegation": null
      }
```

src/mina-signer/tests/rosetta.test.ts
robrobbins: i'm going to start down this path now. getting verification into the test. perhaps along the way we can reveal why that test fails?

question: do you know if this hex, the "signature" in the mock - was it generated with the ocaml signer? 

re: we know that the ocaml signer produces output (hex string) that is compatible with the rosetta `construction/submit` enpoint. we know this from the .py scripts written by Ivan that are in the repo as examples 

thus: if we have a mock signature in a known valid format we can proceed in our tests with confidence that when i get to actually trying to POST to /construction/submit we will (likely) succeed. that's my hope anyway

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -19,10 +21,46 @@ describe('Rosetta', () => {
     "stake_delegation": null
   }`;
 
+  // NOTE: copied from the test-vectors legacy signatures file
```
robrobbins, line 24: found the sign legacy unit tests and thought these would make a (better?) test candidate

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -19,10 +21,46 @@ describe('Rosetta', () => {
...
+    // fails
+    // expect(hex).toEqual(rosettaTnxMockSignature);
+
+    let hex = signatureJsonToHex(legacySignatureJson);
+    let sig = signatureFromHex(hex);
+
+    expect(sig.r).toEqual(BigInt(legacySignatureJson.field));
+    expect(sig.s).toEqual(BigInt(legacySignatureJson.scalar));
+  });
+
```
robrobbins, line 63: interestingly this passes as i had expected the in-place rosetta mock to pass. this brings up the possible case that the code is, indeed, correct and that the assumption about the rosetta mock was incorrect (possibly)

there is an alternative to lines 60 and 61 that do not require knowledge of what `Signature.fromJSON` does - that is to also `import { Signature } from ...` and convert via `Signature.fromJSON(legacySignatureJson)` then directly compare to the `sig` from line 58. i don't think it's needed as it is just doing `BigInt(...)` like the test above but - worth pointing out as it might read better 🤷

robrobbins: @mitschabaude see the pushed change with test-vector and that it does pass

src/mina-signer/tests/rosetta.test.ts
robrobbins: note: i did find that the test-vectors used by the unit tests in sign-legacy function correctly..

src/mina-signer/tests/rosetta.test.ts
mitschabaude: Before you tested fromHex -> toHex and now you're going toHex -> fromHex. I would throw in another -> toHex plus consistency check before concluding that the test works now

src/mina-signer/tests/rosetta.test.ts
mitschabaude: Right you don't need the backticks. But read your code again - it has "${blub}" which doesn't work

src/mina-signer/tests/rosetta.test.ts
mitschabaude: I assume I copied the mock from OCaml test vectors but can't remember

src/mina-signer/tests/rosetta.test.ts
robrobbins: you need the backticks for interpolation.

```
// correct this alone will not work
"${blub}"

// however inside the backticks it does
`"${blub}"`

// the double quotes are needed for that to be represented as a json string otherwise it'll appear to be an int - removing them would cause the later calls in the test to JSON.parse to fail

File: src/mina-signer/tests/rosetta.test.ts (hunk 5 above)
Comment (robrobbins) [reply src/mina-signer/tests/rosetta.test.ts:63]: yeah, i need a hex signature already existing that is known to be valid. or confirmation that the in place one is - thus proving there's some cryptog issue at play. 

i figured i'd search for another to compare, or get more info on the existing one

File: src/mina-signer/tests/rosetta.test.ts (hunk 3 above)
Comment (mitschabaude) [reply src/mina-signer/tests/rosetta.test.ts:10]: my bad - I didn't see the entire blob is in backticks 😅

File: src/mina-signer/tests/rosetta.test.ts (hunk 5 above)
Comment (mitschabaude) [reply src/mina-signer/tests/rosetta.test.ts:63]: Just assume that the one in here is valid

Comment (mitschabaude) [reply src/mina-signer/tests/rosetta.test.ts:63]: And test toHex on your working test case

**Comment by robrobbins** [comment]
closing as a new PR is forthcoming

File: src/mina-signer/src/rosetta.ts [hunk 6]
```diff
@@ -30,7 +66,7 @@ function fieldToHex<T extends Field | Scalar>(
   bytes[bytes.length - 1] |= Number(paddingBit) << 7;
   // map each byte to a 0-padded hex string of length 2
   return bytes
-    .map((byte) => byte.toString(16).padStart(2, '0').split('').reverse().join(''))
+    .map((byte) => byte.toString(16).padStart(2, '0'))
```
mitschabaude, line 69: sweet!

src/mina-signer/src/rosetta.ts
```diff
@@ -49,12 +85,93 @@ function fieldFromHex<T extends Field | Scalar>(
...
+}
+
+// create a payload for /construction/combine
+function rosettaCombinePayload(unsignedPayload: UnsignedPayload, privateKey: Json.PrivateKey, network: NetworkId) {
+  let signature = signTransaction(unsignedPayload.unsigned_transaction, privateKey, network);
+  let signatures = [rosettaCombineSignature(signature, unsignedPayload.payloads[0])];
+  return {
+    network_identifier: { blockchain: "mina", network },
+    unsigned_transaction: unsignedPayload.unsigned_transaction, signatures
+  };
```
mitschabaude, line 170: could you run `prettier` on this file?
```suggestion
  return {
    network_identifier: { blockchain: "mina", network },
    unsigned_transaction: unsignedPayload.unsigned_transaction,
    signatures
  };
```

src/mina-signer/src/rosetta.ts
```diff
@@ -124,24 +241,53 @@ function rosettaTransactionToSignedCommand({
...
+}
+
+type RosettaSignature = {
+  hex_bytes: string;
+  public_key: {
+    hex_bytes: string;
+    curve_type: string;
+  };
+  signature_type: string;
+  signing_payload: any;
```
mitschabaude, line 272: can we change all the `any` to `unknown` for type safety?
```suggestion
  signing_payload: unknown;
```

src/mina-signer/src/rosetta.ts
```diff
@@ -124,24 +241,53 @@ function rosettaTransactionToSignedCommand({
   };
 }
 
-type RosettaTransactionJson = {
+type UnsignedPayload = {
+  unsigned_transaction: UnsignedTransaction;
+  payloads: any[];
```
mitschabaude, line 246:
```suggestion
  payloads: unknown[];
```

src/mina-signer/src/rosetta.ts
```diff
@@ -49,12 +85,93 @@ function fieldFromHex<T extends Field | Scalar>(
...
+  if (signedTransaction.data.payment !== null) {
+    return verifyPayment(unsignedTransactionPaymentToPaymentJson(signedTransaction.data.payment), signatureJsonFromHex(signedTransaction.signature), signedTransaction.publicKey, network);
+  } else if (signedTransaction.data.stakeDelegation !== null) {
+    return verifyStakeDelegation(unsignedTransactionStakeDelegationToDelegationJson(signedTransaction.data.stakeDelegation), signatureJsonFromHex(signedTransaction.signature), signedTransaction.publicKey, network);
+  }
+  throw Error('verifyTransaction: Unsupported transaction');
+}
+
+// create a signature for /construction/combine payload
+function rosettaCombineSignature(signature: SignedRosetta<UnsignedTransaction>, signingPayload: any): RosettaSignature {
```
mitschabaude, line 150:
```suggestion
function rosettaCombineSignature(signature: SignedRosetta<UnsignedTransaction>, signingPayload: unknown): RosettaSignature {
```

src/mina-signer/src/rosetta.ts
```diff
@@ -49,12 +85,93 @@ function fieldFromHex<T extends Field | Scalar>(
...
+  }
+  let publicKey = PublicKey.toBase58(PrivateKey.toPublicKey(PrivateKey.fromBase58(privateKey)));
+  return {
+    data: transaction,
+    signature: signatureJsonToHex(signature),
+    publicKey,
+  };
+}
+
+function unsignedTransactionPaymentToPaymentJson(payment: Payment): PaymentJson {
```
mitschabaude, line 109: nit: not super important but could you simplify these long names? e.g. `paymentFromRosetta()`, `delegationFromRosetta()`

src/mina-signer/src/rosetta.ts
```diff
@@ -49,12 +85,93 @@ function fieldFromHex<T extends Field | Scalar>(
...
+    }
+  }
+}
+
+function verifyTransaction(signedTransaction: SignedRosetta<UnsignedTransaction>, network: NetworkId): boolean {
+  if (signedTransaction.data.payment !== null) {
+    return verifyPayment(unsignedTransactionPaymentToPaymentJson(signedTransaction.data.payment), signatureJsonFromHex(signedTransaction.signature), signedTransaction.publicKey, network);
+  } else if (signedTransaction.data.stakeDelegation !== null) {
+    return verifyStakeDelegation(unsignedTransactionStakeDelegationToDelegationJson(signedTransaction.data.stakeDelegation), signatureJsonFromHex(signedTransaction.signature), signedTransaction.publicKey, network);
+  }
```
mitschabaude, line 145: nit: since you do an early return you can flatten the `else if` to another `if` (would be consistent with the final error not being in an `else` block

mitschabaude: also, this needs `prettier` :D

src/mina-signer/src/rosetta.ts
```diff
@@ -49,12 +85,93 @@ function fieldFromHex<T extends Field | Scalar>(
...
+    },
+    signature_type: "schnorr_poseidon",
+    signing_payload: signingPayload
+  }
+}
+
+// create a payload for /construction/combine
+function rosettaCombinePayload(unsignedPayload: UnsignedPayload, privateKey: Json.PrivateKey, network: NetworkId) {
+  let signature = signTransaction(unsignedPayload.unsigned_transaction, privateKey, network);
+  let signatures = [rosettaCombineSignature(signature, unsignedPayload.payloads[0])];
```
mitschabaude, line 166: I don't know how this is supposed to work. but it confuses and worries me that we include a "payload" with the signature which is not related (at least within this function) to what is signed, in any way. You sign `unsigned_transaction` but then include `payloads[0]` in the `RosettaSignature`.

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -1,17 +1,45 @@
...
+  const rosettaUnsignedTxn: UnsignedTransaction = JSON.parse(rosettaUnsignedTxnString);
```
mitschabaude, line 10: nit: could this test hard-code the `UnsignedTransaction` directly instead of the string? would be more readable and type-safe

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -23,6 +51,41 @@ describe('Rosetta', () => {
...
+  });
+
+  it('match signature', () => {
+    const { signature } = signTransaction(rosettaUnsignedTxn, privateKey, 'mainnet');
+    expect(signature).toEqual(mainnetSignatureHex);
+  });
+
+  it('verify transaction', () => {
+    expect(client.verifyRosettaTransaction({ data: rosettaUnsignedTxn, signature: mainnetSignatureHex, publicKey: publicKey })).toBeTruthy();
+  });
```
mitschabaude, line 83: maybe add a negative test here? for example, that a client with `network:  'testnet'` returns `false` from `verifyRosettaTransaction`?

src/mina-signer/tests/rosetta.test.ts
```diff
@@ -23,6 +51,41 @@ describe('Rosetta', () => {
...
+
+  // it('test user command', () => {
+  //   expect(unsignedTransactionToUserCommand(rosettaUnsignedTxn)).toBeDefined();
+  // });
```
mitschabaude, line 87: remove commented stuff

mitschabaude (commented):
awesome that you made this work!! and also fixed our hex conversion 😅 

a few comments

src/mina-signer/src/rosetta.ts
joaosreis: This simply streamlines the offline part of signing Rosetta transactions, this is, it's taking the output from Rosetta's `/construction/payloads` endpoint (`UnsignedPayload`), producing a signature for it, and bundling everything together to produce the payload expected by the `/contruction/combine` endpoint.

mitschabaude: ok I'll take that :)

Issue #824, opened 2023-03-31
RFC: Transaction Flow (parity of behavior between localBlockchain and Testnet)


rpanic:
### Motivation & Current State

The sending of Transactions and subsequent status retrieval is inconsistent between local and remote implementations and additionally re-uses the TransactionId and re-uses properties for different functionality.
As an example, in remote networks, `TransactionId.isSuccess` is used for both the indication that the sending of the transaction has succeeded and does not work correctly after calling `.wait()`. See [#819](https://github.com/o1-labs/snarkyjs/issues/819)
In comparison, LocalBlockchain throws at `.send()` if the transaction cannot be applied to the ledger, and does not implement .wait().

Altough the interface for these implementation is the same, the behaviour is completely different and leads to worse developer experience and extra effort to check for those things in scripts. 

### Proposed changes

I propose an extension to the current API which aligns both implementations and seperates concerns into different classes.

```
Transaction.send() : Promise<TransactionId>
```
Where TransactionId uniformly implements properties as:
  - `isSuccess`: The Transaction is well-formed and passes initial checks by the node (I think the mina node primarily checks some nonce preconditions and the fee payer)
  - `errors`: Error[] that occurred
  - `hash()`: Returns the hash of the txid
  - `wait()`: Waits for the tx to be mined (details below)
  

```
TransactionId.wait() : Promise<IncludedTransaction>
```
The first implementation spec only needs an default mina node
 - `isIncluded`: The tx got included in a block (is only false if errors occurred)
 - `errors`: Error[] that occurred

The second possibility would extend the above and involve the new archive-node api:
 - `getInfo() : Promise<IncludedTransactionInfo>`: Retrieves information about the mined transaction from an archive node.
 IncludedTransactionInfo would include inclusion-time errors, block info. This method would throw if no archive node url has been specified for remote networks. 
This should be an extra method in order to allow operation without an archive-node api, as is the current interface for remote networks. 
 
Reliably usage would depend on [Archive-Node-API#69](https://github.com/o1-labs/Archive-Node-API/issues/69) being implemented. 

### Things I am not sure about

hash() is currently taken from the graphql response. I don't know why but it seems that this should be possible to do in snarkyjs

mitschabaude:
Thanks for the proposal @rpanic! I like it a lot:

* Making the two interfaces consistent will solve a huge pain point for users
[...]

PR #1422, opened 2024-02-08
Fix parity between local blockchain and network APIs

MartinMinkov:
Implements :page_with_curl:  https://github.com/o1-labs/o1js/issues/824#issuecomment-1932831418

mina: https://github.com/MinaProtocol/mina/pull/15137
o1js-bindings: https://github.com/o1-labs/o1js-bindings/pull/245


mitschabaude:
> Can you please also check #1426 and #1427

These are issues to be fixed for sure, but neither of them is in scope for this PR imo

src/tests/transaction-flow.ts
```diff
@@ -0,0 +1,291 @@
+import {
```
MartinMinkov, line 1: This file aims to test the methods exposed in `Mina.ts` for each implementation of the instance (local blockchain and remote). We switch back and forth and call methods and ensure they behave the same way.

src/lib/mina/graphql.ts
```diff
@@ -0,0 +1,492 @@
+import { ActionStatesStringified, removeJsonQuotes } from '../fetch.js';
```
MartinMinkov, line 1: This file was meant to hold all types of queries related to GraphQL requests. I thought `fetch.ts` was starting to get bloated and hard to read, so moving some of the static definitions to a separate module would help.

src/tests/transaction-flow.ts
```diff
@@ -0,0 +1,291 @@
...
+        AccountUpdate.fundNewAccount(zkAppAddress);
+        zkApp.update(Field(1), PrivateKey.random().toPublicKey());
+      }
+    );
+    transaction.sign([senderKey, zkAppKey]);
+    await sendAndVerifyTransaction(transaction);
+  } catch (error) {
+    errorWasThrown = true;
+  }
+  assert(errorWasThrown);
```
mitschabaude, line 230: maybe use `assert.throws(() => {...})` to make this read cleaner?

MartinMinkov: `assert.throws()` doesn't seem to play well with async functions :sob: I think `assert.rejects` fulfills the same role though!

src/lib/mina.ts
```diff
@@ -129,9 +130,204 @@ type Transaction = {
...
+   */
+  sendOrThrowIfError(): Promise<PendingTransaction>;
+};
+
+/**
+ * Represents a transaction that has been submitted to the blockchain but has not yet reached a final state.
+ * The `PendingTransaction` type extends certain functionalities from the base `Transaction` type,
+ * adding methods to monitor the transaction's progress towards being finalized (either included in a block or rejected).
+ */
+type PendingTransaction = Pick<
```
MartinMinkov, line 146: This file is starting to get big :O Should we separate transaction-related logic into its own module and let `mina.ts` just be responsible for the interaction/implementations between `LocalBlockchain` and `Network`?

src/lib/fetch.ts
```diff
@@ -211,24 +229,14 @@ async function fetchAccountInternal(
 }
 
 type FetchConfig = { timeout?: number };
-type FetchResponse = { data: any; errors?: any };
+type FetchResponse<TDataResponse = any> = { data: TDataResponse; errors?: any };
```
MartinMinkov, line 232: We add a generic type here for `FetchResponse` to get better type safety by passing in types on requests we make.

src/lib/fetch.ts
```diff
@@ -450,83 +467,34 @@ async function fetchLastBlock(graphqlEndpoint = networkConfig.minaEndpoint) {
...
-    transactions {
-      zkappCommands {
-        hash
-        failureReason {
-          failures
-          index
-        }
-      }
-    }
+async function fetchCurrentSlot(graphqlEndpoint = networkConfig.minaEndpoint) {
```
MartinMinkov, line 470: This was implemented but isn't used (I was planning to implement `getCurrentSlot`, but it will wait for async circuits), and will most likely revert.

src/lib/mina.ts
```diff
@@ -292,14 +511,21 @@ function newTransaction(transaction: ZkappCommand, proofsEnabled?: boolean) {
...
-        return await sendTransaction(self);
-      } catch (error) {
-        throw prettifyStacktrace(error);
+      return await sendTransaction(self);
+    },
+    async sendOrThrowIfError() {
+      const pendingTransaction = await sendTransaction(self);
+      if (pendingTransaction.errors.length > 0) {
+        throw Error(
+          `Transaction failed with errors:\n- ${pendingTransaction.errors.join(
```
MartinMinkov, line 523: Is this the best way to print errors to the user? Ideas are welcome!

src/lib/mina.ts
mitschabaude: As a first step I'd actually move `LocalBlockchain` to a separate file, since it's kind of a testing utility and not part of the "main" API

maybe into `mina/local-blockchain.ts`

mitschabaude: and yeah, transaction construction e.g. `createTransaction` and the `Transaction` interface could be another separate file

src/lib/mina/mina-instance.ts
```diff
@@ -20,6 +35,11 @@ export {
   activeInstance,
   setActiveInstance,
   ZkappStateLength,
+  reportGetAccountError,
```
MartinMinkov, line 38: Moved these methods here as they are shared between `local-blockchain` and `remote`

src/lib/mina.ts
MartinMinkov: done!

src/examples/zkapps/hello-world/run.ts
```diff
@@ -27,7 +27,7 @@ txn = await Mina.transaction(feePayer1.publicKey, () => {
   AccountUpdate.fundNewAccount(feePayer1.publicKey);
   zkAppInstance.deploy();
 });
-await txn.sign([feePayer1.privateKey, zkAppPrivateKey]).send();
+await txn.sign([feePayer1.privateKey, zkAppPrivateKey]).sendOrThrowIfError();
```
ymekuria, line 30: @MartinMinkov are there any other examples in the the examples folder that should be updated with the changes to the simulated blockchain similar to this?

src/tests/transaction-flow.ts
mitschabaude: oh sorry yeah!

src/lib/mina.ts
```diff
@@ -786,86 +243,123 @@ function Network(
...
-        async wait(options?: { maxAttempts?: number; interval?: number }) {
-          if (!isSuccess) {
-            console.warn(
-              'Transaction.wait(): returning immediately because the transaction was not successful.'
+        transaction: txn.transaction,
+        toJSON: txn.toJSON,
+        toPretty: txn.toPretty,
+        hash() {
+          return hash;
+        },
```
mitschabaude, line 273: since hash is just a value now, maybe we could make this a property instead of function?

src/lib/mina.ts
mitschabaude, line 1: it's so much cleaner with this file broken up into digestible pieces!

src/lib/fetch.ts
mitschabaude: seems useful to me! with async circuits we can just use this new function directly from a method

src/lib/mina/errors.ts
```diff
@@ -60,23 +60,26 @@ function invalidTransactionError(
...
-      transaction,
-      accountUpdateIndex: NaN,
-      isFeePayer: true,
-      ...additionalContext,
-    });
-    if (message) errorMessages.push(message);
+  // Check if the number of errors match the number of account updates. If there are more, then the fee payer has an error.
+  // We do this check because the fee payer error is not included in network transaction errors and is always present (even if empty) in the local transaction errors.
+  if (accountUpdateErrors.length === n) {
+    let errorsForFeePayer = errors[0];
```
mitschabaude, line 69: after staring at this long enough the logic looks correct, because you slice off the account update errors before and the slice starting from 1 can be either `n` or `n-1` entries long.

but would be more readable to do
```ts
if (errors.length > n) {
   let errorsForFeePayer = errors.unshift();
   // ...
```

and then you can reuse the `errors` below!

src/lib/mina/errors.ts
```diff
@@ -60,23 +60,26 @@ function invalidTransactionError(
...
+      if (message) errorMessages.push(message);
+    }
   }
 
-  // handle errors for each account update
-  let n = transaction.accountUpdates.length;
-  for (let i = 0; i < n; i++) {
-    let errorsForUpdate = errors[i + 1];
+  for (let i = 0; i < accountUpdateErrors.length; i++) {
+    let errorsForUpdate = accountUpdateErrors[i];
```
mitschabaude, line 82: I might be confused, but to me this seems to have a bug: you created
```ts
let accountUpdateErrors = errors.slice(1, n + 1);
```
so if there is no fee payer error, then you're ignoring the errors for the 0th account update?

src/lib/mina/local-blockchain.ts
```diff
@@ -0,0 +1,389 @@
...
+          pendingTransaction,
+          pendingTransaction.errors
+        );
+      };
+
+      const waitOrThrowIfError = async (_options?: {
+        maxAttempts?: number;
+        interval?: number;
+      }) => {
+        return createIncludedOrRejectedTransaction(
```
mitschabaude, line 286: for consistency, I think this should behave like the `Network` implementation of `waitOrThrowIfError()`, and throw an error if the pending transaction already failed

src/tests/transaction-flow.ts
mitschabaude: great that you added this!

mitschabaude (commented):
This is fantastic work, huge step towards v1 readiness!

I have 2-3 comments that I think need addressing

src/lib/mina.ts
MartinMinkov: done!

src/lib/mina/errors.ts
MartinMinkov: great idea! fixed!

src/lib/mina/local-blockchain.ts
MartinMinkov: fixed!

src/lib/mina/errors.ts
```diff
@@ -61,12 +61,11 @@ function invalidTransactionError(
...
   let rawErrors = JSON.stringify(errors);
   let n = transaction.accountUpdates.length;
-  let accountUpdateErrors = errors.slice(1, n + 1);
 
   // Check if the number of errors match the number of account updates. If there are more, then the fee payer has an error.
   // We do this check because the fee payer error is not included in network transaction errors and is always present (even if empty) in the local transaction errors.
-  if (accountUpdateErrors.length === n) {
-    let errorsForFeePayer = errors[0];
+  if (errors.length > n) {
+    let errorsForFeePayer = errors.shift() ?? [];
```
mitschabaude, line 68: damn I always mix up shift and unshift

mitschabaude (approved):
From my side, this only needs a changelog entry and then it's good to go!

src/lib/mina/errors.ts
MartinMinkov: not the best naming imo :P

CHANGELOG.md
```diff
@@ -17,6 +17,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/3b5f7c7...HEAD)
 
+### Breaking changes
+
+- Fixed parity between `Mina.LocalBlockchain` and `Mina.Network` to have the same behaviors https://github.com/o1-labs/o1js/pull/1422
+  - Changed the `TransactionId` type to `Transaction`. Additionally added `PendingTransaction` and `RejectedTransaction` types to better represent the state of a transaction.
+  - `transaction.send()` no longer throws an error if the transaction was not successful for `Mina.LocalBlockchain` and `Mina.Network`. Instead, it returns a `PendingTransaction` object that contains the error. Use `transaction.sendOrThrowIfError` to throw the error if the transaction was not successful.
+  - `transaction.wait()` no longer throws an error if the transaction was not successful for `Mina.LocalBlockchain` and `Mina.Network`. Instead, it returns either a `IncludedTransaction` or `RejectedTransaction`. Use `transaction.waitOrThrowIfError` to throw the error if the transaction was not successful.
+  - `transaction.hash()` is no longer a function, it is now a property that returns the hash of the transaction.
```
barriebyron, line 26: do we need updates to the code in https://github.com/o1-labs/docs2/blob/main/examples/zkapps/interacting-with-zkApps-server-side/src/main.ts and https://docs.minaprotocol.com/zkapps/tutorials/interacting-with-zkapps-server-side ?

Issue #1347, opened 2023-12-23
Warn if callbacks are created outside of transactions


BerzanOrg:
Currently, when a callback is created outside of a transaction, it neither throws nor warns the developer by an Eslint rule.
There isn't a runtime error while doing so, but as the callback doesn't work as expected it breaks the logic.
And developers see a completely unrelated error message which is unrelated to callbacks because the logic is not working as expected.
I don't know if there is a use case for callbacks to be created outside of transactions.
If there is none, I recommend adding an Eslint rule to warn users when callbacks are created outside of transactions.
If anyone can tell me where and how can I contribute to official o1js Eslint plugin, then I'll gladly do it.

This doesn't work as expected. (as the callback is created outside)
```ts
const callback = Experimental.Callback.create(zkAppB, 'approveSend', [amount])

const tx = await Mina.transaction(feePayerAddress, () => {
    AccountUpdate.fundNewAccount(feePayerAddress)
    token.transfer(exchangeAddress, tokenAccount1, amount, callback)
})
```

But this works.
```ts
const tx = await Mina.transaction(feePayerAddress, () => {
    const callback = Experimental.Callback.create(zkAppB, 'approveSend', [amount])
    AccountUpdate.fundNewAccount(feePayerAddress)
    token.transfer(exchangeAddress, tokenAccount1, amount, callback)
})
```

mitschabaude:
Thanks for raising this. This could be fixed I think, by running the Callback lazily. But it would be just another instance of internal complication to support magical APIs, which I think we should stop doing.

I'm on the side that we should probably remove `Callback`. It's a trivial wrapper around running the callback method, and passing the account update that it created on to the other method. That's a much less magical way to think about it, and the behaviour of having to do it in your transaction is more obvious

PR #1430, opened 2024-02-13
Remove experimental callback

mitschabaude:
This follows up with removing `Experiment.Callback` and `Experimental.createChildAccountUpdate` as planned in #1402 

closes https://github.com/o1-labs/o1js/issues/1347


src/examples/zkapps/token_with_proofs.ts
mitschabaude, line 1: I refactored this example, which was using callbacks, to use the new `TokenContract` instead

Trivo25 (approved):

PR #1437, opened 2024-02-15
Missed part of the Mainnet support.

shimkiv:


src/lib/account-update.ts
```diff
@@ -1017,7 +1017,12 @@ class AccountUpdate implements Types.AccountUpdate {
...
-      return hashWithPrefix(prefixes.body, packToFields(input));
+      return hashWithPrefix(
+        activeInstance.getNetworkId() === 'mainnet'
+          ? prefixes.zkappBodyMainnet
+          : prefixes.zkappBodyTestnet,
+        packToFields(input)
+      );
     } else {
       let json = Types.AccountUpdate.toJSON(this);
       return Field(Test.hashFromJson.accountUpdate(JSON.stringify(json)));
```
mitschabaude, line 1028: this will also need to take the network id

mitschabaude:
would be nice to add a test which uses mainnet, or change to mainnet in one of the existing intg tests

ymekuria (approved):

src/lib/account-update.ts
shimkiv: Thanks @mitschabaude, can you please assist me where it comes from, the OCaml side I believe, isn't it?

mitschabaude: yes, from `consistency-test.ml`

src/mina-signer/src/sign-zkapp-command.unit-test.ts
```diff
@@ -81,28 +81,32 @@ expect(stringify(dummyInput.packed)).toEqual(
   stringify(dummyInputSnarky.packed)
 );
 
-test(Random.accountUpdate, (accountUpdate) => {
-  fixVerificationKey(accountUpdate);
+test(
+  Random.accountUpdate,
+  RandomTransaction.networkId,
```
mitschabaude, line 86: nice!

shimkiv: Not really, it probably needs to be aligned and not just random.

shimkiv: Or perhaps your previous [point](https://github.com/o1-labs/o1js/pull/1437#discussion_r1491355883) needs to be addressed first...

shimkiv: Or if `Provable.inCheckedComputation()` is always true then `accountUpdateSnarky.hash()` will always fallback to the `testnet`.

mitschabaude: Yes you need to address the other case

shimkiv: Updated it like [this](https://github.com/o1-labs/o1js/pull/1437/commits/1984c25ca58075b6124dbe8c26e3bbe324001fbb). 
@mitschabaude is that true that in order to make it pass we first should address your previous [point](https://github.com/o1-labs/o1js/pull/1437#discussion_r1491355883) above?

mitschabaude: that looks good, and yes you need to address my previous point

src/lib/account-update.ts
shimkiv: - [Test.hashFromJson.accountUpdate()](https://github.com/o1-labs/o1js/pull/1437/commits/1eb25efe8d980ab8784e7ff2d66957f3eeebe9cc) seems to be done.
- Next up `Test.hashFromJson.transactionCommitments()`.

ymekuria (approved):

Issue #1286, opened 2023-12-04
Be stricter about ways to create UInt64/32


mitschabaude:
**Why**: Currently, it's very easy to create `UInt64` or `UInt32` from a `Field` without range-checking it, without given any hint that this is unsafe. For `Bool` we do a much better job: the only way to create it from a `Field` is with `Bool.Unsafe.ofField()`

**How**: 
* Remove `Field` as input type from `UInt64.from()`. Replace with something like `UInt64.Unsafe.fromField()`
* Remove `Field` as possible constructor input. The tricky thing is that `UInt64` really consists of 1 Field, so we need a way to construct it from field at least in internal methods. There are two ways around this:
  * make the constructor take a `FieldVar` instead of a `Field`, so that creating it from a `Field` becomes a less visible and more low-level operation. This is how `Bool` does it.
  * or make the constructor private externally. This makes the breaking change more breaking though.

barriebyron:
Please advise @mitschabaude can you help identify the required doc updates?

mitschabaude:
@barriebyron the UInt64 API changed, so this is about fixing type errors that turn up when switching to the latest o1js version in doc examples

barriebyron:
Can you help me identify the examples that need updating?

PR #1438, opened 2024-02-16
Ensure safe Uint64/32 instanciations, with new Unsafe.fromField()

julio4:
Related to #1286
Wanted to explore the o1js repository and tried to tackle this issue.

- Changed `Field` to `FieldVar` in constructors
- Removed `Field` from `from()` functions
- Added `Unsafe.fromField()`

The `sha256` gadget still rely on a lots of integers directly from field, so this may be a follow up issue as well.


mitschabaude:
Awesome!

mitschabaude:
> The sha256 gadget still rely on a lots of integers directly from field, so this may be a follow up issue as well.
this is fine, it's a low-level gadget and has to reach beyond the safe abstraction of a UInt32

julio4:
> > The sha256 gadget still rely on a lots of integers directly from field, so this may be a follow up issue as well.
> 
> not needed, sha256 is a low-level gadget and has to reach beyond the safe abstraction of a UInt32

Ok, then I can safely use the FieldVar value of the Field (or the new `Unsafe.fromField`).
Could you please review the changes inside the methods of integers to be sure that there is nothing unsafe (As I only used Field.value internally)?

mitschabaude:
Hey @julio4 this is really valuable work that we'd love to include in a release soon. Any way we can help land this?
If you don't have time to finish it, I'm offering to find someone on our team to do so!

julio4:
> Hey @julio4 this is really valuable work that we'd love to include in a release soon. Any way we can help land this? If you don't have time to finish it, I'm offering to find someone on our team to do so!

Hello! I planned to work on it this week. But I think it should be ready, I would just need some help to run pipelines and build locally. I followed the local dev env setup but I encountered some issues.

mitschabaude:
> I would just need some help to run pipelines and build locally.

@julio4 sure, try to run the following steps. does it fail somewhere?

```
# ensure packages are up to date
npm i

# pull in dependencies in git submodules
git submodule update --init --recursive

# ts build of the main code base
npm run build

# check that all examples still build
npm run build:examples

# check that integration tests work (takes long, feel free to stop after a minute or so if it seems to work)
npm run test:integration

# run unit tests (this takes super long unfortunately, feel free to let ci do this if you got this far)
npm run test:unit
```

src/lib/int.ts
```diff
@@ -19,23 +19,41 @@ class UInt64 extends CircuitValue {
...
+  /**
+   * Create a {@link UInt64}.
+   * The max value of a {@link UInt64} is `2^64 - 1 = UInt64.MAXINT()`.
+   *
+   * **Warning**: Cannot overflow, an error is thrown if the result is greater than UInt64.MAXINT()
+   */
+  constructor(x: UInt64 | UInt32 | FieldVar | number | string | bigint) {
+    if (x instanceof UInt64 || x instanceof UInt32) x = x.value.value;
+    super({ value: Field(x) });
+    UInt64.checkConstant(this.value);
```
mitschabaude, line 31: this doesn't seem to work, it throws an error in every CI test
```suggestion
    let value = Field(x);
    super({ value });
    UInt64.checkConstant(value);
```

src/lib/int.ts
```diff
@@ -546,24 +565,42 @@ class UInt32 extends CircuitValue {
...
+  /**
+   * Create a {@link UInt32}.
+   * The max value of a {@link UInt32} is `2^32 - 1 = UInt32.MAXINT()`.
+   *
+   * **Warning**: Cannot overflow, an error is thrown if the result is greater than UInt32.MAXINT()
+   */
+  constructor(x: UInt32 | FieldVar | number | string | bigint) {
+    if (x instanceof UInt32) x = x.value.value;
+    super({ value: Field(x) });
+    UInt32.checkConstant(this.value);
```
mitschabaude, line 577:
```suggestion
    let value = Field(x);
    super({ value });
    UInt32.checkConstant(value);
```

julio4:
> @julio4 sure, try to run the following steps. does it fail somewhere?
> 
> ```
> # ensure packages are up to date
> npm i
> 
> # pull in dependencies in git submodules
> git submodule update --init --recursive
> 
> # ts build of the main code base
> npm run build
> 
> # check that all examples still build
> npm run build:examples
> 
> # check that integration tests work (takes long, feel free to stop after a minute or so if it seems to work)
> npm run test:integration
> 
> # run unit tests (this takes super long unfortunately, feel free to let ci do this if you got this far)
> npm run test:unit
> ```

Thank you! But I already followed the README-dev and it fail at the main ts build:
```
building cjs version of ./src/index.ts
using bindings from ./src/bindings/compiled/node_bindings/
▲ [WARNING] "import.meta" is not available with the "cjs" output format and will be empty [empty-import-meta]

    dist/node/bindings/js/node/node-backend.js:6:10:
      6 │ let url = import.meta.url;
        ╵           ~~~~~~~~~~~

  You need to set the output format to "esm" for "import.meta" to work correctly.
```

Do you know what node version is used?

mitschabaude:
> Thank you! But I already followed the README-dev and it fail at the main ts build:

@julio4 the command line output you show isn't a failure. It's a warning, which does not point to an actual problem

mitschabaude:
you can also see in the CI test that the build works fine, but something about `UInt` construction still fails in every single test, at runtime

julio4:
> @julio4 the command line output you show isn't a failure. It's a warning, which does not point to an actual problem
Oh yes, I didn't noticed but all builds are working. The error comes from `account-update`, but I'm having some troubles finding where's the `UInt` are created.

mitschabaude:
@julio4 it seems that every construction of uints is failing. So if the account update file is hard to parse, I recommend to just create a trivial test or run the integer test with `./jest src/lib/int.test.ts`

mitschabaude:
@julio4 thanks so much for doing most of this heavy change! I pushed the last remaining fixes, it's ready to go!

mitschabaude (approved):

julio4:
> @julio4 thanks so much for doing most of this heavy change! I pushed the last remaining fixes, it's ready to go!

Thank you for fixing the issue. I tried to dig a little bit on it and didn't found what was causing the error.

PR #1442, opened 2024-02-17
Fix for Mina-Signer exports.

shimkiv:


Trivo25 (approved):

src/mina-signer/index.d.ts
```diff
@@ -1,6 +1,7 @@
 // this file is a wrapper for supporting types in both commonjs and esm projects
 
 import Client from './mina-signer.ts';
-export { type NetworkId } from './src/types.ts';
+import { NetworkId } from './src/types.ts';
 
-export = Client;
+export default Client;
```
mitschabaude, line 6: was this change necessary?

shimkiv: It does not impact functionality but since mina-signer is ESM project it removes error on that line if used to target CJS.

PR #1443, opened 2024-02-17
Export improvements.

shimkiv:
- Mina signer's `Client` should now work with default or named imports for both CJS and ESM projects.
- `o1js` also exports the `NetworkId`.

### DIFF
--- a/src/mina-signer/index.d.ts
+++ b/src/mina-signer/index.d.ts
@@ -1,7 +1,6 @@
 // this file is a wrapper for supporting types in both commonjs and esm projects
 
-import Client from './mina-signer.ts';
-import { NetworkId } from './src/types.ts';
+import Client, { type NetworkId } from './mina-signer.ts';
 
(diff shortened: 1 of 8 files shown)


src/mina-signer/mina-signer.ts
```diff
@@ -34,7 +34,7 @@ import {
 import { sign, Signature, verify } from './src/signature.js';
 import { createNullifier } from './src/nullifier.js';
 
-export { Client as default };
+export { Client as default, type NetworkId };
```
mitschabaude, line 37: You have to do the named Client export here as well

src/mina-signer/index.d.ts
mitschabaude, line 1: This is file is just declaring types. It has no influence on the runtime. So if you want a named export, you have to do them in the two JS files

mitschabaude (approved):
lgtm!

Issue #1149, opened 2023-10-03
mina-signer: use string instead of boolean for compilation option


nicc:
We need to make the mina-signer change that goes with this PR: https://github.com/MinaProtocol/mina/pull/13232

bkase:
Context: If we don't expose this in mina-signer, it's not really real. Why is this feature important? It prevents replay attacks from an arbitrary string-labeled network (instead of just two, which isn't enough in a world with projects like [zeko](https://zeko.io/))

PR #1444, opened 2024-02-18
Enable custom network signing 

Trivo25:
This PR enables custom network signing and verification, all across our stack (o1js and mina-signer). This feature unlocks the use of o1js and mina-signer with layer 2s/3s or chains isomorphic to Mina. All while maintaining backwards compatibility

bindings https://github.com/o1-labs/o1js-bindings/pull/247

closes https://github.com/o1-labs/o1js/issues/1149

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Support for custom network identifiers other than `mainnet` or `testnet` https://github.com/o1-labs/o1js/pull/1444
-  - New optional argument `networkId?: string` to `Signature.create()` and `Signature.verify()` to reflect support for custom network identifiers.
 
 ## [0.16.1](https://github.com/o1-labs/o1js/compare/834a44002...3b5f7c7)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Support for custom network identifiers other than `mainnet` or `testnet` https://github.com/o1-labs/o1js/pull/1444
-  - New optional argument `networkId?: string` toin `Signature.create()` and `Signature.verify()` to reflect support for custom network identifiers.
+  - New optional argument `networkId?: string` to `Signature.create()` and `Signature.verify()` to reflect support for custom network identifiers.
 
 ## [0.16.1](https://github.com/o1-labs/o1js/compare/834a44002...3b5f7c7)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/3b5f7c7...HEAD)
 
+### Added
+
+- Support for custom network identifiers other than `mainnet` or `testnet` https://github.com/o1-labs/o1js/pull/1444
+  - New optional argument `networkId?: string` in `Signature.create(privKey: PrivateKey, msg: Field[], networkId?: string)` to reflect support for custom network identifiers.
+
 ## [0.16.1](https://github.com/o1-labs/o1js/compare/834a44002...3b5f7c7)
 
 ### Breaking changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Added
 
 - Support for custom network identifiers other than `mainnet` or `testnet` https://github.com/o1-labs/o1js/pull/1444
-  - New optional argument `networkId?: string` in `Signature.create(privKey: PrivateKey, msg: Field[], networkId?: string)` to reflect support for custom network identifiers.
+  - New optional argument `networkId?: string` toin `Signature.create()` and `Signature.verify()` to reflect support for custom network identifiers.
 
 ## [0.16.1](https://github.com/o1-labs/o1js/compare/834a44002...3b5f7c7)
 
--- a/src/lib/signature.ts
+++ b/src/lib/signature.ts
@@ -236,11 +236,7 @@ class Signature extends CircuitValue {
    * Signs a message using a {@link PrivateKey}.
    * @returns a {@link Signature}
    */
-  static create(
-    privKey: PrivateKey,
-    msg: Field[],
-    networkId?: string
-  ): Signature {
+  static create(privKey: PrivateKey, msg: Field[]): Signature {
     const publicKey = PublicKey.fromPrivateKey(privKey).toGroup();
     const d = privKey.s;
     // we chose an arbitrary prefix for the signature, and it happened to be 'testnet'
@@ -271,7 +267,7 @@ class Signature extends CircuitValue {
    * Verifies the {@link Signature} using a message and the corresponding {@link PublicKey}.
    * @returns a {@link Bool}
    */
-  verify(publicKey: PublicKey, msg: Field[], networkId?: string): Bool {
+  verify(publicKey: PublicKey, msg: Field[]): Bool {
     const point = publicKey.toGroup();
     // we chose an arbitrary prefix for the signature, and it happened to be 'testnet'
     // there's no consequences in practice and the signatures can be used with any network
--- a/src/lib/signature.ts
+++ b/src/lib/signature.ts
@@ -235,7 +235,11 @@ class Signature extends CircuitValue {
    * Signs a message using a {@link PrivateKey}.
    * @returns a {@link Signature}
    */
-  static create(privKey: PrivateKey, msg: Field[]): Signature {
+  static create(
+    privKey: PrivateKey,
+    msg: Field[],
+    networkId?: string
+  ): Signature {
     const publicKey = PublicKey.fromPrivateKey(privKey).toGroup();
     const d = privKey.s;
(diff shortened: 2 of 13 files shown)


src/mina-signer/src/signature.ts
```diff
@@ -311,9 +309,65 @@ function hashMessageLegacy(
...
+    case 'mainnet':
+      return [networkIdMainnet, 8];
+    case 'testnet':
+      return [networkIdTestnet, 8];
+    default:
+      return networkIdOfString(networkId);
+  }
+}
+
+const createCustomPrefix = (prefix: string) => {
```
Trivo25, line 340: OCaml uses a `*` padding to `20` bytes

src/mina-signer/src/signature.unit-test.ts
```diff
@@ -23,26 +23,29 @@ function checkConsistentSingle(
   msg: Field,
   key: PrivateKey,
   keySnarky: PrivateKeySnarky,
-  pk: PublicKey
+  pk: PublicKey,
+  networkId: string
```
Trivo25, line 27: some cleanup, tests all three network cases now

MartinOndejka (commented):
Signature in o1js (not mina-signer) has still hard-coded testnet prefix.

mitschabaude:
> > Signature in o1js (not mina-signer) has still hard-coded testnet prefix.
> 
> good catch!

The network id was purposefully left out as a parameter in the userland functions to create signatures: the in-SNARK `Signature.verify()` and its mina-signer counter-part `signFields()`. These signatures are not designed to be related to any particular network, and are not verified by a Mina node. 

We only happen to reuse the same signing functions also used by transaction signatures, so we had to pick _some_ choice of `networkId`. So we decided to hard-code "testnet".

Unless there is a specific reason to make these signatures network-specific in the PR, I'd be in favor of reverting these changes.

CHANGELOG.md
```diff
@@ -17,6 +17,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/3b5f7c7...HEAD)
 
+### Added
+
+- Support for custom network identifiers other than `mainnet` or `testnet` https://github.com/o1-labs/o1js/pull/1444
+  - New optional argument `networkId?: string` to `Signature.create()` and `Signature.verify()` to reflect support for custom network identifiers.
```
mitschabaude, line 23: revert?

src/lib/signature.ts
```diff
@@ -235,21 +236,25 @@ class Signature extends CircuitValue {
    * Signs a message using a {@link PrivateKey}.
    * @returns a {@link Signature}
    */
-  static create(privKey: PrivateKey, msg: Field[]): Signature {
+  static create(
+    privKey: PrivateKey,
+    msg: Field[],
+    networkId?: string
```
mitschabaude, line 242: revert?

src/mina-signer/mina-signer.ts
```diff
@@ -122,9 +120,13 @@ class Client {
    * @param privateKey The private key used for signing
    * @returns The signed field elements
    */
-  signFields(fields: bigint[], privateKey: Json.PrivateKey): Signed<bigint[]> {
+  signFields(
+    fields: bigint[],
+    privateKey: Json.PrivateKey,
+    network?: NetworkId
```
mitschabaude, line 126: revert?

Trivo25:
> Unless there is a specific reason to make these signatures network-specific in the PR, I'd be in favor of reverting these changes. Adding the networkId parameter just adds confusion IMO.

Hmm given that you can specify a custom message anyway (with a custom nonce), this really can be reverted. But curious to know if @MartinOndejka sees an explicit need for this?

MartinOndejka:
> > Unless there is a specific reason to make these signatures network-specific in the PR, I'd be in favor of reverting these changes. Adding the networkId parameter just adds confusion IMO.
> 
> Hmm given that you can specify a custom message anyway (with a custom nonce), this really can be reverted. But curious to know if @MartinOndejka sees an explicit need for this?

Not at all, I just remember that it was a little confusing a while back when I found out that it has hardcoded `testnet`.

mitschabaude:
> > > Unless there is a specific reason to make these signatures network-specific in the PR, I'd be in favor of reverting these changes. Adding the networkId parameter just adds confusion IMO.
> > 
> > 
> > Hmm given that you can specify a custom message anyway (with a custom nonce), this really can be reverted. But curious to know if @MartinOndejka sees an explicit need for this?
> 
> Not at all, I just remember that it was a little confusing a while back when I found out that it has hardcoded `testnet`.

@Trivo25 maybe it would be good to add a comment to these methods explaining why they hard-code testnet

src/mina-signer/src/types.ts
```diff
@@ -9,7 +9,7 @@ export type Field = number | bigint | string;
 export type PublicKey = string;
 export type PrivateKey = string;
 export type Signature = SignatureJson;
-export type NetworkId = 'mainnet' | 'testnet';
+export type NetworkId = string;
```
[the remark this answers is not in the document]
mitschabaude: I think TS would just collapse that to `string`, because `'mainnet'` and `'testnet'` are contained in string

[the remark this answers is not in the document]
mitschabaude: so after all, we ca probably just remove the `NetworkId` export which caused us to release 2 or 3 new versions in a row 😆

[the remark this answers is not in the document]
mitschabaude: btw, now we also need those runtime checks which I made you remove @shimkiv 😅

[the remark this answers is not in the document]
mitschabaude: damn, I wonder if we should keep the detailed type somehow. Maybe something like
```ts
type NetworkId = 'mainnet' | 'testnet' | { custom: string }
```
@Trivo25 wdyt

src/mina-signer/src/sign-zkapp-command.unit-test.ts
```diff
@@ -82,42 +82,38 @@ expect(stringify(dummyInput.packed)).toEqual(
...
-  let hashTestnet = accountUpdateHash(accountUpdate, 'testnet');
-  let hashMainnet = accountUpdateHash(accountUpdate, 'mainnet');
-  setActiveInstance(testnetMinaInstance);
-  let hashSnarkyTestnet = accountUpdateSnarky.hash();
-  setActiveInstance(mainnetMinaInstance);
-  let hashSnarkyMainnet = accountUpdateSnarky.hash();
-  expect(hashTestnet).toEqual(hashSnarkyTestnet.toBigInt());
-  expect(hashMainnet).toEqual(hashSnarkyMainnet.toBigInt());
-});
+test(
```
shimkiv, line 85: @Trivo25 
Throughout different tests we check that signatures match like `testnet <-> testnet` but should we also check that the resulting signatures are indeed different? Like `testnet <-> mainnet` should always differ.

Trivo25: Yes, we should! And we do
For example, further down we verify against a different network and expect the verification to fail
```ts
expect(
   verify(sigFieldElements, networkId === 'mainnet' ? 'testnet' : 'mainnet')
).toEqual(false);
```

src/mina-signer/mina-signer.ts
```diff
@@ -39,17 +39,10 @@ export { Client, Client as default, type NetworkId };
...
 
 class Client {
-  private network: NetworkId; // TODO: Rename to "networkId" for consistency with remaining codebase.
+  private network: NetworkId;
 
-  constructor(options: { network: NetworkId }) {
-    if (!options?.network) {
-      throw Error('Invalid Specified Network');
-    }
-    const specifiedNetwork = options.network.toLowerCase();
```
Trivo25, line 48: this is no longer needed since the type now is `'testnet' | 'mainnet' ` which already is lower case

src/mina-signer/src/sign-zkapp-command.unit-test.ts
Trivo25: but I added a check that the hash doesnt match another networks hash!

src/mina-signer/src/signature.unit-test.ts
```diff
@@ -17,32 +17,40 @@ import { AccountUpdate } from '../../bindings/mina-transaction/gen/transaction-b
...
   // verify
-  let okTestnetTestnet = verifyFieldElement(sigTest, msg, pk, 'testnet');
-  let okMainnetTestnet = verifyFieldElement(sigMain, msg, pk, 'testnet');
-  let okTestnetMainnet = verifyFieldElement(sigTest, msg, pk, 'mainnet');
-  let okMainnetMainnet = verifyFieldElement(sigMain, msg, pk, 'mainnet');
-  expect(okTestnetTestnet).toEqual(true);
-  expect(okMainnetTestnet).toEqual(false);
-  expect(okTestnetMainnet).toEqual(false);
-  expect(okMainnetMainnet).toEqual(true);
+  expect(verifyFieldElement(sig, msg, pk, networkId)).toEqual(true);
```
Trivo25, line 33: test is now condensed by using `networkId` as input

src/mina-signer/src/types.ts
```diff
@@ -9,7 +9,13 @@ export type Field = number | bigint | string;
 export type PublicKey = string;
 export type PrivateKey = string;
 export type Signature = SignatureJson;
-export type NetworkId = 'mainnet' | 'testnet';
+export type NetworkId = 'mainnet' | 'testnet' | { custom: string };
+
+export const NetworkId = {
```
Trivo25, line 14: internal helper to get the plain `networkId`

mitschabaude: nicely done!

mitschabaude (approved):
awesome!

PR #1446, opened 2024-02-19
Replace `this.token` with methods on `TokenContract`

### LINKED ISSUE #1439: Assert `isNew` in the deploy method of a token contract
## The problem

Token contracts are able to ensure validity of token interactions by the requirement that account updates of them always have to be included as a parent, for every token interaction.

This only works because we have the 'access' permission!

For default accounts, everyone can include one of their account updates without authorization (common use case: sending someone money without their authorization). The 'access' permission in that case is "none".

Token contracts set their access permission to "proof" or "proofOrSignature". That way, an account update of them _always_ requires a proof. Which ensures that you have to successfully execute a token contract method to approve any token interaction. Usually, the access permission is set to "proof" in the `deploy()` method.

There is a potential security hole. Imagine someone creates an account, and some time later deploys a token contract to it. In the meantime, the access permission is still "none". An attacker who anticipates that this will become a token contract could mint tokens in the meantime. For example, such an attacker could monitor the mempool for any transactions that look like token contract deployments, and try to frontrun them with a high-fee transaction minting themselves tokens.

## The mitigation

A solid defense against this attack is to require the `isNew` precondition on the same account update that sets the access permission. That means the token contract _must_ be secured from the very start of its account's existence. (The worst case is that the token contract can't be deployed, because an attacker created the account earlier.)

This is a subtle issue, so we should mitigate it by default, otherwise many token contracts would be vulnerable to this attack in practice.

The measure I recommend is to `this.account.isNew.requireEquals(Bool(true))` inside the default `deploy()` on `TokenContract`.

mitschabaude:
closes #1439 

Removes `this.token` from `SmartContract` and `AccountUpdate`, for two reasons:
- Reduce API bloat
- Discourage use of token APIs on non-token contracts, because there are too many security pitfalls (for example, the main DEX example before this PR allowed arbitrary minting of liquidity tokens)

Instead, we reintroduce `mint()`, `burn()` and `send()` as helper methods under `TokenContract.internal`. The intention of this name is to make it clearer that these methods are for use _within_ token contract methods, not from the outside (because they wouldn't be authorized when used from outside)

`this.token.id` is reintroduced as `this.deriveTokenId()` on token contracts. The reason to change the API from a property access to a function is that it's actually a function, which computes a hash in the circuit, and it's weird if a property access secretly adds constraints.


src/lib/signature.ts
```diff
@@ -32,16 +32,31 @@ class PrivateKey extends CircuitValue {
...
+  /**
+   * Create a random keypair `{ privateKey: PrivateKey, publicKey: PublicKey }`.
+   *
+   * Note: This uses node or browser built-in APIs to obtain cryptographically strong randomness,
+   * and can be safely used to generate a real keypair.
+   */
+  static randomKeypair() {
+    let privateKey = PrivateKey.random();
+    return { privateKey, publicKey: privateKey.toPublicKey() };
+  }
```
mitschabaude, line 58: quality of life improvement, it's annoying to always need two commands to create a keypair

src/examples/zkapps/dex/arbitrary-token-interaction.ts
mitschabaude, line 1: this example was meant as a reproduction at a time where the missing 'access' permission caused a security hole for token contracts, and is long obsolete

src/examples/zkapps/dex/dex-with-actions.ts
```diff
@@ -71,7 +79,7 @@ class Dex extends SmartContract {
   }
 
   @method createAccount() {
-    this.token.mint({ address: this.sender, amount: UInt64.from(0) });
+    this.internal.mint({ address: this.sender, amount: UInt64.from(0) });
```
MartinMinkov, line 82: `this.internal` is a great way to communicate usage within smart contract methods, awesome!

src/lib/mina/token/token-contract.ts
```diff
@@ -27,10 +29,31 @@ abstract class TokenContract extends SmartContract {
...
+    // set access permission, to prevent unauthorized token operations
     this.account.permissions.set({
       ...Permissions.default(),
       access: Permissions.proofOrSignature(),
     });
+
+    // assert that this account is new, to ensure unauthorized token operations
+    // are not possible before this contract is deployed
+    // see https://github.com/o1-labs/o1js/issues/1439 for details
+    this.account.isNew.requireEquals(Bool(true));
```
MartinMinkov, line 42: Is there a way we can provide documentation via intellisense on this rather than in code documentation? Is it even necessary to communicate this? Just spitting ideas here.

src/lib/signature.ts
```diff
@@ -32,16 +32,31 @@ class PrivateKey extends CircuitValue {
...
   /**
-   * You can use this method to generate a private key. You can then obtain
-   * the associated public key via {@link toPublicKey}. And generate signatures
-   * via {@link Signature.create}.
+   * Generate a random private key.
+   *
+   * You can obtain the associated public key via {@link toPublicKey}.
+   * And generate signatures via {@link Signature.create}.
+   *
+   * Note: This uses node or browser built-in APIs to obtain cryptographically strong randomness,
```
MartinMinkov, line 40: Great call out :+1:

src/lib/zkapp.ts
```diff
@@ -846,10 +847,13 @@ super.init();
     return this.self.currentSlot;
   }
   /**
-   * Token of the {@link SmartContract}.
+   * @deprecated `SmartContract.token` will be removed, and token methods will only be available on `TokenContract.internal`.
+   * Instead of `SmartContract.token.id`, use `TokenContract.deriveTokenId()`.
+   *
+   * For security reasons, it is recommended to use `TokenContract` as the base contract for all tokens.
```
MartinMinkov, line 853: Would it be helpful to use the link tag here for `TokenContract`?

MartinMinkov (approved):

src/lib/mina/token/token-contract.ts
mitschabaude: I think this is a good idea! Maybe a doc-comment on `deploy()`, so that when you're going to override it by typing `super.deploy()`, you have documentation about what it does

Issue #1337, opened 2023-12-19
Remove Apple silicon performance workaround as this has been fixed upstream


jackryanservia:
Earlier this year, we found an issue (#491) where o1js performance was much slower on Apple silicon machines than would be expected and fixed it with a workaround in #683.

The Chromium team identified the issue in [1228686](https://bugs.chromium.org/p/chromium/issues/detail?id=1228686&q=reporter%3Arreverser%40google.com&can=1). It was the result of V8 not using LSE instructions even when the underlying hardware supported them and has since been fixed.

Given the finding below, it makes sense to remove the workaround we applied in #683, as doing so would dramatically improve performance on Apple silicon machines.

jackryanservia:
Running `src/examples/crypto/ecdsa/run.ts` with the current configuration takes 108 seconds to compile and 41 seconds to prove. With `getEfficientNumWorkers` disabled, it takes 16 seconds to compile and  28 seconds to prove! 😮 

Profiling with `getEfficientNumWorkers` reveals that almost no ticks are used on `rdl_dealloc` and `rdl_alloc`:
##### Preworkaround profile:
```
ticks  total  nonlib   name
79876   41.8%   41.8%  Function: *__rdl_dealloc
...
65241   34.1%   34.1%  Function: *__rdl_alloc
```
##### Node.js 20 profile with `getEfficientNumWorkers` disabled:
```

mitschabaude:
Fantastic news, thanks for investigating!!

dfstio:
See also https://github.com/o1-labs/o1js/pull/1456#issuecomment-1980994115

PR #1456, opened 2024-02-22
Improve performance by reverting Apple silicon workaround (root problem is fixed upstream)

jackryanservia:
Closes: #1337 ;)
Bindings PR: o1-labs/o1js-bindings#251

* Spawns one worker for each physical core by default (for some reason, `numCpus - 1` was faster a month ago, but when I tested today, `numCores` was slightly faster).
* Exposes `setNumberOfWorkers()` to allow developers to manually set the number of workers.
* Removes detect-gpu dependency

I tested and everything worked well, prover time decrease of ~35% in browser, and ~46% in node. Compile time improvement is even more significant ~60%! (Tested on Apple M1 Pro 10-Core).

I wasn't exactly sure of the best way to implement the `setNumberOfWorkers()` function, and where everything should live/what it should be named. So I would appreciate any feedback on that kinda thing :)

### DIFF
--- a/src/lib/proof-system/workers.ts
+++ b/src/lib/proof-system/workers.ts
@@ -1,4 +1,6 @@
-export const workers = {
+export { workers, setNumberOfWorkers };
+
+const workers = {
   numWorkers: undefined as number | undefined,
 };
 
@@ -10,6 +12,6 @@ export const workers = {
  * setNumberOfWorkers(2); // set the number of workers to 2
  * ```
  */
-export const setNumberOfWorkers = (numWorkers: number) => {
+const setNumberOfWorkers = (numWorkers: number) => {
   workers.numWorkers = numWorkers;
 };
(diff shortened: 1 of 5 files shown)


src/lib/proof-system/workers.ts
mitschabaude, line 1: this looks good to me!

just one nit: we use the convention across the code base to declare exports at the top of a file, would be nice to follow that
```ts
export { workers, setNumberOfWorkers }
```

mitschabaude (approved):
great to have this in!

dfstio:
My tests show that on M2 Max which has 12 cores, the optimal number of workers is 8, it is 35% faster than with 12 workers.
System info shows:
Total Number of Cores:	12 (8 performance and 4 efficiency)

Can you please set the default number of workers for M2 Max to 8? It seems like 12 now.

mitschabaude:
> Can you please set the default number of workers for M2 Max to 8? It seems like 12 now.

I don't think this is easily possible because the API that nodejs gives us is `availableParallelism()` which returns a number of cores. It doesn't distinguish between performance and efficiency cores.

I suggest to just set the number of workers yourself for fine-tuning on particular architectures, seems a bit much for us to test and optimize every possible hardware

Issue #1023, opened 2023-07-10
`Field.toBits()` and `fromBits()` are unsound


mitschabaude:
`Field.toBits()` without the optional length parameter returns 255 Bools to represent a field element. However, the representation of a field element as 255 bits is not necessarily unique!

Namely, for most Pasta field elements $x$, both $x$ and $x + p$ fit within 255 bits. If the decomposition of $x$ into bits is checked like I think it is -- as $x = x_0 + 2 x_1 + ... + 2^{254} x_{254} \bmod{p}$ --, then the prover can choose between two decompositions: that of $x$ and that of $x + p$.

The same vulnerability applies to `fromBits()` which likely uses the same constraints

Possible fix: Make 254 the default bit length parameter and don't allow 255.

Issue #1313, opened 2023-12-13
More efficient generic gate creation for linear combinations


mitschabaude:
## Why

Currently, when doing chains of `.add()` and `.mul(constant)`, snarky doesn't create any generic gates. Instead, it maintains an arbitrary long unreduced AST using the `Add` and `Scale` combinators.

The AST is reduced to generic gates whenever the variable that holds it is included in a gate.

This can lead to a bad explosion of gates used, because the same unreduced AST can be included _multiple times_ in gates.
Example:
```ts
// u = 5x + 10y + 2z
let u = x.mul(5).add(y.mul(10)).add(z.mul(2));
Gadgets.rangeCheck64(u); // causes 1st chain of generic gates from u
let v = u.mul(u); // causes 2nd and 3rd chain of generic gates from u
```
The example creates 6 generic gates (3 rows) just from doing the same reduction of $5x + 10y + 2z$ (which costs 2 addition gates) _three times_. On top of that, it creates 1.5 rows for the range check and multiplication. **Total cost: 4.5 rows**

PLONK can only do 1 addition (with optional multiplying constants) per generic gate: $ax + by$. So, there seems to be zero value in maintaining unreduced ASTs from addition chains with 3 or more variables.

As soon as the third distinct variable is added, it's definitely time to reduce it!

## How

In the above example, we would add a generic gate for $5x + 10y$ as soon as doing `.add(z.mul(2))`. Then, u would be an unreduced AST containing a single addition. The subsequent inclusions in gates would reduce the remaining addition 3 times at a cost of 1 generic gate each. **Total cost: 3.5 rows**

If we implement #1312 as well, the multiplication would be intelligent enough to only use 1 generic gate for everything, so we're down to: **Total cost: 2.5 rows**

This is a simple change which seems to have no negative impact in practice.

mitschabaude:
this is a zkDSL level change that I think we should make wherever our DSL core will live in the future

PR #1461, opened 2024-02-26
Implement to/fromBits in TS, and set max length to 254 bits

jackryanservia:
Closes: #1023

* Implements `toBits` and `fromBits` methods in TS.
* Sets length limit 254bits for both to prevent issue outlined in #1023


src/lib/field.ts
```diff
@@ -908,34 +909,34 @@ class Field {
...
+    checkBitLength('Field.toBits()', length, 254);
+    let bits = Provable.witness(Provable.Array(Bool, length), () => {
+      let f = this.toBigInt();
+      return Array.from(
+        { length },
+        (_, k) => new Bool(!!((f >> BigInt(k)) & 0xffn))
+      );
+    });
+    if (bits.slice(length).some((bit) => bit.toBoolean()))
+      throw Error(`Field.toBits(): ${this} does not fit in ${length} bits`);
```
mitschabaude, line 931: we can't do `toBoolean()` in circuit - but the constraint below should throw an error in exactly the same case? so maybe put this error in the `assertEquals` message below?

jackryanservia (commented):
I will think more about the `to/fromBits()` implementation and check the constraint count before I merge, but for now I just have a few questions about other things to make sure I'm on the right track! :)

src/lib/field.ts
```diff
@@ -908,34 +909,43 @@ class Field {
...
+    let bits = Provable.witness(Provable.Array(Bool, length), () => {
+      let f = this.toBigInt();
+      return Array.from(
+        { length },
+        (_, k) => new Bool(!!((f >> BigInt(k)) & 0x1n))
+      );
+    });
+    Field.fromBits(bits).assertEquals(
+      this,
+      `Field.toBits(): ${Provable.asProver(() => {
```
jackryanservia, line 938: Is `Provable.asProver()` the correct way to log this?

jackryanservia (commented):
Ok, it looks like the number of rows for both methods is exactly the same as before (and the digest is identical for `fromBits()`).

Let me know if I should add vk regression test for these methods; aside from that I think this is ready to merge if looks ok to reviewer :)

src/lib/nullifier.ts
```diff
@@ -70,7 +71,10 @@ class Nullifier extends Struct({
 
     // shifted scalar see https://github.com/o1-labs/o1js/blob/5333817a62890c43ac1b9cb345748984df271b62/src/lib/signature.ts#L220
     // pk^c
-    let pk_c = scaleShifted(this.publicKey, Scalar.fromBits(c.toBits()));
+    let pk_c = scaleShifted(
+      this.publicKey,
+      Scalar.fromBits(c.toBits().concat(new Bool(false)))
```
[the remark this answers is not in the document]
jackryanservia: Never mind all of this! I am still curious about the shiftiness, but now that I look at the `Scalar` code it's seems like there's nothing wrong with just having `Scalar.fromBits()` pad the end of any input less than 255 bits with `new Bool(false)`.

src/lib/field.ts
```diff
@@ -908,17 +909,17 @@ class Field {
...
-   * which by default is 255 bits. Prefer to pass a smaller `length` if possible.
+   * which by default is 254 bits. Prefer to pass a smaller `length` if possible.
    *
    * @param length - the number of bits to fit the element. If the element does not fit in `length` bits, the functions throws an error.
    *
    * @return An array of {@link Bool} element representing little endian binary representation of this {@link Field}.
    */
-  toBits(length?: number) {
-    if (length !== undefined) checkBitLength('Field.toBits()', length);
+  toBits(length: number = 254) {
```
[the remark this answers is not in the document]
mitschabaude: we could provide a secure, 255-bits version by treating the bits as a `ForeignField` element and asserting that it's less than the field size. That would add some overhead so it's better to not do it by default. It could be a separate method `toFullBits()` or similar. But not urgent IMO

src/lib/nullifier.ts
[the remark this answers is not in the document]
mitschabaude: yes agreed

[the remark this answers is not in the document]
mitschabaude: `Scalar.from()` doesn't take a `Field`. It takes a `bigint` but we can't move out and in of provable types like that

src/lib/field.ts
mitschabaude: what lol no, this will log `undefined` because `asProver()` doesn't return anything.
I would leave out the value,
```suggestion
      `Field.toBits(): Input does not fit in ${length} bits`
```

src/lib/field.ts
```diff
@@ -944,20 +954,20 @@ class Field {
...
-      return b.toField().value;
-    });
-    let x = Snarky.field.fromBits([0, ...bitsVars]);
-    return new Field(x);
+    return bits
+      .map((b) => new Bool(b))
+      .reduce((acc, bit, idx) => {
+        const shift = 1n << BigInt(idx);
+        return acc.add(bit.toField().mul(shift));
+      }, Field.from(0));
```
mitschabaude, line 970: could you add `.seal()` at the end of the returned expression? that way we return a fresh variable instead of an AST of 254 summands. the latter could create an enourmous amount of unnecessary constraints when used twice in different places. see https://github.com/o1-labs/o1js/issues/1313 for more background

tests/vk-regression/vk-regression.json
```diff
@@ -50,16 +50,16 @@
...
   },
   "HelloWorld": {
-    "digest": "11c21044c0ab297f29b386d1d04d0318cff4effa75191371dd576282882c2a8d",
+    "digest": "103e8bd9af45de70a1b214377e71bd05ae464e932dbe48aeada70362fa38b05",
     "methods": {
       "update": {
-        "rows": 826,
-        "digest": "5215ab9c7474b26f3a9073d821aef59b"
+        "rows": 825,
+        "digest": "226ff56f432f39d8056bf7f1813669f5"
```
mitschabaude, line 57: 1-row reduction makes sense because the hello-world example does `privateKey.toPublicKey()` which as the final step converts the `Group` (uncompressed EC point `(x,y)`) into a `PublicKey` (compressed EC point `(x, isOdd)`) and to do that it needs to get all of `y`'s bits with `toBits()`. `toBits()` needs 1 row less now because it adds up 254 bits, not 255. 

It does mean that certain public keys are no longer supported by this method. I think we should address that separately by introducing a dedicated `parity()` gadget using range checks, which makes it more efficient as well

mitschabaude (commented):
Looks great, just a few comments that should be addressed

Also @jackryanservia please add a changelog! Breaking change

src/lib/field.ts
jackryanservia: Ahhh, very interesting! :)

src/lib/field.ts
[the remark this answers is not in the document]
jackryanservia: This makes sense. I will implement tomorrow if there is time.

tests/vk-regression/vk-regression.json
jackryanservia: I had wondered about this. Would it make sense to have this use `toFullBits()` here in the meantime?

mitschabaude: Yes that would make sense

mitschabaude (approved):

ymekuria (approved):
Nice work @jackryanservia !

PR #1464, opened 2024-02-27
Feat: safe sender

### LINKED ISSUE #1425: Replace `this.sender` API with something that is secure by default
## Why

`this.sender` is an unconstrained value. The prover can choose its value. I originally thought of it as a convenient shortcut to letting a user that creates a transaction pass in their address. Also, this behaviour is documented: https://docs.minaprotocol.com/zkapps/o1js-reference/classes/SmartContract#sender

But I severely underestimated how important it is for many applications that they are interacting with _some particular_ address.  In this scenario, the natural pattern is to compare  `this.sender` with the address that you require:
```ts
this.sender.assertEquals(addressAllowedToInteract);
```
Which is a critical security bug. That I have seen multiple developers create in the wild.

I also understand now that documentation can't be enough to prevent this kind of usage. After all, our entire goal is to make o1js be intuitive and discoverable, so that you _don't_ always have to look up the docs. If the intuitive usage of an API is insecure, the API must change.

## How

The way to use `this.sender` securely is to also require a signature by the sender:
```ts
AccountUpdate.createSigned(this.sender);
```
That way, while it's not guaranteed that `this.sender` is really the transaction's fee payer, it _is_ guaranteed that the transaction's creator owns the `this.sender` account, which seems enough.

One idea is to keep the `this.sender` API and always let it add that extra account update. However, I think we shouldn't do that, because account updates have a cost -- there's a limit to them -- and because we don't like magic behaviour.

Instead, I propose to remove `this.sender` and add two different APIs which cover the two use cases for it:
* `this.sender.getAndRequireSignature()` -- explicitly require a signature, for use cases where you really need to know that the sender owns this account
* `this.sender.getUnconstrained()` -- explicitly get the sender as an unconstrained value, for (the likely majority of) zkapps where you could just as well accept the "sender" as method input, and its fine for the user to choose an address they don't actually own, if they want

julio4:
Fix #1425


mitschabaude:
@julio4 thanks for starting this! I went ahead and finished this PR since some of the decisions whether to use unconstrained vs requiring a signature are a bit subtle, and since landing this has become one of the last dependencies of an upcoming major version upgrade

Trivo25 (approved):
Thank you so much for your help, @julio4!

PR #1467, opened 2024-02-28
Diverse vk regression test

mitschabaude:
This just adds another zkprogram to the suite of verification keys checked for regression, which is diverse enough to be affected by #1466 

This provides a sanity check that #1466 really has an effect, and does not break proving of such diverse zkprograms


src/tests/transaction-flow.ts
mitschabaude, line 1: changes here unrelated and done to see the error in CI in case of a test flake

Trivo25 (approved):

PR #1468, opened 2024-02-28
Async circuits, pt 2 - witness and transaction callback

mitschabaude:
This PR adds the two remaining dependencies to support async contract methods:
* introducing `Provable.asyncWitness()`
  * this depends on the bindings and snarky changes
  * needed because we run methods in a witness block when calling contracts from other contracts
* make the `Mina.transaction` callback async
  * note: I _require_ the callback to be async. The reasoning is that 90% of uses need to be async anyway, because they call contracts. If we supported non-async callbacks, it would just become easier for developers to overlook that they have to await their contract calls now (with now, I mean with the next PR)
  
Since these two changes ended up creating a huge diff, and the next PR will do so again, I chose to defer actual async methods to yet another PR

snarky: https://github.com/o1-labs/snarky/pull/839
mina: to be seen if this can be included in an existing PR
bindings: https://github.com/o1-labs/o1js-bindings/pull/253


src/lib/mina/transaction.ts
mitschabaude, line 1: some actual changes here

src/lib/provable-core/exists.ts
mitschabaude, line 1: introducing async witness

src/lib/provable.ts
mitschabaude, line 1: introducing async witness

mitschabaude (commented):
most of the PR is just adding `asnyc` - a few comments to highlight actual changes to review:

src/snarky.d.ts
mitschabaude, line 1: additions to snarky-ml interface

src/lib/mina/transaction.ts
```diff
@@ -443,21 +446,21 @@ function transaction(f: () => void): Promise<Transaction>;
  */
 function transaction(
   sender: DeprecatedFeePayerSpec,
-  f: () => void
+  f: () => Promise<void>
 ): Promise<Transaction>;
 function transaction(
-  senderOrF: DeprecatedFeePayerSpec | (() => void),
-  fOrUndefined?: () => void
+  senderOrF: DeprecatedFeePayerSpec | (() => Promise<void>),
```
Trivo25, line 452: Might be about time that we remove the old deprecated private key as fee payer spec

Trivo25 (approved):

src/lib/mina/transaction.ts
mitschabaude: good idea. I'll do an API cleanup PR before v1 that removes all the long deprecated stuff

PR #1475, opened 2024-03-02
Example for using Merkle map with actions

mitschabaude:


dfstio:
This is a great example!

@barriebyron probably it makes sense to make a tutorial on docs.minaprotocol.com for it, what do you think?

@mitschabaude When recursive actions are available, will it be possible to make proofs in parallel for actions inside the reduce()? To be able to process 1000 actions fast.

Trivo25:
> @mitschabaude When recursive actions are available, will it be possible to make proofs in parallel for actions inside the reduce()? To be able to process 1000 actions fast.

In theory, it will be possible and that's something that should definitely be available in some shape or form. However, the initial recursive reducer will not have parallel processing functionality - I don't think it would make a big difference on most consumer hardware at this point and likely require outsourced computation

dfstio:
> In theory, it will be possible and that's something that should definitely be available in some shape or form. However, the initial recursive reducer will not have parallel processing functionality - I don't think it would make a big difference on most consumer hardware at this point and likely require outsourced computation

Fully agree, it would make sense only in the case of outsourced proving. Outsourced proving is what we will do during zkIgnite cohort 3 in zkCloudWorker project, so any information on future possible designs would be very helpful. Right now it is clear how to do it with ZkProgram recursion, and having nice async API to be called from method for parallel proving would be great.

mitschabaude:
this example is deprecated, possible in a much nicer way nowadays with IndexedMerkleMap and BatchReducer, see the batch reducer unit test

PR #1450, opened 2024-02-20
Async circuits, pt. 1 - async circuit runners

mitschabaude:
Starts to address https://github.com/MinaFoundation/Core-Grants/pull/4

This is the first step of supporting async circuits. It does not yet support async methods in contracts / zkprogram, but brings significant internal changes towards that support:
* Switch to a version of Pickles/Snarky which support async circuits under the hood. Internally, make the circuits that are passed to Pickles async functions
* Change low-level circuit runners -- `constraintSystem()` and `runAndCheck()` -- to be async, and support async callbacks in them. This is a necessary step because we run contract methods through these runners

bindings: https://github.com/o1-labs/o1js-bindings/pull/248
mina: https://github.com/MinaProtocol/mina/pull/15084
snarky: https://github.com/o1-labs/snarky/pull/836

PR #1477, opened 2024-03-04
Async circuit, pt. 3 - contract methods

### LINKED ISSUE #735: Make circuits async: JS side
Why:
* @mrmr1993 might work on allowing async circuits in Pickles
* Smart contracts need the result of async computations inside circuits (fetch accounts / best chain), and currently have error-prone workarounds; huge simplification if we could just fetch
* Users would also like to do async operations, and have no workarounds currently

mitschabaude:
Closes #735 

Most breaking change ever -- requires all smart contract and zkprogram methods to be async.

This PR is stacked behind #1450 and #1468 where most of the internal changes were already made

Here, there aren't much changes apart from
* The type signatures of `@method` and `ZKProgram`; introducing `@method.returns(..)` for methods with a return value
* Ensuring methods are awaited in the method wrapper
* Fixing all examples

TODO:
* [x] changelog (will add after merging main post release)
* [x] update bindings to make tests succeed (will do after ocaml side is final and reviewed)


tests/integration/inductive-proofs.js
mitschabaude, line 1: I deleted this entire folder, it was a left-over from CI tests in the Mina repo that don't exist anymore

src/lib/caller.unit-test.ts
mitschabaude, line 1: deleted this test, don't think it was still useful

src/examples/zkapps/reducer/reducer.ts
mitschabaude, line 1: this example was too similar to `reducer-composite.ts`, I'm trying to purge examples that don't really add anything, as they are a burden to maintain

src/lib/mina/transaction.ts
mitschabaude, line 1: an actual change here. I think the try-catch was unused because it didn't await the promise, so removed it

src/lib/zkapp.ts
mitschabaude, line 1: most of the actual changes in this file!

src/lib/zkapp.ts
```diff
@@ -72,28 +70,41 @@ import {
...
  * ```
  */
-function method<T extends SmartContract>(
-  target: T & { constructor: any },
-  methodName: keyof T & string,
-  descriptor: PropertyDescriptor
+function method<K extends string, T extends SmartContract>(
+  target: T & {
+    [k in K]: (...args: any) => Promise<void>;
+  },
```
mitschabaude, line 102: the decorator type now enforces a particular function signature of the method that is decorated:
```ts
(...args: any) => Promise<void>
```

src/lib/zkapp.ts
```diff
@@ -107,11 +118,6 @@ function method<T extends SmartContract>(
     target,
     methodName
   );
-  let returnType: Provable<any> = Reflect.getMetadata(
-    'design:returntype',
-    target,
-    methodName
-  );
```
mitschabaude, line 114: we no longer infer the method return type, instead there is `@method.returns(...)`

src/lib/zkapp.ts
```diff
@@ -203,7 +237,10 @@ function wrapMethod(
             let result: unknown;
             try {
               let clonedArgs = actualArgs.map(cloneCircuitValue);
-              result = method.apply(this, clonedArgs);
+              result = await assertPromise(
+                method.apply(this, clonedArgs),
+                noPromiseError
+              );
```
mitschabaude, line 243: throughout the wrapped method, we are now awaiting the method result and also throwing a helpful runtime error if it's not returning a promise

src/lib/zkapp.ts
```diff
@@ -655,12 +687,12 @@ class SmartContract extends SmartContractBase {
...
    * let tx = await Mina.transaction(sender, async () => {
    *   AccountUpdate.fundNewAccount(sender);
-   *   zkapp.deploy();
+   *   await zkapp.deploy();
    * });
    * tx.sign([senderKey, zkAppKey]);
    * ```
    */
-  deploy({
+  async deploy({
```
mitschabaude, line 695: `deploy()` needs to be async because it calls `init()`, which can be a method so might be async

src/lib/proof-system.ts
mitschabaude, line 1: ZKProgram was already supporting async methods at runtime, but not at the type level. This here is the change that forces the method type to be async.

Trivo25 (approved):
Awesome! We should definitely make sure all the tutorial are up to date as soon as we release this. Happy to help

PR #1480, opened 2024-03-05
Change `Transaction/PendingTransaction` behaviors

MartinMinkov:
# Description

Change the default behavior of `Transaction.send()` to throw an error if any errors are detected during processing. Additionally, we change `isSuccess` from `Transaction` and use `status` instead to indicate the transaction's current status. `status` can either be `pending` or `rejected`.

Also changes the tests and documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Breaking changes
 
+- Improved on parity fix of `Mina.Localblockchain` and`Mina.Network` to make `.send()` and `.wait()` throw an error by default if the transaction was not successful. https://github.com/o1-labs/o1js/pull/1480
+  - Changed `Transaction.isSuccess` to `Transaction.status` to better represent the state of a transaction.
+  - `Transaction.safeSend()` and `PendingTransaction.safeWait()` are introduced to return a `IncludedTransaction` or `RejectedTransaction` object without throwing errors.
 - Fixed parity between `Mina.LocalBlockchain` and `Mina.Network` to have the same behaviors https://github.com/o1-labs/o1js/pull/1422
   - Changed the `TransactionId` type to `Transaction`. Additionally added `PendingTransaction` and `RejectedTransaction` types to better represent the state of a transaction.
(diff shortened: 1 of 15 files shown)


shimkiv:
@MartinMinkov should we maybe also change the https://github.com/o1-labs/o1js/blob/feat/safe-send/src/examples/zkapps/hello-world/run-live.ts in order to check txn status instead of `pendingTx.hash !== undefined`?

MartinMinkov:
> @MartinMinkov should we maybe also change the https://github.com/o1-labs/o1js/blob/feat/safe-send/src/examples/zkapps/hello-world/run-live.ts in order to check txn status instead of pendingTx.hash !== undefined?

Great callout! I'll run through all the examples for usage of `.hash` to change that. We'll need to change the [zkapp-cli PR](https://github.com/o1-labs/zkapp-cli/pull/595/files#diff-22583e12bccc8c870434c21b89d740d4f716b54b2dd81bd22214dac95bb23903R85) you made as well, if you don't mind!

shimkiv:
Thank you, if you don't have time I can take care of that zkapp-cli PR once o1js released, no worries.

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Breaking changes
 
+- Improved on parity fix of `Mina.Localblockchain` and`Mina.Network` to make `.send()` and `.wait()` throw an error by default if the transaction was not successful. https://github.com/o1-labs/o1js/pull/1480
+  - Changed `Transaction.isSuccess` to `Transaction.status` to better represent the state of a transaction.
+  - `Transaction.safeSend()` and `PendingTransaction.safeWait()` are introduced to return a `IncludedTransaction` or `RejectedTransaction` object without throwing errors.
```
mitschabaude, line 24: since both of them are in the same release, it would be nice to collapse that changelog with the changes from #1422 mentioned below

for example not have both
* `.send()` no longer throws an error ...
* `.send()` throws an error by default ...

but just
* `.send()` when using `Mina.Network` is now consistent with `LocalBlockchain` in that it throws an error

mitschabaude (approved):

ymekuria (approved):

CHANGELOG.md
MartinMinkov: great point! done!

PR #1483, opened 2024-03-06
feat(group.ts): add static method `toInput` for Group instance

MartinMinkov:


mitschabaude:
Ultimately it's the fault of the generic provable types that they just assume there's a toInput

PR #1485, opened 2024-03-06
TS core gadget cleanup and implementations

mitschabaude:
- Review core gadgets for security
- Improve documentation of some range check gadgets
- Change `ofField()` to `fromField()`
- Add TS implementations of most of the logic that creates generic gates, to improve auditability and visibility of how the DSL works
- Remove some unused bindings
- Replace bindings to bool gadgets with (trivial) TS implementations

bindings: https://github.com/o1-labs/o1js-bindings/pull/254


jackryanservia (approved):
This is awesome! :)

I'm not sure how, but I didn't realize that the witness lives in OCaml until now!

PR #1493, opened 2024-03-11
Refactor pickles compile

Trivo25:
refactors `pickles_compile` to witness proofs differently and enable explicit proof witnessing

companion https://github.com/o1-labs/o1js-bindings/pull/255

one of the prerequisites for #661


Trivo25:
Will be superseded

PR #1499, opened 2024-03-12
Add method to derive private keys outside the expected domain range

MartinMinkov:
# Description

Adds a method to derive public keys from private keys generated from the normal private key range. This sort of behaviour was allowed by the old [client_sdk](https://www.npmjs.com/package/@o1labs/client-sdk), but new checks added by [mina-signer](https://www.npmjs.com/package/mina-signer) disallowed this action. To offer a backwards-compatible way to derive old private keys, we add a new function which takes the mod of the private key to ensure it's within the domain

# Testing Methods

Wrote a little script which used a test keypair which worked with the old client_sdk

```js
import { Client } from "mina-signer";
import * as CodaSDK from "@o1labs/client-sdk";

const client = new Client({
  network: "mainnet",
});

const sk = "..."; // old private key
const pk = "B62qqraqohhfoo8zAbv8UGkZ758mrBRQLw7aKcQuYEd8GfwaqfSXUym"; // corresponding public key used as a test

const oldPk = CodaSDK.derivePublicKey(sk);
const newPk = client.derivePublicKeyUnsafe(sk);

console.log(pk);
console.log(oldPk);
console.log(newPk);
console.log(oldPk === newPk);
```


src/provable/curve-bigint.ts
```diff
@@ -143,4 +143,30 @@ const PrivateKey = {
...
+ * This is done to keep backwards compatibility with the previous version of the [client_sdk](https://www.npmjs.com/package/@o1labs/client-sdk), which did the same thing when converting a private key to base58.
+ * @param keyBase58 - The private key that is out of the domain of the Pallas curve
+ * @returns The private key that is in the domain of the Pallas curve
+ * @remarks
+ * This function is particularly useful when migrating old keys to be used by the current [mina-signer](https://www.npmjs.com/package/mina-signer) library,
+ * which may reject keys that do not fit the domain of the Pallas curve.
+ * By performing a modulus operation on the key, it ensures that keys
+ * from the older client_sdk can be made compatible.
+ */
+function convertPrivateKeyToBase58WithMod(keyBase58: string): string {
```
mitschabaude, line 168: I think we should also expose this function on `Client`, because people need to use it when creating signatures with their out of domain private keys

src/mina-signer/mina-signer.ts
```diff
@@ -99,6 +99,30 @@ class Client {
     return PublicKey.toBase58(publicKey);
   }
 
+  /**
+   * Derives the public key corresponding to a given private key, ensuring compatibility with keys from the older [client_sdk](https://www.npmjs.com/package/@o1labs/client-sdk) libraries by converting the private key into the domain of the Pallas curve if necessary.
```
mitschabaude, line 103: hey @MartinMinkov I think it was other tools, not our own client sdk, that created out of domain keys, so would reword this comment to definitely not imply that

mitschabaude: the old client sdk just accepted them

src/provable/curve-bigint.ts
MartinMinkov: sounds good, done!

src/mina-signer/mina-signer.ts
MartinMinkov: good point, fixed!

mitschabaude (approved):

PR #1500, opened 2024-03-12
Simplify bindings import

mitschabaude:
Remove the proxying of bindings imports which was not necessary ever since we got rid of `isReady`

The hope is also to reduce the perceived complexity of the code base.

bindings: https://github.com/o1-labs/o1js-bindings/pull/257


MartinMinkov (approved):
love to see this :tada:

PR #1507, opened 2024-03-14
Refactor provable core

### LINKED ISSUE #1504: Fix dependency cycles between field and core gadgets

mitschabaude:
The general goal here is to make the structure of core provable code more logical.

closes #1504

- get rid of several import cycles surrounding `Field`, by introducing a low-level stub module `field-constructor.ts`
- move `FieldVar` into its own lower level module
- define the one and only internal version of `exists()`
- Fix vulnerability: missing boolean check in `Field.isEven()`
- move `Provable<T>` to its own file
  - Simplify doccomments on `Provable<T>` https://github.com/o1-labs/o1js/pull/1507/commits/48b897ce3b775638a94f933c32da15da26adf071
  - Make `Provable<T>` a type instead of interace, because we don't use interfaces in general and they behave slightly different than types in subtle ways https://github.com/o1-labs/o1js/pull/1507/commits/82fe43bb505e5254d583c23fd25f641c1505569c
  - also make other interfaces types, for consistency
- move some provable types into provable-types/
- move some provable types logic from bindings to provable-types/
- move some logic mostly used by mina-signer into mina-signer
- break up `circuit-value.ts` into separate files for `CircuitValue`, `Struct` and `Unconstrained`

bindings: https://github.com/o1-labs/o1js-bindings/pull/258


src/mina-signer/src/derivers-bigint.ts
mitschabaude, line 1: this was moved here from `bindings/lib/provabe-bigint`

src/lib/provable-types/provable-derivers.ts
mitschabaude, line 1: this was moved here from `bindings/lib/provabe-snarky`

src/lib/provable-types/circuit-value.ts
mitschabaude, line 1: this code is the `CircuitValue` part of what used to be `src/lib/circuit-value.ts`

src/lib/provable-core/fieldvar.ts
mitschabaude, line 1: this whole code used to be in `field.ts`

src/lib/provable-types/provable-intf.ts
mitschabaude, line 1: these interfaces used to be `snarky.d.ts`

src/lib/provable-types/unconstrained.ts
mitschabaude, line 1: this code is the `Unconstrained` part of what used to be `src/lib/circuit-value.ts`

src/lib/provable-core/exists.ts
mitschabaude, line 1: apart from `existsAsync()`, this now also has the `exists()` function we use internally a lot and that `Provable.witness` uses

src/lib/provable-types/unconstrained.ts
ymekuria: Grouping things into the `provable-types` folder and creating the `unconstrained` file makes things easier to understand.

ymekuria (approved):

src/lib/field.unit-test.ts
```diff
@@ -1,12 +1,12 @@
-import { ProvablePure } from '../snarky.js';
 import { Field } from './core.js';
-import { Field as Fp } from '../provable/field-bigint.js';
+import { Fp } from '../bindings/crypto/finite-field.js';
```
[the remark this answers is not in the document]
mitschabaude: Good question! The `Field` module from `field-bigint.ts` is a wrapper around the `Fp` module which was always in `bindings/crypto`. `field-bigint.ts` adds other stuff to it like `toBytes` and `toJSON` methods which have nothing to do with the core finite field arithmetic. It's basically the version of `Field` used by `mina-signer`.

It's hard to track dependencies when logic is wrapped and reexported multiple times. I wanted to make it more explicit where this `Fp` (bigint field arithmetic) logic actually comes from, and remove the dependency on the mina-signer middle-man code where not necessary.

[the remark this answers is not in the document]
ymekuria: Ok great. That logic makes a lot of sense.

PR #1511, opened 2024-03-19
Fix build and run on Windows (+ additional benchmarks CI target platform).

### LINKED ISSUE #1502: Support benchmarks on Windows
https://github.com/o1-labs/o1js/pull/1498

shimkiv:
Although not sure if we need to benchmark on all 3 platforms (+ additional ARM arch in future).
Closes #1502

### DIFF
--- a/src/build/build-example.js
+++ b/src/build/build-example.js
@@ -10,13 +10,9 @@ async function buildAndImport(srcPath, { keepFile = false }) {
   let absPath = await build(srcPath);
   let importedModule;
   try {
-    let absPathForImport = absPath
-    if (platform === 'win32') {
-      absPathForImport = 'file:///' + absPathForImport;
-    }
-    importedModule = await import(absPathForImport);
+    importedModule = await import(absPath);
   } finally {
-    if (!keepFile) await fs.unlink(absPath);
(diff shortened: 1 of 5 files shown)


ymekuria (approved):

src/build/build-example.js
```diff
@@ -2,14 +2,19 @@ import fs from 'fs/promises';
...
 export { buildAndImport, build, buildOne };
 
 async function buildAndImport(srcPath, { keepFile = false }) {
   let absPath = await build(srcPath);
   let importedModule;
   try {
-    importedModule = await import(absPath);
+    let absPathForImport = absPath
+    if (platform === 'win32') {
+      absPathForImport = 'file:///' + absPathForImport;
```
mitschabaude, line 15: Feels like the path should already be fixed in build() and buildOne()

src/build/run.js
```diff
@@ -18,6 +19,9 @@ npx snarky-run [file]`);
 if (!bundle) {
   let absPath = await buildOne(filePath);
   console.log(`running ${absPath}`);
+  if (platform === 'win32') {
+    absPath = 'file:///' + absPath;
```
mitschabaude, line 23: Feels like the path should already be fixed in build() and buildOne()

PR #1523, opened 2024-03-26
Efficient comparisons

### LINKED ISSUE #1434: Make relative Field comparisons efficient
`x.assertLessThan(y)` currently takes 509 constraints.

By using the multi-range check and FFadd gates it should be doable in 20 constraints:

* convert x and y to `ForeignField` (3 limbs) representation. 2 x 1 double generic gate = **2**
* multi-range check x and y. 2 x 4 MRC gates = **8**
* assert $y < p$ (canonical). 4 MRC gates + 1 FFadd = **5**
* bigint subtraction $z = y - x - 1$ and MRC z. 4 MRC + 1 FFadd = **5**
  * This proves that the subtraction doesn't underflow < 0, so $x + 1 \le y$  
  * also, proves that x is the canonical representation, because $x < y < p$

mitschabaude:
closes #1434

- Implement foreign field comparison assertions on two variables
- Use those to implement full-field comparisons which support 255-bit field elements
  - constraints **reduced by up to 20x**
- Unify small field comparisons by defining generic versions that work for any _range check_ (which is an input).
  - These are now used for all uints and more efficient than the previous boolean comparisons
- Reimplement `isEven()` and `isOdd()` using full field comparison, which makes them much cheaper and also handle 255-bit field elements
  - very nice effect on signature methods, which use `isEven()` for converting between compressed and uncompressed public keys
- To be able to use foreign field gadgets inside the `Field` class, and prevent against import cycle problems in the future, I got rid of the dependency on `Field` in many of the gadgets and a few other places


src/lib/provable/crypto/signature.ts
mitschabaude, line 1: here we use the new efficient `isOdd()`, plus some boyscouting

src/lib/provable/field.ts
```diff
@@ -308,42 +313,46 @@ class Field {
...
+
+    // witness a bit b such that x = b + 2z for some z <= (p-1)/2
+    // this is always possible, and unique _except_ in the edge case where x = 0 = 0 + 2*0 = 1 + 2*(p-1)/2
+    // so we can compute isOdd = b AND (x != 0)
+    let [b, z] = exists(2, () => {
+      let x = this.toBigInt();
+      return [x & 1n, x >> 1n];
+    });
+    let isOdd = b.assertBool();
+    z.assertLessThan((Field.ORDER + 1n) / 2n);
```
mitschabaude, line 332: using the new full field comparison!

src/lib/provable/field.ts
```diff
@@ -560,17 +569,14 @@ class Field {
    *
    * @example
    * ```ts
-   * Field(2).lessThan(3).assertEquals(Bool(true));
+   * let isTrue = Field(2).lessThan(3);
```
mitschabaude, line 572: some comments boyscouting in this file. I didn't like that the example for `lessThan()` was one that should actually have used `assertLessThan()`

src/lib/provable/field.ts
```diff
@@ -670,14 +667,12 @@ class Field {
 
   /**
    * Assert that this {@link Field} is less than another "field-like" value.
-   * Calling this function is equivalent to `Field(...).lessThan(...).assertEquals(Bool(true))`.
+   *
+   * Note: This uses fewer constraints than `x.lessThan(y).assertTrue()`.
```
mitschabaude, line 671: didn't like that this said it "was equivalent", people should use the assertion version if possible because they are cheaper

src/lib/provable/gadgets/basic.ts
mitschabaude, line 1: removing the import cycles was mainly about importing `Field` only as a type, and using what we need from `field-constructor.ts` instead

src/lib/provable/gadgets/comparison.ts
mitschabaude, line 1: this file contains the meat

src/lib/provable/gadgets/foreign-field.ts
```diff
@@ -720,3 +709,86 @@ class Sum {
...
+// helpers
+
+/**
+ * Version of `multiRangeCheck` which does the check on a truncated version of the input,
+ * so that it always succeeds, and then checks equality of the truncated and full input.
+ *
+ * This is a hack to get an error when the constraint fails, around the fact that multiRangeCheck
+ * is not checked by snarky.
+ */
+function indirectMultiRangeChange(
```
mitschabaude, line 782: added this to not make the (negative) unit tests break. it's also better DX

src/lib/provable/int.ts
```diff
@@ -401,56 +408,55 @@ class UInt64 extends CircuitValue {
...
-        yMinusX
-      );
-
-      xMinusYFits.or(yMinusXFits).assertEquals(true);
-      // x <= y if y - x fits in 64 bits
-      return yMinusXFits;
     }
+    return lessThanOrEqualGeneric(this.value, y.value, 1n << 64n, (v) =>
+      RangeCheck.rangeCheckN(UInt64.NUM_BITS, v)
+    );
```
mitschabaude, line 414: this used to be less efficient because it called `rangeCheckHelper()` twice. the new version only calls it once

src/lib/provable/provable.ts
```diff
@@ -17,12 +17,11 @@ import {
 import {
   inCheckedComputation,
   inProver,
-  snarkContext,
   asProver,
   constraintSystem,
   generateWitness,
 } from './core/provable-context.js';
-import { exists, existsAsync } from './core/exists.js';
+import { witness, witnessAsync } from './types/witness.js';
```
mitschabaude, line 24: moved `witness()` to its own file to make it not depend on `Field`

src/lib/provable/field.ts
```diff
@@ -798,12 +786,14 @@ class Field {
...
-        }
-        return;
+        assert(
+          x === 0n || x === 1n,
+          `Field.assertBool(): expected ${x} to be 0 or 1`
+        );
+        return new Bool(x === 1n);
       }
       assertBoolean(this);
+      return Bool.Unsafe.fromField(this);
```
mitschabaude, line 796: I made `assertBool()` return the Bool because this is a common use case: assert that some witnessed field element is boolean and then use the bool

src/lib/provable/test/foreign-field-gadgets.unit-test.ts
```diff
@@ -92,7 +92,8 @@ for (let F of fields) {
   let big264 = unreducedForeignField(264, F); // this is the max size supported by our range checks / ffadd
   let big258 = unreducedForeignField(258, F); // rough max size supported by ffmul
 
-  equivalentProvable({ from: [big264, big264], to: big264 })(
+  // addition can fail on two unreduced inputs because we can get x + y - f > 2^264
+  equivalentProvable({ from: [big264, f], to: big264 })(
```
mitschabaude, line 96: we didn't catch this before because of the missing errors on multi range check failures 😬

CHANGELOG.md
```diff
@@ -30,6 +30,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - Remove `CircuitValue`, `prop`, `arrayProp` and `matrixProp` https://github.com/o1-labs/o1js/pull/1507
   - Remove `Mina.accountCreationFee()`, `Mina.BerkeleyQANet`, all APIs which accept private keys for feepayers, `Token`, `AccountUpdate.tokenSymbol`, `SmartContract.{token, setValue, setPermissions}`, "assert" methods for preconditions, `MerkleTee.calculateRootSlow()`, `Scalar.fromBigInt()`, `UInt64.lt()` and friends, deprecated static methods on `Group`, utility methods on `Circuit` like `Circuit.if()`, `Field.isZero()`, `isReady` and `shutdown()` https://github.com/o1-labs/o1js/pull/1515
 - Remove `privateKey` from the accepted arguments of `SmartContract.deploy()` https://github.com/o1-labs/o1js/pull/1515
+- **Efficient comparisons**. Support arbitrary bit lengths for `Field` comparisons and massively reduce their constraints https://github.com/o1-labs/o1js/pull/1523
+  - `Field.assertLessThan()` goes from 510 to 24 constraints, `Field.lessThan()` from 509 to 38
```
Trivo25, line 34: nice!

CHANGELOG.md
```diff
@@ -30,6 +30,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+  - Moderately improve other comparisons: `UInt64.assertLessThan()` from 27 to 14,, `UInt64.lessThan()` from 27 to 15, `UInt32` similar.
```
Trivo25, line 35: double `,,`

Trivo25 (approved):

PR #1530, opened 2024-04-01
Efficient scalar mul and other Scalar improvements

### LINKED ISSUE #980: rework `Group.scale` to accept a non-shifted scalar and values of `0` and `1`
Currently, `Group.scale` only takes shifted scalar values -- it would be nice to add an implementation that works with non-shifted values. Additionally, `scale` does not accept `0` or `1` (`g * 1 = g` and `g * 0 = zero`) but throws an error instead.

### LINKED ISSUE #984: add an unsafe `Group.add` with less constraints that does not handle zero

### LINKED ISSUE #1440: Group.add() throws an asProver error when an input is the zero point

mitschabaude:
This revises the scalar multiplication gadgets and internal representation of `Scalar`.

BEFORE:
* `Scalar` used to be represented as 255 bits, which imposed an unavoidable cost of 255 rows on any gadget that wanted to scale a group element by a `Field`. This is a very typical use case for scaling, see `Signature.verify()` and `Nullifier.verify()`
* On top of that, the 255-bit representation was shifted and we lacked a good way of handling that. As a hack, we used a `scaleShifted()` helper which unshifts "in the exponent" at the cost of doing 3 full scalar muls instead of 1.
* `Group.scale()` didn't support -1, 0 and 1 as scalars

AFTER:
* `Scalar` uses the representation that Pickles uses as well and which our scalar mul gadgets were designed for: 1 low bit and 254 high bits. The high bits are a single field element. The representation is still shifted, but in a slightly different way than before.
* We add efficient gadgets for scaling by a `Field`, and for converting a `Field` to a `Scalar`
* We no longer expose shifting in any API: `Scalar.fromBits()` performs the shift behind the scenes in provable code now, as do `Group.scale()` and `Scalar.fromField()`
* `Group.scale()` supports all scalars, and there's a new test to check that
* Some results:
  * `Signature.verify()` now uses 350 instead of 800 rows
  * The nullifier example zkapp uses 9790 instead of 10774 rows

Closes #1440
Closes #984 
Closes #980


src/lib/provable/field.ts
```diff
@@ -319,24 +320,7 @@ class Field {
    * See {@link Field.isEven} for examples.
    */
   isOdd() {
-    if (this.isConstant()) return new Bool((this.toBigInt() & 1n) === 1n);
```
mitschabaude, line 322: I moved this gadget into `gadgets/comparison.ts` to reuse it for `scaleField()`

src/lib/provable/gadgets/native-curve.ts
mitschabaude, line 1: this is where they new scaling gadgets live! needs careful crypto review

src/lib/provable/group.ts
```diff
@@ -101,67 +106,49 @@ class Group {
         return fromProjective(g_proj);
       }
     } else {
-      const { x: x1, y: y1 } = this;
```
mitschabaude, line 104: this code was moved to `native-group.ts` to reuse there. it was also rewritten a bit and a bug fixed

src/lib/provable/group.ts
```diff
@@ -101,67 +106,49 @@ class Group {
...
+  /**
+   * Lower-level variant of {@link add} which doesn't handle the case where one of the operands is zero, and
+   * asserts that the output is non-zero.
+   *
+   * Optionally, zero outputs can be allowed by setting `allowZeroOutput` to `true`.
+   *
+   * **Warning**: If one of the inputs is zero, the result will be garbage and the proof useless.
+   * This case has to be prevented or handled separately by the caller of this method.
+   */
+  addNonZero(g2: Group, allowZeroOutput = false): Group {
```
mitschabaude, line 136: I added this, I think it's actually a bit more common to use this than fully complete addition

src/lib/provable/test/foreign-field.unit-test.ts
```diff
@@ -108,83 +102,3 @@ equivalent({ from: [f], to: f })(
     return ForeignScalar.fromBits(bits);
   }
 );
-
-// scalar shift in foreign field arithmetic vs in the exponent
```
mitschabaude, line 112: this test was no longer valid, so I removed it

src/lib/provable/test/test-utils.ts
```diff
@@ -62,6 +66,26 @@ function bytes(length: number) {
   });
 }
 
+function pointSpec<T>(field: ProvableSpec<bigint, T>, Curve: CurveAffine) {
```
mitschabaude, line 69: just a testing utility, ends up unused (I used it during development to monitor the size of some circuits)

src/snarky.d.ts
```diff
@@ -347,7 +347,19 @@ declare const Snarky: {
...
-    scale(p: MlGroup, s: MlArray<BoolVar>): MlGroup;
+    /**
+     * Computes `(2*s + 1 + 2^numBits) * P` and also returns the bits of s (which are proven correct).
+     *
+     * `numBits` must be a multiple of 5, and s must be in the range [0, 2^numBits).
+     * The [soundness proof](https://github.com/zcash/zcash/issues/3924) assumes
+     * `numBits <= n - 2` where `n` is the bit length of the scalar field.
+     * In our case, n=255 so numBits <= 253.
+     */
+    scaleFastUnpack(
```
mitschabaude, line 358: this is the scaling gadget that Pickles uses as well

tests/vk-regression/plain-constraint-system.ts
```diff
@@ -173,6 +174,16 @@ const BasicCS = constraintSystem('Basic', {
   },
 });
 
+const CryptoCS = constraintSystem('Crypto', {
```
mitschabaude, line 177: new vk regression test

src/lib/provable/gadgets/scalar.ts
```diff
@@ -0,0 +1,114 @@
...
+  // => split t0 into [5, 83]
+  let tLo = exists(5, () => {
+    let t = t0.toBigInt();
+    return [bit(t, 0), bit(t, 1), bit(t, 2), bit(t, 3), bit(t, 4)];
+  });
+  let tLoBools = TupleN.map(tLo, (x) => x.assertBool());
+  let tHi0 = existsOne(() => t0.toBigInt() >> 5n);
+
+  // prove split
+  // since we know that t0 < 2^88, this proves that t0High < 2^83
```
querolita: tHi0 you mean?

src/lib/provable/gadgets/scalar.ts
```diff
@@ -0,0 +1,114 @@
...
+    .add(t2.mul(1n << (2n * l - 5n)))
+    .seal();
+
+  // return (t + 2^254)*P = (s - 2^254 + 2^254)*P = s*P
+  return scaleShiftedSplit5(P, tHi, tLoBools);
+}
+
+/**
+ * Internal helper to compute `(t + 2^254)*P`.
+ * `t` is expected to be split into 250 high bits (t >> 5) and 5 low bits (t & 0xf1).
```
querolita: 0xf1 = 241, where does it come from?

src/lib/provable/gadgets/scalar.ts
```diff
@@ -61,7 +61,7 @@ function scale(P: { x: Field; y: Field }, s: Field): Group {
 
 /**
  * Internal helper to compute `(t + 2^254)*P`.
- * `t` is expected to be split into 250 high bits (t >> 5) and 5 low bits (t & 0xf1).
+ * `t` is expected to be split into 250 high bits (t >> 5) and 5 low bits (t & 0x1f).
```
querolita: ah okay, that's now 5 bits

src/lib/provable/gadgets/scalar.ts
mitschabaude: This code should no longer exist 🤔 
Anyway, this was supposed to mean 0x1f, which is a mask for the lowest 5 bits

src/lib/provable/gadgets/scalar.ts
mitschabaude: I'm afraid you're reviewing outdated code :O 
yeah I meant tHi0

querolita: yeah, yesterday I was commit-reviewing to gain more context. Today I will focus on the updated `native-curve.ts` file

src/lib/provable/gadgets/native-curve.ts
```diff
@@ -0,0 +1,313 @@
...
+  let [tLo, tHi00, tHi01] = exists(3, () => {
+    let t = t0.toBigInt();
+    return [bit(t, 0), bitSlice(t, 1, 64), bitSlice(t, 65, 23)];
+  });
+  let tLoBool = tLo.assertBool();
+  rangeCheck64(tHi00);
+  rangeCheck64(tHi01);
+
+  // prove (tLo, tHi0) split
+  // since we know that t0 < 2^88 and tHi0 < 2^128, this even proves that t0Hi < 2^87
```
querolita, line 239: isn't it `tHi0` in the comment?

querolita (approved):
Very clear comments to understand the technique for an outsider. Couldn't find anything fishy after review.

src/lib/provable/gadgets/native-curve.ts
mitschabaude: ah, right!

PR #1543, opened 2024-04-09
Release 0.18.0

mitschabaude:
- changelog
- 0.18.0
- changelog


mitschabaude:
note: I've triggered ci against berkeley to check for compatibility against the berkeley node

shimkiv (approved):
If devnet HF will still operate with artifacts for the testnets (e.g. built with `devnet` Dune profile).

Issue #1547, opened 2024-04-09
implicitly prove txs


harrysolovay:
Are there use cases in which the creation of a transaction, targeting a non-test chain, is **not** to be `prove`n? If so, the need to explicitly call `prove` makes sense. Otherwise, perhaps the `prove` method can happen implicitly in the initial `transaction` creation call?

Some additional thoughts:

## Backward Compatibility

This could be implemented such that any subsequent `prove` calls return the initial (cached) result. Would this preserve backward compatibility?

```ts
const unsent = Mina.transaction(sender, async () => {
  // `prove` is implicitly called
})
// the following `prove` call is a noop
await unsent.prove()
```

## Top-level Prove Toggle

Meanwhile, the proving could be disabled by the top-level `proofsEnabled` field in the case of the in-mem test chain.

```ts
let Local = Mina.LocalBlockchain({ proofsEnabled: false })
Mina.setActiveInstance(Local)

const unsent = Mina.transaction(sender, async () => {
  // `prove` is NOT called
})
```

## Pending Proof Access

One could access the pending proof state on the pending transaction creation.

```ts
const unsent = Mina.transaction(sender, async () => {
  // ...
}) // `CreateTransactionPromise`

const proof = await unsent.proofPromise // Promise<(Proof<ZkappPublicInput, undefined> | undefined)[]>
```

mitschabaude:
I think I'm ultimately opposed to this proposal.
Some objections:
* Proving is a very slow and clunky operation. I feel like operations like that should generally be triggered explicitly.
* When I debug my transactions, including errors when proving them, I usually log the transaction after it was constructed, but before proving, so that I can double-check that it's valid. This would no longer be possible when proving happened inside `transaction()`
* It's not always the ideal architecture to prove transactions right in the process where they are created by a user. In the future, we want to enable outsourcing the proving to a server / cloud service. It therefore feels premature to entangle the proving step with the step of calling the contract / computing the proof inputs.

harrysolovay:
Makes sense complete sense.

Issue #1514, opened 2024-03-25
`Transaction.prove` and `Transaction.send` result method chaining


harrysolovay:
Hi o1js community and maintainers. I'm new to this library + zkApps in general, so please take my thoughts with a grain of salt.

I'm contemplating the following code.

```ts
let tx = await Mina.transaction(id, () => { /* ... */ })
await tx.prove()
let pending = await tx
  .sign([key])
  .send()
await pending.wait()
```

Could we enable a fully-chained experience?

```ts
await Mina
  .transaction(id, () => { /* ... */ })
  .prove()
  .sign([key])
  .send()
  .wait()
```

I suppose `prove` and `send` would need to return promise subtypes (`ProveResult` and `SendResult` respectively).

```ts
type ProveResult = Promise<(Proof<ZkappPublicInput, undefined> | undefined)[]> & {
  sign: (additionalKeys?: PrivateKey[] | undefined) => Promise<Transaction> & {
    send(): SendResult
  }
}

type SendResult = Promise<Mina.PendingTransaction> & {
  wait: (options?: {
    maxAttempts?: number | undefined;
    interval?: number | undefined;
  } | undefined) => Promise<Mina.IncludedTransaction>
}
```

Some follow-up thoughts + questions:

- this would be non-breaking
- are there reasons to keep these steps non-chainable?
- a given promise subtype could contain a reference to the step/promise off of which it was chained, incase a direct reference is required

mitschabaude:
to me this seems to add complexity without a clear benefit. With `prettier` formatting your non-chained example even uses less lines of code than the chained one

harrysolovay:
Apologies for not clearly articulating the benefits:

- 1 `await` instead of 4 `await`s.
- No need to define (and therefore come up with) variable names for (A) the tx and (B) the pending counterpart. This would be especially useful in scopes that create/prove/sign/send/wait multiple txs, as devs wouldn't need to spend time considering names for variables that are not subsequently referenced.
- Legibility gains

This last benefit I listed is subjective... but I do believe the chained/fluent equivalent would be––for most developers––more legible/preferable.

I can close this issue if you still disagree. Thank you for hearing me out + considering this approach.

harrysolovay:
That utility would work great! Alternatively, we could create dedicated promise subclasses, as to avoid bloating the return type (the intersection of `Promise<T>` with the mapped type might be a bit intimidating).

```ts
[...]

Issue #1221, opened 2023-11-02
transaction.sign produces an invalid signature after import from JSON


dfstio:
In case the transaction is exported to JSON, then imported from JSON, signed after import, and sent, the “Check signature: Invalid signature” error is thrown, and the authorization field of the transaction is undefined.

In the case of exporting already signed transaction, everything is working as expected.

The code to reproduce the error:
```typescript
import { describe, expect, it } from "@jest/globals";
import fs from "fs/promises";
import {
  AccountUpdate,
  PrivateKey,
  Mina,
  PublicKey,
  UInt64,
  Types,
} from "o1js";

jest.setTimeout(1000 * 60 * 60 * 1); // 1 hour
const transactionFee = 150_000_000;
let senderPrivateKey: PrivateKey | undefined = undefined;
let senderPublicKey: PublicKey | undefined = undefined;

beforeAll(async () => {
  const Local = Mina.LocalBlockchain({ proofsEnabled: true });
  Mina.setActiveInstance(Local);
  const { privateKey } = Local.testAccounts[0];
  senderPrivateKey = privateKey;
  senderPublicKey = senderPrivateKey.toPublicKey();
  expect(senderPublicKey).not.toBeUndefined();
  expect(senderPrivateKey).not.toBeUndefined();
});

describe("Sign, export, and import transaction", () => {
  it("should sign and export transaction", async () => {
    if (senderPublicKey === undefined || senderPrivateKey === undefined) return;
    const sender: PublicKey = senderPublicKey;
    const transaction = await Mina.transaction(
      { sender, fee: transactionFee },
      () => {
        AccountUpdate.fundNewAccount(sender);
        const senderUpdate = AccountUpdate.create(sender);
        senderUpdate.requireSignature();
        senderUpdate.send({
          to: PrivateKey.random().toPublicKey(),
          amount: UInt64.from(1_000_000_000n),
        });
      }
    );
    // Sign BEFORE exporting
    transaction.sign([senderPrivateKey]);
    await fs.writeFile("./json/tx-signed.json", transaction.toJSON());
  });

  it("should send a signed transaction", async () => {
    const transaction: Mina.Transaction = Mina.Transaction.fromJSON(
      JSON.parse(
        await fs.readFile("./json/tx-signed.json", "utf8")
      ) as Types.Json.ZkappCommand
    ) as Mina.Transaction;
    console.log("transaction signed before export:", transaction.toPretty());
    const tx = await transaction.send();
[...]

PR #1548, opened 2024-04-09
chaining `sign`/`send`/`wait` off `transaction` result

harrysolovay:
Partially addresses #1514 (does not touch `prove` / see #1547).

This PR enables chaining `sign`, `send` and `wait` calls off of the result of `transaction`.

One could express this...

```ts
const tx = await Mina.transaction(sender, async () => {
  // ...
});
tx.sign([...keys]);
const txPending = await tx.send();
await txPending.wait();
```

... like so:

```ts
await Mina.transaction(sender, async () => {
  // ...
})
  .sign([...keys])
  .send()
  .wait();
```


src/examples/chaining-tx-methods.ts
```diff
@@ -0,0 +1,61 @@
...
+let initialBalance = 10_000_000_000;
+let zkapp = new SimpleZkapp(zkappAddress);
+await SimpleZkapp.analyzeMethods();
+
+const deployTx = Mina.transaction(sender, async () => {
+  let senderUpdate = AccountUpdate.fundNewAccount(sender);
+  senderUpdate.send({ to: zkappAddress, amount: initialBalance });
+  await zkapp.deploy();
+}).sign([senderKey, zkappKey]);
+await deployTx.then((v) => v.prove());
```
harrysolovay, line 49: Was seeing errors throw from within the ocaml bindings whenever I didn't call `prove`, which is strange considering proofs are disabled in this example. Nevertheless, left it in the long-form. Will adjust this depending on whether we tackle #1547.

mitschabaude: but you can make the example nicer by removing `@method` from `init()`, then this tx won't require a proof

src/lib/mina/transaction.ts
```diff
@@ -276,6 +280,40 @@ type RejectedTransaction = Pick<
   errors: string[];
 };
 
+interface TransactionPromise extends Promise<Transaction> {
+  sign(...args: Parameters<Transaction['sign']>): TransactionPromise;
+  send(): PendingTransactionPromise;
+}
```
mitschabaude, line 286: could you do this with types, not interfaces? like `Promise<Transaction> & ...`? I'd prefer that because we use types for everything else and we understand them well

src/lib/mina/transaction.ts
```diff
@@ -276,6 +280,40 @@ type RejectedTransaction = Pick<
...
+
+function toTransactionPromise(
+  getPromise: () => Promise<Transaction>
+): TransactionPromise {
+  const pending = getPromise().then();
```
mitschabaude, line 291: why are you doing `.then()` here?

mitschabaude (commented):
looks mostly good to me, let's also get another independent review of the API

src/lib/mina/transaction.ts
harrysolovay: Incase the `getPromise` returns a promise that was defined out of the scope of the `getPromise` fn, this ensures that we don't mutate / attach the methods to that original promise. Given how this code uses `toTransactionPromise`/`toPendingTransactionPromise`, forgoing the `then` wouldn't problematic. Still may be a worthwhile safeguard. Should I remove?

mitschabaude: aah I get it now. don't mutate your input values, excellent practice 👍🏻 since the overhead is negligible, leave it in!

src/lib/mina/transaction.ts
```diff
@@ -276,6 +280,40 @@ type RejectedTransaction = Pick<
   errors: string[];
 };
 
+type TransactionPromise = Promise<Transaction> & {
```
MartinMinkov, line 283: quick question: just wanted to check, are these new types exposed to the user or would it be helpful to do so? or would it be better to just keep this internal?

harrysolovay: Good thinking! They should indeed be exposed to the user. Just added their reexports.

ymekuria (commented):
I love this. Do you plan on tackling prove as well?

mitschabaude (approved):

harrysolovay:
> Do you plan on tackling prove as well?

Indeed! Will tackle in a subsequent PR.

dfstio:
Can you also make it possible to call .sign several times?
``` typescript
tx.sign([zkAppPrivateKey]);
tx.sign([deployer]);
```

It will allow to call the function that signs the tx without disclosing the privateKey:
``` typescript
tx.sign([zkAppPrivateKey]);
await signWithDeployerKey(tx);
```
I guess that now the code at the addSignature prevents doing it:
https://github.com/o1-labs/o1js/blob/main/src/lib/mina/account-update.ts#L1867

harrysolovay:
Let me see if I understand your use case:

- the `tx` requires multiple signatures
- you want to apply those signatures in different envs (one in the front-end code, and one in the wallet)
- you therefore need to be able to use `sign` multiple times

Is this correct?

dfstio:
Yes, I'm speaking about 
- the tx that requires several signatures (like deploying or creating a token account or any other custom tx)
- I want to be able to apply those different signatures in the different environments:
   - frontend
   - wallet
   - backend
   - two backend environments

Specifically, I'm interested now in being able to run the code at zkCloudWorker AWS instance and provide from the zkCloudWorker to running code the possibility to sign the txs without disclosing the privateKey (as even the dust funds will be stolen sooner or later):
https://github.com/zkcloudworker/zkcloudworker-lib/blob/main/src/cloud/cloud.ts#L72

We've successfully resolved the challenge to serialize txs to move them between the various environments; now the blocker is to be able to sign the same transaction in the different environments:
https://github.com/o1-labs/o1js/issues/1221
https://discord.com/channels/484437221055922177/1228326948078489642

shimkiv:
Folks, despite the very valuable input, can we please `STOP` discussion of future updates within the `closed` PRs/Issues.
I'd strongly advise to create `new` GH issue, copy/paste relevant content and continue there.
Thank you!

harrysolovay:
Just replied in #1582

PR #1553, opened 2024-04-10
Always test against berkeley in CI

mitschabaude:
This makes CI always test against lightnet built from berkeley, to make sure we stay compatible with devnet

### DIFF
--- a/.github/workflows/live-tests.yml
+++ b/.github/workflows/live-tests.yml
@@ -16,7 +16,6 @@ jobs:
   berkeley:
     timeout-minutes: 45
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main' || (github.event_name == 'pull_request')
     services:
       mina-local-network:
         image: o1labs/mina-local-network:berkeley-latest-lightnet
--- a/.github/workflows/live-tests.yml
+++ b/.github/workflows/live-tests.yml
@@ -13,7 +13,7 @@ on:
   workflow_dispatch: {}
 
 jobs:
-  main-branch:
+  berkeley:
     timeout-minutes: 45
     runs-on: ubuntu-latest
     if: github.ref == 'refs/heads/main' || (github.event_name == 'pull_request')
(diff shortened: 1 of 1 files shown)


.github/workflows/live-tests.yml
```diff
@@ -13,36 +13,10 @@ on:
...
-      - name: Use shared steps for live testing jobs
-        uses: ./.github/actions/live-tests-shared
-        with:
-          mina-branch-name: o1js-main
-
-  berkeley-branch:
-    timeout-minutes: 45
-    runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/berkeley' || (github.event_name == 'pull_request' && github.base_ref == 'berkeley')
+    if: github.ref == 'refs/heads/main' || (github.event_name == 'pull_request')
```
shimkiv, line 19: Why condition was changed by removing `&& github.base_ref == 'main'`?
If the goal is to always run the same job then no reason to have any conditions. But if you want to run it on PR against `main` branch then you need the complete condition.

.github/workflows/live-tests.yml
```diff
@@ -64,29 +38,3 @@ jobs:
         uses: ./.github/actions/live-tests-shared
```
shimkiv, line 38: Up to you but if we will only have single job here then no reasons to extract steps into shared actions file.

.github/workflows/live-tests.yml
mitschabaude: Ah ok . I guess the current conditions would run the job on `main`, and on PRs against `main`, `berkeley` and `develop`? That actually seems fine.
But I can also remove all conditions and run it on all 3 branches

.github/workflows/live-tests.yml
mitschabaude: hm ok, but maybe we need different versions again in the future, so let's leave it as is

.github/workflows/live-tests.yml
shimkiv: Without `&& github.base_ref == 'main'` in current state it wil run on any PR

shimkiv (approved):

.github/workflows/live-tests.yml
mitschabaude: @shimkiv I don't think so because the intro has this:
```
  pull_request:
    branches:
      - main
      - berkeley
      - develop
```

Issue #1552, opened 2024-04-10
MerkleMapWitness.computeRootAndKey gives wrong key in o1js 0.18.0


dfstio:
The following test pass when using o1js 0.17.0 and fail when using o1js 0.18.0 with the error:
```
 expect(received).toEqual(expected) // deep equality

    Expected: "2"
    Received: "4"

      11 |     const [, calculatedKey] = witness.computeRootAndKey(Field(0));
      12 |     console.log("calculatedKey", calculatedKey.toJSON());
    > 13 |     expect(calculatedKey.toJSON()).toEqual(key.toJSON());
         |                                    ^
      14 |   });
      15 | });
      16 |

      at Object.<anonymous> (tests/map-key.test.ts:13:36)
```

```typescript
import { describe, expect, it } from "@jest/globals";
import { Field, MerkleMap } from "o1js";

describe("Map key", () => {
  it(`should get witness`, async () => {
    const map = new MerkleMap();
    const key = Field(2);
    const value = map.get(key);
    console.log("value", value.toJSON());
    const witness = map.getWitness(key);
    const [, calculatedKey] = witness.computeRootAndKey(Field(0));
    console.log("calculatedKey", calculatedKey.toJSON());
    expect(calculatedKey.toJSON()).toEqual(key.toJSON());
  });
});

```

PR #1559, opened 2024-04-10
Fix Merkle map keys

mitschabaude:
fixes https://github.com/o1-labs/o1js/issues/1552

the issue was a bit subtle - the previous code assumed that `field.toBits()` returned 255 bits, which it no longer does


dfstio:
Thank you!

MartinMinkov (approved):

PR #1560, opened 2024-04-11
Pass empty hash correctly to iterator 

Trivo25:
previously the user defined `emptyHash` for a `MerkleList` was not correctly passed down to the iterator when calling `list.startIterating();`, which resulsted in the iterator using the default hash of value `0`

### DIFF
--- a/src/lib/provable/merkle-list.ts
+++ b/src/lib/provable/merkle-list.ts
@@ -326,12 +326,17 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
   // fixed parts
   readonly data: Unconstrained<WithHash<T>[]>;
   readonly hash: Field;
+  readonly emptyHash: Field;
 
   // mutable parts
   currentHash: Field;
   currentIndex: Unconstrained<number>;
 
-  constructor(value: MerkleListIteratorBase<T>) {
+  constructor(
+    value: MerkleListIteratorBase<T> & {
+      emptyHash: Field;
+    }
+  ) {
     Object.assign(this, value);
   }
 
@@ -400,6 +405,7 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
       hash: this.hash,
       currentHash: this.currentHash,
       currentIndex,
+      emptyHash,
     });
   }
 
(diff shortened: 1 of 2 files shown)


src/lib/provable/merkle-list.ts
```diff
@@ -468,7 +474,8 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
   static createFromList<T>(merkleList: typeof MerkleList<T>) {
     return this.create<T>(
       merkleList.prototype.innerProvable,
-      merkleList._nextHash
+      merkleList._nextHash,
+      merkleList.emptyHash
```
mitschabaude, line 478: isn't this change all that is needed?

src/lib/provable/merkle-list.ts
```diff
@@ -331,7 +331,11 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
   currentHash: Field;
   currentIndex: Unconstrained<number>;
 
-  constructor(value: MerkleListIteratorBase<T>) {
+  constructor(
+    value: MerkleListIteratorBase<T> & {
+      emptyHash: Field;
+    }
+  ) {
```
mitschabaude, line 338: this doesn't make sense to me, because the empty hash is supposed to be a static property of the Iterator, and not an instance property

src/lib/provable/merkle-list.ts
```diff
@@ -400,6 +404,7 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
       hash: this.hash,
       currentHash: this.currentHash,
       currentIndex,
+      emptyHash,
```
mitschabaude, line 407: doesn't make sense to me

src/lib/provable/merkle-list.ts
```diff
@@ -448,6 +453,7 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
           hash,
           currentHash: hash,
           currentIndex: Unconstrained.from(0),
+          emptyHash: emptyHash_,
```
mitschabaude, line 456: doesn't make sense to me

PR #1562, opened 2024-04-11
Improvements to `MerkleList` and `Iterator` 

Trivo25:
- `startIteratingFromOldest()` to add an iterator which iterates from oldest to newest elements (`n_i` to `n_0`), helpful for reducer-like applications that need to iterate through historical data
- add `MerkleList.fromReverse` creates a `MerkleList` from an array but reverses it internally
- demonstrates how to use an iterator for actions and reducers that reduce historical data


src/lib/provable/merkle-list.ts
```diff
@@ -506,6 +530,200 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
   }
 }
 
+class MerkleListIteratorFromEnd<T> implements MerkleListIteratorBase<T> {
```
mitschabaude, line 533: I expected this to be doable as part of the same iterator class 🤔

Trivo25: yep - done, I wanted to maintain the same API without introducing a new equivalent for `next`, such as `nextReverse`, I don't like the workaround  but lmk what you think, maybe this can also be improved

src/lib/provable/merkle-list.ts
```diff
@@ -233,6 +239,7 @@ class MerkleList<T> implements MerkleListBase<T> {
       }
 
       static from(array: T[]): MerkleList<T> {
+        array = [...array].reverse();
```
mitschabaude, line 242: let's bring the internal representation in line with the outside in a future PR

src/lib/provable/merkle-list.ts
```diff
@@ -335,19 +342,43 @@ class MerkleListIterator<T> implements MerkleListIteratorBase<T> {
     Object.assign(this, value);
   }
 
+  assertAtEnd() {
+    return this.currentHash.assertEquals(this.Constructor.emptyHash);
+  }
+
+  isAtStart() {
+    return this.currentHash.equals(this.hash);
+  }
```
mitschabaude, line 351: `assertAtEnd()` / `isAtEnd()` etc now don't align with `next()` / `previous()`. i.e. we call `isAtEnd()` inside `previous()` to check is we're done. it's the correct implementation but the wrong name: `isAtEnd()` should now be called `isAtStart()`

src/lib/provable/test/merkle-list.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        let l2Hash = l2.hash;
+
+        l1Hash.assertEquals(l2Hash);
+      });
+    });
+
+    describe('pop and previous should behave the same', () => {
+      it('empty hash', () => {
+        let l1 = List.from([Field(1), Field(2)]);
+        let l2 = l1.clone();
```
mitschabaude, line 37: is this cloning necessary 🤔

src/lib/provable/test/merkle-list.test.ts
mitschabaude, line 1: awesome tests!!

mitschabaude (approved):
I think you also have to replace `next()` with `previous()` and `startIterating()` with `startIteratingFromLast()` in all of our library code

src/lib/provable/merkle-list.ts
mitschabaude: we can push this to a follow-up PR though

src/lib/provable/test/merkle-list.test.ts
Trivo25: yea, otherwise `pop` would literally steal the elements form the iterator :D I could have also just initiated a second array, but I thought I would clone it to make sure it actually does clone

PR #1567, opened 2024-04-12
represent whether signed and/or proven within `Transaction` type + rework `prove` method

### LINKED ISSUE #1514: `Transaction.prove` and `Transaction.send` result method chaining

harrysolovay:
Closes #1514

Updates the `Transaction` and `TransactionPromise` types with two type params `Proven extends boolean` and `Signed extends boolean`. This enables us to do the following.

## 1. conditionally include/exclude members based on stage of method chaining.

```ts
const a = Mina.transaction(sender, async () => {
  await zkapp.increment()
})
a satisfies TransactionPromise<false, false>
a.prove // exists
a.sign // exists
a.proofs // does not yet exist

const b = a.prove()
b satisfies TransactionPromise<true, false>
b.prove // no longer exists
b.sign // exists
b.proofs // now exists

const c = b.sign([senderKey])
c satisfies TransactionPromise<true, true>
c.prove // does not exist
b.sign // no longer exists
b.proofs // exists
```

## 2. Constrain transaction-accepting methods based on the transaction stage.

```ts
declare function acceptsSignedTxsOnly(transaction: Transaction<boolean, true>): void
```

---

An additional change that was needed to make this possible: the `prove` method no longer directly returns the proofs, but rather a `Transaction<true, Signed>`, where `true` is the boolean type param representing whether the transaction has been proven. `TransactionPromise` has a corresponding `prove` method, which produces a `TransactionPromise<Signed, true>`.

In order to access the actual proofs, one can utilize the `proofs` method of a given `Transaction<true, boolean>` or `TransactionPromise<true, boolean>`.

```ts
const proofs = await Mina.transaction(sender, async () => {
  await zkapp.increment();
})
  .sign([...keys])
  .prove()
  .proofs();

proofs; // `(Proof<ZkappPublicInput, undefined> | undefined)[]`
```


src/lib/mina/transaction.ts
```diff
@@ -280,13 +280,32 @@ type RejectedTransaction = Pick<
...
+   * new `TransactionPromise` with the field `proofPromise` containing
+   * a promise which resolves to the proof array.
+   */
+  prove(): TransactionPromise;
+  /**
+   * If the chain of method calls that produced the current `TransactionPromise`
+   * contains a `prove` call, then this field contains a promise resolving to the
+   * proof array which was output from the underlying `prove` call.
+   */
+  proofPromise?: Promise<(Proof<ZkappPublicInput, undefined> | undefined)[]>;
```
mitschabaude, line 303: slightly suboptimal IMO that this makes the the return of `prove()` only optionally have the `proofPromise` property, maybe you can tweak the return type of `prove()`?

harrysolovay: Just updated the `TransactionPromise` type such that it has a type param `Proven extends boolean`, which defaults to `false`. This type param determines whether the developer sees the `prove` method or the `proofPromise` member. Calling `prove` returns a `TransactionPromise<false>`.

mitschabaude (approved):
@harrysolovay it would be nice if you added an entry about all of the new chaining methods in `CHANGELOG.md`

The `prove()` change in particular belongs under breaking changes because `.prove()` returns the proofs in a different format now

src/lib/mina/local-blockchain.ts
```diff
@@ -117,147 +119,152 @@ function LocalBlockchain({
...
       return networkState;
     },
-    async sendTransaction(txn: Transaction): Promise<PendingTransaction> {
-      let zkappCommandJson = ZkappCommand.toJSON(txn.transaction);
-      let commitments = transactionCommitments(
-        TypesBigint.ZkappCommand.fromJSON(zkappCommandJson),
-        minaNetworkId
-      );
+    sendTransaction(
+      txn: Transaction<boolean, boolean>
```
harrysolovay, line 123: Do we want to constrain `sendTransaction` to only proven and signed txs? We could provide means of overriding this for tests.

src/lib/mina/transaction.ts
```diff
@@ -41,12 +41,7 @@ export {
   createIncludedTransaction,
 };
 
-/**
- * Defines the structure and operations associated with a transaction.
- * This type encompasses methods for serializing the transaction, signing it, generating proofs,
- * and submitting it to the network.
- */
-type Transaction = {
+type TransactionCommon = {
```
harrysolovay, line 44: Had to be extracted to avoid circular type reference

harrysolovay:
> it would be nice if you added an entry about all of the new chaining methods in CHANGELOG.md

Given that the changelog is part of the codebase, the commit range would be self-referential. I'm not sure what the o1js workflow is for including this range. Any guidance for where to / how to set up the contents for the next changelog?

mitschabaude:
> Given that the changelog is part of the codebase, the commit range would be self-referential.

Don't change any commit range - just include a bullet point with your change and a link to this PR

src/lib/mina/local-blockchain.ts
mitschabaude: I wouldn't, not every tx needs a proof

src/lib/mina/mina.ts
```diff
@@ -94,8 +96,8 @@ setActiveInstance({
   },
 });
 
-const Transaction = {
-  fromJSON(json: Types.Json.ZkappCommand): Transaction {
+const TransactionUtil = {
```
mitschabaude, line 99: Was this necessary? Would prefer to leave it the same name

src/lib/mina/transaction.ts
```diff
@@ -119,7 +103,39 @@ type Transaction = {
...
+         * @example
+         * ```ts
+         * await transaction.prove();
+         * ```
+         */
+        prove(): Promise<Transaction<true, Signed>>;
+      }
+    : {
+        /** The proofs generated as the result of calling `prove`. */
+        proof: (Proof<ZkappPublicInput, Empty> | undefined)[];
```
mitschabaude, line 121: This should be called proofs (plural), there is one proof per zkapp method call

mitschabaude (changes requested):

CHANGELOG.md
```diff
@@ -15,6 +15,17 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
     _Security_ in case of vulnerabilities.
  -->
 
+## [0.19.0](TODO)
```
harrysolovay, line 18: Saving this for the final commit.

src/lib/mina/mina.ts
mitschabaude: I think it's exported as Mina.Transaction, and people use it

I always like to keep the number of exported names small, and use this pattern a lot

But I'm interested in the auto importing issue, what do you mean?

CHANGELOG.md
mitschabaude: I recommend to merge main and leave the Unreleased header as is

src/lib/mina/mina.ts
mitschabaude: It's also the same as if Transaction were a class - then the Transaction name would serve as a type as well as a collection of static methods as well as a constructor

harrysolovay: I often use auto-imports so that I don't have to manually import symbols. I just use the symbol, and the import statement is generated. This poses some risk / it's easy to accidentally import from the wrong source. But it's a productivity gain imo. Having two `Transaction` symbols in separate files poses confusion about autoimport source.

https://github.com/o1-labs/o1js/assets/4893548/d424bb0a-6810-4634-9c18-47fbe43f08a4

To get around this––while keeping the shadowing––I've moved the runtime "shadow" into `lib/mina/transaction.ts`. Users can still make use of the type-only import if they'd like. Meanwhile `Transaction` is also a namespace, off of which the `fromJSON` method can be accessed.

> Sidenote: shall we rename `fromJSON` to `fromJson`? The `"JSON"` piece is––although an abbreviation––a common term, and thus can be treated as a full segment of the camelcasing.

CHANGELOG.md
harrysolovay: Just to confirm, it's good as is? / you or other team members will update this before the next release?

MartinMinkov: Revert the changes to the original commit :D 

The workflow to update the CHANGELOG goes like so:
1. For each PR, add the functionality under the appropriate section with a link to your PR
2. When it comes to release, we use a bi-weekly cadence with this [CI job running](https://github.com/o1-labs/o1js/blob/main/.github/workflows/release.yml) which will automate the commit range for you. 
3. If you want to do it manually, you can also use this script [here](https://github.com/o1-labs/o1js/blob/main/update-changelog.sh)

That is to say, in normal development, you don't have to worry about it. It will be handled by CI and reviewed when a release PR is made. Like [this one for example](https://github.com/o1-labs/o1js/pull/1510). It would be nice to document this in our README-dev since it's not super clear at this point.

src/lib/mina/mina.ts
mitschabaude: Got it, thanks for moving it (I also use auto imports of course :D)

Json would be more logical but I think JSON might look prettier / more familiar. In any case but something I want to change 5 min before v1

mitschabaude:
Hey @harrysolovay, as a next step it would be nice to merge main, so that we can run the ci pipeline again

src/lib/mina/transaction.ts
```diff
@@ -280,26 +306,74 @@ type RejectedTransaction = Pick<
...
     },
-  }) as TransactionPromise;
+    prove() {
+      return toTransactionPromise(() =>
+        pending.then((v) => (v as never as Transaction<false, Signed>).prove())
+      );
+    },
+    proofs: pending
+      .then((v) => (v as never as Transaction<true, Proven>).proofs)
+      .catch(
```
harrysolovay, line 365: @MartinMinkov and I were wondering whether this is acceptable?: if the transaction fails, then `proofs` resolves to an empty array.

This would solve the problem that was causing the `hello-world/run.ts` integration test to fail: the `pending.then` created a new promise, which escaped the `try`. Aka. the issue didn't have to do with the integration test, but rather an unsuppressed throw here.

harrysolovay: I don't think this would be an issue, as it's really the only case in which you have a promise that can't be contained by the caller's try/catch.

mitschabaude (approved):

MartinMinkov (approved):

harrysolovay:
@MartinMinkov thank you for getting this passed the finish line!

PR #1574, opened 2024-04-13
implement `TestPublicKey` util and refactor examples

### LINKED ISSUE #1516: `Local.testAccounts` vs. `noUncheckedIndexAccess`
With [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) enabled (best practice), the TS checker errors on the inner destructuring of `testAccount` elements. 

```ts
let [{ privateKey, publicKey }] = Local.testAccounts
//   ~~~~~~~~~~~~~~~~~~~~~~~~~
//   ^
//   Property 'privateKey' does not exist on type '{ publicKey: PublicKey; privateKey: PrivateKey; } | undefined'. ts(2339)
```

Is there a reason not to type the length of `Local.testAccounts`? We could do so with [the `TupleN` type](https://github.com/o1-labs/o1js/blob/95e5d419e86be95c5fd4803d6679e2ea6aaaaea3/src/lib/util/types.ts#L28-L32) right [here](https://github.com/o1-labs/o1js/blob/95e5d419e86be95c5fd4803d6679e2ea6aaaaea3/src/lib/mina/local-blockchain.ts#L61-L64). This would spare the developer of needing to override the checker.

```ts
let [{ privateKey, publicKey }] = Local.testAccounts // no TS error, sweet!
```

harrysolovay:
Closes #1516

In addition to solving the false positive / unchecked indexing error of the local blockchain's `testAccounts`, this PR introduces `TestPublicKey`, a subclass of `PublicKey` off of which `key` (the corresponding `PrivateKey`) can be accessed. This leads to a reduction in test code boilerplate, as one doesn't need to define the corresponding private key under its own name.

```diff
- let feePayerPrivateKey = Local.testAccounts[0]?.privateKey!;
- let feePayerPublicKey = Local.testAccounts[0]?.publicKey!;
+ let [feePayer] = Local.testAccounts

// ...

- await Mina.transaction(feePayerPublicKey, () => {
+ await Mina.transaction(feePayer, () => {
    // ...
- }).sign(feePayerPrivateKey)
+ }).sign(feePayer.key)
```

For cases in which a non-pre-funded account is needed (for instance, to be used as a contract account), one can use `TestPublicKey.random`.

```ts
const a = Mina.TestPublicKey.random()
a // `TestPublicKey`

const [b, c] = Mina.TestPublicKey.random(2)
b // `TestPublicKey`
c // `TestPublicKey`
```

---

Note to reviewers re. naming convention: I addressed some (but not all) of the test naming inconsistencies (#1578).


src/lib/mina/local-blockchain.ts
```diff
@@ -39,7 +40,23 @@ import {
   verifyAccountUpdate,
 } from './transaction-validation.js';
 
-export { LocalBlockchain };
+export { LocalBlockchain, TestAccount };
+
+class TestAccount extends PublicKey {
```
mitschabaude, line 45: I think this should be called `TestPublicKey`, not `TestAccount`, because `Account` suggests much more than just a keypair and we already have an `Account` type represents an onchain Mina account with many account fields.

tests/artifacts/javascript/on-chain-state-mgmt-zkapp-ui.js
```diff
@@ -20,39 +20,35 @@ logEvents(
...
 Mina.setActiveInstance(Local);
 // Test account that pays all the fees
-const feePayerKey = Local.testAccounts[0].privateKey;
-const feePayer = Local.testAccounts[0].publicKey;
-// zkApp account
-const zkAppPrivateKey = PrivateKey.random();
-const zkAppAddress = zkAppPrivateKey.toPublicKey();
-const zkAppInstance = new HelloWorld(zkAppAddress);
+const [feePayer] = Local.testAccounts;
+const contractAccount = Mina.TestAccount.random();
```
mitschabaude, line 24: I think this should be
```ts
const contractAddress = Mina.TestPublicKey.random();
```

src/lib/mina/local-blockchain.ts
mitschabaude: Or  `TestAddress`

src/lib/mina/token.test.ts
```diff
@@ -120,86 +116,73 @@ class ZkAppC extends SmartContract {
...
-
-  zkAppBKey = Local.testAccounts[1].privateKey;
-  zkAppBAddress = zkAppBKey.toPublicKey();
-  zkAppB = new ZkAppB(zkAppBAddress, tokenId);
-
-  zkAppCKey = Local.testAccounts[2].privateKey;
-  zkAppCAddress = zkAppCKey.toPublicKey();
-  zkAppC = new ZkAppC(zkAppCAddress, tokenId);
+  [feePayer, bAccount, cAccount] = Local.testAccounts;
+  tokenAccount = Mina.TestAccount.random();
```
mitschabaude, line 138:
```suggestion
  tokenAddress = Mina.TestPublicKey.random();
```

mitschabaude (commented):
This is great!

Re naming, "account" for a public key (with attached private key) feels not optimal, there is the distinct concept of an account. So it should stay at "public key" or "address"

src/lib/mina/local-blockchain.ts
```diff
@@ -39,7 +40,23 @@ import {
...
+  ): N extends 1 ? TestAccount : TupleN<TestAccount, N> {
+    if (count === 1) return new this(PrivateKey.random()) as never;
+    return Array.from(
+      { length: count as number },
+      () => new this(PrivateKey.random())
+    ) as never;
+  }
+
+  constructor(readonly key: PrivateKey) {
+    super(key.toPublicKey());
```
mitschabaude, line 57: this is broken, the constructor has the two properties as arguments, you can use
```suggestion
    super.from(key.toPublicKey());
```

harrysolovay: This change is giving me the following error:

```
'super' must be called before accessing a property of 'super' in the constructor of a derived class.
```

mitschabaude: oh shoot, then just
```ts
let pk = key.toPublicKey();
super(pk.x, pk.isOdd);
```

mitschabaude: forgot that you must always call `super()`

harrysolovay: Is this okay?

harrysolovay:
> Re naming, "account" for a public key (with attached private key) feels not optimal

Can you please add your thoughts in #1578 and I'll fix all occurrences in a follow-up PR?

src/lib/mina/local-blockchain.ts
mitschabaude: yep lgtm. the namespace doesn't conflict type-wise with the type of the same name?

mitschabaude (approved):

src/examples/commonjs.cjs
```diff
@@ -33,39 +32,37 @@ class SimpleZkapp extends SmartContract {
...
 let Local = Mina.LocalBlockchain();
 Mina.setActiveInstance(Local);
 
-let feePayerKey = Local.testAccounts[0].privateKey;
-let feePayer = Local.testAccounts[0].publicKey;
+const [feePayer] = Local.testAccounts
 
-let zkappKey = PrivateKey.random();
-let zkappAddress = zkappKey.toPublicKey();
+let contractAccount = new Mina.random();
```
mitschabaude, line 43: whoops!

src/lib/mina/local-blockchain.ts
harrysolovay: Correct

mitschabaude: and TS just merges the runtime function `TestPublicKey` and object `TestPublicKey` into the same value? I'm amazed

fwiw it also works to assign the function to a property (it modifies the type),
```ts
TestPublicKey.random = ...
```

src/lib/mina/local-blockchain.ts
```diff
@@ -63,18 +81,14 @@ function LocalBlockchain({
...
   }
 
-  let testAccounts: {
-    publicKey: PublicKey;
-    privateKey: PrivateKey;
-  }[] = [];
+  let testAccounts = [] as never as TupleN<TestPublicKey, 100>;
 
-  for (let i = 0; i < 10; ++i) {
+  for (let i = 0; i < 100; ++i) {
```
mitschabaude, line 86: @harrysolovay I just realized that this takes 2ms per public key (it does an elliptic curve scaling), so 200ms for 100 accounts feels a bit excessive to do every time we instantiate local blockchain.

Maybe we can create these test accounts lazily and still preserve the same API with some trickery. If not, I'd prefer to create only 20 or so.

All of this can be a separate PR

harrysolovay, line 86:
```suggestion
  for (let i = 0; i < 10; ++i) {
```

harrysolovay, line 86:
```suggestion
  for (let i = 0; i < 10; ++i) {
```

src/lib/mina/local-blockchain.ts
```diff
@@ -63,18 +81,14 @@ function LocalBlockchain({
     }
   }
 
-  let testAccounts: {
-    publicKey: PublicKey;
-    privateKey: PrivateKey;
-  }[] = [];
+  let testAccounts = [] as never as TupleN<TestPublicKey, 100>;
```
harrysolovay, line 84:
```suggestion
  let testAccounts = [] as never as TupleN<TestPublicKey, 10>;
```

mitschabaude (approved):

Issue #1575, opened 2024-04-14
Can't use `PrivateKey` on server in a NextJS 14 app


sam-goodwin:
I am having trouble using `o1js` in a simple NextJS 14 app. 

You can find my repo here: https://github.com/sam-goodwin/private-wallet

Check out the repo and run:
```sh
pnpm i
pnpm next build
```

It will just hang:
```
next build
   ▲ Next.js 14.1.4
   - Environments: .env

   Creating an optimized production build ...
 ✓ Compiled successfully
 ✓ Linting and checking validity of types    
 ⚠ Sending SIGTERM signal to static worker due to timeout of 60 seconds. Subsequent errors may be a result of the worker exiting.
 ⚠ Restarted collecting page data for undefined because it took more than 60 seconds
 ⚠ See more info here https://nextjs.org/docs/messages/static-page-generation-timeout
   Collecting page data  .
```

Comment out the `PrivateKey.random()` and the error goes away.

When looking at the tutorials I spotted this bizarre code:

```ts
export default function Home() {

  useEffect(() => {
    (async () => {
      const { Mina, PrivateKey } = await import('o1js');
      const { Add } = await import('../../../contracts/build/src/');
    })();
  }, []);
```

This raises some red flags. Is `o1js` not designed to work as a normal module that can be imported?

sam-goodwin:
Ok, I just found a NextJS config that might be something I am missing.

```ts
/** @type {import('next').NextConfig} */
const nextConfig = {
  reactStrictMode: false,


  webpack(config) {
    config.resolve.alias = {
      ...config.resolve.alias,
      o1js: require('path').resolve('node_modules/o1js')
    };
    config.experiments = { ...config.experiments, topLevelAwait: true };
    return config;
  },
  // To enable o1js for the web, we must set the COOP and COEP headers.
  // See here for more information: https://docs.minaprotocol.com/zkapps/how-to-write-a-zkapp-ui#enabling-coop-and-coep-headers
  async headers() {
    return [
      {
        source: '/(.*)',
        headers: [
          {
            key: 'Cross-Origin-Opener-Policy',
            value: 'same-origin',
          },
          {
            key: 'Cross-Origin-Embedder-Policy',
            value: 'require-corp',
          },
        ],
      },
    ];
  }
};

module.exports = nextConfig

```

Why does `o1js` require so much hackery to work? Can't it be an ordinary module? These kinds of things are what lead to people abandoning.

[...]

Issue #1205, opened 2023-10-30
Make every API that needs OCaml/Rust async instead of awaiting at the top level


mitschabaude:
The web version of o1js still has a fairly large unavoidable import time (~0.5s). This used to happen in the first asynchronous microtask (and leaked to users through `await isReady`) and is now happening immediately with top-level await.

To enable quicker loading as well as prepare for https://github.com/o1-labs/o1js/issues/1192, the big async task should be awaited inside the functions that need it, not at the top level. There are just a few of those functions which aren't already async:

* `Provable.runAndCheck()` and `runUnchecked()` -- mostly used as part of `Mina.transaction()` which is already async
* `SmartContract.analyzeMethods()`
* `Provable.constraintSystem()` -- mostly used in `analyzeMethods()`
* `Mina.LocalBlockchain()`

mitschabaude:
The change to async APIs will be a natural side effect of #735, after which this issue is no longer breaking

mitschabaude:
Update: all of the APIs above, except Mina.LocalBlockchain, are async now.

We should make LocalBlockchain async, and move the top level await away from the top level and into functions that depend on it. That might also allow us to simplify Integration with legacy web bundlers like webpack, see #1575

PR #1583, opened 2024-04-15
Remove top-level await

mitschabaude:
* Addresses #1205 
* Removes top-level await to help simplifying integration in UI frameworks
  * see https://github.com/o1-labs/o1js/issues/1575
* Side-effect: `Mina.LocalBlockchain()` becomes async

bindings: https://github.com/o1-labs/o1js-bindings/pull/268

### Details

Instead of awaiting bindings imports at the top-level, we hide them behind a new async function `initializeBindings()`.

`initializeBindings()` does _not_ have to be called by users of o1js, because this PR takes care to call it first wherever bindings are needed.

Almost all user-facing APIs which need to call `initializeBindings()` were already async -- making the remaining two async constitutes the two breaking changes of this PR:
* `LocalBlockchain()`
* `Proof.fromJSON()`

We also export `initializeBindings()` for explicit control, which can be useful to front-load the bindings import in web apps.

### Loading time improvements

Because loading of the bindings is now deferred until it's needed, this comes with a great improvement for initial loading time of o1js. It will enable us to remove hacks to load o1js in a `useEffect()` or similar.

```
$ ./run-in-browser.js src/examples/benchmarks/import.web.ts
```
logs in the browser console:
```
import o1js: 29ms
initialize bindings: 318ms
```


mitschabaude:
> Thanks for putting this together. I'm a little concerned that we haven't totally confirmed this as the root cause for next. I think there's also a bug in package.json and exports.

I agree that it might seem like the wrong order, but
* we want to get rid of top level await anyway
* we are currently very focused on delivering v1.0, the first stable version of o1js
* which means we need to make all breaking changes now, asap
* and also means I don't have time this week to debug the Next setup, which should be fixable without other breaking changes

sam-goodwin:
No I think it's the right order. This is very important. My main point is that we should add a test for client and server side components in next. Full steam ahead on removing top level await! Thanks for getting that going.

sam-goodwin:
But if you wanna follow that up in a later PR, that also makes sense.

src/lib/mina/account-update.ts
```diff
@@ -875,28 +875,11 @@ class AccountUpdate implements Types.AccountUpdate {
   }
 
   hash(): Field {
-    // these two ways of hashing are (and have to be) consistent / produce the same hash
```
mitschabaude, line 878: finally had to remove this, because it would have meant account update hashing depends on the bindings

src/lib/mina/local-blockchain.ts
```diff
@@ -45,10 +45,11 @@ export { LocalBlockchain };
 /**
  * A mock Mina blockchain running locally and useful for testing.
  */
-function LocalBlockchain({
+async function LocalBlockchain({
```
mitschabaude, line 48: `LocalBlockchain` -> async

src/lib/mina/zkapp.ts
```diff
@@ -704,6 +704,7 @@ class SmartContract extends SmartContractBase {
       ._verificationKey;
     if (verificationKey === undefined) {
       if (!Mina.getProofsEnabled()) {
+        await initializeBindings();
         let [, data, hash] = Pickles.dummyVerificationKey();
```
mitschabaude, line 708: awaiting bindings in `deploy()`, to use `Pickles`

src/lib/proof-system/circuit.ts
mitschabaude, line 1: awaiting bindings in `Circuit` APIs

src/lib/proof-system/zkprogram.ts
```diff
@@ -110,18 +110,21 @@ class Proof<Input, Output> {
       proof: Pickles.proofToBase64([this.maxProofsVerified, this.proof]),
     };
   }
-  static fromJSON<S extends Subclass<typeof Proof>>(
+  static async fromJSON<S extends Subclass<typeof Proof>>(
```
mitschabaude, line 113: `Proof.fromJSON()` becomes async because it needs `Pickles` and there's the chance people call this without calling `compile()` first

src/lib/proof-system/zkprogram.ts
```diff
@@ -192,6 +195,7 @@ async function verify(
   proof: Proof<any, any> | JsonProof,
   verificationKey: string | VerificationKey
 ) {
+  await initializeBindings();
```
mitschabaude, line 198: awaiting bindings in `verify()`

src/lib/proof-system/zkprogram.ts
```diff
@@ -614,6 +618,7 @@ async function compileProgram({
   forceRecompile: boolean;
   overrideWrapDomain?: 0 | 1 | 2;
 }) {
+  await initializeBindings();
```
mitschabaude, line 621: awaiting bindings in `compile()`

note: we don't need it in any of the provers because you have to call `compile()` before running them

src/lib/provable/bool.ts
```diff
@@ -198,14 +198,13 @@ class Bool {
    * This can only be called on non-witness values.
    */
   toBoolean(): boolean {
```
mitschabaude, line 200: I restructured this so that `toBoolean()` doesn't need `Snarky` to be defined when the input is constant.

When the input is a variable, it's fine to rely on `Snarky` because one of the circuit running methods must have been called first, and they are all guarded by `initializeBindings()`

src/lib/provable/core/provable-context.ts
mitschabaude, line 1: we're awaiting bindings in every method that runs provable code, so that this code is able to use the `Snarky` import

src/lib/provable/crypto/poseidon.ts
```diff
@@ -33,6 +33,7 @@ type ProvableHashable<T> = Provable<T> & Hashable<T>;
 class Sponge {
   #sponge: unknown;
 
+  // TODO: implement constant version in TS. currently, you need to call `initializeBindings()` before successfully calling this
```
mitschabaude, line 36: I left this as TODO. `Poseidon.Sponge()` cannot be used before initializing bindings first, which could lead to users experiencing errors. I think the function is rarely used though, and there's an easy workaround: call `initializeBindings()`. The eventual solution is to implement `Sponge()` in TS (should be similar to `Poseidon.hash()`)

src/snarky.js
mitschabaude, line 1: this is how the Node.js bindings work now

src/snarky.web.js
mitschabaude, line 1: this is how the web bindings work now

mitschabaude (commented):
most of the PR is fixing tests -- highlighted some of the actual changes

src/lib/provable/crypto/poseidon.ts
ymekuria: Would it be easy/worthwhile to add an error message to call `initializeBindings()` until `Sponge()` is implemented in TS?

ymekuria (approved):
I'm excited for theses changes! Nice work! I had some minor comments and questions.

MartinMinkov (approved):

sam-goodwin:
So glad these changes were merged. Is there a timeline for release to NPM?

ymekuria:
> So glad these changes were merged. Is there a timeline for release to NPM?

@sam-goodwin There is a plan to release early next week. We will let you know if things change.

Issue #1506, opened 2024-03-14
Add more benchmarks.


shimkiv:
https://github.com/o1-labs/o1js/pull/1498

shimkiv:
[Internal discussion](https://o1-labs.slack.com/archives/C028Q27R8UC/p1712575516480379).

PR #1591, opened 2024-04-18
Benchmark o1js startup time.

shimkiv:
- Part of https://github.com/o1-labs/o1js/issues/1506.


benchmark/benchmarks/init.ts
```diff
@@ -0,0 +1,22 @@
...
+    tic('o1js import');
+    const { initializeBindings } = await import('o1js');
+    toc();
+
+    tic('bindings initialization');
+    await initializeBindings();
+    toc();
+  },
+  // Run once with no warmups to get the worst-case scenario metrics
+  { numberOfWarmups: 0, numberOfRuns: 1 }
```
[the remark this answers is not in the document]
mitschabaude: we can't, because the import will be zero cost on the second run!

benchmark/runners/with-cloud-history.ts
```diff
@@ -6,18 +6,23 @@
  * ./run benchmark/runners/with-cloud-history.ts --bundle
  * ```
  */
+
 import { initializeBindings } from 'o1js';
 import { logResult } from '../benchmark.js';
 import { EcdsaBenchmark } from '../benchmarks/ecdsa.js';
+import { InitBenchmark } from '../benchmarks/init.js';
```
mitschabaude, line 13: I think this doesn't work, because we're already importing o1js here before trying to measure o1js import time. Node.js will be smart enough to not compile and run the module a second time, so we're not getting the proper benchmark

mitschabaude (commented):
This is awesome @shimkiv, I also thought about adding this benchmark!
I just think we can't include the `.run()` call in the same script that imports o1js at the beginning (even indirectly, via `EcdsaBenchmark`)

So, there are two options I think:
* either use a completely separate script for the import benchmarks
* or run the import benchmark at the beginning and then import all other benchmarks with `await import(...)` _after_ the import benchmark has finished

benchmark/runners/with-cloud-history.ts
mitschabaude: also `EcdsaBenchmark` imports o1js

shimkiv:
Thank you @mitschabaude, I then will extract it into separate script.

mitschabaude (approved):

PR #1594, opened 2024-04-18
More reducer stuff

mitschabaude:
- `iterate()` abstraction
- add scary comments


Trivo25:
some more prototyping here as well https://github.com/o1-labs/o1js/pull/1605

PR #1595, opened 2024-04-18
feat(typedoc): update config

MartinMinkov:


Trivo25 (approved):

Trivo25 (approved):

typedoc.json
```diff
@@ -1,16 +1,50 @@
...
   "githubPages": false,
-  "exclude": ["dist/**/*", "src/mina-signer/**/*", "src/examples/**/*"],
+  "exclude": [
+    "dist/**/*",
+    "src/mina-signer/**/*",
+    "src/bindings/**/*",
+    "src/examples/**/*",
+    "src/lib/provable/+(core|test)/**/*",
+    "src/**/*(test|unit-test).ts",
+    "src/lib/provable/provable.ts"
```
45930, line 16: @Trivo25 I don't suppose you have any recollection about why `src/lib/provable/provable.ts` was added to the exclude list eh?  I am probably going to remove it from this list, since we do want the provable.ts file to generate docs.

mitschabaude: @45930 As I recall it, typedoc was messing up docs for exports correctly that are defined as both an object and a type.
I.e. we have both `type Provable = ...` and `const Provable = ...`.

By excluding `provable.ts`, which contains the object (but not the type) definition, at least the type `Provable` and its docstring are displayed correctly in our docs.

It was always a non-satisfactory workaround, we _would_ like to have all our docsstrings in the reference obviously.

45930: @mitschabaude Cheers thanks for the context! 🍻

Issue #1426, opened 2024-02-11
Mina.fetchActions returns an empty list of actions when endActionState is defined


dfstio:
On Berkeley testnet, calling 
```typescript
let actions = await Mina.fetchActions(publicKey, {
      fromActionState: startActionsState,
      endActionState: endActionState,
    });
```
returns an empty list when endActionState is not undefined. It works as expected when endActionState is undefined.

This behavior is the same on minascan and minaexplorer endpoints.

The test code:
https://github.com/dfstio/merkle-map-demo/blob/endActionState/tests/actions.test.ts

The test code log:
```
merkle-map-demo % yarn test actions
[2:39:11 PM] number of all actions: 12
[2:39:11 PM] lastActionState: 4671254814962134621561188624509467899566137729694462218629375270473276286241
[2:39:11 PM] number of actions starting at startActionsState: 6
[2:39:11 PM] actions starting at startActionsState: [
  {
    actions: [ [Array] ],
    hash: '2889372988449481232750872203377957230274508842265070242272487645409136733839'
  },
  {
    actions: [ [Array] ],
    hash: '15147297301450527213667297732587033679468451271733367936786243634936439048634'
  },
  {
    actions: [ [Array] ],
    hash: '5766519161874551854426173362531259652853737507928229135738402350087629707642'
  },
  {
    actions: [ [Array] ],
    hash: '3472596428023171912405119750508284466233959846144065437966752178811252259985'
  },
  {
    actions: [ [Array] ],
    hash: '5072649747128810814145361414529837940474819464966837380953978607528499679008'
  },
  {
    actions: [ [Array] ],
    hash: '4671254814962134621561188624509467899566137729694462218629375270473276286241'
  }
]
[2:39:11 PM] number of actions starting from startActionsState and ending on lastActionState: 0
[2:39:11 PM] number of actions starting from startActionsState and ending on endActionState: 0
 FAIL  tests/actions.test.ts
  Actions
    ✓ should get all actions (499 ms)
    ✓ should get actions starting from startActionsState (138 ms)
    ✕ should get actions starting from startActionsState and ending on lastActionState (172 ms)
    ✕ should get actions starting from startActionsState and ending on endActionState (163 ms)

  ● Actions › should get actions starting from startActionsState and ending on lastActionState

    expect(received).toBeGreaterThan(expected)

    Expected: > 0
    Received:   0

      67 |         actions.length
      68 |       );
[...]

PR #1597, opened 2024-04-18
Test remote reducer

mitschabaude:


shimkiv:
In summary failed job has logs attached: https://github.com/o1-labs/o1js/actions/runs/8743722468?pr=1597
These logs are from the Lightnet processes (mina daemon, archive node, archive-node-api). This will let you debug further. I can help you with that if you'd like to. 
Also might be helpful to run this tests of this branch on local instance of the lightnet and follow the docker container processes logs to see in real time what is going on.

dfstio:
The same problem exists on the Berkeley testnet:
https://github.com/o1-labs/o1js/issues/1426

dfstio:
> * Also not sure if this is specific to the archive node - Mina interaction on lightnet. For example, with what latency can we expect actions to arrive in the archive node API after the node published them? Are the two even properly connected?

On Berkeley, I've seen latency of up to 2 hours. It means that I should wait two hours to start reducing the actions after the actions have been sent: https://github.com/o1-labs/o1js/issues/1426#issuecomment-1938787799

PR #1598, opened 2024-04-19
Remove Account constructor

mitschabaude:
small breaking change, removes `Account()` API which didn't add anything IMO, instead exposes derived `Account` type which helps with low-level things

also added changelog for https://github.com/o1-labs/o1js/pull/1271


ymekuria (approved):
LGTM.

PR #1599, opened 2024-04-20
RSA example fix & Test coverage

Shigoto-dev19:
## Description 
This PR builds upon the changes introduced in #1229

## Changes

- Move the existing files `./src/examples/crypto/rsa.ts` to a new directory `./src/examples/crypto/rsa/rsa.ts`

-   Fix carry computation bug in multiplication for provable Bigint2048
    - The constraint `(xy - qp - r)_i + c_(i-1) === c_i * 2^116` was not satisfied 
    - Calculating the carry with `carry = res_i.div(2n ** 116n); ` prevents field overflow while satisfying the constraint mentioned above.

- Quick fix  `rangeCheck128Signed` function
   - The function returns a weird `Snarky.run` error.
   - Replace `Provable.witnessFields(2, callback)` with `Provable.witness(Provable.Array(Field, 2), callback)`
   
## Additions

- Add a new file for test utilities: `testUtils.ts`
   - Contains utility functions for random RSA parameter generation.
   - Depends on the external package [big-integer](https://www.npmjs.com/package/big-integer) for faster and more scalable bigint arithmetic.
   
- Add a new file for provable RSA65537 signature verification: `rsa.test.ts`
   - Contains test cases for:
      - Simple RSA verification
      - Hardcoded large RSA parameters imported from an external resource.
      - Large, randomly generated RSA parameters with modulus sizes of 512, 1024, and 2048 bits
      - Unhappy test cases for non-compliant RSA parameters, including modulus sizes larger than 2048 bits
      
 - Add a new file for RSA ZkProgram benchmarks: `run.ts`
   - Move the existing RSA ZkProgram from `rsa.ts` to `run.ts`
   - Add execution timing for proof generation & verification.

## Notes
- Usage of [big-integer](https://www.npmjs.com/package/big-integer) is justified for the following reasons:
   - `modInv` & `modPow` exceed local memory limit when operating on 2048-bit native bigints
   - The package is crucial to generate random large primes for RSA parameters.
      - It has a method to return primality of a big integer based on the **Miller–Rabin primality test**

## Build
First of all: Build `./src/examples` directory
 ````bash 
npm run build:examples 
````
Run RSA local tests: Skipped tests with increased iterations
````bash
NODE_OPTIONS=--experimental-vm-modules npx jest rsa.test.ts
````
Run RSA ZkProgram benchmarks
````bash
node ./dist/examples/crypto/rsa/run.js
````

### RSA ZkProgram Benchmarks Preview 

|              | Total rows | RangeCheck0 | Generic |
|------------|---------------|--------------------|-------------|
| Total      | 12401       | 2488               | 9913      |

compile: 2.368s
generate RSA parameters: 2048 bits: 13.549ms
prove: 17.015s
verify: 983.869ms

## TODO
- Refactor RSA tests.
- Adapt bigint arithmetic to native bigint to omit using an external dependency.
- Optimize RSA ZkProgram to address large circuit size causing 'Runtime Error: unreachable' in 'run.ts' execution (solved).

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -94,6 +94,7 @@
     "typescript": "5.1"
   },
   "dependencies": {
+    "big-integer": "^1.6.52",
```
Trivo25, line 97: Why did you choose this over native bigint support?

Trivo25:
This is great! Thank you so much for taking the time :)

package.json
Shigoto-dev19: Because generating RSA parameters of size=2048 bits exceed memory limit when operating with native bigints.
Let me tag you on Discord to provide more context on the PR.

mitschabaude:
> Refactor rangeCheck128Signed function
> Use Gadgets.isDefinitelyInRangeN(128, x).assertTrue(); instead of doing custom range check on two 64-bit limbs.

I think we should actually revert this change, it's less efficient this way

src/examples/crypto/rsa/rsa.ts
```diff
@@ -97,11 +101,7 @@ function multiply(
   for (let i = 0; i < 2 * 18 - 2; i++) {
     let res_i = res[i].add(carry);
 
-    [carry] = Provable.witnessFields(1, () => {
-      let res_in = res_i.toBigInt();
-      if (res_in > 1n << 128n) res_in -= Field.ORDER;
-      return [res_in >> 116n];
-    });
+    carry = res_i.div(2n ** 116n);
```
mitschabaude, line 104: awesome fix! since this already proves ` res_i.assertEquals(carry.mul(1n << 116n));`, we can remove the line below

src/examples/crypto/rsa/rsa.ts
```diff
@@ -160,41 +160,7 @@ function rangeCheck116(x: Field) {
...
-
-console.log(verify.summary());
-console.log('rows', verify.rows);
-
-tic('compile');
-await rsa.compile();
-toc();
+  Gadgets
+    .isDefinitelyInRangeN(128, x)
+    .assertTrue("BigInt carry should not exceed 128 bits!");
```
mitschabaude, line 165: I think this should be changed back, `isDefinitelyInRangeN()` uses a different gate which uses more constraints than `rangeCheck64()`, and the extra creation of a `Bool` instead of asserting directly also wastes constraints

mitschabaude:
> Total 	245481 	1332 	239525 	4624

this seems a lot 🤔 for me it was only 12k constraints, and this circuit just adds a couple of range checks on the inputs

src/examples/crypto/rsa/rsa.ts
Shigoto-dev19: `.div(2n ** 116n)` solved the bug it added a ton of constraints.

I fixed the problem again using 
````typescript
carry = Provable.witness(Field, () => {
     let res_in = res_i.toBigInt();
     return Field(res_in * (Fp.inverse(2n ** 116n) ?? 0n));
});
````

src/examples/crypto/rsa/rsa.ts
Shigoto-dev19: I am happy to learn about gates intricacies.
True - this change alone added 4000 rows :open_mouth: 
- Reverting the changes causes a weird `Snarky.run` error :thinking: 
- I did a workaround by using `Provable.witness(Provable.Array(Field, 2), () => {})` instead of `witnessFields(2, callback)`
- Now everything pass and  rsa has the following summary:
   -  Total rows: 12401 
   - RangeCheck0: 2488
   -  Generic: 9913

mitschabaude (approved):
Thanks so much for adding these fixes!

PR #1605, opened 2024-04-22
`Symbol.iterator` for `MerkleList`

Trivo25:
maybe?


mitschabaude:
the same convenience would also be achieved if we have a `.toArray(length)` method on `MerkleList` which returns a static-sized array. And that would be more secure because we could call `assertAtEnd()` before returning the array

Issue #673, opened 2023-01-06
Expose side loaded verification key to snarkyjs


mitschabaude:
"Side-loaded" verification keys are used by Pickles to recursively verify proofs which aren't baked into the verifying circuit at compile time. This is for example necessary for zkApp proofs, which are verified in the transaction snark but can't be baked in at compile time. This seems to be an alternative way of using proof recursion, which is more flexible w.r.t. what proofs can be verified. A user reports needing this feature:

> A few weeks back we talked about verifying Proofs inside ZkPrograms / Contracts with a dynamically passed verification-keys (not baked in ones). This is a feature I need for a project I want to develop

This would also be needed to develop ZkPrograms which are mutually recursive (they can verify the proofs of each other, not just in one direction). Right now, in ZkProgram there is a strict order in which the verification keys have to be compiled: If Program A wants to verify proofs of Program B, then compiling Program A requires the verification key / tag of Program B to be already available. So, there is no way that Program B could depend on Program A.

The goal is to expose the more flexible recursion style to snarkyjs.

rpanic:
This would enable more general-style rollups which could have a "network"-state on their own, which would enable the development of more efficient and distributed rollups versus the current trustless but centralized per-contract idea. 

Also, more contract patterns could emerge, like upgradable proxy contracts, etc.

mitschabaude:
That would be a nice API @maht0rz! I imagine something like this should be possible, I think the OCaml code that verifies zkApp proofs in a circuit looks similar

maht0rz:
@mitschabaude i'd love to contribute by implementing this but i am currently bottlenecked by my knowledge of Kimchi, and how it could be exposed into snarkyjs. I personally don't know why exactly this doesnt work already, so that makes it difficult to try and figure out what needs to change in the first place.

rpanic:
I'd be very interested in the reasons why that is not exposed by default. Is it just a matter of API or is there a different mechanism for the side-loaded keys under the hood? How does the side-loading mechanism actually work on a abstract level?

mitschabaude:
> Is it just a matter of API or is there a different mechanism for the side-loaded keys under the hood?

There's a different mechanism! Pickles supports two kinds of verification keysL "compiled" (these are what snarkyjs currently uses) and "side-loaded":
https://github.com/MinaProtocol/mina/blob/dfc3a21ce43904bd9d23a3d3ce2c3a2c570071f2/src/lib/pickles/tag.ml#L5

PR #1606, opened 2024-04-22
Expose sideloaded verification keys

rpanic:
This PR implements sideloaded verification keys as outlined in this RFC: https://github.com/MinaFoundation/Core-Grants/pull/3

Bindings: https://github.com/o1-labs/o1js-bindings/pull/269

Replacement to https://github.com/o1-labs/o1js/pull/1238
Closes https://github.com/o1-labs/o1js/issues/673


src/examples/zkprogram/mututal-recursion.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Every multiplication or addition step consumes a previous proof from the other circuit to verify prior state. 
+ */
+
+class DynamicMultiplyProof extends DynamicProof<Undefined, Field> {
+  static publicInputType = Undefined;
+  static publicOutputType = Field;
+  static maxProofsVerified = 1 as const;
+}
+
+const add = ZkProgram({
```
mitschabaude, line 23: this example makes me a bit uncomfortable, since `add` isn't proving that `DynamicMultiplyProof` came from the multiply program, and that might not be obvious to users seeing this example.

I think adding a TODO comment where you highlight this caveat would be enough to prevent this pattern accidentally ending up in a user app and creating a vulnerability

src/index.ts
```diff
@@ -1,5 +1,5 @@
 export type { ProvablePure } from './lib/provable/types/provable-intf.js';
-export { Ledger, initializeBindings } from './snarky.js';
+export { Ledger, initializeBindings, Pickles } from './snarky.js';
```
mitschabaude, line 2: wait, why are we exporting the `Pickles` namespace? if there's no specific reason, I'd rather avoid exporting an undocumented and easy-to-misuse low-level API

src/lib/proof-system/zkprogram.ts
```diff
@@ -787,14 +919,43 @@ function picklesRuleFromFunction(
       let input = fromFieldVars(publicInputType, publicInput);
       result = await func(input, ...finalArgs);
     }
+
+    proofs.forEach(({ proofInstance, classReference }, index) => {
+      if (proofInstance instanceof DynamicProof) {
```
mitschabaude, line 924: nit: early return if `!(proofInstance instanceof DynamicProof)` would be cleaner

src/snarky.d.ts
```diff
@@ -741,6 +741,19 @@ declare const Pickles: {
...
+    create:(name: string, numProofsVerified: 0 | 1 | 2, publicInputLength: number, publicOutputLength: number) => unknown /* tag */,
+    // Instantiate the verification key inside the circuit (required).
+    inCircuit:(tag: unknown, verificationKey: unknown) => undefined,
+    // Instantiate the verification key in prover-only logic (also required).
+    inProver:(tag: unknown, verificationKey: string) => undefined,
+    // Create an in-circuit representation of a verification key
+    vkToCircuit:(verificationKey: () => string) => unknown /* verificationKeyInCircuit */,
+    // Get the digest of a verification key in the circuit
+    vkDigest:(verificationKeyInCircuit: unknown) => MlArray<FieldVar>,
+};
```
mitschabaude, line 755: nit: this doesn't look code-formatted

src/lib/proof-system/sideloaded.unit-test.ts
```diff
@@ -0,0 +1,225 @@
...
+    program1Proof = proof1;
+
+    const proof2 = await program2.foo(
+      { field1: Field(1), field2: Field(2) },
+      Field(3)
+    );
+    program2Proof = proof2;
+
+    await sideloadedProgram.compile();
+  });
```
mitschabaude, line 149: I find this `before()` pattern with variable reassignments unreadable and error-prone. I think it's the legacy of Jest + missing top-level await. Can you just move all that preparation logic to before the `describe()` block?

mitschabaude (commented):
Awesome, thanks so much for contributing this @rpanic!

I have just a few suggestions

src/index.ts
rpanic: Oh, that is a leftover from testing, thanks for catching that!

src/lib/proof-system/zkprogram.ts
```diff
@@ -787,14 +919,43 @@ function picklesRuleFromFunction(
...
+        if (vk === undefined) {
+          throw new Error(
+            'proof.verify() not called, call it at least once in your circuit'
+          );
+        }
+
+        if (Provable.inProver()) {
+          Pickles.sideLoaded.inProver(computedTag, vk.data);
+        }
+        const circuitVk = Pickles.sideLoaded.vkToCircuit(() => vk.data);
```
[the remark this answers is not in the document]
rpanic: Very good question, to be honest, I got the instructions on how to use the exposed pickles calls from matthew and just accepted them as-is. Would also be interested in what the background on those is @mrmr1993

mitschabaude (approved):

Issue #1615, opened 2024-04-24
`UInt64.rightShift()` does a left shift


mitschabaude:
https://github.com/o1-labs/o1js/blob/ea199efea8f393fa72a5d672762812a88989efec/src/lib/provable/int.ts#L373-L375

Issue #1616, opened 2024-04-24
`ForeignField.toBits()` discards high limbs of unreduced elements for input bit lengths <  176


mitschabaude:
https://github.com/o1-labs/o1js/blob/ea199efea8f393fa72a5d672762812a88989efec/src/lib/provable/foreign-field.ts#L356-L360

This behaviour is incorrect since we don't prove the higher limbs are zero! The simple fix is to add assertions that they are.

mitschabaude:
closed by #1617

PR #1617, opened 2024-04-24
Fix shift direction and `toBits` constraints 

Trivo25:
closes https://github.com/o1-labs/o1js/issues/1615 https://github.com/o1-labs/o1js/issues/1616


src/lib/provable/int.ts
```diff
@@ -371,7 +371,7 @@ class UInt64 extends CircuitValue {
    * ```
    */
   rightShift(bits: number) {
```
shimkiv, line 373: Worth to also update the comment.

PR #1633, opened 2024-05-02
fix rosettaCombinePayload

qwadratic:


src/mina-signer/src/rosetta.ts
```diff
@@ -282,7 +282,7 @@ function rosettaTransactionToSignedCommand({
 }
 
 type UnsignedPayload = {
-  unsigned_transaction: UnsignedTransaction;
+  unsigned_transaction: string;
```
MartinMinkov, line 285: What is the issue with `unsigned_transaction` being of type `UnsignedTransaction`? Is there no way to keep it typed here?

qwadratic:
because actually it is a string

чт, 2 мая 2024 г. в 19:05, Martin Minkov ***@***.***>:

MartinMinkov:
Seems like there's some tests to fix:
https://github.com/o1-labs/o1js/actions/runs/8926010581/job/24521454236?pr=1633

qwadratic:
@MartinMinkov can you please run tests again?

qwadratic:
because I get   `Preset ts-jest/presets/js-with-ts not found.` when trying to run tests locally

mitschabaude:
> @MartinMinkov can you please run tests again? why do we need aprroval to run the workflow?

If we allowed all external contributors to run workflows, that could lead to exploits because workflows can access important secrets like our npm token

mitschabaude:
> because I get `Preset ts-jest/presets/js-with-ts not found.` when trying to run tests locally

@qwadratic does this work?
```
git submodule update --init --recursive
npm i
npm run build
./jest src/mina-signer/tests/rosetta.test.ts
```
if no, where does it fail?

qwadratic:
Fixed!

qwadratic:
@mitschabaude @MartinMinkov what can we do with failing benchmark check? this PR doesn't touch o1js

mitschabaude:
> @mitschabaude @MartinMinkov what can we do with failing benchmark check? this PR doesn't touch o1js

You can ignore that failure @qwadratic, it just doesn't work for external PRs

Looks like this is good to go once @MartinMinkov approves!

mitschabaude (commented):
reviewed, have same question as @MartinMinkov asked: https://github.com/o1-labs/o1js/pull/1633/files#r1587988536

qwadratic:
the typing here was just wrong, the rosetta response has string typed unsigned_transaction (which is json payload for signing)

I fixed it and now its right

mitschabaude (approved):
with some extra context the change makes sense

Issue #1632, opened 2024-05-02
Side loading verification keys give "proof.verify() not called, call it at least once in your circuit" error in SmartContract


dfstio:
https://discord.com/channels/484437221055922177/1235505048537010257

The following code
```typescript
export class MyContract extends SmartContract {
  @state(Field) value = State<Field>();

  async deploy(args: DeployArgs) {
    super.deploy(args);
    this.account.permissions.set({
      ...Permissions.default(),
      editState: Permissions.proof(),
    });
  }

  @method async setValue(
    value: Field,
    proof: SideloadedProgramProof,
    vk: VerificationKey
  ) {
    proof.verify(vk);
    proof.publicInput.assertEquals(value);
    this.value.set(value);
  }
}
```
gives an error
```
    proof.verify() not called, call it at least once in your circuit

      at node_modules/o1js/dist/node/index.cjs:9797:15
          at Array.forEach (<anonymous>)
      at main (o1js/dist/node/index.cjs:9789:12)
```
when compiling SmartContract

PR #1634, opened 2024-05-02
Fixed bug when using sideloaded keys in SmartContracts

rpanic:
Fixes https://github.com/o1-labs/o1js/issues/1632

Adds regression test


Trivo25:
need to run `./run ./tests/vk-regression/vk-regression.ts --bundle --dump` for the new vk tests

mitschabaude (approved):
Nice!

PR #1638, opened 2024-05-06
Dynamic call unit test

mitschabaude:
Adds a test which demonstrates that we can switch out which contract is called without breaking the caller contract

### DIFF
--- a/src/lib/mina/test/dynamic-call.unit-test.ts
+++ b/src/lib/mina/test/dynamic-call.unit-test.ts
@@ -43,7 +43,7 @@ class Caller extends SmartContract {
   }
 
   // subcontract to call. this property is changed below
-  // TODO: having to set this property is a hack, it would be nice to
+  // TODO: having to set this property is a hack, it would be nice to pass the contract as parameter
   static Subcontract: new (...args: any) => Subcontract = SubcontractA;
 }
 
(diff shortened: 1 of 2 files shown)


MartinMinkov (approved):

src/lib/mina/test/dynamic-call.unit-test.ts
```diff
@@ -0,0 +1,92 @@
...
+class Caller extends SmartContract {
+  @method
+  async call(a: UInt64, b: UInt64, address: PublicKey) {
+    const subcontract = new Caller.Subcontract(address);
+    await subcontract.submethod(a, b);
+  }
+
+  // subcontract to call. this property is changed below
+  // TODO: having to set this property is a hack, it would be nice to
+  static Subcontract: new (...args: any) => Subcontract = SubcontractA;
```
MartinMinkov, line 47: Is this meant to be part of the above comment? The sentence just cuts off :sweat_smile:

mitschabaude: Heh thanks. Got distracted

src/lib/mina/test/dynamic-call.unit-test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+class Caller extends SmartContract {
+  @method
+  async call(a: UInt64, b: UInt64, address: PublicKey) {
+    const subcontract = new Caller.Subcontract(address);
+    await subcontract.submethod(a, b);
+  }
+
+  // subcontract to call. this property is changed below
+  // TODO: having to set this property is a hack, it would be nice to
```
mitschabaude, line 46:
```suggestion
  // TODO: having to set this property is a hack, it would be nice to pass the contract as parameter
```

PR #1639, opened 2024-05-08
Verification key permissions: better-named wrappers

mitschabaude:
This PR adds new wrappers for verification key permissions, which better convey their meaning.

This was suggested in the original VK permissions RFC: https://github.com/MinaProtocol/mina/pull/13994#discussion_r1321758306

It also makes setting VK permissions slightly nicer

### DIFF
--- a/src/examples/zkapps/zkapp-self-update.ts
+++ b/src/examples/zkapps/zkapp-self-update.ts
@@ -9,7 +9,6 @@ import {
   Mina,
   AccountUpdate,
   Provable,
-  TransactionVersion,
 } from 'o1js';
 
 class SelfUpdater extends SmartContract {
@@ -17,10 +16,7 @@ class SelfUpdater extends SmartContract {
     super.init();
     this.account.permissions.set({
       ...Permissions.default(),
-      setVerificationKey: {
-        auth: Permissions.proof(),
-        txnVersion: TransactionVersion.current(),
-      },
+      setVerificationKey: Permissions.VerificationKey.proofUntilHardfork(),
     });
   }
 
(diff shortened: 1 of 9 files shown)


src/examples/zkapps/zkapp-self-update.ts
```diff
@@ -9,18 +9,14 @@ import {
...
 class SelfUpdater extends SmartContract {
   init() {
     super.init();
     this.account.permissions.set({
       ...Permissions.default(),
-      setVerificationKey: {
-        auth: Permissions.proof(),
-        txnVersion: TransactionVersion.current(),
-      },
+      setVerificationKey: Permissions.VerificationKey.proofUntilHardfork(),
```
stevenpack, line 19: I appreciate the creativity to try get it in as a non-breaking change. It looks kind of awkward though to set that particularly permission differently to the rest.

WDYT about deprecating Impossible (so it still works, and the deprecation points to the new one. E.g.)

/**
 * @deprecated `Impossible` has been deprecated and will be removed in a future version. You should use ImpossibleUntilHardFork
 */
Permissons.Impossible

stevenpack: And add Permissions.ImpossibleUntilHardFork

mitschabaude: It was already looking differently than the rest though!

For any other permission, it's
```ts
Permission.impossible()
```
For verification keys, it used to be
```ts
{
  auth: Permission.impossible(),
  txnVersion: Transaction.current()
}
```

so I think it becomes less awkward with this change

stevenpack: Per slack, we resolved on `ImpossibleDuringCurrentVersion` as the most accurate and most descriptive (following the existing casing standards of course)

Trivo25 (approved):

PR #1660, opened 2024-05-20
Fix Int64 sign checks and mod

mitschabaude:
- add Int64 check method which doesn't allow ambiguous 0, to be enabled in v2
- fix `neg()` in the unique representation
- deprecate misleading `isPositive()` and replace with to v2 methods
- fixed version of Int64.mod()

### DIFF
--- a/src/lib/provable/int.ts
+++ b/src/lib/provable/int.ts
@@ -1227,9 +1227,24 @@ class Int64 extends CircuitValue implements BalanceChange {
     this.toField().assertEquals(y_.toField(), message);
   }
   /**
-   * Checks if the value is positive.
+   * @deprecated Use {@link isPositiveV2()} or {@link isNonNegativeV2()} instead.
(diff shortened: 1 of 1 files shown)


src/lib/provable/int.ts
```diff
@@ -1142,14 +1146,27 @@ class Int64 extends CircuitValue implements BalanceChange {
   }
 
   /**
-   * Negates the value.
-   *
-   * `Int64.from(5).neg()` will turn into `Int64.from(-5)`
+   * @deprecated Use {@link negV2()} instead.
```
Trivo25, line 1149: should we add this part to the changelog as well?

Trivo25 (approved):

src/lib/provable/int.ts
mitschabaude: ok we can do that

PR #1661, opened 2024-05-22
Add dynamic SHA256 update method

Shigoto-dev19:
## Description

- This implementation enables streaming preimage bytes into a provable SHA256 hash function.
- It is primarily an adaptation of [StableLib's](https://github.com/StableLib/stablelib/blob/a7bac13/packages/sha256/sha256.ts) SHA256 class, modified to be provable.

## Changes

- Added a `SHA256Hasher` class.
- Introduced an optional boolean `reverseEndianness` parameter to the `bytesToWords`, `wordsToBytes`, `wordToBytes`, and `bytesToWord` functions in the `bit-slices.ts` file.
- Created a `zkProgram` that demonstrates similar functionality to [SHA256Program](https://github.com/o1-labs/o1js/blob/main/src/examples/crypto/sha256/sha256.ts#L7-L18) but dynamically.
- Added sliding window tests inspired by [noble-hashes test](https://github.com/paulmillr/noble-hashes/blob/main/test/hashes.test.js#L472-L497) in accordance with the [noble approach to testing](https://github.com/paulmillr/noble-hashes/blob/main/test/README.md).

## Notes

- There is some redundant code in the `hashBlocks` method of the hasher class since the core logic is copied from the original `SHA256.hash()` function.
- The dynamic SHA256 is slightly less efficient than the original SHA256 hash function, likely due to byte processing:
  - For `Bytes12` input:
    - `SHA256.hash` --> 5036 rows
    - `SHA256.update` --> 5241 rows
- It's noteworthy that a fluent interface or method chaining works as provable in o1js, which opens up possibilities for various future implementations.
- Attempting to witness a hasher instance with `Provable.witness` is currently not possible, but it would be a valuable feature.
- Observing how noble-hashes use abstract classes with method chaining suggests that it might be possible to migrate many binary hash functions to be provable in o1js, similar to [noble-hashes](https://github.com/paulmillr/noble-hashes).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+- Added `SHA256Hasher` class that allows streaming preimage bytes into a provable SHA256 hash function.
+  - exposed via `Gadgets.SHA256.update(data)`
+
+### Added
+
 - Added `base64Encode()` and `base64Decode(byteLength)` methods to the `Bytes` class.
 
 ### Fixes
--- a/src/examples/crypto/sha256/run.ts
+++ b/src/examples/crypto/sha256/run.ts
@@ -1,5 +1,11 @@
-import { Bytes12, SHA256Program } from './sha256.js';
+import {
+  Bytes12,
+  SHA256Program,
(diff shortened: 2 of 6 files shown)


Shigoto-dev19:
This PR doesn't offer a solid addition to o1js.

PR #1665, opened 2024-05-24
docs: In the example for Field.inv() the variable `someField` has to be used instead of the non-existent variable `example`.

### LINKED ISSUE #1664: docs: In the example for Field.inv() the variable `someField` has to be used instead of the non-existent variable `example`
In the example for Field.inv() the variable `someField` has to be used instead of the non-existent variable `example`

nickfreeman-de:
docs: In the example for Field.inv() the variable `someField` has to be used instead of the non-existent variable `example`.

Fixes #1664


Trivo25:
Thank you!

nicc (approved):

PR #1671, opened 2024-06-03
IndexedMerkleMap: Support 0 and -1 keys

mitschabaude:
Stacked after https://github.com/o1-labs/o1js/pull/1666

## Why

This refactors how `IndexedMerkleMap` represents the maximum possible key. Previously, there was a hard-coded entry of `nextKey: -1`, i.e. the largest possible value, that was present from the beginning.

This had two limitations:
* you couldn't ever insert -1 as a key because according to the `nextKey` it was already included
* you couldn't properly use 0 as a key (e.g., you couldn't `set()` its value), because it didn't have a low node associated with it.

The first point is minor (we probably don't need -1 as a key) but the second point is very annoying, because 0 is the natural value for a dummy key. Setting key=0 to value=0 (the initial value) is a very convenient way to represent a dummy update, because it doesn't affect the tree at all. So with this PR, dummy updates of the Merkle map are supported out of the box without any additional logic.

## How

The changes here make both 0 and -1 usable like any other keys. (As before, the Merkle map is initially populated with (0, 0) as the only key-value pair.)

The issue is solved by doing the same [as the Aztec implementation](https://docs.aztec.network/aztec/concepts/storage/trees/indexed_merkle_tree) and making 0 act as both the smallest _and_ the largest key simultaneously. So, we always have `nextKey: 0` in the node with the largest currently stored key, and that node acts as a valid low node for the 0 key.

Implementation-wise, it just means adding a couple of special cases to the key comparison functions. The constraints increase but not significantly:
```
indexed merkle map (get) 461
indexed merkle map (get option) 990
sparse merkle map (get) 4208

indexed merkle map (insert) 1706
indexed merkle map (update) 905
indexed merkle map (set) 1893
sparse merkle map (set) 8160

indexed merkle map (assert included) 460
indexed merkle map (assert not included) 517
indexed merkle map (is included) 523
```

### DIFF
--- a/src/lib/provable/merkle-tree-indexed.ts
+++ b/src/lib/provable/merkle-tree-indexed.ts
@@ -102,19 +102,6 @@ class IndexedMerkleMapBase {
     readonly sortedLeaves: StoredLeaf[];
   }>;
 
-  clone() {
-    let cloned = new (this.constructor as typeof IndexedMerkleMapBase)();
-    cloned.root = this.root;
-    cloned.length = this.length;
-    cloned.data.updateAsProver(({ nodes, sortedLeaves }) => {
-      return {
-        nodes: nodes.map((row) => [...row]),
-        sortedLeaves: [...sortedLeaves],
-      };
-    });
-    return cloned;
-  }
-
   // we'd like to do `abstract static provable` here but that's not supported
   static provable: Provable<
     IndexedMerkleMapBase,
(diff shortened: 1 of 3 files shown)


src/lib/provable/merkle-tree-indexed.ts
```diff
@@ -701,3 +709,57 @@ function bisectUnique(
...
+
+// custom comparison methods where 0 can act as the min and max value simultaneously
+
+/**
+ * Assert that `x in (low, high)`, i.e. low < x < high, with the following exceptions:
+ *
+ * - high=0 is treated as the maximum value, so `x in (low, 0)` always succeeds if only low < x; except for x = 0.
+ * - x=0 is also treated as the maximum value, so `0 in (low, high)` always fails, because x >= high.
+ */
+function assertStrictlyBetween(
```
Trivo25, line 721: should we move that to somewhere close to the range check gadgets?

Trivo25 (approved):

src/lib/provable/merkle-tree-indexed.ts
mitschabaude: I don't know, because it's so very specialized to this use case

mitschabaude:
I cherry-picked a fix here so that `IndexedMerkleTree` is fully functional when merged

PR #1672, opened 2024-06-04
Use IndexedMerkleMap for OffchainState

mitschabaude:
This follows #1666 and #1671 and uses the new `IndexedMerkleMap` for `OffchainState`.

Also:
* ~~Fixes `IndexedMerkleMap.clone()`~~ (cherry-picked in previous PR)
* Adds `IndexedMerkleMap.{overwrite, overwriteIf}`

TODO:
* [x] Make tree height, actions per update and actions per batch parameters that can be overriden
* [x] Fix proving when there are multiple settlement proofs in a row

### DIFF
--- a/src/lib/mina/actions/offchain-contract.unit-test.ts
+++ b/src/lib/mina/actions/offchain-contract.unit-test.ts
@@ -2,7 +2,6 @@ import {
   SmartContract,
   method,
   Mina,
-  State,
   state,
   PublicKey,
   UInt64,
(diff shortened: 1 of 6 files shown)


src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -12,23 +11,22 @@ import assert from 'assert';
...
-const offchainState = OffchainState({
-  accounts: OffchainState.Map(PublicKey, UInt64),
-  totalSupply: OffchainState.Field(UInt64),
-});
+const offchainState = OffchainState(
+  {
+    accounts: OffchainState.Map(PublicKey, UInt64),
+    totalSupply: OffchainState.Field(UInt64),
+  },
+  { logTotalCapacity: 10, maxActionsPerProof: 5 }
```
mitschabaude, line 21: we change the defaults, mainly to check that this works and to exercise the case that settling needs multiple proofs

src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -12,23 +11,22 @@ import assert from 'assert';
...
 
 // example contract that interacts with offchain state
 
 class ExampleContract extends SmartContract {
-  // TODO could have sugar for this like
-  // @OffchainState.commitment offchainState = OffchainState.Commitment();
-  @state(OffchainStateCommitments) offchainState = State(
-    OffchainStateCommitments.empty()
-  );
+  @state(OffchainState.Commitments) offchainState = offchainState.commitments();
```
mitschabaude, line 29: this looks a bit nicer now, and was necessary because the empty state now depends on the tree size which is part of the offchain state config

Trivo25 (approved):

PR #1676, opened 2024-06-06
Batch Reducer

mitschabaude:
Introduces a new kind of reducer for the case that you only want to reduce actions in small batches.
Avoids deadlock when more actions than your batch size are pending.

This targets zkapps which want to create account updates for each action.

Example application: A token airdrop, where users submit "claim" actions concurrently, and we process them in a reducer by paying out the claims and simultaneously removing the claims from our storage (to avoid double-claims). 

The batch reducer unit test implements such an airdrop.

## Batch reducer API

A batch reducer is declared by specifying an action type and a batch size:
```ts
let batchReducer = new BatchReducer({ actionType: Action, batchSize: 5 });
```

The action class is a generic type argument:
```ts
batchReducer satisfies Experimental.BatchReducer<typeof Action>;
```

There are essentially 3 methods for interacting with `batchReducer`, the first of which is familiar

```ts
// dispatch an action
batchReducer.dispatch(action: Action): void;
```

Reducing actions comes with a separate preparation step outside your contract:

```ts
// outside contract: prepare array of batches of actions
batchReducer.prepareBatches(): Promise<Array<{ batch: ActionBatch<Action>; proof: Proof }>>;

// inside contract: process one batch of actions
batchReducer.processBatch({ batch, proof }, callback: (action: Action, isDummy: Bool) => void): void;
```

`processBatch()` is what you would call in your "reducer" method. If we want to catch up with the current chain, we have to call that method, in a separate transaction, for each of the batches returned by `prepareBatches()`.

> Note: I went for a `forEach()` type of API instead of `reduce()`, because the API surface is simpler and usually allows you to easily implement `reduce()` like flows anyway, by mutating a variable in scope.

A batch reducer uses 2 onchain state fields (one more than a classical reducer):
```ts
class Contract extends SmartContract {
  @state(Field)
  actionState = State(BatchReducer.initialActionState);
  @state(Field)
  actionStack = State(BatchReducer.initialActionStack);
}
```

## Batch reducer implementation

Under the hood, `processBatch()` creates a recursive proof of _reversing the action order_: Its output is a stack of actions where the top-most action is the next pending one, and the bottom-most action is the last one submitted to the chain.

The reason for this reversal is that popping the last element of a (Merkle) list can be implemented as a provable operation, which proves that the popped element was indeed part of the original list. So, once we created a reversed stack of actions, we can pop off the same stack for multiple transactions, without having to create the recursive proof again.

Reusing the same action stack for many transactions means that we have to store it on the contract. The `actionStack` stored on the contract is always guaranteed to consist only of pending actions in the correct (reversed) order; so, popping off this stack will never yield an invalid action. Once the stack runs out of actions, we have to call `prepareBatch()` again, to give us a fresh stack.

Note that I said that `prepareBatch()` creates a recursive proof of creating a new action stack. It being recursive means it can handle an arbitrary amount of pending actions, so we will always be able to catch up with the chain (note: stack proving is very efficient, it can do hundreds of action lists per proof).
As an optimization, however, we don't actually prove the full stack offchain. We leave a final chunk to be proved inside `processBatch()` itself. If this final chunk is large enough to build the entire stack, then we don't even need a recursive proof. `prepareBatch()` will figure out all of that and return quickly if no proof is needed.

To understand the `processBatch()` logic, note that it handles two different cases:
1. We are calling it with a freshly created stack of actions (to be completed within the method itself), which successfully connects us to the onchain action state. In this case, we overwrite the previous stack stored onchain (which typically will be exhausted or at least smaller than our new stack).
   - This is what happens in the first batch of the `batches` that `prepareBatches()` returns
2. We are calling it with no new stack but instructing it to use the onchain stack
   - This happens on all subsequent `batches`  that `prepareBatches()` returns

There is a lot of robustness in this design. We don't _have_ to finish publishing transactions for all the `batches` before calling `prepareBatches()` again. We could _always_ create a new stack before calling `processBatch()` -- worst case, the onchain stack still had some actions in it, so we wasted a small amount of recursive proving (note: this is only a waste if we even need a recursive proof).
Also, we will never break anything by instructing `processBatch()` to use the onchain stack -- worst case, it's empty and no progress is made.

## zkApp Testing framework

As an aside, this also introduces a little testing framework intended to make writing local blockchain tests easier and quicker. A "test" is modeled as a list of "instructions" that either:
* modify the current state of accounts in some way, for example by creating a transaction, or
* do an assertion on current state of accounts, for example checking balances or the contents of offchain state fields

## Other additions

In the spirit of always adding the methods to o1js that we need ourselves, I added a whole bunch of new API in this PR. See the changelog. For example, conditional versions of many existing APIs:
- `.requireEqualsIf()` for conditional preconditions
- `IndexedMerkleMap.setIf()`
- `Provable.assertEqualsIf()` // I use this all the time now!
- `AccountUpdate.createIf()`

These kind of APIs are especially useful for reducer-type logic where we can never do assertions and always have to handle invalid cases gracefully: by not doing anything. We need more of them!


src/lib/mina/actions/batch-reducer.ts
mitschabaude, line 1: main addition of this PR - `BatchReducer`!

src/lib/mina/actions/batch-reducer.unit-test.ts
mitschabaude, line 1: BatchReducer unit test, written with the new testing framework

src/lib/mina/actions/offchain-contract.unit-test.ts
mitschabaude, line 1: Offchain state unit test, rewritten to the framework

src/lib/mina/actions/offchain-state.ts
```diff
@@ -47,9 +55,15 @@ type OffchainState<Config extends { [key: string]: OffchainStateKind }> = {
...
-  ): void;
+  setContractInstance(contract: OffchainStateContract): void;
+
+  /**
+   * Set the smart contract class that this offchain state is connected with.
+   *
+   * This is an alternative for `setContractInstance()` which lets you compile offchain state before knowing an contract instance.
+   * However, you must call `setContractInstance()` before calling `createSettlementProof()`.
+   */
+  setContractClass(contract: OffchainStateContractClass): void;
```
mitschabaude, line 66: this is needed by the testing framework (and presumably other users as well), to compile the offchain state program before having a contract instance at hand

src/lib/mina/local-blockchain.ts
```diff
@@ -419,6 +420,9 @@ async function LocalBlockchain({
     setProofsEnabled(newProofsEnabled: boolean) {
       this.proofsEnabled = newProofsEnabled;
     },
+    resetProofsEnabled() {
+      this.proofsEnabled = originalProofsEnabled;
+    },
```
mitschabaude, line 425: easier to reset `proofsEnabled` to its original value than to figure out what the original value was and set it

src/lib/mina/test/test-contract.ts
mitschabaude, line 1: new zkapp testing framework

src/lib/mina/smart-contract-context.ts
```diff
@@ -23,3 +29,17 @@ function accountUpdateLayout() {
 
   return layout;
 }
+
+function contract<S extends SmartContract>(
+  expectedConstructor?: new (...args: any) => S
```
mitschabaude, line 34: convenience function for getting the currently active smart contract

src/lib/testing/equivalent.ts
```diff
@@ -84,9 +84,11 @@ type FuncSpec<In1 extends Tuple<any>, Out1, In2 extends Tuple<any>, Out2> = {
   to: ToSpec<Out1, Out2>;
 };
 
+type AnyTupleFunction = (...args: AnyTuple) => any;
+
 type SpecFromFunctions<
-  F1 extends AnyFunction,
-  F2 extends AnyFunction
+  F1 extends AnyTupleFunction,
+  F2 extends AnyTupleFunction
```
mitschabaude, line 91: only change that was needed because of the TS upgrade

package.json
```diff
@@ -91,7 +91,7 @@
     "typedoc": "^0.25.8",
     "typedoc-plugin-markdown": "^4.0.0-next.53",
     "typedoc-plugin-merge-modules": "^5.1.0",
-    "typescript": "5.1"
+    "typescript": "^5.4.5"
```
mitschabaude, line 94: upgraded TS because the latest version is much smarter with preserving narrowed types in callbacks, and that helped me avoid some awkward code structure in this PR

src/lib/mina/account-update.ts
```diff
@@ -1008,9 +1008,24 @@ class AccountUpdate implements Types.AccountUpdate {
     return new AccountUpdateTree({ accountUpdate, id, children });
   }
 
+  /**
+   * @deprecated Use {@link AccountUpdate.default} instead.
+   */
   static defaultAccountUpdate(address: PublicKey, tokenId?: Field) {
+    return AccountUpdate.default(address, tokenId);
```
mitschabaude, line 1015: it always annoyed me that `AccountUpdate.defaultAccountUpdate()` was so verbose, so I renamed it to `AccountUpdate.default()`

src/lib/mina/account-update.ts
```diff
@@ -1056,6 +1071,22 @@ class AccountUpdate implements Types.AccountUpdate {
...
     return accountUpdate;
   }
+
+  /**
+   * Create an account update that is added to the transaction only if a condition is met.
+   *
+   * See {@link AccountUpdate.create} for more information. In this method, you can pass in
+   * a condition that determines whether the account update should be added to the transaction.
+   */
+  static createIf(condition: Bool, publicKey: PublicKey, tokenId?: Field) {
```
mitschabaude, line 1081: exposing a well-known trick as easy to understand method

src/lib/mina/actions/batch-reducer-program.unit-test.ts
mitschabaude, line 1: lower-level unit test for the recursive part of batch action reducing

src/lib/util/global-context.ts
mitschabaude, line 1: Better error feedback, added for myself when I accidentally forgot to await a promise.

I think this massively improves the DX when hitting
```
It seems you're running multiple provers concurrently within the same JavaScript thread, which, at the moment, is not supported and would lead to bugs.
```

mitschabaude (commented):
Some comments to help reviewing!

mitschabaude:
btw I had some explanatory comments here: https://github.com/o1-labs/o1js/pull/1676#pullrequestreview-2125551470

src/lib/mina/account-update.ts
```diff
@@ -1056,6 +1071,22 @@ class AccountUpdate implements Types.AccountUpdate {
...
+    return AccountUpdate.create(
+      // if the condition is false, we use an empty public key, which causes the account update to be ignored
```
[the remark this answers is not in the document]
mitschabaude: not the protocol, just o1js. it will skip "dummy" account updates both when hashing the public input:
https://github.com/o1-labs/o1js/blob/6c52fea324cbe37607c88d03908157149165dac8/src/lib/mina/account-update.ts#L1540-L1542

and when forming the transaction:
https://github.com/o1-labs/o1js/blob/6c52fea324cbe37607c88d03908157149165dac8/src/lib/mina/account-update.ts#L1597-L1598

it's a bespoke system of enabling optional account updates that I added at some point because it was needed for something

src/lib/mina/actions/batch-reducer.ts
```diff
@@ -0,0 +1,768 @@
...
+
+  /**
+   * Process a batch of actions which was created by `prepareBatches()`.
+   *
+   * **Important**: The callback exposes the action's value along with an `isDummy` flag.
+   * This is necessary because we process a dynamically-sized list in a fixed number of steps.
+   * Dummies will be passed to your callback once the actual actions are exhausted.
+   * Make sure to write your code to account for dummies.
+   *
+   * **Warning**: Don't call `processBatch()` on two _different_ batches within the same method. The second call
```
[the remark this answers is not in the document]
mitschabaude: I think Martin's work on preventing accidental precondition overwriting will enforce it!

src/lib/mina/actions/batch-reducer.ts
```diff
@@ -0,0 +1,768 @@
...
+    let newActions = Actions.pushEvent(update.body.actions, fields);
+    update.body.actions = Provable.if(
+      condition,
+      Actions,
+      newActions,
+      update.body.actions
+    );
+  }
+
+  // TODO: `dispatchIf()` would be nice
```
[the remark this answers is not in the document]
mitschabaude: oops

Trivo25 (commented):
What do you think about adding a variant that does this without recursion for smaller batches, directly in the smart contract method that calls `reduce` - essentially a low level variation of the recursion one with less abstractions?

mitschabaude:
> What do you think about adding a variant that does this without recursion for smaller batches, directly in the smart contract method that calls `reduce` - essentially a low level variation of the recursion one with less abstractions?

The current implementation already does this! The smart contract method does the final "chunk" of the recursive work, which will be all of the work if there are few pending actions.

But we could have a variant which doesn't even need external input. Are you suggesting that?
On the other hand, I don't see how this would improve the DX because you still need the most general method with recursion either way.

Trivo25:
> But we could have a variant which doesn't even need external input. Are you suggesting that?

Yea, that's what I was referring to! A lightweight version that doesn't require any external inputs or proof/compilation dependencies.

mitschabaude:
> > But we could have a variant which doesn't even need external input. Are you suggesting that?
> 
> Yea, that's what I was referring to! A lightweight version that doesn't require any external inputs or proof/compilation/zkProgram dependencies.

Right, good point about saving the compilation step as well! In that case, it will have to be on a separate smart contract method though. But I can imagine people making use of that.

I'm going to revisit the PR next week to see if I can break up the reducer logic into more modular steps. then it should be very easy to add variations of `processBatch()` and even let people write their own variations with more low-level methods

mitschabaude:
well @Trivo25, on the OTHER hand.. I just remembered why I was convinced we would always need an external call 😅 

people need an external call to figure out how many batches there are == how many transactions to create.

the lightweight method doesn't really have a way to tell you that, or even tell you that no transactions are currently needed.

I'm still in favor of adding more flexibility but I can't see when you _wouldn't_ want to call an external method 🤔

CHANGELOG.md
```diff
@@ -17,6 +17,34 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/ed198f305...HEAD)
 
+### Breaking changes
+
+- Fixed a vulnerability in `OffchainState` where it didn't store the `IndexedMerkleTree` length onchain and left it unconstrained https://github.com/o1-labs/o1js/pull/1676
```
Trivo25, line 22: What were the consequences of not constraining the length?

mitschabaude: the `length` only constrains where people are allowed to insert new values. insertion prevents overwriting an existing value even if the length is unconstrained though. but, people could insert at random positions instead of linearly from the start, which could make the merkle tree harder to use and break assumptions of other users of it.

what's probable worst is that people could _hide_ where they are inserting leaves into the tree. so, they update the onchain root but then no one else can reconstruct that root because they don't know where new leafs were inserted. so it would make the contract unusable.

for example, in the offchain state implementation, everyone re-builds the tree on their own, assuming new leaves are inserted at the `length`, and it being unconstrained could have enabled an actor who calls `settle()` in a malicious way to break the offchain state root unrecoverable for anyone else.

mitschabaude: I thought it might be nice to hash the root and length together into a single commitment, maybe also called `root` for familiarity, so that there is only one onchain state to store and people can't forget about the length

Trivo25: Oh that makes sense! I was somehow under the assumption that length meant height and the height is already embedded in the witness structure and verification, but that's obviously something different.

MartinMinkov (commented):
Bigger review incoming! But in the meantime, do we not intend to make the changes for testing public? Is this intended to be an internal only tool? If not, what would you want to further build out to make it more publicity accessible (if it should)? Additionally, would you want to see this eventually include network support?

src/lib/mina/test/test-contract.ts
```diff
@@ -0,0 +1,319 @@
...
+            signers?: PrivateKey[];
+          }
+        | {
+            type: 'deploy';
+            options?: {
+              contract?: typeof SmartContract | SmartContract;
+              account?: Mina.TestPublicKey;
+            };
+          }
+        | { type: 'expect-state'; state: State; expected: Expected<State> }
```
MartinMinkov, line 245: What do you think about adding more `expect` cases here? A few ideas I can think of a few that could be helpful:

1. Expect token balance
- `{ type: 'expect-token-balance'; address: PublicKey; tokenId: Field; expected: bigint }`
2. Expect event
- `{ type: 'expect-event'; eventName: string; data?: any }`
3. Expect action
- `{ type: 'expect-action'; data?: any }`
4. Expect error
- `{ type: 'expect-error'; callback: () => Promise<void>; errorType?: typeof Error }`

Feel free to ignore if they don't seem interesting/worthwhile. Could always be added later as well

MartinMinkov (approved):
Looks good to me! I'll defer final review to @Trivo25 . As a side note, I'm attempting to run the example using lightnet, but having issues with the tool as of right now. If I can get it working, I'll post my findings if it works smoothly!

mitschabaude:
> Bigger review incoming! But in the meantime, do we not intend to make the changes for testing public? Is this intended to be an internal only tool?

I think we should make them public. It would be nice to to some polishing on it though,
* consider integration with node.js test runner
* support multiple contracts (the current version "natively" supports a single contract)
* ability to check for expected _on-chain_ states (currently you can only check offchain states and balances)
* in general better dogfooding: use it for more tests and see how it scales / what is missing
* think about a good name 😅 

I don't want to put the burden of polishing the testing framework to be ready for release on this already huge PR

src/lib/mina/test/test-contract.ts
mitschabaude: great point

expect error especially, that's completely missing from the current framework

src/lib/mina/actions/batch-reducer.unit-test.ts
```diff
@@ -0,0 +1,259 @@
...
+  maxUpdatesFinalProof: 4,
+  maxUpdatesPerProof: 4,
+});
+class Batch extends batchReducer.Batch {}
+class BatchProof extends batchReducer.BatchProof {}
+
+/**
+ * Contract that manages airdrop claims.
+ *
+ * WARNING: This airdrop design is UNSAFE against attacks by users that set their permissions such that sending them MINA is impossible.
```
ymekuria, line 51: Thanks for adding this warning!

src/lib/mina/test/test-contract.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    for (let instruction of instructions) {
+      await runInstruction(spec, instruction);
+    }
+  }
+
+  // create local instance and execute test
+  // if proofsEnabled is 'both', run the test with AND without proofs
+
+  console.log();
```
ymekuria, line 135: Is there a purpose for this empty `console.log`?

ymekuria (approved):
Nice work! That was a big PR. I added some questions and comments.

src/lib/mina/test/test-contract.ts
mitschabaude: but let's defer making the test framework production-ready to a different PR

src/lib/mina/actions/batch-reducer-program.unit-test.ts
```diff
@@ -0,0 +1,79 @@
...
+  actionStackProgram,
+  proveActionStack,
+} from './batch-reducer.js';
+import { Field } from '../../../index.js';
+import { expect } from 'expect';
+import { describe, it } from 'node:test';
+import { Actions as ActionsBigint } from '../../../bindings/mina-transaction/transaction-leaves-bigint.js';
+
+// analyze program with different number of actions
+for (let actionsPerProof of [10, 30, 100, 300, 1000]) {
```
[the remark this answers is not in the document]
mitschabaude: this is just there to see how many constraints various numbers of actions use, to inform how to set the defaults for `maxUpdatesPerProof` and `maxUpdatesFinalProof`

src/lib/mina/actions/batch-reducer.ts
```diff
@@ -0,0 +1,768 @@
...
+
+  /**
+   * Process a batch of actions which was created by `prepareBatches()`.
+   *
+   * **Important**: The callback exposes the action's value along with an `isDummy` flag.
+   * This is necessary because we process a dynamically-sized list in a fixed number of steps.
+   * Dummies will be passed to your callback once the actual actions are exhausted.
+   * Make sure to write your code to account for dummies.
```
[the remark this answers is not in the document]
mitschabaude: good idea, will do

src/lib/mina/actions/batch-reducer.ts
```diff
@@ -0,0 +1,768 @@
...
+  Action = InferProvable<ActionType>
+> {
+  batchSize: BatchSize;
+  actionType: Actionable<Action>;
+  Batch: ReturnType<typeof ActionBatch>;
+
+  program: ActionStackProgram;
+  BatchProof: typeof Proof<Field, ActionStackState>;
+
+  maxUpdatesFinalProof: number;
```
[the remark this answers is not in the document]
mitschabaude: good call, I'm adding lots of comments

src/lib/mina/actions/batch-reducer.ts
```diff
@@ -0,0 +1,768 @@
...
+  maxActionsPerUpdate: number;
+
+  constructor({
+    actionType,
+    batchSize,
+    maxUpdatesPerProof = 300,
+    maxUpdatesFinalProof = 100,
```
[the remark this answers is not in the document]
mitschabaude: I think the new doccomments shed some light on this. basically, trying to balance the following:
* should be able to process a lot of pending actions per proof, so we don't need many recursive proofs
* shouldn't make the circuit _too_ large, so we don't pay with extremely slow proofs even if we barely have any pending actions

src/lib/mina/test/test-contract.ts
mitschabaude: I thought it makes the log output look better with the extra line break :D 
Not sure if we want to leave the testing framework as verbose as it is currently when we release it though, probably should have a `verbose` flag

mitschabaude:
@Trivo25 I started this draft PR which exposes more modularity and a way to do everything without external inputs (marked as "unsafe" though).

Does this go in the direction of what you were thinking?

src/lib/util/assert.ts
```diff
@@ -10,3 +10,14 @@ function assertPromise<T>(value: Promise<T>, message?: string): Promise<T> {
   assert(value instanceof Promise, message ?? 'Expected a promise');
   return value;
 }
+
+/**
+ * Assert that the value is not undefined, return the value.
+ */
+function assertDefined<T>(
```
Trivo25, line 17: actually sad that type guards arent really useful in this scenario :(

Trivo25 (approved):

src/lib/util/assert.ts
mitschabaude: not sure if you meant that but this comment made me realized that I can remove `as T`!

PR #1687, opened 2024-06-16
Mina-signer: mention private key conversion function when PrivateKey.fromBase58() fails

### LINKED ISSUE #1568: Mina-signer: mention private key conversion function when `PrivateKey.fromBase58()` fails
Several users of mina-signer have been trying to use it with out of range private keys and ran into an error.

That's why we introduced a conversion function to support these private keys.
That function should be mentioned in the error message, whenever `PrivateKey.fromBase58()` fails with an error containing
```
Scalar: inputs larger than XXX are not allowed, got YYY
```

emirsoyturk:
closes #1568

### DIFF
--- a/src/mina-signer/src/field-bigint.ts
+++ b/src/mina-signer/src/field-bigint.ts
@@ -155,11 +155,15 @@ function checkRange(lower: bigint, upper: bigint, name: string) {
   return (x: bigint) => {
     if (x < lower)
       throw Error(
-        `${name}: inputs smaller than ${lower} are not allowed, got ${x}`
+        `${name}: inputs smaller than ${lower} are not allowed, got ${x} \
+        ${name === 'Scalar' ? 'You may want to use `convertPrivateKeyToBase58WithMod` before using this key.' : ''}
+        `
       );
     if (x >= upper)
       throw Error(
-        `${name}: inputs larger than ${upper - 1n} are not allowed, got ${x}`
+        `${name}: inputs larger than ${upper - 1n} are not allowed, got ${x} \
+        ${name === 'Scalar' ? 'You may want to use `convertPrivateKeyToBase58WithMod` before using this key.' : ''}
+        `
       );
   };
 }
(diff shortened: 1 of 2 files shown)


src/mina-signer/src/field-bigint.ts
```diff
@@ -155,11 +155,15 @@ function checkRange(lower: bigint, upper: bigint, name: string) {
   return (x: bigint) => {
     if (x < lower)
       throw Error(
-        `${name}: inputs smaller than ${lower} are not allowed, got ${x}`
+        `${name}: inputs smaller than ${lower} are not allowed, got ${x} \
+        ${name === 'Scalar' ? 'You may want to use `convertPrivateKeyToBase58WithMod` before using this key.' : ''}
+        `
```
mitschabaude, line 160: nice!
could you just remove the `\` and make it so that the error message starts at the beginning of the new line?
same for the "larger" case below

```suggestion
        `${name}: inputs smaller than ${lower} are not allowed, got ${x}.
${name === 'Scalar' ? 'You may want to use `convertPrivateKeyToBase58WithMod` before using this key.' : ''}
`
```

src/mina
mitschabaude, line 1: please revert the submodule change!

mitschabaude (changes requested):

src/mina-signer/src/field-bigint.ts
```diff
@@ -155,11 +155,21 @@ function checkRange(lower: bigint, upper: bigint, name: string) {
   return (x: bigint) => {
     if (x < lower)
       throw Error(
-        `${name}: inputs smaller than ${lower} are not allowed, got ${x}`
+        `${name}: inputs smaller than ${lower} are not allowed, got ${x}. ${
+          name === 'Scalar'
+            ? 'You may want to use `convertPrivateKeyToBase58WithMod` before using this key.'
+            : ''
```
mitschabaude, line 161: After thinking about it again, checking for `name == 'Scalar'` in this function doesn't seem to be a very precise condition for informing the user about `convertPrivateKeyToBase58WithMod()`. 

the way it's written it will show up on any scalar, not just private keys (for example, scalars that are part of signatures when deserializing a signature).

I would prefer an implementation which catches the error in functions that try to serialize private keys, and if that doesn't work because the input private keys doesn't fall within the range, add the info about `convertPrivateKeyToBase58WithMod()` to the error message

Issue #1686, opened 2024-06-14
Support feature flags for side-loaded keys


mitschabaude:
in o1js side-loaded keys, feature flags get hard-coded to None: https://github.com/o1-labs/o1js-bindings/blob/10e33c03648fa6d2201d2665cd2fde7064525ac8/ocaml/lib/pickles_bindings.ml#L763
which means they can't use optional custom gates!

the transaction snark, which verifies zkapp proofs which might use these gates as well, instead hard-codes all feature flags to be "Maybe":
https://github.com/MinaProtocol/mina/blob/8d45fa970786e58fc6778ec3bdbe5fec1b361b27/src/lib/transaction_snark/transaction_snark.ml#L610-L627

as a quick fix we could do the same in o1js. However, Matthew says that "maybe" incurs a lot of overhead and we should expose customizing that.

note how o1js computes the feature flags itself when compiling circuits. we would just have to save them and (optionally) provide them alongside the side-loaded key
https://github.com/o1-labs/o1js/blob/2b187fbe9c8d727b71e987b0477e98a2082fa54f/src/lib/proof-system/zkprogram.ts#L1040

Trivo25:
I need to confirm my assumption but I think we have to provide the developer with these configuration options:
- all `maybe` - can verify _any_ proof, no matter what custom gates it uses
- user defined compile constants - let the developer decide the shape of custom proofs their wrap circuit should be able to define

This is because we cannot provide feature flags dynamically to Pickles (at runtime), as doing so results in a change to the verification key of the wrap circuit. Instead of providing the feature flags as meta data alongside the verification key or proof, the developer would have to define this as part of the circuit configuration at compile time

Trivo25:
We could also require these constants as part of either the custom proof class, or additional meta data of the verification - but for side loading I believe it feels more natural to define these constants as part of the zkProgram, rather than "pseudo" dynamically in the verification key or proof

mitschabaude:
> I did some thinking, providing these constants as part of the custom proof class probably does feel better, since this way the developer describes the shape of the proof (or circuit) that their programs can side load - this is also quicker to implement. I dont think we can provide them alongside the verification key, as the feature flags need to be define on the branch/proof level

this sounds like the right approach to me!

we could also expose 3 ways to easily create the feature flags that the user puts on the dynamic proof class:
* all maybe
* all none
* compute from a given ZkProgram

PR #1688, opened 2024-06-17
Specify feature flags for `Pickles.sideLoaded`

Trivo25:
Allows of custom specification of feature flags under the `DynamicProof` class 

bindings https://github.com/o1-labs/o1js-bindings/pull/277

closes https://github.com/o1-labs/o1js/issues/1686

### DIFF
--- a/src/examples/zkprogram/dynamic-keys-merkletree.ts
+++ b/src/examples/zkprogram/dynamic-keys-merkletree.ts
@@ -42,6 +42,7 @@ const sideloadedProgram = ZkProgram({
   },
 });
 
+// given a zkProgram, we compute the feature flags that we need in order to verify proofs that were generated
 const featureFlags = await FeatureFlags.fromZkProgram(sideloadedProgram);
 
 class SideloadedProgramProof extends DynamicProof<Field, Field> {
@@ -49,7 +50,7 @@ class SideloadedProgramProof extends DynamicProof<Field, Field> {
   static publicOutputType = Field;
   static maxProofsVerified = 0 as const;
 
-  // we configure this side loaded proof to only accept proofs that doesn't use any feature flags (custom gates)
+  // we use the feature flags that we computed from the `sideloadedProgram` ZkProgram
   static featureFlags = featureFlags;
 }
 
--- a/src/examples/zkprogram/dynamic-keys-merkletree.ts
+++ b/src/examples/zkprogram/dynamic-keys-merkletree.ts
@@ -1,5 +1,6 @@
 import {
   DynamicProof,
+  FeatureFlags,
   Field,
   MerkleTree,
   MerkleWitness,
@@ -37,6 +38,9 @@ class SideloadedProgramProof extends DynamicProof<Field, Field> {
   static publicInputType = Field;
   static publicOutputType = Field;
   static maxProofsVerified = 0 as const;
+
+  // we configure this side loaded proof to only accept proofs that doesn't use any feature flags (custom gates)
+  static featureFlags = FeatureFlags.allNone;
 }
 
 const tree = new MerkleTree(64);
--- a/src/examples/zkprogram/dynamic-keys-merkletree.ts
+++ b/src/examples/zkprogram/dynamic-keys-merkletree.ts
@@ -31,8 +31,18 @@ const sideloadedProgram = ZkProgram({
         return publicInput.add(privateInput);
       },
     },
+    assertAndAdd: {
+      privateInputs: [Field],
+      async method(publicInput: Field, privateInput: Field) {
+        // this uses assert to test range check gates and their feature flags
+        publicInput.assertLessThanOrEqual(privateInput);
+        return publicInput.add(privateInput);
+      },
+    },
   },
 });
+
+const featureFlags = await FeatureFlags.fromZkProgram(sideloadedProgram);
 
 class SideloadedProgramProof extends DynamicProof<Field, Field> {
   static publicInputType = Field;
@@ -40,7 +50,7 @@ class SideloadedProgramProof extends DynamicProof<Field, Field> {
   static maxProofsVerified = 0 as const;
 
   // we configure this side loaded proof to only accept proofs that doesn't use any feature flags (custom gates)
-  static featureFlags = FeatureFlags.allNone;
+  static featureFlags = featureFlags;
 }
 
 const tree = new MerkleTree(64);
@@ -120,7 +130,7 @@ console.log('Compiling circuits...');
 const programVk = (await sideloadedProgram.compile()).verificationKey;
 const mainVk = (await mainProgram.compile()).verificationKey;
 
-console.log('Proving deployment of sideloaded key');
+console.log('Proving deployment of side-loaded key');
 const rootBefore = tree.getRoot();
 tree.setLeaf(1n, programVk.hash);
 const witness = new MerkleTreeWitness(tree.getWitness(1n));
@@ -147,4 +157,19 @@ const proof2 = await mainProgram.validateUsingTree(
 );
 
 const validProof2 = await verify(proof2, mainVk);
+console.log('ok?', validProof2);
+
+console.log('Proving different method of child program');
+const childProof2 = await sideloadedProgram.assertAndAdd(Field(0), Field(10));
+
+console.log('Proving verification inside main program');
+const proof3 = await mainProgram.validateUsingTree(
+  proof1.publicOutput,
+  proof1,
+  programVk,
+  witness,
+  SideloadedProgramProof.fromProof(childProof)
+);
+
+const validProof3 = await verify(proof2, mainVk);
 console.log('ok?', validProof2);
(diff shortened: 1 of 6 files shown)


src/lib/proof-system/zkprogram.ts
```diff
@@ -1230,17 +1363,16 @@ function computeFeatureFlags(gates: Gate[]): MlFeatureFlags {
...
-    MlBool(flags.foreignFieldMul),
-    MlBool(flags.xor),
-    MlBool(flags.rot),
-    MlBool(flags.lookup),
-    MlBool(flags.runtimeTables),
-  ];
+    ...Object.entries(flags).map(([_, flag]) =>
+      MlBool(typeof flag === 'boolean' ? flag : false)
+    ),
+  ] as MlFeatureFlags;
```
mitschabaude, line 1375: wait how can we just ignore the keys of the `flags` object? are we relying on a particular order?

oh I see that order is enforced by creating the full object in a separate function. that seems a bit brittle to me.

I also don't like how this refactor sacrifices type safety for a few saved lines of code.

IMO revert and adapt original to the undefined case, or at least make it independent of the order of the keys on `flags`

src/lib/proof-system/zkprogram.ts
```diff
@@ -83,6 +85,113 @@ const Empty = Undefined;
...
+  fromGates: (gates: Gate[]) => featureFlagsFromGates(gates),
+
+  /**
+   * Given a ZkProgram, return the feature flag configuration that fits the given program.
+   * This function considers all methods of the specified ZkProgram and finds a configuration that fits all.
+   */
+  fromZkProgram: fromZkProgram,
+};
+
+async function fromZkProgram<
```
mitschabaude, line 145: nit: give this function a name that refers to feature flags

src/lib/proof-system/zkprogram.ts
```diff
@@ -83,6 +85,113 @@ const Empty = Undefined;
...
+  rot: boolean | undefined;
+  lookup: boolean | undefined;
+  runtimeTables: boolean | undefined;
+};
+/**
+ * Feature flags indicate what custom gates are used in a proof of circuit.
+ * Side loading, for example, requires a set of feature flags in advance (at compile time) in order to verify and side load proofs.
+ * If the side loaded proofs and verification keys do not match the specified feature flag configurations, the verification will fail.
+ * Flags specified as `undefined` are considered as `maybe` by Pickles. This means, proofs can be sided loaded that can, but don't have to, use a specific custom gate.
+ * _Note:_ `Maybe` feature flags incur a small proving overhead.
```
mitschabaude, line 103:
```suggestion
 * _Note:_ `Maybe` feature flags incur a proving overhead.
```
matthew actually wrote "a lot of overhead"

src/lib/proof-system/zkprogram.ts
```diff
@@ -236,6 +345,22 @@ class DynamicProof<Input, Output> extends ProofBase<Input, Output> {
...
+  /**
+   * As the name indicates, feature flags are features of the proof system.
+   *
+   * If we want to side load proofs and verification keys, we first have to tell Pickles what _shape_ of proofs it should expect.
+   *
+   * For example, if we want to side load proofs that use foreign field arithmetic custom gates, we have to make Pickles aware of that by defining
+   * these custom gates.
+   *
+   * _Note:_ Only proofs that use the exact same composition of custom gates which were expected by Pickles can be verified using side loading.
+   * If you want to verify _any_ proof, no matter what custom gates it uses, you can use {@link FeatureFlags.allMaybe}. Please note that this might incur a small overhead.
```
mitschabaude, line 357:
```suggestion
   * If you want to verify _any_ proof, no matter what custom gates it uses, you can use {@link FeatureFlags.allMaybe}. Please note that this might incur significant overhead.
```

src/lib/proof-system/zkprogram.ts
```diff
@@ -1014,11 +1139,19 @@ function picklesRuleFromFunction(
       let computedTag: unknown;
       // Only create the tag if it hasn't already been created for this specific Proof class
       if (SideloadedTag.get(tag.name) === undefined) {
+        let featureFlags = [
+          0,
+          ...Object.entries(Proof.featureFlags).map(([_, flag]) =>
+            MlOption.mapTo(flag, MlBool)
+          ),
```
mitschabaude, line 1146: we can't rely on the order of keys in `Proof.featureFlags` here, because it can be user input. I would make a second version of `featureFlagsToMl` below with a different return type

src/snarky.d.ts
```diff
@@ -598,17 +598,6 @@ type Test = {
   };
 };
 
-type FeatureFlags = {
```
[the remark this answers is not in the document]
mitschabaude: good idea!

src/lib/proof-system/zkprogram.ts
```diff
@@ -236,6 +345,22 @@ class DynamicProof<Input, Output> extends ProofBase<Input, Output> {
...
+   * For example, if we want to side load proofs that use foreign field arithmetic custom gates, we have to make Pickles aware of that by defining
+   * these custom gates.
+   *
+   * _Note:_ Only proofs that use the exact same composition of custom gates which were expected by Pickles can be verified using side loading.
+   * If you want to verify _any_ proof, no matter what custom gates it uses, you can use {@link FeatureFlags.allMaybe}. Please note that this might incur a small overhead.
+   *
+   * You can also toggle specific feature flags manually by specifying them here.
+   * Alternatively, you can use {@FeatureFlags.fromZkProgram} to compute the set of feature flags that are compatible with a given program.
+   */
+  static featureFlags: FeatureFlags = FeatureFlags.allMaybe;
```
mitschabaude, line 362: isn't changing the default from `None` to `Maybe` a breaking change?

I do agree that maybe makes sense as the default

src/lib/proof-system/zkprogram.ts
Trivo25: But can we trust Matthew? :D

src/lib/proof-system/zkprogram.ts
Trivo25: yep thats right, didn't consider the ordering.. I reverted it, after all its not needed to massively change this anymore (a previous approach required a different structure)

src/lib/proof-system/zkprogram.ts
Trivo25: Yea it does, sadly.. I'll set it to `none` for now to maintain backwards compatibility :/ The good thing is developers will notice that they need to change something when proofs with different feature flags wont verify

src/lib/proof-system/zkprogram.ts
```diff
@@ -1230,16 +1353,33 @@ function computeFeatureFlags(gates: Gate[]): MlFeatureFlags {
     let flag = gateToFlag[gate.type];
     if (flag !== undefined) flags[flag] = true;
   }
+  return flags;
+}
+
+function featureFlagsToMlOption(
```
Trivo25, line 1359: I removed the old binary `true | false` feature flag data structure and replaced it entirely with the new option one, also in `picklesRuleFromFunction` so we dont have to maintain multiple of these conversion helpers

src/examples/zkprogram/dynamic-keys-merkletree.ts
```diff
@@ -30,13 +31,27 @@ const sideloadedProgram = ZkProgram({
...
+// given a zkProgram, we compute the feature flags that we need in order to verify proofs that were generated
+const featureFlags = await FeatureFlags.fromZkProgram(sideloadedProgram);
+
 class SideloadedProgramProof extends DynamicProof<Field, Field> {
   static publicInputType = Field;
   static publicOutputType = Field;
   static maxProofsVerified = 0 as const;
+
+  // we use the feature flags that we computed from the `sideloadedProgram` ZkProgram
+  static featureFlags = featureFlags;
```
MartinMinkov, line 54: Why must we pass this in if we are computing the feature flags based on the `ZkProgram`? Why wouldn't we just generate at runtime after compiling?

Trivo25: Good question! We have to know the feature flag configuration at compile time, in advance. 
We use this for side loading, which means we can, in theory, verify proofs from any zkProgram, no matter how exactly the circuit looks like. The ONLY thing that we need to know in advance are the feature flags! Hence we give the developer to either specify them manually, or compute them from _any_ zkProgram as sort of "blueprint", if that makes sense!

Trivo25: Oh, and the ZkProgram the developer wants to use as a blueprint could be any!

MartinMinkov: Makes sense to me! For some reason when I was reading this PR, I thought we were attaching this to the `ZkProgram` itself, not the side loaded proof!

mitschabaude (approved):
One final comment, otherwise lgtm!

PR #1689, opened 2024-06-17
Expose SHA256 low-level API

Shigoto-dev19:
## Description

- This PR exposes the low-level API of the SHA-256 hash function in o1js.
- It supports the development of `partialSHA` and `dynamicSHA` as external packages while leveraging the efficiencies of the low-level API.
- Closes #1668.

## Changes

- Separated the low-level SHA-256 functions and exported them as part of the SHA-256 gadgets in o1js.
- Added 2 importable functions and 1 constant:
  - **SHA-256 Compression Function**
    - Enables flexible block hashing with better control of `initial` and `intermediate` hashes.
  - **Message Schedule Computation Function** 
    - Computes the message schedule from `message blocks` to prepare it as input to the compression function.
  - **SHA256 Initial State Constant**
    - The initial hash values, also known as the SHA256 nothing-up-my-sleeve constants, are now exposed for enhanced flexibility in constructing SHA256 hashes.

### DIFF
--- a/src/lib/provable/gadgets/sha256.ts
+++ b/src/lib/provable/gadgets/sha256.ts
@@ -88,7 +88,7 @@ const SHA256 = {
     // padding the message $5.1.1 into blocks that are a multiple of 512
     let messageBlocks = padding(data);
 
-    let H = SHA256Constants.H.map((x) => UInt32.from(x));
+    let H = SHA256.initialState;
     const N = messageBlocks.length;
 
     for (let i = 0; i < N; i++) {
@@ -102,6 +102,9 @@ const SHA256 = {
   },
   compression: sha256Compression,
   createMessageSchedule,
+  get initialState() {
+    return SHA256Constants.H.map((x) => UInt32.from(x));
+  },
 };
 
 function Ch(x: UInt32, y: UInt32, z: UInt32) {
--- a/src/lib/provable/gadgets/sha256.ts
+++ b/src/lib/provable/gadgets/sha256.ts
@@ -94,22 +94,7 @@ const SHA256 = {
     const N = messageBlocks.length;
 
     for (let i = 0; i < N; i++) {
-      const M = messageBlocks[i];
-      // for each message block of 16 x 32bit do:
-      const W: UInt32[] = [];
-
-      // prepare message block
-      for (let t = 0; t <= 15; t++) W[t] = M[t];
-      for (let t = 16; t <= 63; t++) {
-        // the field element is unreduced and not proven to be 32bit, we will do this later to save constraints
-        let unreduced = DeltaOne(W[t - 2])
-          .value.add(W[t - 7].value)
-          .add(DeltaZero(W[t - 15]).value.add(W[t - 16].value));
-
-        // mod 32bit the unreduced field element
-        W[t] = UInt32.Unsafe.fromField(divMod32(unreduced, 16).remainder);
-      }
-
+      const W = computeMessageSchedule(messageBlocks[i]);
       H = sha256Compression(H, W, K);
     }
 
@@ -118,6 +103,7 @@ const SHA256 = {
     return Bytes.from(H.map((x) => wordToBytes(x.value, 4).reverse()).flat());
   },
   compression: sha256Compression,
+  computeMessageSchedule,
 };
 
 function Ch(x: UInt32, y: UInt32, z: UInt32) {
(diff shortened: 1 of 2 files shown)


mitschabaude (commented):
for visibility - some feedback given via DM
image

src/lib/provable/gadgets/sha256.ts
```diff
@@ -88,80 +88,22 @@ const SHA256 = {
     // padding the message $5.1.1 into blocks that are a multiple of 512
     let messageBlocks = padding(data);
 
-    const H = SHA256Constants.H.map((x) => UInt32.from(x));
+    let H = SHA256Constants.H.map((x) => UInt32.from(x));
```
Trivo25, line 91: Since these are constants, we should keep them that - you can create a second variable further down that takes the return value from `computeMessageScheduele`

src/lib/provable/gadgets/sha256.ts
```diff
@@ -88,80 +88,22 @@ const SHA256 = {
...
-      H[0] = H[0].addMod32(a);
-      H[1] = H[1].addMod32(b);
-      H[2] = H[2].addMod32(c);
-      H[3] = H[3].addMod32(d);
-      H[4] = H[4].addMod32(e);
-      H[5] = H[5].addMod32(f);
-      H[6] = H[6].addMod32(g);
-      H[7] = H[7].addMod32(h);
+      const W = computeMessageSchedule(messageBlocks[i]);
+      H = sha256Compression(H, W, K);
```
Trivo25, line 98: see comment above; do we have to K and H  in here as well? Why not get these constants in the method directly?

src/lib/provable/gadgets/sha256.ts
```diff
@@ -287,3 +229,87 @@ function sigma(u: UInt32, bits: TupleN<number, 3>, firstShifted = false) {
...
+
+/**
+ * Performs the SHA-256 compression function on the given hash values and message schedule.
+ *
+ * @param H - The initial or intermediate hash values (8-element array of UInt32).
+ * @param W - The message schedule (64-element array of UInt32).
+ * @param K - The constants used in the SHA-256 algorithm (64-element array of UInt32).
+ * @returns The updated intermediate hash values after compression.
+ */
+function sha256Compression(H: UInt32[], W: UInt32[], K: UInt32[]) {
```
Trivo25, line 241: note: this is the original code just moved

Trivo25 (commented):
Looks good so far, no major concerns except cleaning up and moving around the otherwise unused constants in the original method

Trivo25:
And, feel free to take the honor of adding your changes to the changelog!

src/lib/provable/gadgets/sha256.ts
Shigoto-dev19: Yes it is, should I add a notice in JSDoc?
It's also the case for `computeMessageSchedule`

src/lib/provable/gadgets/sha256.ts
Shigoto-dev19: Yes - we have `K` always which will be used internally but `H` only in the case of being an `initialState`

Trivo25: oh thats right, yep!

src/lib/provable/gadgets/sha256.ts
Trivo25: nope all good!

src/lib/provable/gadgets/sha256.ts
Shigoto-dev19: Before I push again, what do you think about the changes below

```ts
 const SHA256 = {
  hash(data: FlexibleBytes) {
    // preprocessing §6.2
    // padding the message $5.1.1 into blocks that are a multiple of 512
    let messageBlocks = padding(data);

    let H = SHA256.initialState;
    const N = messageBlocks.length;

    for (let i = 0; i < N; i++) {
      const W = createMessageSchedule(messageBlocks[i]);
      H = sha256Compression(H, W);
    }

    // the working variables H[i] are 32bit, however we want to decompose them into bytes to be more compatible
    // wordToBytes expects little endian, so we reverse the bytes
    return Bytes.from(H.map((x) => wordToBytes(x.value, 4).reverse()).flat());
  },
  compression: sha256Compression,
  createMessageSchedule,
  initialState: SHA256Constants.H.map((x) => UInt32.from(x)),
};
```

Trivo25: That works for me!

Shigoto-dev19: Exported `initialState` as a getter instead

```ts
get initialState() {
    return SHA256Constants.H.map((x) => UInt32.from(x));
  },
 ```

mitschabaude (approved):
great work!

PR #1699, opened 2024-06-25
Two new zkApp examples

mitschabaude:
This PR just adds two new examples for things that are not documented well
- example for withdrawing custom tokens from a zkapp
- example for linking of state between two zkapps

### DIFF
--- a/src/examples/zkapps/joint-update.ts
+++ b/src/examples/zkapps/joint-update.ts
@@ -1,0 +1,76 @@
+/**
+ * This is an example for two zkapps that are guaranteed to update their states _together_.
+ *
+ * So, A's state will updated if and only if B's state is updated, and vice versa.
+ *
+ * The difficulty here is that, while zkApps know and prove which other zkApps they call themselves,
+ * there's nothing in the protocol that lets them know which other zkApps are calling _them_.
+ *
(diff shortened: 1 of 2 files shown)


boray (approved):

src/examples/zkapps/joint-update.ts
```diff
@@ -0,0 +1,76 @@
...
+  SmartContract,
+  State,
+  method,
+  state,
+} from 'o1js';
+
+const aPubKey = PrivateKey.randomKeypair().publicKey;
+const bPubKey = PrivateKey.randomKeypair().publicKey;
+
+class A extends SmartContract {
```
45930, line 36: [NIT] May be more clear to name these something like `Parent` and `Child`, or if that's too overloaded `Main` and `Dependent`, rather than A and B.  A and B makes me imagine siblings with which I can interact one or the other.  In fact, we can only interact with B.

45930 (approved):

src/examples/zkapps/joint-update.ts
mitschabaude: Good point!

Trivo25 (approved):
just stamping so we can merge

PR #1700, opened 2024-06-25
Compute feature flag from list of zkProgram

Trivo25:
Many use cases require side loading to load proofs from a list of various zkPrograms. In order to do that, we have to provide the correct feature flags that fit all these zkPrograms - this PR introduces a function that does that

### DIFF
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -154,7 +154,7 @@ const FeatureFlags = {
 async function fromZkProgramList(programs: Array<AnalysableProgram>) {
   let flatMethodIntfs: Array<UnwrapPromise<ReturnType<typeof analyzeMethod>>> =
     [];
-  for await (const program of programs) {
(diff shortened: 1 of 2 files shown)


src/lib/proof-system/zkprogram.ts
```diff
@@ -135,23 +141,34 @@ const FeatureFlags = {
...
 
-async function featureFlagsfromZkProgram<
-  P extends {
-    analyzeMethods: () => Promise<{
-      [I in keyof any]: UnwrapPromise<ReturnType<typeof analyzeMethod>>;
-    }>;
+async function fromZkProgramList(programs: Array<AnalysableProgram>) {
+  let flatMethodIntfs: Array<UnwrapPromise<ReturnType<typeof analyzeMethod>>> =
+    [];
+  for await (const program of programs) {
```
mitschabaude, line 157: I don't think we need `for await` here do we?

mitschabaude (approved):
approved with one comment

PR #1703, opened 2024-06-25
Add multi-range check to `assertAlmostReduced`

MartinMinkov:
## Summary

Addressing an issue with `EcsdaSignature` and `ForeignCurve` overriding `check()` to use `assertAlmostReduced`. This skips the option to multi-range check all the input variables, which can lead to passing assertions if a malicious prover wishes to do so.

## Changes
- Added a range check on inputs within the `check` functions of `EcdsaSignature` and `ForeignCurve` 
- Update the regression verification key artifacts


src/lib/provable/foreign-field.ts
```diff
@@ -180,7 +180,7 @@ class ForeignField {
     Gadgets.ForeignField.assertAlmostReduced(
       xs.map((x) => x.value),
       this.modulus,
-      { skipMrc: true }
```
Trivo25, line 183: This is quite a deep change that might impact more than just ECDSA, what do others think about going one level higher and applying the multi range checks in the modified `check` methods of ECDSA? So, for example here

```diff
  static check(signature: EcdsaSignature) {
+   // add mrc on signature.r and signature.s
    // more efficient than the automatic check, which would do this for each scalar separately
    this.Curve.Scalar.assertAlmostReduced(signature.r, signature.s);
  }
  ```

mitschabaude: Yes agree, this is the wrong place

mitschabaude: The MRC check is done on each foreign field element automatically already, so it will always be a waste of constraints - _except_ inside `check()` methods

MartinMinkov: >The MRC check is done on each foreign field element automatically already, so it will always be a waste of constraints - except inside check() methods

Ah, TIL! Thanks for the input everyone, made those changes!

mitschabaude (commented):
@MartinMinkov the change looks good to me now, but it's breaking and we should ship it in a backwards-compatible way. `check()` is called internally just by using the overall class as method input. So it's _the class_ that we have to create an alternative version for.

How about this:
* Leave `ForeignCurve` class untouched
* New `ForeignCurveV2` which extends `ForeignCurve` and just overrides the `check()` method
* new `createForeignCurveV2()` which is exactly like `createForeignCurve()` but uses `ForeignCurveV2` 
* deprecate `createForeignCurve()`, move doccomments over to `createForeignCurveV2()`

And same for `EcdsaSignature`!

mitschabaude (approved):
Perfect!

PR #1704, opened 2024-06-26
Add frame bit to encryption method

Trivo25:
This PR introduces a naive `frame bit` to the `encrypt` and `decrypt` method. The last message block is marked by an additional `1`

### DIFF
--- a/src/lib/provable/crypto/encryption.ts
+++ b/src/lib/provable/crypto/encryption.ts
@@ -74,9 +74,6 @@ function encryptInternal(
   return { publicKey, cipherText };
 }
 
-/**
- * Decrypts a {@link CipherText} using a {@link PrivateKey}.
- */
 function decryptInternal(
   { publicKey, cipherText }: CipherText,
   privateKey: PrivateKey,
--- a/src/lib/provable/crypto/encryption.ts
+++ b/src/lib/provable/crypto/encryption.ts
@@ -27,6 +27,7 @@ function encrypt(message: Field[], otherPublicKey: PublicKey) {
   for (let i = 0; i < message.length; i++) {
     let keyStream = sponge.squeeze();
     let encryptedChunk = message[i].add(keyStream);
+    if (message.length - 1 === i) encryptedChunk = encryptedChunk.add(1);
     cipherText.push(encryptedChunk);
     // absorb for the auth tag (two at a time for saving permutations)
     if (i % 2 === 1) sponge.absorb(cipherText[i - 1]);
@@ -58,6 +59,7 @@ function decrypt(
   for (let i = 0; i < cipherText.length; i++) {
     let keyStream = sponge.squeeze();
     let messageChunk = cipherText[i].sub(keyStream);
+    if (cipherText.length - 1 === i) messageChunk = messageChunk.sub(1);
     message.push(messageChunk);
     if (i % 2 === 1) sponge.absorb(cipherText[i - 1]);
     if (i % 2 === 1 || i === cipherText.length - 1)
--- a/src/lib/provable/crypto/encryption.ts
+++ b/src/lib/provable/crypto/encryption.ts
@@ -42,6 +42,36 @@ function encrypt(message: Field[], otherPublicKey: PublicKey) {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +72,64 @@ function decrypt(
...
+
+  let sponge = new Poseidon.Sponge();
+  sponge.absorb(sharedSecret.x); // don't think we need y, that's enough entropy
+
+  // encryption
+  let cipherText = [];
+  for (let i = 0; i < message.length; i++) {
+    let keyStream = sponge.squeeze();
+    let encryptedChunk = message[i].add(keyStream);
+    if (message.length - 1 === i) encryptedChunk = encryptedChunk.add(1);
```
Trivo25, line 93: this is the only line added

src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +72,64 @@ function decrypt(
...
+  let sponge = new Poseidon.Sponge();
+  sponge.absorb(sharedSecret.x);
+  let authenticationTag = cipherText.pop();
+
+  // decryption
+  let message = [];
+  for (let i = 0; i < cipherText.length; i++) {
+    let keyStream = sponge.squeeze();
+    let messageChunk = cipherText[i].sub(keyStream);
+    if (cipherText.length - 1 === i) messageChunk = messageChunk.sub(1);
```
Trivo25, line 125: this is the only line added

src/lib/provable/crypto/encryption.ts
```diff
@@ -40,6 +42,38 @@ function encrypt(message: Field[], otherPublicKey: PublicKey) {
...
+
+  let sponge = new Poseidon.Sponge();
+  sponge.absorb(sharedSecret.x); // don't think we need y, that's enough entropy
+
+  // encryption
+  let cipherText = [];
+  for (let i = 0; i < message.length; i++) {
+    let keyStream = sponge.squeeze();
+    let encryptedChunk = message[i].add(keyStream);
+    if (message.length - 1 === i) encryptedChunk = encryptedChunk.add(1);
```
Trivo25, line 61: this is the only line added

src/lib/provable/gadgets/foreign-field.ts
```diff
@@ -739,8 +739,7 @@ class Sum {
 function assertLessThan(x: Field3, y: bigint | Field3) {
   let y_ = Field3.from(y);
 
-  // constant case
-
+  // constant case, y = constant, x = constant
```
Trivo25, line 742: this is just a left over recommendation to adjust the comments as well

mitschabaude (commented):
What you're doing is a bit different than the recommendation.
I think the goal of the rec is to make the final message chunk unambiguously different from the other ones. To do so, the final chunk is (0, msg[i)) while the other chunks are (1, msg[i]).

I don't think it's a problem to swap 0 and 1 here, but from what I can tell, just adding 1 doesn't achieve the goal of distinguishing the final chunk.

Caveat: I didn't take the time to understand the math of the attack and remediation.

mitschabaude:
To replicate the frame bit exactly, we'd need to completely redo the encryption method since the inputs can't be full field elements; they'd need to be at most 253-bit chunks.

Another option could be to understand the motivation for the frame bit at a fundamental level and find an alternative remediation to the attack which is more circuit-friendly.

mitschabaude:
I'd probably do a new version where the message is required to be `Bytes`, and then accumulate 31 bytes = 248 bits into one chunk. That still leaves room for the frame bit without overflowing the field.

And arguably, bytes is a more natural representation of an arbitrary message than field elements

mitschabaude:
> I'd probably do a new version where the message is required to be `Bytes`, and then accumulate 31 bytes = 248 bits into one chunk.

Or maybe not bytes but something more flexible like `(Bool | UInt8 | UInt32 | UInt64)[]`. that makes it easier to use (and more efficient) with large elements like individual uint64s. And the encryption method packs as many inputs as possible into 1 chunk such that there is still room for a frame bit.

Trivo25:
https://github.com/o1-labs/o1js/pull/1729

PR #1705, opened 2024-06-27
ScalarField and ShiftedScalar conversions

rpanic:
This PR adds `ScalarField`, which enabled ForeignField-based representation of Pallas scalars and implements conversion functions between `ScalarField` and `Scalar` (the shifted representation)

### DIFF
--- a/src/lib/provable/scalar-field.ts
+++ b/src/lib/provable/scalar-field.ts
@@ -1,0 +1,44 @@
+import { createField } from './core/field-constructor.js';
+import { createForeignField } from './foreign-field.js';
+import { field3ToShiftedScalar } from './gadgets/native-curve.js';
+import { Provable } from './provable.js';
+import { Scalar } from './scalar.js';
+
+export { ScalarField };
+
+/**
+ * ForeignField representing the scalar field of Pallas and the base field of Vesta
+ */
+class ScalarField extends createForeignField(
+  28948022309329048855892746252171976963363056481941647379679742748393362948097n
+) {
+  /**
+   * Provable method to conver a {@link ScalarField} into a {@link Scalar}
+   *
+   * This is always possible and unambiguous, since the scalar field is larger than the base field.
+   */
+  public toScalar(): Scalar {
+    const field3 = this.value;
+    const shiftedScalar = field3ToShiftedScalar(field3);
+    return Scalar.fromShiftedScalar(shiftedScalar);
+  }
+
+  /**
+   * Converts this {@link Scalar} into a {@link ScalarField}
+   */
+  static fromScalar(s: Scalar): ScalarField {
+    if (s.lowBit.isConstant() && s.high254.isConstant()) {
+      return new ScalarField(
+        ScalarField.fromBits(createField(s.toBigInt()).toBits())
+      );
+    }
+    const field = Provable.witness(ScalarField.provable, () => {
+      return ScalarField.fromBits(createField(s.toBigInt()).toBits())
+    });
+    const foreignField = new ScalarField(field);
+    const scalar = foreignField.toScalar();
+    Provable.assertEqual(Scalar, s, scalar);
+
+    return foreignField;
+  }
+}
--- a/src/lib/provable/scalar-field.ts
+++ b/src/lib/provable/scalar-field.ts
@@ -1,6 +1,5 @@
 import { Fq } from '../../bindings/crypto/finite-field.js';
-import { createField } from './core/field-constructor.js';
-import { createForeignField } from './foreign-field.js';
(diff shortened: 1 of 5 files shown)


src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+import { Provable } from './provable.js';
+import { Scalar } from './scalar.js';
+
+export { ScalarField };
+
+/**
+ * ForeignField representing the scalar field of Pallas and the base field of Vesta
+ */
+class ScalarField extends createForeignField(
+  28948022309329048855892746252171976963363056481941647379679742748393362948097n
```
mitschabaude, line 13: can you import this constant? e.g. `Fq.modulus` where `Fq` is imported like in `scalar.ts`

src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+) {
+  /**
+   * Provable method to conver a {@link ScalarField} into a {@link Scalar}
+   *
+   * This is always possible and unambiguous, since the scalar field is larger than the base field.
```
mitschabaude, line 18: fix doccomments
```suggestion
   * Provable method to convert a {@link ScalarField} into a {@link Scalar}
```

src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+    return Scalar.fromShiftedScalar(shiftedScalar);
+  }
+
+  /**
+   * Converts this {@link Scalar} into a {@link ScalarField}
+   */
+  static fromScalar(s: Scalar): ScalarField {
+    if (s.lowBit.isConstant() && s.high254.isConstant()) {
+      return new ScalarField(
+        ScalarField.fromBits(createField(s.toBigInt()).toBits())
```
mitschabaude, line 32: I think this should work, and be simpler and more efficient:
```suggestion
        ScalarField.from(s.toBigInt())
```

src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+      );
+    }
+    const field = Provable.witness(ScalarField.provable, () => {
+      return ScalarField.fromBits(createField(s.toBigInt()).toBits())
```
mitschabaude, line 36: I think this should be enough, since `bigint` is the "JS value" representation of `ForeignField`:
```suggestion
      return s.toBigInt()
```

src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+ */
+class ScalarField extends createForeignField(
+  28948022309329048855892746252171976963363056481941647379679742748393362948097n
+) {
+  /**
+   * Provable method to conver a {@link ScalarField} into a {@link Scalar}
+   *
+   * This is always possible and unambiguous, since the scalar field is larger than the base field.
+   */
+  public toScalar(): Scalar {
```
mitschabaude, line 20: maybe there should be a static version of this method, which can be used on instances of the base `ForeignField`?
so we don't have to convert with `new ScalarField()` before doing `.toScalar()`?

src/lib/provable/scalar-field.ts
```diff
@@ -0,0 +1,44 @@
...
+import { field3ToShiftedScalar } from './gadgets/native-curve.js';
+import { Provable } from './provable.js';
+import { Scalar } from './scalar.js';
+
+export { ScalarField };
+
+/**
+ * ForeignField representing the scalar field of Pallas and the base field of Vesta
+ */
+class ScalarField extends createForeignField(
```
mitschabaude, line 12: to make this class more usable, I recommend adding wrappers for the most important static methods, at least `.from()`, to make it return a `ScalarField`

src/lib/provable/scalar-field.ts
rpanic: I don't fully understand how putting that function into `ForeignField` would work, since this method can only exist if the modulus is exactly Fq.modulus, which is only true in the `ScalarField`.

mitschabaude: I mean put a static version of this method on this class

src/lib/provable/scalar-field.ts
rpanic: Unfortunately, that is not possible due the overloading of the base `.from()`. I added the static method with a `ForeignField` as parameters, I think that should give enough convenience in constructing and working with them.

mitschabaude: right, because the `.from()` return is canonical! ok

mitschabaude (approved):

PR #1711, opened 2024-06-28
Fix for mina-signer pkg-pr-new-publish CI job

shimkiv:


.github/workflows/pkg-pr-new-publish.yml
```diff
@@ -29,24 +29,24 @@ jobs:
...
-      - name: Build mina-signer
-        run: |
-          git submodule update --init --recursive
-          npm ci
-          cd src/mina-signer
-          npm ci
-          npm run prepublishOnly
-      - name: Publish mina-signer on pkg-pr-new
-        run: npx pkg-pr-new publish # Enable `--compact` once published to NPM with `repository` in package.json
+  # pkg-pr-new-mina-signer:
```
MartinMinkov, line 32: Can you leave a comment on why this is commented out and any actions that should take place before enabling it?

shimkiv: Actually let's try to release sequentially.

shimkiv: Seems to work better, but for some reason it releases `o1js` instead of `mina-signer`
https://github.com/o1-labs/o1js/actions/runs/9722423299
I'll raise this issue with the tool maintainers.

shimkiv: And final build with fixes https://github.com/o1-labs/o1js/actions/runs/9723324883

PR #1712, opened 2024-06-28
Assert Preconditions are not already set when trying to set their values

MartinMinkov:


src/lib/mina/precondition.ts
```diff
@@ -553,6 +555,27 @@ function getPreconditionContextExn(accountUpdate: AccountUpdate) {
   return c;
 }
 
+function assertPreconditionNotSet(property: any) {
+  if ('isSome' in property && property.isSome.equals(true).toBoolean()) {
```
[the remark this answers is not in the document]
mitschabaude: for example, the batch reducer PR introduces `requireEualsIf(condition, ...)` which sets `isSome = condition` i.e. a variable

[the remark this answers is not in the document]
mitschabaude: and with that check, `requireNothing()` works as an escape hatch because it resets `isSome` to `Bool(false)`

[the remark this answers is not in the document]
MartinMinkov: Fixed! https://github.com/o1-labs/o1js/pull/1712/commits/8171b5382267b4860e63d74fcc2c1987bb367642

src/lib/mina/precondition.ts
```diff
@@ -553,6 +558,49 @@ function getPreconditionContextExn(accountUpdate: AccountUpdate) {
...
+      Example of Correct Usage:
+      // Incorrect Usage:
+      timestamp.requireBetween(newUInt32(0n), newUInt32(2n));
+      timestamp.requireBetween(newUInt32(1n), newUInt32(3n));
+
+      // Correct Usage:
+      timestamp.requireBetween(newUInt32(0n), newUInt32(3n));
+    `;
+  if (!property.isSome.isConstant() || property.isSome.toBoolean()) {
+    property.isSome.assertEquals(Bool(true), errorMessage);
```
mitschabaude, line 594: To be more general I think you should also pass in the new `isSome` to this method, and here you should do
```suggestion
    property.isSome.assertEquals(isSome, errorMessage);
```

(this is currently not needed, but needed after the batch reducer PR which adds conditional precondition)

src/lib/mina/precondition.ts
```diff
@@ -553,6 +558,49 @@ function getPreconditionContextExn(accountUpdate: AccountUpdate) {
...
+      consider refactoring your code to consolidate all assertions for the same field before setting the precondition.
+
+      Example of Correct Usage:
+      // Incorrect Usage:
+      timestamp.requireBetween(newUInt32(0n), newUInt32(2n));
+      timestamp.requireBetween(newUInt32(1n), newUInt32(3n));
+
+      // Correct Usage:
+      timestamp.requireBetween(newUInt32(0n), newUInt32(3n));
+    `;
```
mitschabaude, line 592: nice error message!

two comments:
* it would help with debugging if the error message could include something about where this is called on.
  * i.e. name of the state field it's called on, or name of the precondition field
  * to make it work for onchain state, you might need to move the logic outside `AccountUpdate.assertEquals()`
* can you remove the indent?

src/lib/mina/precondition.ts
```diff
@@ -364,13 +366,16 @@ function preconditionSubclass<
...
-          property.value.upper = value;
+        let isInterval = 'lower' in property.value && 'upper' in property.value;
+        let newValue = isInterval ? { lower: value, upper: value } : value;
+        ensureConsistentPrecondition(property, newValue);
+        if (isInterval) {
+          (property.value as { lower: U; upper: U }).lower = value;
+          (property.value as { lower: U; upper: U }).upper = value;
         } else {
           property.value = value;
         }
```
mitschabaude, line 377: could be just this right?
```suggestion
        property.value = newValue;
```

src/lib/mina/precondition.ts
MartinMinkov: fixed in https://github.com/o1-labs/o1js/pull/1712/commits/9196fd50110847206f61dfd1f1b27f0f5a3c6efd and https://github.com/o1-labs/o1js/pull/1712/commits/e5350926c0d99f1e3a45ec6c10851d2eae30a173#diff-e4f43c493710dee3354495e0b987791eeb910288e8faae32094af6bd17bcb5e6R225!

src/lib/mina/precondition.ts
MartinMinkov: fixed here! https://github.com/o1-labs/o1js/pull/1712/commits/d3b99acbbfb79c47db28d98b9a0df6d3b3541548

MartinMinkov:
Question for the reviewer, am  I missing anywhere that `ensureConsistentPrecondition` that should be called?

src/lib/mina/precondition.ts
```diff
@@ -553,6 +553,56 @@ function getPreconditionContextExn(accountUpdate: AccountUpdate) {
...
+Ensure that preconditions for '${name}' are set in a single place and are not overwritten. If you need to update a precondition,
+consider refactoring your code to consolidate all assertions for '${name}' before setting the precondition.
+
+Example of Correct Usage:
+// Incorrect Usage:
+timestamp.requireBetween(newUInt32(0n), newUInt32(2n));
+timestamp.requireBetween(newUInt32(1n), newUInt32(3n));
+
+// Correct Usage:
+timestamp.requireBetween(newUInt32(0n), newUInt32(3n));
```
mitschabaude, line 594: small nit: this is what I would've expected to be the single `requireBetween()` call equivalent to the two above:
also, you're missing the space in `new UInt32`
```suggestion
timestamp.requireBetween(new UInt32(1n), new UInt32(2n));
```

src/lib/mina/precondition.ts
```diff
@@ -553,6 +553,56 @@ function getPreconditionContextExn(accountUpdate: AccountUpdate) {
...
+`;
+    property.isSome.equals(newIsSome).assertTrue(errorMessage);
+    if ('lower' in property.value && 'upper' in property.value) {
+      property.value.lower.equals(value.lower).assertTrue(errorMessage);
+      property.value.upper.equals(value.lower).assertTrue(errorMessage);
+    } else {
+      property.value.equals(value).assertTrue(errorMessage);
```
mitschabaude, line 601: `assertEquals()` shouldn't be changed to `equals().assertTrue()`, because that uses more constraints (0 vs 2)

Why is the `assertTrue()` error better? `assertEquals()` does include the input error message?

mitschabaude (approved):
great stuff, the only important comment I have left is about `assertEquals()` vs `equals().assertTrue()`. approving in advance!

EDIT: see another comment after this review about a missing `ensureConsistentPrecondition()` call

mitschabaude:
> Question for the reviewer, am I missing anywhere that `ensureConsistentPrecondition` that should be called?

@MartinMinkov Actually, now that you asked I had a second look: You're missing it in `GlobalSlot` -> `requireBetween()` (it's a precondition separate from network and account, also defined in `precondition.ts`)

mitschabaude:
@MartinMinkov I merged the batch reducer PR, would be nice if you can merge main and also call `ensureConsistentPrecondition()` in `requireEqualsIf()`, in both `state.ts` and `precondition.ts`

src/lib/mina/precondition.ts
MartinMinkov: > assertEquals() does include the input error message?

It seemingly doesn't in some cases. Due to the weak `any` typing, it seems like the `assertEquals` being used differs for certain types, like `CircuitValue`, where the error message isn't implemented. Using the `assertTrue` method made sure to always have the error message present.

MartinMinkov:
@mitschabaude Seems like something is broken with the Batch Reducer :O https://github.com/o1-labs/o1js/actions/runs/9879696858/job/27286485188?pr=1712#step:4:263

I'll take a look soonish, just wanted to notify

src/lib/mina/precondition.ts
mitschabaude: right, I just looked at `Field`. this would be nice to fix of course, it should have a good error message everywhere

src/lib/mina/precondition.ts
```diff
@@ -408,25 +408,30 @@ function preconditionSubclass<
...
-        property.value.upper = Provable.if(
+        let upper = Provable.if(
           condition,
           fieldType,
           value,
           defaultUpper(fieldType) as U
         );
+        ensureConsistentPrecondition(property, condition, value, longKey);
+        property.value.lower = lower;
+        property.value.upper = upper;
```
mitschabaude, line 431: this looks problematic to me, you're passing `value` to `ensureConsistentPrecondition()` which is potentially different than `lower` and `upper`

src/lib/mina/precondition.ts
```diff
@@ -408,25 +408,30 @@ function preconditionSubclass<
...
       } else {
-        property.value = Provable.if(
+        ensureConsistentPrecondition(property, condition, value, longKey);
+        let newValue = Provable.if(
```
mitschabaude, line 434: this looks wrong, you should pass `newValue` to `ensureConsistentPrecondition()`

MartinMinkov:
> @mitschabaude Seems like something is broken with the Batch Reducer :O https://github.com/o1-labs/o1js/actions/runs/9879696858/job/27286485188?pr=1712#step:4:263 I'll take a look soonish, just wanted to notify

User error on my part :facepalm:

mitschabaude (approved):

PR #1713, opened 2024-06-28
First Pass at Non-provable `SelfProof` Counting + Event

harrysolovay:
Co-authored by @yunus433

Heavily WIP: updates `ZkProgram` and the `method` decorator such that the number of aggregated proofs is tracked and then emitted within a standard event, of which block explorers and other such services could make use.

### DIFF
--- a/src/lib/mina/zkapp.ts
+++ b/src/lib/mina/zkapp.ts
@@ -166,7 +166,7 @@ function method<K extends string, T extends SmartContract>(
           (v: unknown): v is ZkProgram<any, any> => v instanceof ZkProgram
         )
         .reduce(
-          (acc: number, cur: ZkProgram<any, any>) => acc + cur.callsCount,
+          (acc: number, cur: ZkProgram<any, any>) => acc + cur.getCallsCount(),
           0
         )
     );
@@ -581,6 +581,7 @@ function computeCallData(
 }
 
 export class MetadataEvent extends Struct({ callsCount: Field }) {
+  static index = Field.ORDER - 1n;
   static key = 'unsafe-metadata-event';
 }
 
@@ -990,11 +991,15 @@ super.init();
   /**
    * Emits an event. Events will be emitted as a part of the transaction and can be collected by archive nodes.
    */
-  emitEvent<K extends keyof this['events']>(type: K, event: any) {
-    console.log({ type, event: event.toJSON() });
+  emitEvent<K extends keyof this['events'] | (typeof MetadataEvent)['key']>(
+    type: K,
+    event: any
+  ) {
     let accountUpdate = this.self;
     let eventTypes: (keyof this['events'])[] = Object.keys(this.events);
-    if (eventTypes.length === 0)
+    // Checks for 1 (not 0) because the `{[MetadataEvent.key]: MetadataEvent}` entry is injected
+    // into the `this.events` record (within the constructor).
+    if (type !== MetadataEvent.key && eventTypes.length === 1)
       throw Error(
         'emitEvent: You are trying to emit an event without having declared the types of your events.\n' +
           `Make sure to add a property \`events\` on ${this.constructor.name}, for example: \n` +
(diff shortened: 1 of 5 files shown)


harrysolovay:
This currently results in a `Cannot read properties of undefined (reading 'toFields') error`.
It seems the `MetadataEvent` constructor has the static `toFields` method (given it inherits from the Struct factory call). It is present when logged out as well, so I suspect the offending `toFields`-lacking item is not the `MetadataEvent`. The complete error + stacktrace doesn't make clear where this offending item lives:

```
compile (trivial zkapp)
/Users/harrysolovay/Desktop/o1js/dist/node/bindings/compiled/_node_bindings/o1js_node.bc.cjs:6761
         throw err;
         ^

TypeError: Cannot read properties of undefined (reading 'toFields')
    at assertEqualImplicit (o1js/dist/node/lib/provable/provable.js:203:16)
    at Object.assertEqual (o1js/dist/node/lib/provable/provable.js:195:16)
    at PublicKey.assertEquals (o1js/dist/node/lib/provable/types/circuit-value.js:104:18)
    at Object.wrappedMethod [as value] (o1js/dist/node/lib/mina/zkapp.js:299:42)
    at async TrivialZkapp.wrappedMethod (o1js/dist/node/lib/mina/zkapp.js:144:38)
    at async file:///Users/harrysolovay/Desktop/o1js/dist/node/lib/mina/zkapp.js:853:38
    at async Object.constraintSystem (o1js/dist/node/lib/provable/core/provable-context.js:64:9)
    at async TrivialZkapp.analyzeMethods (o1js/dist/node/lib/mina/zkapp.js:851:60)
    at async TrivialZkapp.compile (o1js/dist/node/lib/mina/zkapp.js:464:27)
    at async file:///Users/harrysolovay/Desktop/o1js/dist/examples/zkapps/simple-zkapp-with-proof.js:65:51

Node.js v21.7.3
```

mitschabaude:
> The complete error + stacktrace doesn't make clear where this offending item lives:

I get the following info from the stacktrace:
* It comes from `PublicKey.assertEquals()`, so the offending item is a public key which is `undefined`
* The `assertEquals()` is called inside `wrappedMethod()` in `zkapp.ts`. with source maps it would be easy to see on which line, but since there is only one `assertEquals()` on public keys in that file it's probably that one

src/lib/mina/zkapp.ts
```diff
@@ -157,7 +159,20 @@ function method<K extends string, T extends SmartContract>(
...
+          (v: unknown): v is ZkProgram<any, any> => v instanceof ZkProgram
+        )
+        .reduce(
+          (acc: number, cur: ZkProgram<any, any>) => acc + cur.getCallsCount(),
+          0
+        )
+    );
+    this.emitEvent(MetadataEvent.key, MetadataEvent.fromValue({ callsCount }));
+    return descriptor.value(...args);
+  } as AsyncFunction;
```
mitschabaude, line 175: * this seems like the wrong place to add wrapper logic that runs on every smart contract call
  * there's already `wrapMethod()` for that, called on the next line - why not add your logic there?
* just to call it out: if you add logic that runs on every smart contract call and influences the circuit and the shape of the transaction, that's a breaking change
* the current logic for finding recursive proof arguments seems invalid, because these proof arguments are not instances of ZkProgram -- ZkProgram isn't even a class

mitschabaude:
> Heavily WIP: updates ZkProgram and the method decorator such that the number of aggregated proofs is tracked and then emitted within a standard event, of which block explorers and other such services could make use.

The way I understood the original idea was that we wanted to track the number of times a proof was recursively verified? like, if there is a chain of proofs P0 (base proof) -> P1 (recursively verifies P1) -> P2 (recursively verifies P2), and P2 gets sent to the chain, we wanted to expose 3 as the number of aggregated proofs.

In the current approach, you're not doing that, as you're just counting `SelfProof` arguments, but a method with such an argument can be called on its own proof again & again.

Another problem I have with the WIP apprach is that you're modifying the default ZkProgram that everybody uses. Would be nicer to implement it as wrapper for ZkProgram so that it's optional to expose something that leaks private inputs.

harrysolovay:
@mitschabaude thank you for the feedback on this draft. It does seem that our brute solution neglects a wide range of considerations, such as the possibility of a recursive proof being created in and loaded in from an out-of-band process. Closing this draft for now.

PR #1715, opened 2024-07-02
Refactor Nullifier methods to use safer MerkleMapWitness operations

MartinMinkov:


mitschabaude (commented):
Wouldn't it cause less code duplication to only deprecate the individual methods that use computeRootAndKey?

Many people might not even use those method so they can just continue to use Nullifier without disruption

mitschabaude:
So I'm saying, I don't see why we have to add another version of the class in this case

MartinMinkov:
> So I'm saying, I don't see why we have to add another version of the class in this case

Yeah, that was my mistake. Changed the PR to this approach instead!

mitschabaude (approved):

PR #1716, opened 2024-07-02
Fix MayUseToken vulnerability in AccountUpdates

MartinMinkov:
## Description:
This PR addresses a security vulnerability in the `MayUseToken` field of AccountUpdate. The issue could potentially lead to miscalculation of total balance changes due to improper handling of token inheritance.
Key changes:

## Changes

- Added a check() method to ensure that parentsOwnToken and inheritFromParent flags are not both set to true simultaneously.

These changes prevent potential misuse of the API where a parent AccountUpdate might incorrectly treat a child as using the parent's own derived token when it's actually inheriting the token ID or using the default token ID.

## Repo updates:
bindings: https://github.com/o1-labs/o1js-bindings/pull/279


Trivo25 (commented):
Fix looks good to me, but I think this is a breaking change since `isParentsOwnToken` did do what it said it would originally, but could be misleading if used improperly. Imo we should consider adding `isParentsOwnTokenV2` and potentially defering the well formedness check in `check` to v2 entirely

src/lib/mina/account-update.ts
```diff
@@ -1193,11 +1193,22 @@ class AccountUpdate implements Types.AccountUpdate {
...
         return a.body.mayUseToken.inheritFromParent;
       },
+      check(mayUseToken: { parentsOwnToken: Bool; inheritFromParent: Bool }) {
+        // Ensure the flags are not both set to true
+        mayUseToken.parentsOwnToken
+          .and(mayUseToken.inheritFromParent)
+          .assertFalse(
+            'MayUseToken: parentsOwnToken and inheritFromParent cannot both be true'
+          );
+      },
```
mitschabaude, line 1211: there are a couple of issues with this `check()` method:
* it's not in the right place, so wouldn't be called automatically to constrain a `MayUseToken` type when we witness it
  * it needs to go into the provable type definition that lives on `.type`: `{ ...provablePure(...), check() { ... } }`
* `MayUseToken.type` is not even the provable type that the auto-derived account update type will use -- it treats the `mayUseToken` property as an anonymous object of two bools, see `gen/transactions.ts` and `js-layout.ts`. So, when witnessing entire account updates, nothing prevents `mayUseToken` from having both props = true
* Even _if_ we modified the account update type, there are many places where we skip _all_ checks on witnessed account updates, on purpose, because of the assumption that well-formedness of them is asserted by the transaction snark on the Mina node. (see `AccountUpdate.witness()` with the `skipCheck` parameter). From the finding, it wasn't clear to me of the OCaml `mayUseToken` field really includes such a wellformedness check. If not, I think we should add it in the necessary places in o1js somehow, for example in the token contract where it walk child account updates.

MartinMinkov: Got it! Do these changes make sense then?
Constructing the [check method properly](https://github.com/o1-labs/o1js/pull/1716/commits/5a27a22cbe35183bda85d1c86ea624c144df801a) and c[hanging the AccountUpdate.check method](https://github.com/o1-labs/o1js/pull/1716/commits/61017895a7f853ea55755c94e216e26547770d02) to also check for mayUseToken?

src/lib/mina/account-update.ts
```diff
@@ -1171,7 +1197,10 @@ class AccountUpdate implements Types.AccountUpdate {
   static empty() {
     return AccountUpdate.dummy();
   }
-  static check = Types.AccountUpdate.check;
+  static check = (au: AccountUpdate) => {
+    Types.AccountUpdate.check(au);
+    Types.MayUseToken.check(au.body.mayUseToken);
+  };
```
mitschabaude, line 1203: this override is not necessary, `Types.AccountUpdate.check()` already works by recursively calling check on all the types it is composed of

MartinMinkov:
Closed for https://github.com/o1-labs/o1js/pull/1741

PR #1728, opened 2024-07-09
Add reducer API production warning

ymekuria:
## Description:
This PR adds a warning about using the current `o1js` reducer API in production applications. This warning was added to the reducer doc comments as well as the examples using the reducer API. A note was also added mentioning that work is actively in progress to mitigate this limitation.

This warning will be removed once all work is done to update the reducer API.


mitschabaude (approved):
thanks, good to add this

PR #1729, opened 2024-07-09
Encryption v2

Trivo25:


src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +73,95 @@ function decrypt(
...
+  let privateKey = Provable.witness(Scalar, () => Scalar.random());
+  let publicKey = Group.generator.scale(privateKey);
+  let sharedSecret = otherPublicKey.toGroup().scale(privateKey);
+
+  let sponge = new Poseidon.Sponge();
+  sponge.absorb(sharedSecret.x);
+
+  // encryption
+  let cipherText = [];
+  for (let [n, chunk] of chunks.entries()) {
```
mitschabaude, line 144: I wasn't aware of this trick 😍

src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +73,95 @@ function decrypt(
...
+    if (i % 2 === 1) sponge.absorb(cipherText[i - 1]);
+    if (i % 2 === 1 || i === cipherText.length - 1)
+      sponge.absorb(cipherText[i]);
+  }
+
+  // authentication tag
+  sponge.squeeze().assertEquals(authenticationTag!);
+
+  // return the message as a flat array of bytes
+  return message.flat();
```
mitschabaude, line 117: hm, we're not stripping off the padding here. it will make the function a bit hard to deal with. but I think it's the right approach, the actual message length needs to be hard-coded in the circuit anyway

Trivo25: yep, still need to do that. I am thinking of adding a property to `CipherText`

src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +73,108 @@ function decrypt(
...
+
+  // pad message to a multiple of 31 so they still fit into one field element
+  const multipleOf = 31;
+  const n = Math.ceil(messageLength / multipleOf) * multipleOf;
+
+  // create the padding
+  const padding = Array.from({ length: n - messageLength }, () =>
+    UInt8.from(0)
+  );
+  message.bytes = bytes.concat(padding);
```
mitschabaude, line 146: this should probably not mutate the input message

src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +73,108 @@ function decrypt(
...
+
+  const sponge = new Poseidon.Sponge();
+  sponge.absorb(sharedSecret.x);
+
+  // encryption
+  const cipherText = [];
+  for (let [n, chunk] of chunks.entries()) {
+    // absorb frame bit
+    if (n === chunks.length - 1) sponge.absorb(Field(1));
+    else sponge.absorb(Field(0));
```
mitschabaude, line 164: ok so now you're simply absorbing the frame bit instead of concatenating it with the message!

in that case,
- the optimization done below to always absorb two ciphertext chunks at a time becomes obsolete, and slower than absorbing them one at a time, because including the frame bit we're doing 2 at a time anyway. so remove that optimization i.e. just `sponge.absorb(encryptedChunk)` below
- since we're using an entire separate field element for the frame bit, we could now again have a version of `encryptV2()` which operates on messages of full field elements. and which doesn't have the issue that the message length is ambiguous.
  - consider this as an optional refactor. in that case, the concatenation of bytes to fields could be a separate step done before calling encrypt to give an alternative API with `Bytes` input, which needs to add the message length to the ciphertext output.

src/lib/provable/crypto/encryption.ts
```diff
@@ -68,3 +73,108 @@ function decrypt(
...
+  for (let i = 0; i < cipherText.length; i++) {
+    // absorb frame tag
+    if (i === cipherText.length - 1) sponge.absorb(Field(1));
+    else sponge.absorb(Field(0));
+
+    const keyStream = sponge.squeeze();
+    const messageChunk = cipherText[i].sub(keyStream);
+
+    // convert to bytes
+    const byteMessage = wordToBytes(messageChunk, 32);
```
mitschabaude, line 105: wait shouldn't this be `wordToBytes(messageChunk, 31)`? I'm confused

src/examples/encryptionv2.ts
```diff
@@ -0,0 +1,21 @@
...
+  Provable,
+} from 'o1js';
+
+await initializeBindings();
+
+class Bytes32 extends Bytes(32) {}
+const priv = PrivateKey.random();
+const pub = priv.toPublicKey();
+
+const plainMsg = 'Hello world';
```
mitschabaude, line 16: you should also have a test with more than 31 characters :D

src/lib/provable/crypto/encryption.ts
Trivo25: upps, thats right!

src/lib/provable/crypto/encryption.ts
Trivo25: Yea, absorbing it directly makes a little more sense now that I got more familiar with the sponge concept, its also a little easier and we can indeed go back to field elements - I think I will do that

mitschabaude (approved):
I still recommend to have the example use more than 31 chars (that are not zero-padding) as a sanity check, otherwise lgtm

PR #1731, opened 2024-07-09
Expose `MAX_ACCOUNT_UPDATES` from `TokenContract.ts`

### LINKED ISSUE #1730: Expose `MAX_ACCOUNT_UPDATES` from `TokenContrac.ts`
We need that in the fungible token standard implementation.

I'll submit a PR for this shortly.

kantp:
We need that constant in the fungible token standard implementation.

Closes #1730.

### DIFF
--- a/src/lib/mina/token/token-contract.ts
+++ b/src/lib/mina/token/token-contract.ts
@@ -23,8 +23,9 @@ export { TokenContract };
 abstract class TokenContract extends SmartContract {
   // change default permissions - important that token contracts use an access permission
 
-  // it's fine to have this restriction, because the protocol also has a limit of ~20
   // TODO find out precise protocol limit
+  /** The maximum number of account updates using the token in a single
+   * transaction that this contract supports. */
   static MAX_ACCOUNT_UPDATES = 20;
 
   /**
--- a/src/lib/mina/token/token-contract.ts
+++ b/src/lib/mina/token/token-contract.ts
@@ -13,7 +13,7 @@ import { DeployArgs, SmartContract } from '../zkapp.js';
 import { TokenAccountUpdateIterator } from './forest-iterator.js';
 import { tokenMethods } from './token-methods.js';
 
-export { TokenContract };
+export { TokenContract, TokenContractV2 };
 
 /**
  * Base token contract which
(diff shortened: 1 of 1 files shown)


src/lib/mina/token/token-contract.ts
```diff
@@ -27,6 +23,10 @@ const MAX_ACCOUNT_UPDATES = 20;
 abstract class TokenContract extends SmartContract {
   // change default permissions - important that token contracts use an access permission
 
+  // it's fine to have this restriction, because the protocol also has a limit of ~20
+  // TODO find out precise protocol limit
+  static MAX_ACCOUNT_UPDATES = 20;
```
Trivo25, line 28: I think we should add a small doc comment explaing what this actually means. It's just a "rough upper limit", which can vary greatly on the type of authorization

mitschabaude: good point!

kantp: Good idea! How does it translate to the number of maximal account updates?

kantp: Also, it would be great if we could determine the TODO item there before we properly release the token standard, as a change to the parameter will require token contracts to be updated.

mitschabaude: ok I ran a script to find out. the actual maximum is 12, and the following configurations fulfill that maximum:
```
  { proof: 0, pair: 5, single: 2, total: 12 },
  { proof: 0, pair: 6, single: 0, total: 12 }
```

where
* `proof` means a proof-authorized account update,
* `single` means a non-proof account update (signature or no auth)
* `pair` means a consecutive pair of no-proof updates

mitschabaude: so.. would be nice to change it to 12

but since a change of the maximum inside `forEachUpdate()` would be a breaking change, we should land differently. for example, add a new `forEachUpdateV2()` which has the constant changed to 12

mitschabaude: will be a nice 1 - 12/20 = 40% reduction in constraints

mitschabaude: actually, arguably the max is 11 since the token contract which approves is an update itself

mitschabaude: actually the max is 10 then 😅 since the two 12 configurations don't involve a proof update

here are the max configs with at least 1 proof update:
```
[
  { proof: 1, pair: 4, single: 2, total: 11 },
  { proof: 1, pair: 5, single: 0, total: 11 }
]
```

mitschabaude: actually I just checked, the fee payer is counted as well, which can't be below the token contract

so the relevant maximum here is 9

mitschabaude (approved):

PR #1733, opened 2024-07-09
Batch Reducer: expose more flexibility

mitschabaude:
- random fix
- method to conditionally set a merkle map key
- clean up conditional merkle map code
- changelog
- minor comment fix
- batch reducer design
- make reducer a class
- make action batch a class
- process next batch
- dispatch, change type stuff
- prove next batch
- fetch action batch
- get actions, more type tweaks
- more docs
- action batch verify
- let's not forget this
- check action batch
- action state program and type refactor
- simplify
- prover
- usability tweaks
- expose proof on the class
- make batch reducer usable without importing reducer
- two account update qol improvements
- unit test for recursive part of batch reducer
- fix handling of non-existing accounts in fetch logic
- don't try to fetch accounts with dummy public keys
- fix event pushing: transform unconstrained data to constants
- move added method to the right changelog version
- testing framework?
- deploy automatically
- make test return a list of actions
- allow getting offchain state outside the contract
- allow returning other actions
- expect state
- add map to expectstate
- shorten
- reenable proofs
- fix a few bugs
- test framework: support batch reducer, expect balance
- move and fix assert defined
- debug context error
- batch reducer unit test
- much slicker global state error message
- actually let's just sacrifice 1ms for having a good error by default
- clean up test and remove debugging
- minor
- support proofs disabled in batch reducer rollup
- add method to reset proofs enabled to the original value
- enable compiling batch reducer methods before having a contract instance
- push down type cast
- refactor: separate instance creation and running tests on it
- make test runner changes work for offchain state as well
- support running a test with AND without proofs, improve expect errors
- upgrade typescript to get rid of an error
- slightly more precise error catching
- allow skipping deploy, pass trace with txn
- make proofs enabled: both work for reducer test
- deploy action
- simplify
- introduce two kinds of accounts, and simplify deploy
- minor
- return instance from test runner, so we can build tests upon each other in the future
- revert events changes
- changelog
- adjust numbers
- call them "instruction" to not overload "actions"
- minor
- add convenience method
- assert equal if
- start refactor to reverse stacking design
- prove action stack
- make action stacking logic reusable
- fix: map can only be witnessed if both the length and root match - otherwise the length is unconstrained
- unit-test offchain stack prover
- add wrapper to provide the inputs needed to complete recursive proof in contract
- start fetching and offchain preparation logic
- start dedicated file for action types
- start refactor to expose individual actions as witness
- requireEqualsIf()
- onchain verification logic to select action stack
- new merkle list convenience methods
- finish reducer logic
- polish reducer logic and input type
- merkle actions from fields
- simplify some types
- finish preparation of batches
- some renames
- and so the debugging begins
- fix inconsistent hashing of stack
- doesn't workwith proofs yet, and weird invalid signature error
- works with proofs
- fix: get correct empty value for action state precondition
- remove debugging
- allow returning a list of test instructions
- test recursion
- polish test
- more example polish, highlight unsafety in a better way
- improve naming
- remove tests for original batch reducer logic
- delete old batch reducer logic
- reduce noise
- changelog
- simplify some generic types
- fix unrelated example
- final optimization: select action before unhashing
- dispatch if
- add more detailed doccomments
- unsafe process action state


Trivo25 (commented):
Yep I think this comes close to what I imagined, thanks!

PR #1737, opened 2024-07-11
Canonical ecdsa check

Trivo25:
check that `signature.s` is canonical, adjust doc comment


src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -236,7 +236,8 @@ function verifyEcdsaGeneric(
     G?: { windowSize: number; multiples?: Point[] };
     P?: { windowSize: number; multiples?: Point[] };
     ia?: point;
-  } = { G: { windowSize: 4 }, P: { windowSize: 4 } }
+  } = { G: { windowSize: 4 }, P: { windowSize: 4 } },
+  checkSCanonical = false
```
Trivo25, line 240: I went with this extra flag to stay backwards compatible.
This breaks `ecdsa.verify` yet again, we still need to figure out how we want to handle this

src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -281,6 +282,9 @@ function verifyEcdsaGeneric(
   // if we allowed non-canonical Rx, the prover could make verify() return false on a valid signature, by adding a multiple of `Curve.order` to Rx.
   ForeignField.assertLessThan(Rx, Curve.order);
 
+  // assert s to be canonical
+  if (checkSCanonical) ForeignField.assertLessThan(s, Curve.order);
```
Trivo25, line 286: I decided to only assert `s` to be canonical and leave `r` as is. Since `r` will be checked to be canonical but just returns `false` instead of throwing an error, I thought this would be the least breaking case and save constraints

src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -324,6 +328,10 @@ function verifyEcdsa(
  *     for variable public key, there is a possible use case: if the public key is a public input, then its multiples could also be.
  *     in that case, passing them in would avoid computing them in-circuit and save a few constraints.
  * - The initial aggregator `ia`, see {@link initialAggregator}. By default, `ia` is computed deterministically on the fly.
+ *
+ *
+ * _Note_: If `signature.r` is a non-canonical element, an error will be thrown.
```
mitschabaude, line 333:
```suggestion
 * _Note_: If `signature.s` is a non-canonical element, an error will be thrown.
```

src/lib/provable/gadgets/elliptic-curve.ts
mitschabaude: makes sense!

mitschabaude (commented):
this does it for me.

an alternative would have been to encode the canonical check in the `ForeignEcdsa` signature type, but this version is safer for the gadgets version.

one question however, there is still a non-canonical representation in the sense that I can negate s and it works as a valid signature (with the same r)

if the goal is to enforce non-malleability of the signature, isn't preventing against that equally important as the assertion you added?

mitschabaude (approved):

mitschabaude:
don't forget to remove the flag since this is now going into `v2`!

PR #1739, opened 2024-07-11
Limit prefixToField to only accept ascii characters

Trivo25:


src/lib/provable/crypto/poseidon.ts
```diff
@@ -160,6 +160,11 @@ function prefixToField(prefix: string) {
       // convert char to 8 bits
       let bits = [];
       for (let j = 0, c = char.charCodeAt(0); j < 8; j++, c >>= 1) {
+        if (j === 7)
```
Trivo25, line 163: change to only support the default ascii set, the character needs to be fully consumed after 7 bits. I think the default ascii set is more than enough for us to support here

mitschabaude (approved):
lgtm, but FYI there is a second `prefixToFields()` method somewhere which already handles non ASCII correctly, maybe we should just remove this redundant version of it

PR #1741, opened 2024-07-11
[v2] Fix MayUseToken vulnerability in AccountUpdates

MartinMinkov:
bindings: https://github.com/o1-labs/o1js-bindings/pull/280
mina: https://github.com/MinaProtocol/mina/pull/15808


src/lib/mina/account-update.unit-test.ts
```diff
@@ -122,3 +131,40 @@ function createAccountUpdate() {
...
+      parentsOwnToken: Bool(true),
+      inheritFromParent: Bool(true),
+    });
+  }).toThrowError(
+    'MayUseToken: parentsOwnToken and inheritFromParent cannot both be true'
+  );
+}
+
+// throws an error when both flags are true in AccountUpdate.check method
+// TODO: This should be throwing but it isn't, need to investigate
```
MartinMinkov, line 148: Under the assumption of this [comment](https://github.com/o1-labs/o1js/pull/1716#discussion_r1673801107), I would suspect that `AccountUpdate.check` should throw, but it isn't here :thinking:

mitschabaude: aren't you missing the Mina changes that you had in the other PR?

MartinMinkov: Ah, I didn't know that was necessary. So for an `AccountUpdate` to recursively call `MayUseToken.check`, we need this [change](https://github.com/MinaProtocol/mina/pull/15808/files#diff-1edc359c0b2f3f296bb16dd374c402ecfb87b2de6d1f0d28968966c26e61fd2bR1141)?

mitschabaude: yeah exactly

mitschabaude: the info goes into `js-layout.ts`

mitschabaude (approved):

PR #1742, opened 2024-07-11
Refactor Int64: Remove v2 methods

MartinMinkov:


src/lib/provable/int.ts
```diff
@@ -1247,41 +1289,22 @@ class Int64 extends CircuitValue implements BalanceChange {
...
-  // }
-
-  // TODO add this when `checkV2` is enabled
-  // then it will be the correct logic; right now it would be misleading
-  /**
-   * Checks if the value is negative (x < 0).
-   */
-  // isNegative() {
-  //   return this.sgn.isNegative();
-  // }
```
mitschabaude, line 1281: I think `isNonNegative()` and `isNegative()` could be useful!

src/lib/provable/int.ts
```diff
@@ -1203,24 +1242,27 @@ class Int64 extends CircuitValue implements BalanceChange {
...
+   * console.log(x1.mod(y1).toString()); // Output: 2
+   * ```
+   *
+   * @throws {Error} Implicitly, if y is zero or negative.
    */
-  modV2(y: UInt64 | number | string | bigint | UInt32) {
+  mod(y: UInt64 | number | string | bigint | UInt32) {
     let y_ = UInt64.from(y);
     let rest = this.magnitude.divMod(y_).rest.value;
     let isNonNegative = this.magnitude
```
mitschabaude, line 1268: this can be changed to just `this.isNonNegative()` now

src/lib/provable/int.ts
```diff
@@ -1048,10 +1048,25 @@ class Int64 extends CircuitValue implements BalanceChange {
...
+   * ```ts
+   * const x = new Int64(0); // canonical representation of zero
+   * ```
+   */
   constructor(magnitude: UInt64, sgn = Sign.one) {
-    super(magnitude, sgn);
+    const mag = UInt64.from(magnitude);
+    const isZero = mag.equals(UInt64.zero);
+    const canonicalSign = Provable.if(isZero, Sign.one, sgn);
+    super(mag, canonicalSign);
```
mitschabaude, line 1068: this shouldn't be in the constructor logic, should be a separate method like `.create()` as before

the Int64 class still needs an unsafe constructor internally which doesn't add any unnecessary constraints

the the constructor should be made `private`

src/lib/provable/int.ts
```diff
@@ -1092,6 +1107,14 @@ class Int64 extends CircuitValue implements BalanceChange {
     }
     return Int64.fromFieldUnchecked(Field(x));
   }
+
+  fromObject(obj: {
+    magnitude: UInt64 | number | string | bigint;
+    sgn: Sign | bigint;
+  }) {
+    return new Int64(UInt64.from(obj.magnitude), Sign.fromValue(obj.sgn));
```
mitschabaude, line 1115: should use `create()`
and we still want `Unsafe.fromObject()` IMO

src/lib/provable/int.ts
```diff
@@ -1188,12 +1215,24 @@ class Int64 extends CircuitValue implements BalanceChange {
...
    *
    * `x.div(y)` returns the floor of `x / y`, that is, the greatest
-   * `z` such that `z * y <= x`.
+   * *`z`* such that *`z * y <= x`.
    * On negative numbers, this rounds towards zero.
+   *
+   * This method guarantees that all results, including zero, have a consistent
+   * representation, eliminating potential ambiguities in zero handling.
    */
   div(y: Int64 | number | string | bigint | UInt64 | UInt32) {
```
mitschabaude, line 1237: if we make the internal constructor unsafe, this needs `.create()` as before

mitschabaude (commented):
would be nice to start a `CHANGELOG-v2.md`

src/lib/provable/int.ts
MartinMinkov: When I attempt to make `Int64`s constructor private, it results in some annoying TS errors due to how `CircuitValue` is set up.

I can look into modifying `CircuitValue` to accommodate for private constructor definition but I wanted to avoid modifying `CircuitValue` due to how it's used in other areas. Do you think I should go down path or just leave the `Int64` public but with a deprecation comment?

```
The 'this' context of type 'typeof Int64' is not assignable to method's 'this' of type 'AnyConstructor'.
  Cannot assign a 'private' constructor type to a 'public' constructor type.
```

mitschabaude: ok got it. let's leave it deprecated for now, I'll look into refactoring it to a Struct

mitschabaude (approved):

PR #1744, opened 2024-07-11
ci: add workflow to automatically merge main into v2 branch

MartinMinkov:


.github/workflows/auto-merge-main-to-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+  push:
+    branches:
+      - main
+
+jobs:
+  auto-merge:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
shimkiv, line 25: Please update this action to use the latest (`v4` I think).

.github/workflows/auto-merge-main-to-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Try to merge main into v2
+        run: |
+          git checkout v2
+          git merge origin/main
+          git push origin v2
+        continue-on-error: true
+
+      - name: Create Pull Request if merge failed
+        if: failure()
+        uses: peter-evans/create-pull-request@v3
```
shimkiv, line 43: Please update this action to use the latest (`v6` I think).

.github/workflows/auto-merge-main-to-v2.yml
```diff
@@ -0,0 +1,48 @@
...
+        with:
+          branch: auto-merge-main-to-v2
```
shimkiv, line 45: > Commit all changes to a new branch, or update an existing pull request branch.

Are concerned about the edge cases when two different merges into the `main` branch will cause this workflow to kick off and they both will push to the same `auto-merge-main-to-v2` branch with possible conflicts?
Or `v2` is not supposed to last long enough to spend time on that kind of concerns?

shimkiv (approved):

.github/workflows/auto-merge-main-to-v2.yml
MartinMinkov: That's a great point! I added this [slight change here ](https://github.com/o1-labs/o1js/pull/1744/commits/15fe7d699833fa44593e09df0d4d2fbe0c09f327)to make `auto-merge-main-to-v2` have a postfix of the git commit. That should fix the issues with two merges clobbering each other.

shimkiv: Thank you!

PR #1745, opened 2024-07-11
Fix: Prevent underconstraint in xor(), and(), and not() functions

MartinMinkov:
## Overview
This PR addresses a potential overflow vulnerability in the bitwise operations (xor, and, not) as identified in a recent audit. The changes reduce the maximum allowed bit length from 254 to 240 bits and refactor the bit length validation logic to prevent underconstraining of circuits.

## Changes
1. Reduced maximum bit length from 254 to 240 for xor, and, and not operations.
1. Extracted bit length validation logic into a separate function validateBitLength() to reduce code duplication and ensure consistent checks across all bitwise operations.
1. Updated error messages to provide more context and clarity.
1. Adjusted test cases to reflect the new 240-bit limit.

## Motivation
The audit revealed that for input lengths between 240 and 254 bits, the padLength calculation could result in a 256-bit value, potentially causing an overflow when constraining the weighted sum of 16-bit limbs. This could lead to underconstrained circuits, posing a security risk.

## Technical Details
- xor(), and(), and not() functions now use a maximum length of 240 bits.
- A new validateBitLength() function handles input validation for all bitwise operations.
- The padLength calculation remains unchanged, but with the new 240-bit limit, it will never exceed 240 bits (next multiple of 16).

## Testing

Updated existing test cases in bitwise.unit-test.ts to use 240 bits instead of 254.
Added new test cases to verify behavior at the new maximum length.

### DIFF
--- a/src/lib/provable/gadgets/bitwise.ts
+++ b/src/lib/provable/gadgets/bitwise.ts
@@ -19,7 +19,7 @@ export {
 };
 
 function not(a: Field, length: number, checked: boolean = false) {
-  validateBitLength(length, Field.sizeInBits, 'not');
+  validateBitLength(length, 240, 'not');
 
   // obtain pad length until the length is a multiple of 16 for n-bit length lookup table
   let padLength = Math.ceil(length / 16) * 16;
@@ -140,7 +140,7 @@ function buildXor(a: Field, b: Field, out: Field, padLength: number) {
 }
 
 function and(a: Field, b: Field, length: number) {
-  validateBitLength(length, Field.sizeInBits, 'and');
+  validateBitLength(length, 240, 'and');
 
   // obtain pad length until the length is a multiple of 16 for n-bit length lookup table
   let padLength = Math.ceil(length / 16) * 16;
--- a/src/lib/provable/gadgets/bitwise.ts
+++ b/src/lib/provable/gadgets/bitwise.ts
@@ -19,14 +19,7 @@ export {
 };
 
 function not(a: Field, length: number, checked: boolean = false) {
-  // check that input length is positive
-  assert(length > 0, `Input length needs to be positive values.`);
-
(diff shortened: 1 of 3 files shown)


src/lib/provable/gadgets/bitwise.ts
```diff
@@ -19,14 +19,7 @@ export {
...
 function not(a: Field, length: number, checked: boolean = false) {
-  // check that input length is positive
-  assert(length > 0, `Input length needs to be positive values.`);
-
-  // Check that length does not exceed maximum field size in bits
-  assert(
-    length < Field.sizeInBits,
-    `Length ${length} exceeds maximum of ${Field.sizeInBits} bits.`
-  );
+  validateBitLength(length, Field.sizeInBits, 'not');
```
mitschabaude, line 22: `not()` also pads the length so let's do 240 here as well

src/lib/provable/gadgets/bitwise.ts
```diff
@@ -150,14 +140,7 @@ function buildXor(a: Field, b: Field, out: Field, padLength: number) {
...
 function and(a: Field, b: Field, length: number) {
-  // check that both input lengths are positive
-  assert(length > 0, `Input lengths need to be positive values.`);
-
-  // check that length does not exceed maximum field size in bits
-  assert(
-    length <= Field.sizeInBits,
-    `Length ${length} exceeds maximum of ${Field.sizeInBits} bits.`
-  );
+  validateBitLength(length, Field.sizeInBits, 'and');
```
mitschabaude, line 143: `and()` also pads the length so let's do 240 here as well

mitschabaude:
> Should probably be targetting v2

not necessary IMO

Trivo25 (approved):
let's add a small note to the doc comments as well

ymekuria (approved):

PR #1751, opened 2024-07-16
Accept .provable and some other type improvements

mitschabaude:
This PR makes `Type` work where previously only `Type.provable` worked:

```diff
- let bytes = Provable.witness(Bytes32.provable, () => Bytes32.random());
+ let bytes = Provable.witness(Bytes32, () => Bytes32.random());
```

Other small improvements:
* `Unconstrained` gets unwrapped automatically by `toValue` / `fromValue` because it doesn't make sense to have an unconstrained value outside the circuit
* Add `provableExtends()` to cleanly extend a provable class (like `MerkleList`) with additional instance methods. This is used for `AccountUpdateForest` as an example. It's pretty easy to use and could be a mechanism to define types with methods in the future.

bindings: https://github.com/o1-labs/o1js-bindings/pull/284


Trivo25 (approved):
we should add a changelog entry

PR #1752, opened 2024-07-16
Fix: Do not support small foreign curves

mitschabaude:
Don't support curves on base fields smaller than 2^176, because the implementation of addition is unsound

These curves won't be needed in practice as far as I can tell.

I think this is effectively not a breaking change because nobody will have used those curves, and addition might have thrown an error for them

### DIFF
--- a/src/lib/provable/gadgets/elliptic-curve.ts
+++ b/src/lib/provable/gadgets/elliptic-curve.ts
@@ -3,7 +3,7 @@ import { Field } from '../field.js';
 import { Provable } from '../provable.js';
 import { assert } from './common.js';
 import { Field3, ForeignField, split, weakBound } from './foreign-field.js';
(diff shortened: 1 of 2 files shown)


Trivo25 (approved):

src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -65,6 +65,11 @@ function add(p1: Point, p2: Point, Curve: { modulus: bigint }) {
     return Point.from(p3);
   }
 
+  assert(
+    Curve.modulus > l2Mask + 1n,
+    'Base field moduli smaller than 2^176 are not supported'
+  );
```
mitschabaude, line 71: this is needed because we can use the addition gadget directly, without the `ForeignCurve` wrapper

PR #1753, opened 2024-07-16
Fix: Add account update checks not done in Mina

mitschabaude:
This is a follow-up to https://github.com/o1-labs/o1js/pull/1749 where we ensured the `Int64` class, including its `check()` method, was used for the balance change type on account updates.

The mentioned PR in combination with https://github.com/o1-labs/o1js/pull/1742 which changes the `check()` method to prevent a non-unique zero representation means that for account updates witnessed normally, the `Int64` check is done on the balance change.

However, this check is not done by the Mina node (it doesn't affect security there), so to protect zkApp circuits this PR makes sure the check is done on _all_ account updates.


src/lib/mina/account-update.ts
```diff
@@ -1184,19 +1184,34 @@ class AccountUpdate implements Types.AccountUpdate {
     return new AccountUpdate(accountUpdate.body, accountUpdate.authorization);
   }
 
+  /**
+   * This function acts as the `check()` method on an `AccountUpdate` that is sent to the Mina node as part of a transaction.
```
MartinMinkov, line 1188: Great to document this, thanks!

MartinMinkov (approved):

PR #1757, opened 2024-07-16
Remove insecure v1 versions of ECDSA and curve ops

mitschabaude:
This finalizes the ECDSA and curve fixes that went into v1 in a compatible way


src/lib/provable/gadgets/foreign-field.ts
```diff
@@ -432,11 +432,7 @@ function equals(x: Field3, c: bigint, f: bigint) {
   }
 }
 
-const provableLimb = modifiedField({
-  toInput(x) {
-    return { packed: [[x, Number(l)]] };
-  },
-});
+const provableLimb = modifiedField({});
```
mitschabaude, line 435: having a `toInput` method but no `check` method is a major footgun that caused the ecdsa vulnerability, because fileds are packed but unpacking is not constrained.
Since we don't want `check()` here I'm removing `toInput`.

MartinMinkov (approved):
:tada:

src/lib/provable/gadgets/foreign-field.ts
MartinMinkov: Dang! Sneaky footgun. Are there other potential areas this is occurring? 

Should we also update the `provable` interface or somewhere relevant to mention the case of leaving out certain provable methods?

mitschabaude: I did add docs on `Hashed` and `Packed` specifically about this issue. Other than that, I think it's fine because this is more low-level than users typically operate in.

PR #1759, opened 2024-07-16
Canonical representation of provable types

mitschabaude:
This adds another (optional) method to the `Provable<T>` interface,
```ts
toCanonical(t: T): T;
```

## Motivation

The method allows a provable type to define its own specific notion of canonical representation. This is useful for types with non-unique representations, currently `ForeignField` and `Scalar` (and `Int64` used to be another example but we changed that to be unique).

For most provable types, `toCanonical()` is not defined; in that case, we fall back to the identity function (i.e., treating every value as its own canonical representation).

The addition of `toCanonical()` is used to solve two issues:

1. Non-canonical representations are often associated with completeness gaps. This is the case for both `ForeignField` and `Scalar`, some of whose operations are incomplete (not able to create a proof) on unreduced inputs (`scale()` for scalar, various arithmetic for foreign fields).  
  In _most_ situations, this incompleteness is not a problem because witness generators will ensure canonical representations anyway, so an honest prover can succeed. _However_, there are situations, especially with actions and reducer, where one (malicious) party can _force_ another (honest) party to process their inputs: Say, if an action contains a Scalar that you can't call `scale()` on, but the reducer logic involves scaling by it, then a malicious input can break the reducer.
2. Second, `Provable.equal()` has unintuitive (or, you might say unsound) behavior when there are multiple representations of the "same" value: The naive equality check that just compares field elements will return `false` even if the values represent the same (foreign field, scalar, ..) element.

We solve these issues as follows:
* Before dispatching an action, call `toCanonical()` on it, so only canonical elements ever end up in the reducer
* In `Provable.equal(type, x, y)`, call `type.toCanonical()` on both inputs `x` and `y` before comparing them. That is, they are counted as equal using the notion of equality particular to that type
  * Example: Foreign fields that are the same modulo the field size are reported as equal

## Discussion

There are two possible versions of `toCanonical()`:

* **Weak version:** Don't transform the input, but just _assert_ that the input is canonical, and return it again. In other words, it allows you to prove that a given value is canonical before using it in places where non-canonical inputs are not supposed to be allowed
* **Strong version:** Is able to take any input that satisfies the usual (`check()`) properties of the type, and transform it to a canonical version.

For solving the actions-reducer issue, the weak version is sufficient, because we call `toCanonical()` at dispatch (as opposed to in the reducer), where the proof is allowed to fail.

However, this PR implements the strong version of `toCanonical()` which actually works on all non-canonical inputs. This will make it easier to use and reason about when used in other protocols that rely on canonical values.
  
## PR Details

* Add `toCanonical()` to the provable interface
* Infer `toCanonical()` for composed types like `Struct` from its base types. So, e.g. the `toCanonical()` method of a `Struct({ x: ForeignField, y: ForeignField })` maps both `x` and `y` to their canonical version.
* Implement `toCanonical()` for both `Scalar` and `ForeignField`. All other base types use the identity.
  * Add tests that demonstrate both the existing incompleteness of scalar multiplication, and how `toCanonical()` protects against it
* Change `Provable.equal()` to perform `toCanonical()` on both inputs. For this to work in a comprehensive way, I had to remove the "implicit" version of `equal()`. So now you have to always use the "explicit" version where you pass in the type along with the values to be compared.
* Change `Reducer` and `BatchReducer` to do `toCanonical()` before dispatching
* **Caveat**: I skipped the implementation of `toCanonical()` in `from-layout.ts`, which is used for protocol types like account updates. Instead I just used the identity function. Currently this is fine because none of the protocol types have non-canonical versions. However, for forward compatibility and safety, this should probably be added in the future.


ymekuria (approved):
LGTM

PR #1763, opened 2024-07-18
Update quotientBits bound too low in divMod32

ymekuria:
## Summary

This PR addresses an issue with the the `divMod32()` gadget that allowed potential user error. In `divMod32()`, user may provide a `quotientBits` parameter indicating the maximum allowed (logarithm) of the quotient. The bound on `quotientBits` is too low to prevent an overflow. 

## Impact

If `quotientBits = 240` or `quotientBits = 224` is used, then an overflow may occur. 

## Changes

- `divMod32()` was modified to accept `nBits` and assert it is in the range `[0, 2**255)` instead of `quotientBits`.
- `quotientBits` is computed from the `nBits` parameter. 
-  Calls to `divMod32()` were modified to accept 'nBits' in the `SHA256 ` implementation and `addMod32()`


src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -49,5 +54,5 @@ function divMod32(n: Field, quotientBits = 32) {
 }
 
 function addMod32(x: Field, y: Field) {
-  return divMod32(x.add(y), 1).remainder;
+  return divMod32(x.add(y), 32).remainder;
```
boray, line 57: I think this should be 64. Otherwise, rangeCheckN fails since N is zero.

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -49,5 +54,5 @@ function divMod32(n: Field, quotientBits = 32) {
 }
 
 function addMod32(x: Field, y: Field) {
-  return divMod32(x.add(y), 1).remainder;
+  return divMod32(x.add(y), 64).remainder;
```
mitschabaude, line 57: this seems unnecessary, why are we doing this?

mitschabaude: oh I just saw that quotient bits is now this minus 32. so it should be 33 to match the previous version

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -6,11 +6,16 @@ import { rangeCheck32, rangeCheckN } from './range-check.js';
 
 export { divMod32, addMod32 };
 
-function divMod32(n: Field, quotientBits = 32) {
+function divMod32(n: Field, nBits = 64) {
+  assert(
+    nBits >= 0 && nBits <= 255,
+    `nBits must be in the range [0, 255], got ${nBits}`
+  );
```
mitschabaude, line 13: the check here should be for `nBits < 255` not `<= 255`!

src/lib/provable/gadgets/arithmetic.ts
boray: Why should the quotient bits actually be 1?

mitschabaude: because the addition of two 32 bit numbers yields a 33 bit number

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -6,11 +6,16 @@ import { rangeCheck32, rangeCheckN } from './range-check.js';
 
 export { divMod32, addMod32 };
 
-function divMod32(n: Field, quotientBits = 32) {
+function divMod32(n: Field, nBits = 64) {
+  assert(
+    nBits >= 0 && nBits < 255,
+    `nBits must be in the range [0, 255], got ${nBits}`
```
mitschabaude, line 12:
```suggestion
    `nBits must be in the range [0, 255), got ${nBits}`
```

mitschabaude (approved):

ymekuria:
Thanks for the review @mitschabaude. Do you think this should be in the V2 branch because the API changed ( `quotientBits` -> `nBits`)?

PR #1767, opened 2024-07-23
BLAKE2b gadget

### LINKED ISSUE #1754: [Prototype] Implement Blake Hashing for Mina Consensus Selection in o1js Circuit
LambdaClass needs to implement the Mina consensus selection algorithm in an o1js circuit. To achieve this, they require the Blake hashing function, which needs to be implemented as a gadget. We need to prototype this implementation and assess its feasibility, particularly focusing on performance and memory constraints.

**Objective:**

Prototype Blake hashing in o1js, evaluate its performance, and determine if it's suitable for use in the consensus selection algorithm implementation.

**Key Points to Address:**

1. Implement Blake hashing as a gadget in o1js.
1. Assess performance constraints, with a particular focus on memory usage.
1. Determine if the o1js implementation is feasible or if we need to switch to Snarky for native proving.

boray:
Closes #1754 

Ingredients:
- BLAKE2b hash function implementation with arbitrary digest length https://www.blake2.net/blake2.pdf
- `divMod64` and `addMod64` needed for compression function
- `Gadgets.or()`
- `UInt64.or()`
- Add zero and one accessors to UInt8


src/lib/provable/gadgets/bitwise.ts
```diff
@@ -173,6 +174,35 @@ function and(a: Field, b: Field, length: number) {
   return outputAnd;
 }
 
+function or(a: Field, b: Field, length: number) {
+  // Validate at 240 bits to ensure padLength (next multiple of 16) doesn't exceed 254 bits,
+  // preventing potential underconstraint issues in the circuit
+  validateBitLength(length, 240, 'not');
```
MartinMinkov, line 180:
```suggestion
  validateBitLength(length, 240, 'or');
```

src/examples/crypto/blake2b/blake2b.ts
```diff
@@ -0,0 +1,18 @@
...
+class Bytes12 extends Bytes(12) {}
+
+let BLAKE2BProgram = ZkProgram({
+  name: 'blake2b',
+  publicOutput: Bytes(32),
+  methods: {
+    blake2b: {
+      privateInputs: [Bytes12],
+      async method(xs: Bytes12) {
+        return Gadgets.BLAKE2B.hash(xs,32);
```
Trivo25, line 14: super tiny nit but lets add a space between the arguments :D `(xs,32)` -> `(xs, 32)`

src/lib/provable/crypto/hash.ts
```diff
@@ -2,6 +2,7 @@ import { Gadgets } from '../gadgets/gadgets.js';
 import { Poseidon } from './poseidon.js';
 import { Keccak } from './keccak.js';
 import { Bytes } from '../wrapped-classes.js';
+import { BLAKE2 } from 'node_modules/@noble/hashes/_blake2.js';
```
Trivo25, line 5: wrong import?

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,189 @@
...
+    [2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9],
+    [12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11],
+    [13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10],
+    [6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5],
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+  ],
+};
+
+const BLAKE2B = {
+  hash(data: FlexibleBytes, digest_length = 64) {
```
Trivo25, line 35: nit: `camelCase`, `digestLength`

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,189 @@
...
+    const state = initialize(digest_length);
+    update(state, Bytes.from(data).bytes);
+    const out = final(state);
+    return Bytes.from(out);
+  },
+  get IV() {
+    return BLAKE2BConstants.IV.map((x) => UInt64.from(x));
```
Trivo25, line 42: nit: consider avoiding computing this multiple times and only do it once

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,189 @@
...
+function G(
+  v: UInt64[],
+  a: number,
+  b: number,
+  c: number,
+  d: number,
+  x: UInt64,
+  y: UInt64
+) {
+  v[a] = v[a].addMod64(v[b].addMod64(x));
```
Trivo25, line 75: I think we could save a good chunk of constraints on some of these by using addition on field elements, ignoring the overflow and only at the very end reducing them mod64, eg

```diff
- v[a] = v[a].addMod64(v[b].addMod64(x));
 
+ const unreducedVA = v[a].value.add(v[b].value.add(x.value));
+ v[a] = UInt64.Unsafe.fromField(Gadgets.divMod64(unreducedVA, 1).remainder);
```

We can do this because `divMod64` range checks the `remainder` to 64 bit `rangeCheck64(remainder);` already, so we don't need to do this at every addition step in the chain -- this way we save at least one call to `addMod64`

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,189 @@
...
+    f: UInt64[];
+    buf: UInt8[];
+    buflen: number;
+    outlen: number;
+  },
+  input: UInt8[]
+): void {
+  for (let i = 0; i < input.length; i++) {
+    if (state.buflen === 128) {
+      state.t[0] = state.t[0].add(128); // ?
```
Trivo25, line 152: what's the question mark for?

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,189 @@
...
+
+function final(state: {
+  h: UInt64[];
+  t: UInt64[];
+  f: UInt64[];
+  buf: UInt8[];
+  buflen: number;
+  outlen: number;
+}): UInt8[] {
+  state.t[0] = state.t[0].add(state.buflen); // ?
```
Trivo25, line 171: question mark?

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -448,6 +450,29 @@ const Gadgets = {
...
+   *
+   * **Note:** Both {@link Field} elements need to fit into `2^paddedLength - 1`. Otherwise, an error is thrown and no proof can be generated.
+   * For example, with `length = 2` (`paddedLength = 16`), `and()` will fail for any input that is larger than `2**16`.
+   *
+   * @example
+   * ```typescript
+   * let a = Field.from(3);    // ... 000011
+   * let b = Field.from(5);    // ... 000101
+   *
+   * let c = Gadgets.or(b);    // ... 000111
```
Trivo25, line 469: this should be `Gadgets.or(a, b, 16);`

src/lib/provable/test/blake2b.unit-test.ts
```diff
@@ -0,0 +1,84 @@
...
+
+
+
+for (let { digest_length ,preimage, hash } of testVectors()) {
+  let actual = Gadgets.BLAKE2B.hash(Bytes.fromString(preimage), digest_length);
+  expect(actual.toHex()).toEqual(hash);
+}
+
+
+function testVectors() {
```
Trivo25, line 21: where are some of these test vectors from? we should use some official ones also (if they arent the official ones!)

tests/vk-regression/vk-regression.json
```diff
@@ -288,16 +296,29 @@
     }
   },
   "sha256": {
-    "digest": "38eaa6264b1e4c1c0b5787c06efd92aa7855890aeeafed91818d466f1b109ca5",
+    "digest": "2502d4e6456fcb0b405f2bf995b34bd1d703951bcd0d461247327d3c86335691",
```
Trivo25, line 299: oh! This shouldn't have changed, let's investigate

Trivo25 (commented):
Great job! I left some initial comments after a first pass

src/lib/provable/crypto/hash.ts
boray: good catch! wrong and unused.

src/lib/provable/test/blake2b.unit-test.ts
boray: [Official test vectors](https://github.com/BLAKE2/BLAKE2/blob/master/testvectors/blake2b-kat.txt) have  key parameter which I didn't implement. Therefore, I had to create test vectors manually.

src/lib/provable/gadgets/blake2b.ts
boray: I left that while debugging and forgot to remove it.

src/lib/provable/gadgets/blake2b.ts
boray: same as above. removed.

src/lib/provable/gadgets/blake2b.ts
boray: Yeah, we can do that. Thanks for pointing that out. Second argument of `divMod64` should be 64 for the new implementation of it.

mitschabaude:
@boray since this relies on the changes in #1763, you should change the base branch to `feature/divmod32-quotientbits-bound` to make this easier to review

mitschabaude:
changed the base branch, which reveals that you didn't merge that branch but  added it as a single squashed commit.

I highly recommend to never do that! it means we have the same changes in two different commits, which git can't resolve and instead will mark them as conflicts. (see "This branch has conflicts that must be resolved" on this PR now)

the preferred way to use git at o1Labs is to basically always join branches using `git merge`, which reuses the original commits and so doesn't create any fake conflicts

tests/vk-regression/vk-regression.json
boray: Now, this doesn't change.

mitschabaude (commented):
a few quick comments on the gadgets additions!

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,53 @@ function divMod32(n: Field, nBits = 64) {
...
+      let r = nBigInt - (q << 64n);
+      return [q, r] satisfies [bigint, bigint];
+    }
+  );
+
+  if (quotientBits === 1) {
+    quotient.assertBool();
+  } else {
+    rangeCheckN(quotientBits, quotient);
+  }
```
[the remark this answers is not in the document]
boray: quotientBits is not necessarily 64 since it's calculated from nBits. So, I think it's better to keep `rangeCheckN()`.

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,53 @@ function divMod32(n: Field, nBits = 64) {
 function addMod32(x: Field, y: Field) {
   return divMod32(x.add(y), 33).remainder;
 }
+
+
+function divMod64(n: Field, nBits = 128) {
+  assert(nBits >= 0 && nBits <= 255, `nBits must be in the range [0, 255], got ${nBits}`)
```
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,53 @@ function divMod32(n: Field, nBits = 64) {
...
+
+  return {
+    remainder,
+    quotient,
+  };
+}
+
+function addMod64(x: Field, y: Field) {
+  return divMod64(x.add(y), 128).remainder;
+}
```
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/gadgets/bitwise.ts
```diff
@@ -173,6 +174,35 @@ function and(a: Field, b: Field, length: number) {
   return outputAnd;
 }
 
+function or(a: Field, b: Field, length: number) {
```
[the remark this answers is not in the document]
boray: makes sense! I replaced the dance with the snippet above.

src/lib/provable/test/bitwise.unit-test.ts
```diff
@@ -183,6 +193,18 @@ await equivalentAsync({ from: [maybeField, maybeField], to: field }, { runs })(
   }
 );
 
+await equivalentAsync({ from: [maybeField, maybeField], to: field }, { runs })(
+  (x, y) => {
+    if (x >= 2n ** 64n || y >= 2n ** 64n)
+      throw Error('Does not fit into 64 bits');
+    return x & y;
```
querolita, line 200: why do you return `&` in this case but `Bitwise.or` in the next?

src/lib/provable/int.ts
```diff
@@ -863,6 +887,23 @@ class UInt32 extends CircuitValue {
...
+   * ```typescript
+   * let a = UInt32.from(3);    // ... 000011
+   * let b = UInt32.from(5);    // ... 000101
+   *
+   * let c = a.or(b);    // ... 000111
+   * c.assertEquals(7);
+   * ```
+   */
+  or(x: UInt32) {
+    return new UInt32(Bitwise.or(this.value, x.value, UInt32.NUM_BITS).value);
```
querolita, line 904: Newbie question here: why are there two identical definitions of these functions in the file? Meaning, this one on line 903 and the other one on line 426. Same for the other function names.

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,55 @@ function divMod32(n: Field, nBits = 64) {
 function addMod32(x: Field, y: Field) {
   return divMod32(x.add(y), 33).remainder;
 }
+
+function divMod64(n: Field, nBits = 128) {
+  assert(
+    nBits >= 0 && nBits < 255,
+    `nBits must be in the range [0, 255), got ${nBits}`
```
querolita, line 63: Is 255 not included because of Pasta curves?

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,55 @@ function divMod32(n: Field, nBits = 64) {
...
+    () => {
+      let nBigInt = n.toBigInt();
+      let q = nBigInt >> 64n;
+      let r = nBigInt - (q << 64n);
+      return [q, r] satisfies [bigint, bigint];
+    }
+  );
+
+  if (quotientBits === 1) {
+    quotient.assertBool();
```
querolita, line 94: Is this a special case for a more efficient circuit in this concrete case?

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,55 @@ function divMod32(n: Field, nBits = 64) {
...
+  n.assertEquals(quotient.mul(1n << 64n).add(remainder));
+
+  return {
+    remainder,
+    quotient,
+  };
+}
+
+function addMod64(x: Field, y: Field) {
+  return divMod64(x.add(y), 128).remainder;
```
querolita, line 109: What's the reasoning behind `nBits` being 128 in this case? (Instead of a larger or smaller value perhaps). Is it to prevent the quotient from being larger than 64 bits upon reduction?

querolita (commented):
Great PR 😄 

For today I am leaving some comments regarding the arithmetic side of this PR. Tomorrow I will look deeper into the actual hash function and leave feedback there as well.

src/lib/provable/int.ts
boray: One is for UInt64 and the other is for UInt32.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+  const v = h.concat(BLAKE2B.IV); // initalize local work vector. First half from state and second half from IV.
+
+  v[12] = v[12].xor(t[0]); // low word of the offset
+  v[13] = v[13].xor(t[1]); // high word of the offset
+  v[14] = v[14].xor(f[0]);
+  v[15] = v[15].xor(f[1]);
+
+  const m: UInt64[] = [];
+  for (let i = 0; i < 16; i++) {
+    m.push(
```
querolita, line 107: I would clarify here that it is little-endian encoding.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+  h: UInt64[];
+  t: UInt64[];
+  f: UInt64[];
+  buf: UInt8[];
+  buflen: number;
+  outlen: number;
+}): UInt8[] {
+  state.t[0] = state.t[0].add(state.buflen);
+  if (state.t[0].equals(UInt64.zero)) {
+    state.t[1] = state.t[1].add(UInt64.zero);
```
querolita, line 174: I don't understand this line. Why do you add zero? I would have said one, as in the `update` function.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+  state.f[0] = UInt64.from('0xFFFFFFFFFFFFFFFF');
+
+  while (state.buflen < 128) {
```
querolita, line 178: I would write a comment here saying that this is the padding to null until reaching a multiple of 128 length.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+    state.buf[state.buflen++] = UInt8.from(0);
+  }
+  compress(state);
+
+  const out: UInt8[] = [];
+  for (let i = 0; i < state.outlen; i++) {
+    out[i] = UInt8.from(
+      state.h[i >> 3].rightShift(8 * (i & 7)).and(UInt64.from(0xff))
```
querolita, line 186: Write a comment with an example showing how the digest is obtained from the state.

querolita (commented):
Finished my first pass, left some comments and questions 👍🏻

src/lib/provable/test/blake2b.unit-test.ts
querolita: I would in any case include tests with longer preimages, to make sure the code works when two blocks are required (meaning t1 is nonzero). I believe the longer preimage in these tests was 43 bytes. Let's try with 128 for example.

src/lib/provable/int.ts
querolita: Ah okay nice, I missed the scope of the methods thank you!

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+function initialize(outlen: number): {
+  h: UInt64[];
+  t: UInt64[];
+  f: UInt64[];
+  buf: UInt8[];
+  buflen: number;
+  outlen: number;
+} {
+  const h = BLAKE2B.IV.slice(); // shallow copy IV to h
+  h[0] = UInt64.from(0x01010000).xor(UInt64.from(outlen)).xor(h[0]);
```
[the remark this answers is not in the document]
mitschabaude: `UInt64.from(0x01010000)` puts a constant Field element inside `UInt64.value`
same for `UInt64.from(outlen)`, because `outLen` is a `number` just like `0x01010000` (the fact that it's a function argument rather than a constant makes no difference)

so here, `h[0]` still gets assigned a constant again!

when that constant is combined with variables in the circuit, it works just like in snarky. i.e. the constant is either folded into gate coefficients, or, most commonly, a dedicated generic gate is added that creates a variable fixed to be equal to this constant

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+    buf: UInt8[];
+    buflen: number;
+    outlen: number;
+  },
+  input: UInt8[]
+): void {
+  for (let i = 0; i < input.length; i++) {
+    if (state.buflen === 128) {
+      state.t[0] = state.t[0].add(128);
+      if (state.t[0].equals(UInt64.zero)) {
```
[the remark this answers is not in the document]
mitschabaude: btw @boray this is a bug. you can't use a `Bool` as condition, it's an object, always truthy!

I highly recommend to do out-of-circuit logic on plain JS values, like bigint to avoid mistakes like this, i.e. here I would use
```
if (state.t[0].toBigint() === 0n) {
```

src/lib/provable/gadgets/arithmetic.ts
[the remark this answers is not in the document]
mitschabaude: I suggested to use `rangeCheck64` **if quotientBits === 64**. so what I mean is, test for that condition, and use the efficient range check in that case

src/lib/provable/gadgets/arithmetic.ts
mitschabaude: yeah if both inputs are 64 bits then the result is 65 bits, that would cause an efficient (boolean) check for the quotient

src/lib/provable/gadgets/arithmetic.ts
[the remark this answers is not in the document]
boray: Okay, got it. This improvement is relevant for divMod32 too!

src/lib/provable/test/bitwise.unit-test.ts
boray: Good catch! That should be `|` not `&`. Interestingly, it doesn't fail the unit test. Any ideas, @mitschabaude?

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -448,6 +450,29 @@ const Gadgets = {
   and(a: Field, b: Field, length: number) {
     return and(a, b, length);
   },
+  /**
+   * Bitwise OR gadget on {@link Field} elements. Equivalent to the [bitwise OR `|` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_OR).
+   * The OR gate works by comparing two bits and returning `1` if at least one bit is `1`, and `0` otherwise.
+   *
+   * The `length` parameter lets you define how many bits should be compared. `length` is rounded to the nearest multiple of 16, `paddedLength = ceil(length / 16) * 16`, and both input values are constrained to fit into `paddedLength` bits. The output is guaranteed to have at most `paddedLength` bits as well.
```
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/test/bitwise.unit-test.ts
mitschabaude: my only idea is that you use such a small number of `runs` that with high likelyhood, one of the two inputs is bigger than $2^64$ in every single run, so you get an error every time.

try it with `uint(64)` as the random generator instead of `maybeField`!

mitschabaude: I think you can actually remove this unit test that creates a proof, and stick with the ones above that only run a circuit and check constraints

The proof one will make CI slower and provides hardly any additional value

src/lib/provable/gadgets/arithmetic.ts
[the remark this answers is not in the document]
mitschabaude: no it's not! there is no dedicated single-gate 32-bit range check, like there is for 64 bits

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -886,4 +948,24 @@ const Gadgets = {
...
+
+  /**
+   * Implementation of the [BLAKE2b hash function.](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2) Hash function with arbitrary length output.
+   *
+   * Applies the BLAKE2b hash function to a list of byte-sized elements.
+   *
+   * The function accepts {@link Bytes} as the input message, which is a type that represents a static-length list of byte-sized field elements (range-checked using {@link Gadgets.rangeCheck8}).
+   * Alternatively, you can pass plain `number[]`, `bigint[]` or `Uint8Array` to perform a hash outside provable code.
+   *
+   * Produces an output of {@link Bytes} that conforms to the chosen bit length.
```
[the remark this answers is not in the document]
boray: Generally, digest lengths are denoted in bits.

src/lib/provable/gadgets/arithmetic.ts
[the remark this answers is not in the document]
boray: I didn't know that! I removed divMod32 change with the latest commit. AFAIU from the kimchi page of the proof systems book, only 64-bit range check has a dedicated single-gate. Quoting from the book: 
> The RangeCheck0 gate can also be used on its own to perform 64-bit range checks by constraining witness cells 1-2 to zero.

src/lib/provable/gadgets/arithmetic.ts
boray: I am not sure why we don't include 255. I followed the range checking practice used in `divMod32`. Maybe @mitschabaude knows the reason?

src/lib/provable/gadgets/arithmetic.ts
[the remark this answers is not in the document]
querolita: It is true that only range checks for 64bits have a dedicated Kimchi gate. However, one can also perform 32bit range checks just less efficiently. Maybe let's not remove the o1js function for it as it might be better from a ux point of view, instead of asking the zkapp developer to use a rc64 for x and another one for x * 2^32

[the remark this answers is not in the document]
mitschabaude: For 32 bits there are two different ways to do it in 2.5 gates, and one of them is already used -- the one that doesn't use the range check gate.
Using that one is better because the range check gate and lookup table are only optionally included in the proof system, and the proof gets slower by including them. So it only pays off if you need a lot of range checks.

src/lib/provable/gadgets/arithmetic.ts
mitschabaude: Because the field size is less than 2^255, there are 255 bit integers that have two different representations in the field (in fact that's true for almost all 255 bit integers since the field size is just slightly over 2^254)

That non-uniqueness would make the divmod gadget unsound: the prover could pick between two different quotient/remainder splits, one of which is wrong.

src/lib/provable/gadgets/arithmetic.ts
boray: yes, it should be 65. It used to be 128 at some point. I adapted the changes made in #1763 and forgot to update nBits.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+    [2, 12, 6, 10, 0, 11, 8, 3, 4, 13, 7, 5, 15, 14, 1, 9],
+    [12, 5, 1, 15, 14, 13, 4, 10, 0, 7, 6, 3, 9, 2, 8, 11],
+    [13, 11, 7, 14, 12, 1, 3, 9, 5, 0, 15, 4, 8, 6, 2, 10],
+    [6, 15, 14, 9, 11, 3, 0, 8, 12, 2, 13, 7, 1, 4, 10, 5],
+    [10, 2, 8, 4, 7, 6, 1, 5, 15, 11, 9, 14, 3, 12, 13, 0],
+  ],
+};
+
+const BLAKE2B = {
+  hash(data: FlexibleBytes, digestLength = 64) {
```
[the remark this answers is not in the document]
boray: Yes, that was missing. I added a check with the latest commit.

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -867,6 +892,43 @@ const Gadgets = {
...
+  divMod64,
+
+  /**
+   * Addition modulo 2^64. The operation adds two {@link Field} elements in the range [0, 2^128] and returns the result modulo 2^64.
+   *
+   * Asserts that the result is in the range [0, 2^64) using {@link Gadgets.rangeCheck64}.
+   *
+   * It uses {@link Gadgets.divMod64} internally by adding the two {@link Field} elements and then decomposing the result into `remainder` and `quotient` and returning the `remainder`.
+   *
+   * **Note:** The gadget assumes both inputs to be in the range [0, 2^64). When called with non-range-checked inputs, be aware that the sum `a + b` can overflow the native field and the gadget can succeed but return an invalid result.
```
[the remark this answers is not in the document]
boray: yes, the inputs are not checked. addMod does field addition first then divMod with the result.

src/lib/provable/test/bitwise.unit-test.ts
boray: While developing smart contracts, sometimes it works without proofs but fails with them. Isn't it beneficial to test with proofs. Is it different for primitives?

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
boray: `state.t[0]` is a variable but I am not sure if that logic should be in-circuit or not. I added the in-circuit version as a comment.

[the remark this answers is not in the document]
boray: I implemented this part with reference to the snippet below from the [BLAKE2B RFC](https://datatracker.ietf.org/doc/html/rfc7693.html). 
```c
      ctx->t[0] += ctx->c;                // mark last block offset
       if (ctx->t[0] < ctx->c)             // carry overflow
           ctx->t[1]++;                    // high word
```

src/lib/provable/gadgets/blake2b.ts
boray: I added a comment!

src/lib/provable/gadgets/blake2b.ts
boray: Yes, that’s my mistake. It should be one according to the RFC as well. I missed that because the test vectors didn’t contain a long preimage. I’ve added a long one to prevent this from happening again.

src/lib/provable/gadgets/blake2b.ts
boray: I added a comment!

src/lib/provable/gadgets/blake2b.ts
boray: I added a comment but I am not sure how an example would look like.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+  f: UInt64[];
+  buf: UInt8[];
+  buflen: number;
+  outlen: number;
+} {
+  const h = BLAKE2B.IV.slice(); // shallow copy IV to h
+  h[0] = UInt64.from(0x01010000).xor(UInt64.from(outlen)).xor(h[0]);
+  return {
+    h,
+    t: [UInt64.zero, UInt64.zero],
```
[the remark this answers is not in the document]
boray: yes, I added a comment about endianness to the top.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+      UInt64.from(
+        buf[i * 8]
+          .toUInt64()
+          .or(buf[i * 8 + 1].toUInt64().leftShift(8))
+          .or(buf[i * 8 + 2].toUInt64().leftShift(16))
+          .or(buf[i * 8 + 3].toUInt64().leftShift(24))
+          .or(buf[i * 8 + 4].toUInt64().leftShift(32))
+          .or(buf[i * 8 + 5].toUInt64().leftShift(40))
+          .or(buf[i * 8 + 6].toUInt64().leftShift(48))
+          .or(buf[i * 8 + 7].toUInt64().leftShift(56))
```
mitschabaude, line 161: this should be done with addition and multiplication instead of `or()` and `leftShift()`

e.g.
```diff
- .or(buf[i * 8 + 1].toUInt64().leftShift(8))
+ .add(buf[i * 8 + 1].value.mul(1n << 8n))
```

in a circuit, addition and multiplication is cheap while bitwise operations are expensive!

you can also leave it and we do a second pass for efficiency, and you focus on correctness for now?

boray: I wasn't aware that add/mul is cheaper than bitwise ops. Let's focus on correctness now and leave this improvement to the efficiency pass.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,190 @@
...
+  h: UInt64[];
+  t: UInt64[];
+  f: UInt64[];
+  buf: UInt8[];
+  buflen: number;
+  outlen: number;
+} {
+  const h = BLAKE2B.IV.slice(); // shallow copy IV to h
+  h[0] = UInt64.from(0x01010000).xor(UInt64.from(outlen)).xor(h[0]);
+  return {
```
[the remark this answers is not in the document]
boray: I added the exact comment from the reference implementation. This line is more understandable when other parameters, like key, salt, and personalization, are present. I didn’t implement those since this implementation was requested by LambdaClass for the bridge effort, which didn’t require those parameters.

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -867,6 +892,43 @@ const Gadgets = {
...
+   * let { remainder, quotient } = Gadgets.divMod64(n);
+   * // remainder = 8, quotient = 1
+   *
+   * n.assertEquals(quotient.mul(1n << 64n).add(remainder));
+   * ```
+   */
+  divMod64,
+
+  /**
+   * Addition modulo 2^64. The operation adds two {@link Field} elements in the range [0, 2^128] and returns the result modulo 2^64.
```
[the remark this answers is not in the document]
boray: TBH, I am not sure which one is right. The ranges in `addMod32` comment block doesn't match too. I took the comment for `addMod64` from `addMod32` since it's the exact same function but modulo 2^64.

src/lib/provable/gadgets/blake2b.ts
querolita: Basically, as a rule of thumb, binary operations which are cheap on CPU are expensive in cryptography (where arithmetic operations in a field are the "cheap" operations instead). Modeling binary operations in a circuit becomes a big overhead, so anytime possible where binary ops can be replaced with arithmetic counterparts, is preferred for efficiency reasons. 

Either way, let's not forget that second pass for efficiency in a later PR.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+}): UInt8[] {
+  state.t[0] = state.t[0].addMod64(UInt64.from(state.buflen)); // mark last block offset
+  if (state.t[0].toBigInt() < state.buflen) {
+    // carry overflow ?
+    state.t[1] = state.t[1].addMod64(UInt64.one); // high word
+  }
+  /*
+  state.t[1] = state.t[1].add(
+    Provable.if(
+      state.t[0].lessThan(UInt64.from(state.buflen)),
```
querolita, line 264: If this needs to go in the circuit, then the same lines of the `update` part should also be written in circuit manner. In any case, would it make sense to have a final constrain checking `t[1]*2^64+t[0]=total_length`? That, together with the modular additions gives you a correct decomposition. 

Now, going back to what @mitschabaude said about the `state.t[0]` being a variable, one cannot decide the design of the circuit depending on the value of a witness cell. The circuit must have the same "universal shape", regardless of the concrete values of the variables. 

So putting all the above together, it's not a matter of whether that logic needs to go into the circuit or not. It's about not being able to decide upon the concrete value a variable takes. Thus, I believe this part should be rewritten, perhaps with the use of `Provable.if` (but I might be missing the right o1js know-how, so maybe ask another team member for further insights). 

At least in Kimchi, one would create a circuit that is already "bifurcated". Meaning, all the constraints hold and they are the same, no matter what concrete values are passed to the prover. So we play with multiplications to "select" one branch or another, but both sides should eventually evaluate to 0 to satisfy the constraint system.

querolita (commented):
Leaving two more comments before acceptance. These are related to the actual constraints being created behind the scenes.

src/lib/provable/gadgets/arithmetic.ts
```diff
@@ -56,3 +56,57 @@ function divMod32(n: Field, nBits = 64) {
...
+
+  if (n.isConstant()) {
+    assert(
+      n.toBigInt() < 1n << BigInt(nBits),
+      `n needs to fit into ${nBits} bits, but got ${n.toBigInt()}`
+    );
+    let nBigInt = n.toBigInt();
+    let q = nBigInt >> 64n;
+    let r = nBigInt - (q << 64n);
+    return {
```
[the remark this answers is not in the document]
Trivo25: Yes! For constant cases (where all variables of a computation are constant), we calculate the result without creating constraints and return another constant value

src/lib/provable/gadgets/blake2b.ts
mitschabaude: @boray you can think of `add()`, `mul()` and `assertEqual()` as the primitive operations out of which everything else is built. nothing else is as efficient, and most things are far less efficient

mitschabaude: after all this all turns into polynomial equations, and polynomials are made up of addition and multiplication

src/lib/provable/test/blake2b.unit-test.ts
mitschabaude: > Therefore, I had to create test vectors manually.

How?

src/lib/provable/gadgets/gadgets.ts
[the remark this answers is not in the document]
mitschabaude: first of all, all ranges are always of the form `[a, b)` i.e. open on the upper end

second, 128 bits is wrong. it uses a 1-bit range check on the quotient. for that to work, the addition result has to be at most 65 bits

also, you say further below that "the gadget assumes both inputs to be in the range [0, 2^64)". that will definitely work an is the use case, so stick with that here in the description as well

src/lib/provable/test/bitwise.unit-test.ts
```diff
@@ -183,6 +193,18 @@ await equivalentAsync({ from: [maybeField, maybeField], to: field }, { runs })(
...
+    if (x >= 2n ** 64n || y >= 2n ** 64n)
+      throw Error('Does not fit into 64 bits');
+    return x | y;
+  },
+  async (x, y) => {
+    let proof = await Bitwise.or(x, y);
+    return proof.publicOutput;
+  }
+);
+
```
mitschabaude, line 207: I'd say remove this, the in-circuit equivalence tests above are much more thorough and creating a proof here just makes CI slower

tests/vk-regression/vk-regression.json
```diff
@@ -204,6 +208,10 @@
       "Poseidon": {
         "rows": 208,
         "digest": "afa1f9920f1f657ab015c02f9b2f6c52"
+      },
+      "BLAKE2B": {
+        "rows": 6012,
```
mitschabaude, line 213: nice!

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,282 @@
...
+  for (let i = 0; i < input.length; i++) {
+    if (state.buflen === 128) {
+      // buffer full ?
+      state.t[0] = state.t[0].addMod64(UInt64.from(state.buflen)); // add counters
+      if (state.t[0].toBigInt() < state.buflen) {
+        // carry overflow ?
+        state.t[1] = state.t[1].addMod64(UInt64.one); // high word
+      }
+      state = compress(state, false); // compress (not last)
+      state.buflen = 0; // counter to zero
```
mitschabaude, line 236: Shouldn't this also reset `state.buf` to an empty array? If you don't do this, you'll start overwriting `buf` at the start but the rest of it will still contain the old entries.

and if yes, wouldn't it be less confusing in general to always use `state.buf.length` instead of an extra parameter `buflen`, so that they can't get out of sync?

mitschabaude (changes requested):

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,282 @@
...
+  while (state.buflen < 128) {
+    state.buf[state.buflen++] = UInt8.from(0); // fill up with zeroes
+  }
+  compress(state, true);
+
+  // little endian convert and store
+  const out: UInt8[] = [];
+  for (let i = 0; i < state.outlen; i++) {
+    out[i] = UInt8.from(
+      state.h[i >> 3].rightShift(8 * (i & 7)).and(UInt64.from(0xff))
```
mitschabaude, line 278: if you aim for efficiency: here you can also use arithmetic instead of bitwise operations

boray: I tried to replace this line with `state.h[i >> 3].div(2 ** (8 * (i & 7))).mod(0xff)` but it increased the row number a lot. I am not sure if I'm missing something.

src/lib/provable/test/bitwise.unit-test.ts
mitschabaude: It's different here yeah, for various reasons.

For one, we are already testing many extremely similar methods with proofs in this file, the additional surface covered is funny

src/lib/provable/gadgets/blake2b.ts
mitschabaude: You can check out the bytesToWord function that's used in the keccak and sha2 implementations!
Probably a good idea to use that function

src/lib/provable/gadgets/blake2b.ts
boray: I refactored in the last commit!

src/lib/provable/test/bitwise.unit-test.ts
boray: removed!

src/lib/provable/test/blake2b.unit-test.ts
```diff
@@ -0,0 +1,89 @@
...
+} from '../../testing/equivalent.js';
+import { Random, sample } from '../../testing/random.js';
+import { expect } from 'expect';
+
+
+
+for (let { digest_length ,preimage, hash } of testVectors()) {
+  let actual = Gadgets.BLAKE2B.hash(Bytes.fromString(preimage), digest_length);
+  expect(actual.toHex()).toEqual(hash);
+}
```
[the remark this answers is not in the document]
boray: I added in-circuit tests in the last commit.

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+ *
+ * All mathematical operations are on 64-bit words in BLAKE2b.
+ *
+ * Byte (octet) streams are interpreted as words in little-endian order,
+ * with the least-significant byte first.
+ */
+
+type State = {
+  h: UInt64[];
+  t: UInt64[];
```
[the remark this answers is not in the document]
boray: no, `t` is set to `[Uint64.zero, Uint64.zero]` at initialisation. `t[0]` is incremented by `buflen` during every `update()` and `final()`.  However, t[1] is only used when input size is larger than 2^64 bytes which is unlikely. We can brake up the `t `into two `Uint64`s instead of an array, but would it improve efficiency?

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+  }
+  /*
+  state.t[1] = state.t[1].add(
+    Provable.if(
+      state.t[0].lessThan(UInt64.from(state.buflen)),
+      UInt64.one,
+      UInt64.zero
+    )
+  );
+*/
```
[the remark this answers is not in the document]
boray: I was confused because the non-provable version of this part didn’t raise any errors. But, I believe this logic needs to be provable.

src/lib/provable/test/blake2b.unit-test.ts
boray: I hashed messages from other hash function unit tests using `nobleBlake2b`. Also, I added longer messages with @querolita 's recommendation.

src/lib/provable/test/arithmetic.unit-test.ts
```diff
@@ -45,7 +60,20 @@ equivalent({
   }
 );
 
-await equivalentAsync({ from: [field], to: divMod32Output }, { runs: 3 })(
+equivalent({
+  from: [field],
+  to: divModOutput,
+})(
+  (x) => {
+    assert(x < 1n << 128n, `x needs to fit in 128bit, but got ${x}`);
```
boray, line 68: This assertion fails because the random field is larger than 2^128. It works for divMod32 above, which is smaller, and that should fail first. Any ideas, @mitschabaude?

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
mitschabaude: I believe it doesn't!

src/lib/provable/test/arithmetic.unit-test.ts
mitschabaude: the test checks whether the two functions are equivalent.
so the **test succeeds if both functions throw an error** for the same input.

so, it's not necessarily the case that "divMod32 should fail first". Maybe, `divMod32()` correctly raises an error if the input is larger than 2^64.

if this test fails, my idea would be to check why `divMod64()` doesn't throw an error on inputs larger than 2^128

src/lib/provable/test/blake2b.unit-test.ts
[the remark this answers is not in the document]
mitschabaude: I see it, but there were two misunderstandings:
* I didn't mean to run them in a ZkProgram with proving, but just in a circuit with checking of constraints. You can use `runAndCheckSync()` for that like in other test files
* I meant run the same test vectors in a circuit, not a small number of random inputs like you did

src/lib/provable/gadgets/blake2b.ts
mitschabaude: I meant `wordToBytes()` btw 😅

src/lib/provable/test/bitwise.unit-test.ts
mitschabaude: Huh, I meant just this test but you removed all of them. I think it's not a bad idea to remove most of them, but can you at least keep 1-2 of them?

src/lib/provable/test/arithmetic.unit-test.ts
boray: I checked both functions, and they throw an error on inputs larger than 2^128. The error disappeared when I either removed the `quotientBits === 64` case in `divMod64` or replaced it with a `rangeCheckN()`.

src/lib/provable/gadgets/blake2b.ts
boray: yeah, I tried that, but it doesn’t work for this case. `wordToBytes()` takes a word as `Field` and converts it into an array of bytes as `UInt8[]`. Here, the logic is more complex. I don’t think a specific function in `bit-slices.ts` is necessary for this case since it would only be used here.

src/lib/provable/gadgets/blake2b.ts
boray: The reference implementation doesn’t do that, so I followed the same approach.

src/lib/provable/test/bitwise.unit-test.ts
boray: Which ones should we keep?

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
boray: The gadget compiles with both. Is there a way to test our beliefs?

src/lib/provable/test/arithmetic.unit-test.ts
mitschabaude: I know what happens now, rangeCheck64 doesn't throw an error in constraint generation mode because snarky doesn't support checking custom gates. 

This isn't a problem for real proofs (because there, the proof would fail even if constraint checking succeeded).

So, I recommend to resolve this by simply not testing large field elements in the test, just make the inputs uint64

src/lib/provable/test/bitwise.unit-test.ts
mitschabaude: Let's keep the first two

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
mitschabaude: Yes there's a way to find out, make 't' consist of bigints, if you still can make everything work without issues then t never interacts with a variable, and so stays constant, and so the logic isn't in-circuit

src/lib/provable/gadgets/blake2b.ts
mitschabaude: Ok interesting, in that case I would try to find a spec and see if it matches the reference impl on this point. It could indeed be the correct behavior, or it could be a bug in the reference!

boray: There are two specifications on the [official BLAKE2B website](https://www.blake2.net). [One of them](https://www.blake2.net/blake2.pdf) doesn’t include a reference implementation, while the [other](https://datatracker.ietf.org/doc/html/rfc7693#section-2.6) does. The reference implementation has a `c` variable that is maintained in the state as a pointer to the buffer. Similarly, the `noble-hashes` implementation of BLAKE2B uses `pos`, which is the same as `buflen`. I’m in favor of following the reference implementation.

mitschabaude: Ok, I checked another implementation and it does the same. Lgtm

src/lib/provable/gadgets/blake2b.ts
mitschabaude: The way I read this logic, it just converts each 64-bit state.h element to 8 bytes (UInt8), in little indian order. So it's the same as wordToBytes, you should be able to use something like
```ts
let out = state.h.slice(0, 8*outlen)
  .flatMap(x => wordToBytes(x.value))
```

src/lib/provable/gadgets/gadgets.ts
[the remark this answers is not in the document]
boray: fixed!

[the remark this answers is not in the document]
mitschabaude: the inputs are in the range [0, 2^64) though, not [0, 2^65)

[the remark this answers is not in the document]
boray: right, it should be correct now.

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
boray: I did that and everything worked but I worry the proof generation would fail when the case t[1] is changed reached.

src/lib/provable/gadgets/blake2b.ts
boray: It works! Thanks.

src/lib/provable/test/blake2b.unit-test.ts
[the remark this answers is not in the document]
boray: Got it. I added a test that uses `equivalentProvable()` to run on test vectors, which revealed a mismatch in the 10k-byte preimage. I’m investigating the why it's happening.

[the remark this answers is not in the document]
boray: sorry, it's a false alarm. It works with 10k-byte preimage too.

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
mitschabaude: then you need to make sure that this case is tested (should be ensured anyway!)

tests/vk-regression/vk-regression.json
boray: It's now down to 4150!

src/lib/provable/gadgets/blake2b.ts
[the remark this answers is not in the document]
mitschabaude: Oh right. Then that case doesn't matter :D 

But again, I went through the code a while ago and concluded that t0 / t1 at different steps of the circuit only depend on the constant input size. 

So they are inherently constant. A constant is just a wrapper around a bigint, and your "provable logic" applied to a constant just produces another constant (= another bigint), and is not proved at all. So it's all the same when you do those operations in plain JS. You're just making it more explicit and easier to see that they are on constants.

It's fine!

mitschabaude (approved):
all my comments are addressed as far as I can see. great job @boray!

querolita (approved):

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,277 @@
...
+  v[b] = v[b].xor(v[c]).rotate(63, 'right');
+}
+
+/**
+ * Compression function. "last" flag indicates last block.
+ * @param {State} state
+ * @param {boolean} last
+ */
+function compress(
+  state: {
```
Trivo25, line 129: nit: You define a `State` type above, might as well use it

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/**
+ * Updates hash state
+ * @param {State} state
+ * @param {UInt8[]} input
+ * @returns {State} updated state
+ */
+function update(
+  state: {
```
Trivo25, line 220: same

src/lib/provable/test/bitwise.unit-test.ts
```diff
@@ -160,82 +170,6 @@ await equivalentAsync({ from: [maybeField], to: field }, { runs })(
     return proof.publicOutput;
   }
 );
-await equivalentAsync({ from: [maybeField], to: field }, { runs })(
```
Trivo25, line 163: why are we getting rid of all these unit tests?

mitschabaude: https://github.com/o1-labs/o1js/pull/1767#discussion_r1768704766

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+    assert(
+      digestLength >= 1 && digestLength <= 64,
+      `digestLength must be in the range [1, 64], got ${digestLength}`
+    );
+    assert(
+      data.length >= 0 && data.length < 2 ** 128,
+      `data byte length must be in the range [0, 2**128), got ${data.length}`
+    );
+    const state = initialize(digestLength);
+    const updated_state = update(state, Bytes.from(data).bytes);
```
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/gadgets/blake2b.ts
```diff
@@ -0,0 +1,283 @@
...
+  return state;
+}
+
+/**
+ * Finalizes the hash state and returns digest
+ * @param {State} state
+ * @returns {UInt8[]} digest
+ */
+function final(state: {
+  h: UInt64[];
```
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/gadgets/blake2b.ts
boray: fixed!

Trivo25 (approved):

PR #1776, opened 2024-07-26
Update TokenContract to not use depricated methods

ymekuria:
## Summary

This PR removes the use of deprecated APIs within the TokenContract.  

## Changes

- `AccountUpdate.defaultAccountUpdate()` was replaced with `AccountUpdate.default()`.
- `Int64.neg()` was replaced with `Int64.negV2()`.


Trivo25 (changes requested):
We can't use `negV2` in the V1 token contract, this will break the existing constraint system

mitschabaude:
Should go to v2!

Trivo25:
done

PR #1779, opened 2024-07-29
ForeignField reduction necessity comments.

shimkiv:
- https://app.zenhub.com/workspaces/sdk-6130fedb3b0fc600123d8796/issues/zh/114
  - We decided to simply add more comments.

### DIFF
--- a/src/lib/provable/crypto/foreign-curve.ts
+++ b/src/lib/provable/crypto/foreign-curve.ts
@@ -42,7 +42,7 @@ class ForeignCurve {
   /**
    * Create a new {@link ForeignCurve} from an object representing the (affine) x and y coordinates.
    *
-   * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
+   * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Please refer to the {@link ForeignField} constructor comments for more details.
    * 
    * @example
    * ```ts
--- a/src/lib/provable/crypto/foreign-curve.ts
+++ b/src/lib/provable/crypto/foreign-curve.ts
@@ -12,7 +12,6 @@ import { assert } from '../gadgets/common.js';
 import { Provable } from '../provable.js';
 import { provableFromClass } from '../types/provable-derivers.js';
 import { l2Mask, multiRangeCheck } from '../gadgets/range-check.js';
-import type { ForeignField } from '../foreign-field.js';
 
 // external API
 export {
--- a/src/lib/provable/crypto/foreign-curve.ts
+++ b/src/lib/provable/crypto/foreign-curve.ts
@@ -12,6 +12,7 @@ import { assert } from '../gadgets/common.js';
 import { Provable } from '../provable.js';
 import { provableFromClass } from '../types/provable-derivers.js';
 import { l2Mask, multiRangeCheck } from '../gadgets/range-check.js';
+import type { ForeignField } from '../foreign-field.js';
 
 // external API
 export {
@@ -42,6 +43,8 @@ class ForeignCurve {
   /**
    * Create a new {@link ForeignCurve} from an object representing the (affine) x and y coordinates.
    *
+   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
+   * 
    * @example
    * ```ts
    * let x = new ForeignCurve({ x: 1n, y: 1n });
--- a/src/lib/provable/crypto/foreign-curve.ts
+++ b/src/lib/provable/crypto/foreign-curve.ts
@@ -43,7 +43,7 @@ class ForeignCurve {
   /**
    * Create a new {@link ForeignCurve} from an object representing the (affine) x and y coordinates.
    *
-   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
+   * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
    * 
    * @example
    * ```ts
--- a/src/lib/provable/crypto/foreign-ecdsa.ts
+++ b/src/lib/provable/crypto/foreign-ecdsa.ts
@@ -18,6 +18,7 @@ import { Keccak } from './keccak.js';
 import { Bytes } from '../wrapped-classes.js';
 import { UInt8 } from '../int.js';
 import type { Bool } from '../bool.js';
+import type { ForeignField } from '../foreign-field.js';
 
 // external API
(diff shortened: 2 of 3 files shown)


src/lib/provable/foreign-field.ts
```diff
@@ -92,6 +92,27 @@ class ForeignField {
    * ```ts
    * let x = new ForeignField(5);
    * ```
+   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants.
+   * 
+   * - When constructing from another {@link ForeignField} instance, ensure the modulus matches. If not, check the modulus using `Gadgets.ForeignField.assertLessThan()` and handle appropriately.
+   *   @example
+   *   ```ts
+   *   let smallField = new SmallField(0);
```
Trivo25, line 100: is the `0` a typo here? I would use something other than 0 for thie example

src/lib/provable/crypto/foreign-curve.ts
```diff
@@ -42,6 +43,8 @@ class ForeignCurve {
   /**
    * Create a new {@link ForeignCurve} from an object representing the (affine) x and y coordinates.
    *
+   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
```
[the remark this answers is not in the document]
ymekuria: Nit: To match comment style for adding notes elsewhere in the file.

src/lib/provable/crypto/foreign-ecdsa.ts
```diff
@@ -38,7 +39,8 @@ class EcdsaSignature {
 
   /**
    * Create a new {@link EcdsaSignature} from an object containing the scalars r and s.
-   * @param signature
+   * 
+   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
```
ymekuria, line 43:
```suggestion
   *Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Refer to {@link ForeignField} constructor comments for more details.
```

src/lib/provable/foreign-field.ts
```diff
@@ -92,6 +92,31 @@ class ForeignField {
    * ```ts
    * let x = new ForeignField(5);
    * ```
+   * 
+   * _Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants.
```
ymekuria, line 96:
```suggestion
   * Note:_ Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants.
```

ymekuria (approved):
I'm approving with minor comments.

src/lib/provable/foreign-field.ts
```diff
@@ -92,6 +92,31 @@ class ForeignField {
...
+   * 
+   * - When constructing from another {@link ForeignField} instance, ensure the modulus matches. If not, check the modulus using `Gadgets.ForeignField.assertLessThan()` and handle appropriately.
+   *   @example
+   *   ```ts
+   *   class SmallField extends createForeignField(17n) {}
+   *   class LargerField extends createForeignField(19n) {}
+   * 
+   *   let smallField = new SmallField(10);
+   *   let largerField = new LargerField(17);
+   *   Gadgets.ForeignField.assertLessThan(largerField.value, SmallField.modulus);
```
mitschabaude, line 106: I'm not a fan of using constants (10, 17) here, because there's no need to prove that constants are reduced (and the assertion is a no-op). I think you should rather leave open how the fields are constructed, or show how they are witnessed as variables:

```ts
let smallField: SmallField = ...
```

```ts
let smallField = Provable.witness(SmallField, () => 10n);
```

src/lib/provable/foreign-field.ts
```diff
@@ -92,6 +92,31 @@ class ForeignField {
...
+   * - When constructing from a {@link Field3} array, ensure all elements are valid Field elements and range checked.
+   *   @example
+   *   ```ts
+   *   let field3Array = [Field(15), Field(0), Field(0)];
+   *   let x = new ForeignField(field3Array);
+   *   ```
+   * - Ensure constants are correctly reduced to the modulus of the field.
+   *   @example
+   *   ```ts
+   *   let x = new ForeignField(20); // Automatically reduced to the field's modulus
```
mitschabaude, line 118: you can't use the base constructor `ForeignField` directly, so this example should changed to something like `MyField`

mitschabaude:
> We decided to simply add more comments.

Honestly I think it can't hurt to throw an error when a `ForeignField` with a different `.modulus` is passed in, because that would be almost certainly a mistake

shimkiv:
> Honestly I think it can't hurt to throw an error when a `ForeignField` with a different `.modulus` is passed in, because that would be almost certainly a mistake

Thank you @mitschabaude I'll add it back. Should we also throw in `Field3` case:

```typescript
    import { isField } from './core/field-constructor.js';
    ...
    // Field3
    if (Array.isArray(x)) {
      if (x.some((limb) => !isField(limb))) {
        throw new Error(
          `ForeignField constructor: invalid Field3 element. Please provide valid Field elements.`
        );
      }
      this.value = x;
      return;
    }
```

src/lib/provable/foreign-field.ts
```diff
@@ -101,26 +100,33 @@ class ForeignField {
    *   class SmallField extends createForeignField(17n) {}
    *   class LargerField extends createForeignField(19n) {}
    * 
-   *   let smallField = new SmallField(10);
-   *   let largerField = new LargerField(17);
+   *   let smallField: SmallField = ...;
+   *   let largerField: LargerField = ...;
    *   Gadgets.ForeignField.assertLessThan(largerField.value, SmallField.modulus);
    *   let x = new SmallField(largerField);
```
mitschabaude, line 106: this whole example doesn't work anymore with the new error. you can remove it, it's really not a common use

mitschabaude:
> Although I'm not sure if that check makes sense since types should be checked.

absolutely, we don't check things at runtime that are guaranteed by types. note that the different modulus was not part of the type, that's why we needed an error there

src/lib/provable/foreign-field.ts
```diff
@@ -92,15 +92,31 @@ class ForeignField {
...
       this.value = x.value;
       return;
     }
     // Field3
     if (Array.isArray(x)) {
+      if (x.some((limb) => !isField(limb))) {
+        throw new Error(
+          `ForeignField constructor: invalid 'Field3' element. Please provide valid Field elements.`
+        );
+      }
```
mitschabaude, line 119: as I wrote [here](https://github.com/o1-labs/o1js/pull/1779#issuecomment-2262267473), we don't need these kinds of checks. please revert

mitschabaude (changes requested):

src/lib/provable/foreign-field.ts
shimkiv: My apologies, I read it incorrectly.

mitschabaude (approved):

PR #1783, opened 2024-07-31
Packed character improvement

boray:
## Summary
This PR implements the `toInput()` method on `Character` to declare the bit length and reduce the number of fields used when packed.

## Changes

- Implemented `toInput()` on `Character`.


src/lib/provable/option.ts
```diff
@@ -60,7 +60,7 @@ function Option<A extends ProvableType>(
   } {
   type T = InferProvable<A>;
   type V = InferValue<A>;
-  let strictType: Provable<T, V> = ProvableType.get(type);
+  let strictType: Provable<T, V> & { toCanonical: undefined } = type as any;
```
mitschabaude, line 63: this looks like a merge artifact, remove!

src/bindings
mitschabaude, line 1: you shouldn't change the bindings commit!

boray: I did it by mistake at merge. I'll fix it.

mitschabaude (approved):

PR #1784, opened 2024-07-31
Correcting order for actions occuring within the same block for the same account

45930:
The community has raised this issue where their reducer cannot correctly derive the action hash given the response from the archive node.  This only happens when multiple actions are dispatched within the same block.

The zk noid team shared this minimal repro: https://github.com/aii23/actions_issue_poc

The issue is that events from the archive node arrive in the wrong order for producing the actions hash.

For example, running their script against a light node can produce a graphQL response like:

```gql
[
  {
    accountUpdateId: '3',
    data: [
      '23293912831932499872034135232525254163343702531942971872613724597787050731163',
      '1'
    ]
  }
]
[
  {
    accountUpdateId: '5',
    data: [
      '19698539030616983614904223910467017498097850769349730675723779555348012696719',
      '0'
    ]
  }
]
[
  {
    accountUpdateId: '7',
    data: [
      '23344670544599826568585832059189005555232761671965680539590523406038183782503',
      '1'
    ]
  },
  {
    accountUpdateId: '6',
    data: [
      '23379676726753250474018172284972553782392077412995035117267089127972991216710',
      '1'
    ]
  }
]
```

accountUpdateIds 3 and 5 were dispatched in separate blocks, while accountUpdateIds 6, and 7 were dispatched within the same block.  Note the order of the actions top to bottom is [3, 5, 7, 6].

This PR corrects the order by sorting the array by account update id within a single block of actions to correctly represent the order as [3, 5, 6, 7].  I confirmed this fix by running the zk noid script locally against this build of o1js.


src/lib/mina/fetch.ts
```diff
@@ -769,7 +805,6 @@ async function fetchActions(
       );
     }
   });
-  addCachedActions({ publicKey, tokenId }, actionsList, graphqlEndpoint);
```
asimaranov, line 772: Is it required to remove addCachedActions? If yes, could you please explain the reason

45930: Yes good callout.  I forgot to replace this after moving some code around.

asimaranov:
We found out that fix works not in all the scenarios. We used this patch on our backend. It started working. Then it thrown the same issue. 
We checked and the previous fetch function started to work while the patched stopped. 
Address on devnet: `B62qmbX7xkXxLhKQpi4iabLzEtzyXi9uA5pFU1PbACoXCMt6DEC3t1g`
Not patched fetch works on it
Patched fetch throws
```
Error: Failed to derive correct actions hash for B62qmbX7xkXxLhKQpi4iabLzEtzyXi9uA5pFU1PbACoXCMt6DEC3t1g.
        Derived hash: 18711700257838647448684440256433891881288423153285399691473064599228530899996, expected hash: 23151667407356824933654059015358591226495725733024978946167172163412083267949).
```

45930:
@asimaranov could you please recreate the graphQL JSON response that fails and the order that the patched o1js arrives at?

asimaranov:
> @asimaranov could you please recreate the graphQL JSON response that fails and the order that the patched o1js arrives at?

[archive_gql_resp.json](https://github.com/user-attachments/files/16457077/archive_gql_resp.json)

You mean this?

src/lib/mina/fetch.ts
45930: Oh, I did not forget.  `addCachedActions` is still present in this commit, you can see it on line 738.

45930:
@asimaranov No, I mean can you log out the response from `makeGraphqlRequest` in `fetchActions`, then log out the `actionsList` generated by `createActionsList` in this patched version of o1js.  If we're still ordering the actions incorrectly, it's helpful to know what the inputs and outputs are.

There are other reasons why we may fail to derive the hash, so this will help us pinpoint the issue.

src/lib/mina/fetch.ts
```diff
@@ -77,6 +78,26 @@ type NetworkConfig = {
   lightnetAccountManagerEndpoint: string;
 };
 
+type ActionsQueryInputs = {
+  publicKey: string;
+  actionStates: ActionStatesStringified;
+  tokenId?: string;
+}
```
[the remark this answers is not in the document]
45930: @mitschabaude, this type is only used in the file `fetch.ts`.  The graphQL query accepts the destructured inputs.

I made this type because previously we had the function param:

```
accountInfo: {
    publicKey: string;
    actionStates: ActionStatesStringified;
    tokenId?: string;
  },
```

When I made the new function `createActionList`, I duplicated this, so I made a custom type.  I didn't think `AccountInfo` was a good type name, but I see how the one I chose is also confusing.  WDYT?  Is there a better type name, or is it better to just leave the explicit definition in 2 function definitions?

src/lib/mina/graphql.ts
```diff
@@ -250,22 +250,24 @@ type EventQueryResponse = {
...
-    };
-    actionState: {
-      actionStateOne: string;
-      actionStateTwo: string;
-    };
-    actionData: {
-      accountUpdateId: string;
-      data: string[];
-    }[];
+export type FetchedAction = {
```
mitschabaude, line 253: nit: we usually do all exports at the top of a file, so we see what a file exports immediately

src/lib/mina/fetch.ts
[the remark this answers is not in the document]
mitschabaude: hm ok, you can leave it here as well :D

aii23:
Tried to recreate issue, that @asimaranov mentioned with current fix. Run on background script that dispatch several actions at once first, and then dispatch 40 more actions one by one. 
So after dispatching several actions fetch is working, but after 40 more actions it is starting to fail with the same error.
Script is located here https://github.com/aii23/actions_issue_poc/blob/fix_test/lightnet_reproduce_2.sh. 
I will investigate it further later, to find a which number of transactions it starts to fail(Because 40 is obviously overkill)

45930:
Thanks @aii23!  40 should still be a processable number so we want to get to the bottom of it.  When you do the test with 40, do all 40 actions show up in the graphql response?

aii23:
I will check this later today, or maybe tomorrow, to find exact condition for it to happen.

dfstio:
> Tried to recreate issue, that @asimaranov mentioned with current fix. Run on background script that dispatch several actions at once first, and then dispatch 40 more actions one by one. So after dispatching several actions fetch is working, but after 40 more actions it is starting to fail with the same error. Script is located here https://github.com/aii23/actions_issue_poc/blob/fix_test/lightnet_reproduce_2.sh. I will investigate it further later, to find a which number of transactions it starts to fail(Because 40 is obviously overkill)

I believe from my experience that this number of actions depends upon the time when you check it again, for 128 actions you will see this problem even after an hour, for few actions it should go away almost immediately, so it is better to check the actions immediately after the block is created.

dfstio:
I would also propose adding pagination to the action fetching logic, as it will be impossible to fetch a large number of actions from the archive node at once. 

You can see it, for example, trying to fetch events on mainnet for the contract B62qjwDWxjf4LtJ4YWJQDdTNPqZ69ZyeCzbpAFKN7EoZzYig5ZRz8JE:

```typescript
import { fetchEvents, UInt32 } from "o1js";

const contractAddress =
  "B62qjwDWxjf4LtJ4YWJQDdTNPqZ69ZyeCzbpAFKN7EoZzYig5ZRz8JE";

describe("Should get events", () => {
  it(`should get events`, async () => {
    console.log("Fetching events for", contractAddress);
    let blockHeight = 375561;
    const interval = 1000;
    while (blockHeight > 359609) {
      try {
        const events1 = await fetchEvents(
          { publicKey: contractAddress },
          "https://api.minascan.io/archive/mainnet/v1/graphql",
          {
            from: UInt32.from(blockHeight - interval),
            to: UInt32.from(blockHeight),
          }
        );
        console.log("Events1:", blockHeight, events1?.length);
      } catch (e) {
        console.error("Error fetching events for blocks", blockHeight, e);
      }
      blockHeight -= interval;
      await sleep(1000);
    }
  });
});

function sleep(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

```
Increasing the interval to 2000 gives the error:
```
[11:45:42 AM] Fetching events for B62qjwDWxjf4LtJ4YWJQDdTNPqZ69ZyeCzbpAFKN7EoZzYig5ZRz8JE
Error fetching events for blocks 375561 Error: Gateway Time-out
    at fetchEvents2 (/Users/mike/Documents/DeFi/MinaNFT/minanft-tools/node_modules/o1js/dist/node/index.cjs:22299:11)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Object.<anonymous> (/Users/mike/Documents/DeFi/MinaNFT/minanft-tools/tests/punk.test.ts:14:25)

```

aii23:
Still trying to find exact conditions on witch all of this happening. Is there a way to configure lightnode, so it produce block faster. Because now to ensure that all transaction goes into different blocks, I need to wait hell lot of time.

45930:
@aii23 

I'd appreciate if you would give the latest commit a spin: [d2c9320](https://github.com/o1-labs/o1js/pull/1784/commits/d2c9320eb44cd9157f8e2059c325292c22f4a84e)

Instead of always reversing the order, this commit attempts to find the right order by sorting on account update id.  It may solve the edge cases you were hitting.

run-ci-tests.sh
```diff
@@ -14,8 +14,7 @@ case $TEST_TYPE in
 
 "Reducer integration tests")
   echo "Running reducer integration tests"
-  ./run src/examples/zkapps/reducer/actions-as-merkle-list.ts --bundle
-  ./run src/examples/zkapps/reducer/actions-as-merkle-list-iterator.ts --bundle
+  ./run src/examples/zkapps/reducer/run.ts --bundle
```
45930, line 17: Change here is because previously these files would run a local test as they were imported.  I wrapped the tests in an export and call them explicitly in this run file.  This allows the run-live file to import the same smart contracts without too many excess diffs.

src/examples/zkapps/reducer/run-live.ts
```diff
@@ -0,0 +1,143 @@
...
+});
+pendingTx = await deployTx.sign([senderKey, keys.contract]).send();
+
+await pendingTx.wait();
+toc();
+console.log('working so far');
+// push some actions
+
+tic('dispatch transactions');
+let dispatchTx = await Mina.transaction(senderSpec, async () => {
```
45930, line 68: In this transactions, one sender creates many actions to reduce in the same block

src/examples/zkapps/reducer/run-live.ts
```diff
@@ -0,0 +1,143 @@
...
+await containsTx.prove();
+pendingTx = await containsTx.sign([senderKey]).send();
+await pendingTx.wait();
+toc();
+
+tic('dispatch transactions (multi-transaction)');
+
+tic('building');
+const txs = [];
+let dispatchTx1 = await Mina.transaction(senderSpec, async () => {
```
45930, line 100: In these transactions, different senders create actions to reduce within the same block

run-ci-live-tests.sh
```diff
@@ -21,6 +21,8 @@ DEX_PROC=$!
 FETCH_PROC=$!
 ./run src/tests/transaction-flow.ts --bundle | add_prefix "TRANSACTION_FLOW" &
 TRANSACTION_FLOW_PROC=$!
+./run src/examples/reducer/run-live.ts --bundle | add_prefix "REDUCER" &
+REDUCER_FLOW_PROC=$!
```
45930, line 25: New reducer test in CI handles some action serialization cases using lightnet graphql API.  The coverage on this is not ultra high, but it's a starting point, and covers the issue that zk noid originally isolated in their repo.

Trivo25 (commented):
make sure to update the bindings to the proper commit from main

src/examples/zkapps/reducer/actions-as-merkle-list-iterator.ts
```diff
@@ -18,7 +18,9 @@ import {
   SmartContract,
   method,
   assert,
-} from 'o1js';
+} from '../../../../dist/node/index.js';
```
Trivo25, line 21: `.. from 'o1js'`

src/examples/zkapps/reducer/run-live.ts
```diff
@@ -0,0 +1,156 @@
...
+const useCustomLocalNetwork = process.env.USE_CUSTOM_LOCAL_NETWORK === 'true';
+
+if (!useCustomLocalNetwork) {
+  throw 'Only Lightnet is currently supported';
+}
+
+tic('Run reducer examples against real network.');
+console.log();
+const network = Mina.Network({
+  mina: useCustomLocalNetwork
```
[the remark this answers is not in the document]
Trivo25: actually, we could just let it default to lightnet every time

src/examples/zkapps/reducer/run-live.ts
```diff
@@ -0,0 +1,156 @@
...
+  lightnetAccountManager: 'http://localhost:8181',
+});
+Mina.setActiveInstance(network);
+
+let { keys, addresses } = randomAccounts('contract', 'user1', 'user2');
+
+let pendingTx: Mina.PendingTransaction;
+
+// compile contracts & wait for fee payer to be funded
+const senderKey = useCustomLocalNetwork
```
Trivo25, line 35: same as above, I think this would simplify and improve readability by a ton (until we run it with a live public network)

src/lib/mina/fixtures/fetchActionsResponse.ts
```diff
@@ -0,0 +1,114 @@
+export const mockFetchActionsResponse =
```
Trivo25, line 1: nit: rename to `fetch-action-response.ts` and export as `export { ... }`

src/lib/mina/mina.ts
```diff
@@ -448,7 +448,7 @@ function Network(
         if (actions !== undefined) return actions;
       }
       throw Error(
-        `getActions: Could not find actions for the public key ${publicKey}`
+        `getActions: Could not find actions for the public key ${publicKey.toBase58()}`
```
Trivo25, line 451: nice!

src/examples/utils/randomAccounts.ts
```diff
@@ -0,0 +1,19 @@
+import { PrivateKey, PublicKey } from '../../../dist/node/index.js';
```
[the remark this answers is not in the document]
45930: I specifically changed this to test against the code in the current git branch of o1js.  When importing `from 'o1js'` I did not have the behavior present in my local files.

45930:
> make sure to update the bindings to the proper commit from main

@Trivo25 I don't know what this means.  Is there a script to run or something?

src/examples/utils/randomAccounts.ts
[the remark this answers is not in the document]
Trivo25: importing from o1js in the examples should proxy to `./dist/node/index.js` automatically, did you run `npm run build` before running your example `./run ./path/to/example.ts --bundle`?

src/examples/zkapps/reducer/run-live.ts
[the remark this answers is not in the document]
Trivo25: I think for tests that do have an implementation for `useCustomLocalNetwork = true` we should keep it that way, whereas tests that dont can just remove `useCustomLocalNetwork` entirely

Trivo25:
> > make sure to update the bindings to the proper commit from main
> @Trivo25 I don't know what this means. Is there a script to run or something?

`git submodule update --init --recursive` initialises all submodules (recursively), you can then go into the bindings submodule via `cd src/bindings` and check out specific commits like normal `git checkout ..` and then add that commit to the parent in o1js 


but it looks like in this particular instance we can just merge main into this branch to fix the discrepancy

src/examples/utils/randomAccounts.ts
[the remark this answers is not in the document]
45930: Maybe I was just using a stale build 🤷

45930:
> do you think we should test some other action-ranges?

In general, that seems reasonable.  I don't want to bloat the scope of this fix though.  This fix is currently blocking usage of actions for devs,so I want to get the corrected functionality out ASAP.

Trivo25 (approved):

Issue #1777, opened 2024-07-27
zkApp.fetchEvents() throws Gateway Time-out error on devnet


dfstio:
On mainnet, both zkApp.fetchEvents() and fetchEvents({ publicKey: contractAddress }) work.
On devnet, fetchEvents({ publicKey: contractAddress }) work, while zkApp.fetchEvents() throws Gateway Time-out error.

The code to reproduce:
```typescript
import { NameContractV2 } from "minanft";
import { PublicKey, Mina, fetchEvents } from "o1js";

const contractAddress =
  "B62qs2NthDuxAT94tTFg6MtuaP1gaBxTZyNv9D3uQiQciy1VsaimNFT";

const Devnet1 = {
  mina: ["https://api.minascan.io/node/devnet/v1/graphql"],
  archive: ["https://api.minascan.io/archive/devnet/v1/graphql"],
};

describe("Should get events", () => {
  it(`should get events`, async () => {
    const networkInstance = Mina.Network({
      mina: "https://api.minascan.io/node/devnet/v1/graphql",
      archive: "https://api.minascan.io/archive/devnet/v1/graphql",
    });
    Mina.setActiveInstance(networkInstance);

    const address = PublicKey.fromBase58(contractAddress);
    const zkApp = new NameContractV2(address);
    console.log("Fetching events for", address.toBase58());
    try {
      const events1 = await fetchEvents({ publicKey: contractAddress });
      console.log("Events1:", events1?.length);
      const events2 = await zkApp.fetchEvents();
      console.log("Events2:", events2?.length);
    } catch (e) {
      console.error("Error fetching events", e);
    }
  });
});
```
The log:
```
[2:38:49 PM] Fetching events for B62qs2NthDuxAT94tTFg6MtuaP1gaBxTZyNv9D3uQiQciy1VsaimNFT
[2:38:50 PM] Events1: 145
Error fetching events Error: Gateway Time-out
    at fetchEvents2 (/Users/mike/Documents/DeFi/MinaNFT/minanft-tools/node_modules/o1js/dist/node/index.cjs:22299:11)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at fetchEvents (/Users/mike/Documents/DeFi/MinaNFT/minanft-tools/node_modules/o1js/dist/node/index.cjs:19878:10)
[...]

PR #1794, opened 2024-08-20
Update default query for zkapp fetch actions to not add params

45930:
## Summary

DFST reported that fetching events for a given account was working via the `fetch` module, but not via the built-in `zkapp.fetch` method.

I discovered that the zkapp version always adds the `from: 0` parameter, which I believe is executing a poorly optimized query on the archive node.  I brought this issue up with the node operators team for further review.

This PR removes the `from: 0` parameter by default and only adds parameters to the query when explicitly provided by the user.

## Context

https://github.com/o1-labs/o1js/issues/1777


src/lib/mina/zkapp.ts
```diff
@@ -1058,21 +1059,18 @@ super.init();
...
     // filters all elements so that they are within the given range
     // only returns { type: "", event: [] } in a flat format
     let events = (
-      await Mina.fetchEvents(this.address, this.self.body.tokenId, {
-        from: start,
-        to: end,
-      })
+      await Mina.fetchEvents(this.address, this.self.body.tokenId, queryFilterOptions)
     )
-      .filter((eventData) => {
```
45930, line 1069: Removing this filter because it is a duplicate of the filter applied by the archive node.  Was it intentional to leave this as insurance?

src/lib/mina/zkapp.ts
```diff
@@ -1088,8 +1086,23 @@ super.init();
     let sortedEventTypes = Object.keys(this.events).sort();
 
     return events.map((eventData) => {
+      // If we don't know what types of events this zkapp may emit, then return the raw data
+      if (sortedEventTypes.length === 0) {
```
45930, line 1090: There may be something I don't understand here.  I ran an example script with an initialized smart contract and got the events payload from the graphQL query, then the script failed here because `this.events` was equal to `{}`, and there is no case to support that in this branching return.

How does a smart contract get its `this.events` hydrated?  Does that happen in a compile step or something?

Trivo25: It does it right here

https://github.com/o1-labs/o1js/blob/a8df62c57bb0d59b1a685ad61c59f4d3faae82fd/src/lib/mina/zkapp.ts#L1019

https://github.com/o1-labs/o1js/blob/a8df62c57bb0d59b1a685ad61c59f4d3faae82fd/src/lib/mina/zkapp.ts#L1121

Trivo25 (approved):

src/lib/mina/zkapp.ts
mitschabaude: yeah, this implementation was written when the archive node API was still in an early WIP phase and simply always returned the full list of events

src/lib/mina/zkapp.ts
mitschabaude: `this.events` is set in the user code

```ts
class Contract extends SmartContract {
   events = { ... }  // HERE (note: this is executed when a new instance is created)
}
```

I think you could also just throw an error in this function if the `events` property wasn't changed, I don't see what `zkapp.fetchEvents()` gives us over `Mina.fetchEvents()` in that case

Anyway, if you keep this case @45930, it would be nice to have it share most of the code with the `=== 1` case

mitschabaude: The point of having SmartContract-specific methods instead of generic Mina methods is usually to make use of the custom type layout that a SC defines, for example for 8-field state and here for events

src/lib/mina/zkapp.ts
mitschabaude: so, good to remove

src/lib/mina/zkapp.ts
45930: @mitschabaude so in my example, I am using the mina NFT contract and this logs out `{}`.  A new instance of a smart contract with an empty `events` property.

```
const zkApp = new NFTContractV2(address);
console.log(zkApp.events);
```


But the smart contract does emit events, e.g.

```
    @method async setOracle(oracle: PublicKey) {
      this.oracle.set(oracle);
      this.emitEvent("oracle", oracle);
    }
  
    @method async setPriceLimit(limit: UInt64) {
      this.priceLimit.set(limit);
      this.emitEvent("limit", limit);
    }
```

Though the developer did not define the events property, the contract may still emit events, so I would expect an instance of it to be able to fetch its own events.  If the user _must_ set this value, or we recommend to use `Mina.fetchEvents()` instead, then I think we still need a `length == 0` case to raise a helpful error explaining the problem.

The original code checks for the `length == 1` case, otherwise assumes that `length > 1`, so I got the error: `Error fetching events TypeError: Cannot read properties of undefined (reading 'fromFields')`, which makes no sense unless you understand that we're trying to access a lookup value that doesn't exist.

45930: I see 2 reasonable options that both involve adding a `length == 0` check:

1. Handle the case as if it is the anonymous fetchEvents function, and just return the raw events
2. Throw an error like `Error: This smart contract does not define any events.  Set the events property or use Mina.fetchEvents instead`

mitschabaude: both options are ok to me

mitschabaude: we should probably make `this.emitEvent()` throw an error if there are no events defined! the event ids won't be set properly

45930: > we should probably make this.emitEvent() throw an error if there are no events defined! 

Ah, we already do this, so that makes the decision easier.  I only got myself into this state by transcribing a smart contract incorrectly, and not including the events definition.  I will update this case to throw an error instead so that it matches, but still gives a better error message than the `Cannot read properties of undefined` one in case someone else makes the same mistake.

45930: New error message would be:

```
Error: fetchEvents: You are trying to fetch events without having declared the types of your events.
Make sure to add a property `events` on NFTContractV2, for example: 
class NFTContractV2 extends SmartContract {
  events = { 'my-event': Field }
}
Or, if you want to access the events from the zkapp account B62qs2NthDuxAT94tTFg6MtuaP1gaBxTZyNv9D3uQiQciy1VsaimNFT without casting their types
then try Mina.fetchEvents('B62qs2NthDuxAT94tTFg6MtuaP1gaBxTZyNv9D3uQiQciy1VsaimNFT') instead.
```

45930: @mitschabaude do the changes in https://github.com/o1-labs/o1js/pull/1794/commits/82dcbe177c3c53c9da7b996983242d5b83459f34 look good to you?

mitschabaude: beautiful!

mitschabaude (approved):

Issue #1717, opened 2024-07-02
Enable private outputs for zkProgram


garwalsh:
Currently zkProgram takes public and private input and returns public output. It does not currently support private output, a result of executing a proof that is not part of the proof statement itself.

This QoL update would enable private outputs such that zkProgram becomes more flexible and more accurately models how a developer would naturally expect such an API to function.

Example use case: a payment app that works via a claim box. The private output is the link to claim the payment that you only want the zkApp to share with the recipient

PR #1795, opened 2024-08-21
zkProgram auxiliary output

Trivo25:
This PR introduces a new return type to `zkProgram` - auxiliary output.
Auxiliary output behaves similarly to public output with the only difference that it is not part of the proof statement. This means, the returned data is _auxiliary_ data that the prover can use for other computation after a proof was successfully generated. 

`zkProgram` gets a new optional config property `auxiliaryOutput?: ProvableTypePure` which allows the developer to specify the auxiliary output type. This behaves identically to `publicOutput`. 

```ts
let MyProgram = ZkProgram({
  name: 'example-with-auxiliary-output',
  auxiliaryOutput: Field, // new optional property auxiliaryOutput
  methods: {
    baseCase: {
      privateInputs: [],
      async method() {
          return Field(9);
      },
    },
  },
});
``` 

Depending on whether `auxiliaryOutput` and/or `publicOutput` is specified, the return type of the provable method will automatically adjust.

If only `auxiliaryOutput` __or__ `publicOutput` is specified, the method implicitly expects the return value to be the defined output kind (either `auxiliaryOutput` __or__ `publicOutput`).

If __both__, `auxiliaryOutput` __and__ `publicOutput`, are specified in the `zkProgram`'s config, the method's return type expects an explicit return value that specifies both `auxiliaryOutput` __and__ `publicOutput`
```ts
return {
   publicOutput: Field(9),
   auxiliaryOutput: Field(5),
};
```

As a result of this, the return type of a provable method changes accordingly. 

If `auxiliaryOutput` is not defined, the return type remains `Proof<PublicInput, PublicOutput>`. However, if `auxiliaryOutput` is specified, the return type changes to an explicit version that returns the proof _and_ the new `auxiliaryOutput`.

```ts
let { proof, auxiliaryOutput } = await MyProgram.baseCase();
```

bindings https://github.com/o1-labs/o1js-bindings/pull/293

closes https://github.com/o1-labs/o1js/issues/1717


mitschabaude:
I have some high-level comments first:

The addition to the zkprogram interface is a bit different than I had expected: You add a single auxiliary output type for the entire program, while I would have expected one for _each method_ (possibly different)

After skimming the PR I understand that this is because auxiliary outputs are threaded through pickles compile, which accepts a single `auxiliary_typ` for the entire program.

However, I don't think that's necessary and to me it would be more natural to handle the aux output on the JS side. That would enable you to implement a different output for every method and would also remove the current severe limitation that the aux output only consists of field elements (`ProvablePure`).

I recommend to make this change before the auxiliary output API stabilizes, to give it maximal usefulness from the outset.

If you look at how pickles itself obtains the aux output, it's pretty simple: basically you just need a "hook" for the circuit to communicate with the code that runs the prover (which, indirectly, runs the circuit)

In ocaml, that "hook" is the `Request` mechanism: In the step circuit, after [running the user circuit](https://github.com/MinaProtocol/mina/blob/2dd698dc8fedf67bc60806598fe05d75413a6021/src/lib/pickles/step_main.ml#L287-L292) which returns the aux output, we issue a "Request" to provide that value to the outside:
https://github.com/MinaProtocol/mina/blob/2dd698dc8fedf67bc60806598fe05d75413a6021/src/lib/pickles/step_main.ml#L301-L305

Note two things:
* this is not really a "request", it's the opposite: providing a value to external code, from inside the circuit
* importantly, the `auxiliary_var` (consisting of variable fields) is transformed to `auxiliary_value` (a constant) before provided to the outside. This is a crucial step that can only be done _inside the circuit_. Looking up the value for a variable is done with `As_prover.read`, in o1js that would be `Provable.toConstant()` and you'd do it in an `asProver()` block
  * the JS function that's still "inside the circuit" and which has the circuit result available is `main()` in `picklesRuleFromFunction()`. that's the place where you would have this "to constant" logic if you'd extract the aux output in JS
  
To obtain the aux output during proving, in the step prover, Pickles first declares an empty reference `auxiliary_value` which is assigned when the circuit calls the "Request" hook:
 https://github.com/MinaProtocol/mina/blob/2dd698dc8fedf67bc60806598fe05d75413a6021/src/lib/pickles/step.ml#L753-L754
 
In JS the equivalent would be defining an unassigned variable and then passing a callback which the circuit code will run that assigns the variable, i.e. something like this:
 ```ts
let auxiliaryOutput;
// ...
await picklesProver(..., {
  hook: (auxValue) => {
    auxiliaryOutput = auxValue;
   }
 });
 // now we have `auxiliaryOutput` assigned, yay!
 ```

This sort of thing would naturally happen in the JS function that wraps the ocaml prover, i.e. `prove_()` in `toProver()` in `ZkProgram`.

To implement the "hook", you could either use a global mutable variable defined in JS, or hook into the ocaml request mechanism (more complicated, in the end also relies on a global mutable variable -- namely, snarky's internal state)

mitschabaude:
so in summary, this is how the prover could obtain the auxiliary output from within the circuit:

```ts
let auxiliaryOutput;

// set a hook for the circuit to give us a value (make sure this is called at the end of the circuit!)
programState.provideAuxOutput = (auxValue) => (auxiliaryOutput = auxValue);

// run prover
await picklesProver(...);

// reset hook
programState.provideAuxOutput = () => {}

// now we have `auxiliaryOutput` assigned, yay!
```

mitschabaude:
A second comment, I would prefer if both the method output and the prover output would be more regular, i.e. _always_ return objects of the form `{ publicOutput: ..., auxiliaryOutput?: ... }` and `{ proof: ..., auxiliaryOutput?: ... }`, where the aux output property is inferred to be undefined if there is no aux output.

I realize that this is backwards breaking so the suggestion would be to change this in a separate v2 PR, not here

The irregular return value optimizes for neatness but it adds confusion (what is returned when a single result is returned?) and it makes the API significantly harder to use in a generic way. I.e. a generic function that calls a prover and obtains `result.proof` is easy to write, but if we have to figure out at runtime if the result is the proof directly or has a property called proof, it becomes tiresome

mitschabaude:
Finally, a discussion was whether to call the aux output `privateOutput` or `auxiliaryOutput`

I don't want to impose one of those but it would be nice to see some reasoning for the naming decision. In my mind, these are arguments in favor of each:

Pro **privateOutput**:
1. "Private" is a property people care about, "auxiliary" is not. So, if we call it `auxiliaryOutput` we will inevitably get the question: "Is the `auxiliaryOutput` private?". That would be self-documenting if we call it `privateOutput`.
2. Also, we do call it `privateInput` on the other end and both refer to the same concept: a witness from the computation trace that is, by default, not revealed by the proof. Also, we have a duality of terms with "public input" + "public output". So for symmetry, it should be "private output".


Pro **auxiliaryOutput**:
The benefit I see here is that this name is more low-key and promises less, it doesn't suggest any powerful properties that people might, in confusion, expect from a "private output". In particular: The private output is not attested to by the proof -- so verifying the proof doesn't demonstrate that this particular private output was produced (indeed, you averted that confusion already quite well by not including the private output in the proof object, and therefore not having it as input to `verify()`)

Fizzixnerd:
I'm in agreement with @mitschabaude, specifically I support the statements summarized below.

1. The return type should be uniform and should not depend on the details of the configuration object passed to it.  This is to allow strongly typed software to _easily_ manipulate these object generically.
2. The name should be `privateOutput` to avoid confusion and favour uniformity/duality with the inputs.

Just my 2cents!

mitschabaude:
@Trivo25 I have another thought regarding naming.

We could stick with "auxiliary" for zkprogram to communicate that the _verifier can not check that this output is correct_. (Alternative that you also mentioned was unconstrainedOutput.)

But for the zkprocessor layer, we will have proper private outputs which are guaranteed by the proof. Using the same mechanism that exists in SmartContract (namely, we put a commitment to private output in the public output).
So, in zkprocessor, we don't need the relativizing "auxiliary" any more, and can use the more confident term"privateOutput". (Or actually, maybe it will be nice to have both!)

Trivo25:
Closing in favor of https://github.com/o1-labs/o1js/pull/1809

PR #1800, opened 2024-08-22
Add support for building with nix

Geometer1729:
Adds a devshell from which the bindings and package can be built.
Mostly taken from @georgeee 's work on o1js-bindings.
The o1js-bindings changes are in https://github.com/o1-labs/o1js-bindings/pull/295
Uses mina from this pr:
https://github.com/MinaProtocol/mina/pull/15848


Geometer1729:
Sorry about all the test reruns, I forgot to push the o1js-bindings changes first so there were a bunch of git errors

flake.nix
```diff
@@ -0,0 +1,116 @@
...
+      in {
+        formatter = pkgs.nixfmt;
+        inherit mina;
+        devShells = {
+          default = pkgs.mkShell {
+            inputsFrom = [ prj.pkgs.o1js_bindings prj.pkgs.__ocaml-js__ ];
+            shellHook =
+            ''
+            rustup update nightly-2023-09-01-x86_64-unknown-linux-gnu
+            rustup component add rust-src --toolchain nightly-2023-09-01-x86_64-unknown-linux-gnu
```
Fizzixnerd, line 85: Is it normal to perform an update as part of a `devShell`?  Normally this would go in a command somewhere, I would think.

Geometer1729: That makes sense, I'll see if I can find another way to get that to work.

Fizzixnerd: Sounds good to me.

Trivo25:
is this ready for review? I think it would be great to land this cc @Geometer1729

Geometer1729:
@Trivo25 Sure, I had left it as a draft because It still uses a the [georgeee/granular-nix](https://github.com/MinaProtocol/mina/tree/georgeee/granular-nix) branch, but I suppose that's no reason not to at least review it.

src/mina-signer/tests/client.test.ts
```diff
@@ -1,4 +1,4 @@
-import Client from '../dist/node/mina-signer/mina-signer.js';
+import Client from '../../../dist/node/mina-signer/mina-signer.js';
```
Trivo25, line 1: did the dist folder move?

Trivo25:
Nice! Let's keep an eye on Georges PR and merge this when its ready, this is going to be super helpful :)

src/mina-signer/tests/client.test.ts
mitschabaude, line 1: This changes the build we use for mina-signer tests from the _build of mina-signer only_ to the _build of o1js + mina-signer_.

We don't want this change: We want to test mina-signer in isolation, in the same version as it is published.

mitschabaude (changes requested):

src/mina-signer/tests/client.test.ts
Geometer1729: Oh, okay. That makes sense. I just tried to run the tests, got a bunch of file not found errors and changed the paths to the files that existed on my machine. I should've put more thought into what was going on with that. I see now it is even documented that you need to build mina-signer for the tests to work. 

How would you feel about adding this:
```bash
if ! [ -f /mina-signer/dist ]
then
  pushd src/mina-signer
    npm run build
  popd
fi
```
to the test script so it just builds mina-signer if you need to?

Geometer1729: No, I thought I had changed something with that but I just didn't realize you were intended to build mina-signer before running the tests. I should've read the docs.

mitschabaude: ok good idea to add that!

src/mina-signer/tests/keypair.test.ts
```diff
@@ -1,4 +1,4 @@
-import Client from '../dist/node/mina-signer/mina-signer.js';
+import Client from '..//dist/node/mina-signer/mina-signer.js';
```
mitschabaude, line 1: accidental diff here

mitschabaude (approved):
lgtm apart from https://github.com/o1-labs/o1js/pull/1800#discussion_r1750473205

Trivo25:
lets consider https://o1-labs.slack.com/archives/C01SJSXSM7H/p1726047094851739?thread_ts=1725974529.488629&cid=C01SJSXSM7H too; we can do that in a follow up PR

pin.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+# Set up flake registry to get o1js with all the submodules
+
+# Find the root of the o1js repo
+ROOT=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+pushd "$ROOT"
```
georgeee, line 6: what's the point of `pushd`/`popd` if we simply could do `cd "$ROOT"` without changing dir back (dir will be changed only within this script, not from above)

pin.sh
```diff
@@ -0,0 +1,13 @@
...
+# Update the submodules
+git submodule sync && git submodule update --init --recursive
+# Add the flake registry entry
+nix registry add o1js "git+file://$ROOT?submodules=1"
+# update mina input to local submodule
+nix flake update mina --override-input mina 'path:src/mina' --flake '.?submodules=1'
```
georgeee, line 12: Shouldn't it be the following? I'm not sure what the command does and why `--flake` argument is required

```suggestion
nix flake update o1js --override-input mina 'path:src/mina' --flake '.?submodules=1'
```

georgeee (commented):
BTW, #15848 is merged, so this PR can also be put forward

pin.sh
Geometer1729: I generally try to use pushd and popd in scripts cause they feel a bit easier to reason about, but I doesn't make a difference here and I don't feel strongly about it if you think cd is better I'm happy to change it.

PR #1804, opened 2024-08-23
Fix cache access structure and update submodules to the latest 

Trivo25:
A handful of changes to the cache system in the Mina repository changed and resulted in o1js accessing an out of bounds elements in the header structure of the cached data, this fixes it.
Additionally, this updates the bindings and mina submodule to the latest commit while making sure everything builds correctly (this wasn't the case previously)

Caching continues to work as intended as the change was only in the structure of the header data 

o1js-bindings https://github.com/o1-labs/o1js-bindings/pull/298

### DIFF
--- a/src/lib/proof-system/prover-keys.ts
+++ b/src/lib/proof-system/prover-keys.ts
@@ -53,7 +53,7 @@ function parseHeader(
   methods: MethodInterface[],
   header: SnarkKeyHeader
 ): CacheHeader {
-  let hash = Pickles.util.fromMlString(header[1][2][7]);
+  let hash = Pickles.util.fromMlString(header[1][2][6]);
   switch (header[0]) {
     case KeyType.StepProvingKey:
     case KeyType.StepVerificationKey: {
--- a/src/lib/proof-system/prover-keys.ts
+++ b/src/lib/proof-system/prover-keys.ts
@@ -211,7 +211,6 @@ type MlSnarkKeysHeader = [
   headerVersion: number,
   kind: [_: 0, type: MlString, identifier: MlString],
   constraintConstants: unknown,
-  commit: MlString,
   length: number,
   constraintSystemHash: MlString,
   identifyingHash: MlString


svv232 (commented):
What is the review methodology here out of curiosity? I am inclined to think this is fine as long as CI passes since we are just modifying the submodule binding. The access seems like a fine solution as well, because I don't know of way to length check/sanity check the header to prevent this issue from happening again.

Trivo25:
> What is the review methodology here out of curiosity? I am inclined to think this is fine as long as CI passes since we are just modifying the submodule binding. The access seems like a fine solution as well, because I don't know of way to length check/sanity check the header to prevent this issue from happening again.

Good question! Generally, CI is a good indicator when it comes to updating submodules and making sure the freshly built bindings don't break anything. Other than that, the cache access change was just a hot fix that I did and since it now "just works" it should be good to go (given it didn't do that before :D)

ymekuria (approved):

src/lib/proof-system/prover-keys.ts
```diff
@@ -53,7 +53,7 @@ function parseHeader(
   methods: MethodInterface[],
   header: SnarkKeyHeader
 ): CacheHeader {
-  let hash = Pickles.util.fromMlString(header[1][2][7]);
+  let hash = Pickles.util.fromMlString(header[1][2][6]);
```
mitschabaude, line 56: so, did the ocaml type change? if yes it would be nice to also change `MlSnarkKeysHeader`

mitschabaude (approved):

src/lib/proof-system/prover-keys.ts
Trivo25: yep https://github.com/o1-labs/o1js/pull/1804/commits/8ab009cfd35154cb10bfca2cb34cef6a3f76f05c
https://github.com/MinaProtocol/mina/commit/93d16e26a437a9366fa40d8d0f422fc00a8a6063#diff-e97cc66a36b931a3bf0e75bcead151b392188d176a3c71e7408756487b8ff5d1L133

PR #1809, opened 2024-08-28
Auxiliary output per method

Trivo25:


mitschabaude:
How about renaming the return object within a method from `{ publicOutput, auxiliaryOutput }` to `{ public, auxiliary }`?

src/lib/proof-system/zkprogram.ts
```diff
@@ -741,18 +790,41 @@ function ZkProgram<
...
+      let auxiliaryType = methodIntfs[i].auxiliaryType;
+      let auxiliaryOutputExists =
+        auxiliaryType && auxiliaryType.sizeInFields() !== 0;
+
+      let auxiliaryOutput;
+      if (auxiliaryOutputExists) {
+        let cachedAuxiliaryValue = programState.getAuxiliaryOutput(
+          methodIntfs[i].methodName
+        );
+        auxiliaryOutput = auxiliaryType?.fromValue(cachedAuxiliaryValue);
```
mitschabaude, line 803: why do we do `fromValue()` here? that seems to be a no-op to me because what is returned will be the provable type. `toValue()` would make sense to me though!

mitschabaude: I changed this

mitschabaude (approved):

CHANGELOG-v2.md
```diff
@@ -24,11 +24,14 @@ Showing all changes since the last release (v.1.5.0)
...
 ### Breaking Changes
 
-- The `divMod32()` gadget was modified to accept `nBits` instead of `quotientBits`, and assert it is in the range [0, 2**255) to address an issue previoulsy where the bound on `quotientBits` was too low https://github.com/o1-labs/o1js/pull/1763.
+- The `divMod32()` gadget was modified to accept `nBits` instead of `quotientBits`, and assert it is in the range [0, 2\*\*255) to address an issue previoulsy where the bound on `quotientBits` was too low https://github.com/o1-labs/o1js/pull/1763.
 
 - `Provable.equal()` now turns both types into canonical form before comparing them https://github.com/o1-labs/o1js/pull/1759
+
   - Removed implicit version `Provable.equal(x, y)` where you didn't have to pass in the type
 
+- `zkProgram` to support `auxiliaryOutput`
```
Trivo25, line 33: Just a reminder; Will go through the changelog before V2 release anyway

mitschabaude:
I updated the branch protection rules to exclude a CI check that no longer exists - ready to merge!

PR #1815, opened 2024-09-09
Enhance Ethereum Signature Verification and Add Curve Parsing Methods

### LINKED ISSUE #1693: Improve DevX for ECDSA Gadget
## Summary

O1JS has implemented the primitives needed to verify ECSDA over Secp256k1 signatures (used in ethereum and other blockchains), but the developer experience leaves room for improvement.

## Current State

The only examples of the ECDSA gadget in o1js use o1js to produce the signature and to verify the signature.  ([example](https://github.com/o1-labs/o1js/blob/main/src/examples/crypto/ecdsa/run.ts))

It's not clear how to bring a signature that was produced elsewhere into o1js to verify in a zkapp.

## Desired State

It should be intuitive to verify signatures produced by other libraries, and we should have examples demonstrating how to do it.

Shigoto-dev19:
### Description:

- This PR introduces several enhancements to improve the handling of Ethereum signature verification and curve parsing. 
- Fixes #1693 

### Changes:

- **Added `verifyEthers` method**: A new method in `foreign-ecdsa.ts` to verify Ethereum signatures using the EIP-191 message hashing standard.
- **Added `fromEthers` method**: A new method in `foreign-curve.ts` to parse and convert Ethereum public keys into ForeignCurve points, supporting compressed and uncompressed formats.
- **Added `fromHex` method**: A new utility method to coerce hexadecimal strings into ForeignCurve points.
- **Upgraded `Ecdsa` class in example**: Updated the `Ecdsa` class to use `createEcdsaV2` for improved compatibility and to replace the deprecated version in the example.

- **Added example**: An example was included to demonstrate how to verify Ethereum signatures using ethers.js with the updated methods.
- Update `vk-regression.json`

### Preview(ecdsa-ethers)
```sh
ethers verify (build constraint system): 3.948s
{
  'Total rows': 46230,
  RangeCheck0: 7602,
  RangeCheck1: 3441,
  Zero: 7851,
  Generic: 14082,
  ForeignFieldMul: 993,
  ForeignFieldAdd: 1845,
  EndoMulScalar: 128,
  Xor16: 9568,
  Rot64: 720
}
ecdsa / ethers verify (compile): 11.174s
ecdsa / ethers verify (prove): 52.133s
```


src/lib/provable/crypto/foreign-curve.ts
```diff
@@ -74,6 +75,116 @@ class ForeignCurve {
...
+  }
+
+  /**
+   * Create a new {@link ForeignCurveV2} instance from an Ethereum public key in hex format, which may be either compressed or uncompressed.
+   * This method is designed to handle the parsing of public keys as used by the ethers.js library.
+   *
+   * The input should represent the affine x and y coordinates of the point, in hexadecimal format.
+   * Compressed keys are 33 bytes long and begin with 0x02 or 0x03, while uncompressed keys are 65 bytes long and begin with 0x04.
+   *
+   * **Note**: Ensure the inputs are range-checked if they originate from a field with a different modulus, or if they are not constants.
```
mitschabaude, line 129: this note doesn't make sense, because the input is a string, so it is always a "constant"

src/examples/crypto/ecdsa/run.ts
```diff
@@ -34,3 +41,42 @@ console.timeEnd('keccak + ecdsa verify (prove)');
...
+
+// compressed public key generated by ethers.js
+const compressedPublicKey =
+  '0x020957928494c38660d254dc03ba78f091a4aea0270afb447f193c4daf6648f02b';
+
+// ECDSA signature generated by ethers.js
+const rawSignature =
+  '0x6fada464c3bc2ae127f8c907c0c4bccbd05ba83a584156edb808b7400346b4c9558598d9c7869f5fd75d81128711f6621e4cb5ba2f52a2a51c46c859f49a833a1b';
+
+const publicKeyE = Secp256k1.fromEthers(compressedPublicKey.slice(2));
```
mitschabaude, line 62: you do `.slice(2)` on the input from ethers.js here

couldn't it be a source of confusion that the input can't be passed in directly? maybe `fromEthers` should look at the first two chars and handle the case when they are "0x"

Shigoto-dev19: yes, I'll add a check to handle both cases: with and without "0x"

Shigoto-dev19: I remember now that I didn’t fix it as you suggested initially because I wanted to keep it consistent with the `Bytes.fromHex` API. I've now addressed it with the following:


```ts
static fromHex(hex: string) {
    // trim the '0x' prefix if present
    if (hex.startsWith('0x')) {
      hex = hex.slice(2);
     }
     // ...
}
```

@mitschabaude Do you think it’s worth tweaking `Bytes.fromHex` in a similar way as well?

src/lib/provable/crypto/foreign-curve.ts
Shigoto-dev19: Removed note :ballot_box_with_check:

src/examples/crypto/ecdsa/run.ts
mitschabaude: hm your call. I think it's fine for `fromEthers()` which takes a string formatted in a particular library's style, but I don't think we need it in our general `Bytes.fromHex()` method 🤔

mitschabaude (approved):

mitschabaude (approved):
reapproving for the CI changes

Issue #1798, opened 2024-08-22
Make useFullCommitment true by default or remove useFullCommitment from o1js and always use full tx while signing the AccountUpdate?


dfstio:
Discussion in discord:
https://discord.com/channels/484437221055922177/1275913380284203009

PR #1818, opened 2024-09-12
Always use full commitment

Trivo25:
Why? This prevents transactions from being replayed. Deletable accounts will also benefit from this change in the future

closes https://github.com/o1-labs/o1js/issues/1798


src/lib/mina/account-update.ts
```diff
@@ -897,8 +901,6 @@ class AccountUpdate implements Types.AccountUpdate {
    */
   requireSignature() {
     let { nonce, isSameAsFeePayer } = AccountUpdate.getSigningInfo(this);
-    // if this account is the same as the fee payer, we use the "full commitment" for replay protection
-    this.body.useFullCommitment = isSameAsFeePayer;
```
mitschabaude, line 901: previously there were two different ways we ensured that it's not replayed:
* if the account was the same as the fee payer, we used the full commitment
* otherwise, we incremented the nonce and set a nonce precondition

I like the simplification of just always using the full commitment (my initial idea on why not to do that was flexibility of signing with one key without yet having the fee payer sign; but now I think the simplicity is more important)

What this also means is that we _don't_ need to do the nonce increase in this function depending on `isSameAsFeePayer`.

Instead, I would do the following:
* add an `boolean` optional parameter to this function `noFullCommitment`, if it's true do the nonce increment instead
* remove the `isSameAsFeePayer` return value from `getSigningInfo`

src/lib/mina/account-update.ts
```diff
@@ -614,6 +614,10 @@ const Body = {
   ): Body {
     let { body } = Types.AccountUpdate.empty();
     body.publicKey = publicKey;
+
+    // useFullCommitment = true by default, prevents replay attacks and similar issues
+    // the developer can still modify this to false if needed later on
+    body.useFullCommitment = Bool(true);
```
mitschabaude, line 620: I'm not sure about this, how about setting this to true in `requireSignature()` instead of always? it's a bit weird to me that the default account update differs from the natural dummy AU with no apparent reason

mitschabaude: or in `setLazySignature`?

mitschabaude: well or maybe we should remove `setLazySignature` since it no longer holds a private key so no real purpose

PR #1826, opened 2024-09-18
run regression tests in parallel

Trivo25:


mitschabaude:
the problem here might be that this chek is required by CI?
image

mitschabaude:
so I'm gonna force merge and then change the requirement

PR #1827, opened 2024-09-18
avoid overwriting `proofsEnabled` through default parameter

Trivo25:
```ts
MyProgram.setProofsEnabled(false);
await MyProgram.compile();
```

Wouldn't work because `compile()` sets the internal flag to `true` no matter what since that's the default parameter. The fix is to specify no default parameter (`undefined` by default) and only setting the internal flag if the developer actually passes in a value. This is backwards compatible because 1) the API doesn't change and 2) the internal flag is still initiated with `true` so its behavior stays the same 

thanks for pointing this out @rpanic


mitschabaude (approved):

mitschabaude:
force merging as well because this now requires regression checks 1 and 2

Issue #1675, opened 2024-06-05
Allow non-pure provable types as inputs/outputs in zkprogram


mitschabaude:
This would be a great DX improvement because it would allow directly updating rich provable types like `MerkleList` or `IndexedMerkleMap` by a zkprogram

mitschabaude:
done

PR #1828, opened 2024-09-18
Allow non-pure provable types as inputs/outputs in zkprogram

ymekuria:
## Summary

This PR updates `ZkProgram` to allow non-pure provable types as inputs and outputs.

Addresses #1675 

## Impact

Currently only pure provable types can be used as inputs and outputs to a `ZkProgram`. This PR  improves the DX with changes that allow a developer to update rich provable types like a `IndexedMerkleMap` or a `MerkleList`  in a `ZkProgram`.


src/lib/proof-system/zkprogram.ts
```diff
@@ -1487,20 +1533,37 @@ function getMaxProofsVerified(methodIntfs: MethodInterface[]) {
...
 }
 
+function toFieldAndAuxConsts<T>(type: Provable<T>, value: T) {
+  return {
+    publicInputFields: MlFieldConstArray.to(type.toFields(value)),
+    publicInputAux: type.toAuxiliary(value),
+  };
+}
+
 ZkProgram.Proof = function <
```
Trivo25, line 1567: This needs to be adjusted too! We need to "derive" the proof class from our zkProgram correctly via `const MyProof = ZkProgram.Proof(MyProgram);` which should not work on the type level right now

src/lib/proof-system/zkprogram.ts
```diff
@@ -161,12 +161,22 @@ const FeatureFlags = {
 };
 
 function createProgramState() {
+  let auxInputCache: Map<string, any[]> = new Map();
```
Trivo25, line 164: I think we should rename this to avoid confusion. The methods below already handle axuililary output, if we name this auxiliary too it might cause some confusion

src/lib/proof-system/zkprogram.ts
```diff
@@ -161,12 +161,22 @@ const FeatureFlags = {
...
   let methodCache: Map<string, unknown> = new Map();
-
   return {
+    setAuxilaryInput(value: any[]) {
```
Trivo25, line 167: see above

src/lib/proof-system/zkprogram.ts
```diff
@@ -161,12 +161,22 @@ const FeatureFlags = {
...
+      auxInputCache.set('auxinput', value);
+    },
+    getAuxilaryInput: () => {
```
Trivo25, line 170: see above

src/lib/proof-system/zkprogram.ts
```diff
@@ -161,12 +161,22 @@ const FeatureFlags = {
...
+      let entry = auxInputCache.get('auxinput');
+      if (entry === undefined) throw Error(`Auxiliary input not defined`);
+      return entry;
+    },
+
     setAuxiliaryOutput(value: unknown, methodName: string) {
```
Trivo25, line 176: see here! already done

src/lib/proof-system/zkprogram.ts
```diff
@@ -804,8 +834,24 @@ function ZkProgram<
...
+        publicOutput = fromFieldAndAuxConsts(
+          publicOutputType,
+          publicOutputFields,
+          publicInputAux
+        );
+
+        programState.resetAuxCache();
+      } else {
+        // create public output when public input is pure
+        publicOutput = fromFieldAndAuxConsts(
```
Trivo25, line 850: this is the pure case, we should be able to use the pure version of the function

src/lib/proof-system/zkprogram.ts
```diff
@@ -1487,20 +1533,37 @@ function getMaxProofsVerified(methodIntfs: MethodInterface[]) {
...
+function fromFieldVars<T>(type: Provable<T>, fields: MlFieldArray) {
+  return type.fromFields(MlFieldArray.from(fields), []);
 }
+
 function toFieldVars<T>(type: ProvablePure<T>, value: T) {
   return MlFieldArray.to(type.toFields(value));
 }
 
-function fromFieldConsts<T>(type: ProvablePure<T>, fields: MlFieldConstArray) {
+function fromFieldConsts<T>(type: any, fields: MlFieldConstArray) {
```
Trivo25, line 1544: Why replace ` ProvablePure<T>` with `any` here?

src/lib/proof-system/zkprogram.ts
```diff
@@ -1487,20 +1533,37 @@ function getMaxProofsVerified(methodIntfs: MethodInterface[]) {
...
   return type.fromFields(MlFieldConstArray.from(fields));
 }
-function toFieldConsts<T>(type: ProvablePure<T>, value: T) {
+
+function fromFieldAndAuxConsts<T>(
+  type: Provable<T>,
+  fields: MlFieldConstArray,
+  aux: any[] = []
```
Trivo25, line 1551: nit: we could do `unknown` too, maybe a bit more fitting

src/examples/zkprogram/program-with-non-pure-input.ts
```diff
@@ -0,0 +1,34 @@
...
+let MyProgram = ZkProgram({
+  name: 'example-with-non-pure-inputs',
+  publicInput: MyStruct,
+  publicOutput: MyStruct,
+
+  methods: {
+    baseCase: {
+      privateInputs: [],
+      async method(input: MyStruct) {
+        return {
```
Trivo25, line 17: This works! But right now we don't handle the case where the aux values are changed _in_ the circuit, like this

```ts
input.label = "in circuit";
return {
   publicOutput: input
}

// after the circuit
assert(input.label == "in circuit")
```


so either we make this work or we explcitily tell developers not to touch the auxiliary values in the circuit; I prefer the first approach

src/lib/proof-system/zkprogram.ts
ymekuria: The reason I made the aux type `any[]` was to match the return value from `toAuxiliary`.
```ts
(property) toAuxiliary: (value?: T | undefined) => any[]
A function that takes value (optional), an element of type T, as argument and returns an array of any type that make up the "auxiliary" (non-provable) data of value.

@param value
the element of type T to generate the auxiliary data array from, optional. If not provided, a default value for auxiliary data is returned.

@return — An array of any type describing how this T element is made up of "auxiliary" (non-provable) data.
```

Trivo25: That's right! Should be okay then

src/examples/zkprogram/program-with-non-pure-input.ts
mitschabaude: We need them in the circuit!

src/examples/zkprogram/program-with-non-pure-input.ts
```diff
@@ -0,0 +1,42 @@
...
+  },
+});
+
+//
+
+console.log('compiling MyProgram...');
+await MyProgram.compile();
+console.log('compile done');
+
+let input = new MyStruct({ label: 'input', value: Field(5) });
```
Trivo25, line 34: this is missing the `description` property

src/examples/zkprogram/program-with-non-pure-input.ts
```diff
@@ -0,0 +1,42 @@
...
+
+let { proof } = await MyProgram.baseCase(input);
+let ok = await MyProgram.verify(proof);
+
+console.log('proof', proof);
+
+assert(ok, 'proof not valid!');
+// assert(proof.publicOutput.label === 'inCircuit');
```
Trivo25, line 42: once the PR is ready we should re-enable this check

src/lib/proof-system/zkprogram.ts
```diff
@@ -161,12 +161,30 @@ const FeatureFlags = {
 };
 
 function createProgramState() {
+  let nonPureDataCache: Map<string, any[]> = new Map();
```
Trivo25, line 164: we could just reuse the method cache for this too, would make it a bit cleaner but we would have to differenciate between nonpure auxiliary data and the auxiliary output its currently keeping track. both are valid imo

src/lib/proof-system/zkprogram.ts
```diff
@@ -1266,10 +1327,22 @@ function picklesRuleFromFunction(
...
       result = (await func(...finalArgs)) as any;
     } else {
-      let input = fromFieldVars(publicInputType, publicInput);
+      console.log('auxData before input', auxInputData);
+      let input = fromFieldVars(publicInputType, publicInput, auxInputData);
       result = (await func(input, ...finalArgs)) as any;
     }
 
+    console.log('result input', result);
+    if (result.publicOutput) {
```
Trivo25, line 1336: this should probably be `result?.publicOutput` since `result` is unfornately typed

Trivo25 (approved):
lmk when the PR is ready for a more thorough review!

Trivo25 (changes requested):
(approved by accident, sorry)

Trivo25:
@ymekuria please rebase too

src/examples/zkprogram/program-with-non-pure-input.ts
```diff
@@ -0,0 +1,39 @@
...
+        input.label = 'in-circuit';
+        return {
+          publicOutput: input,
+        };
+      },
+    },
+  },
+});
+
+//
```
Trivo25, line 27: we can probably remove this line

Trivo25 (commented):
looks good so far! There's one case missing tho, and that's the case where we return a constant non-pure data structure from the method, so something like this

```ts

// ..
  baseCase: {
      privateInputs: [],
      async method() {
        return {
          publicOutput: new MyStruct({ label: 'in-circuit', value: Field(5) }),
        };
      },
    },
```

this yields

`publicOutput.label = undefined` and not `publicOutput.label = "in-circuit"` as expected

src/lib/proof-system/zkprogram.ts
```diff
@@ -391,10 +398,30 @@ function ZkProgram<
...
       }
-      let publicInputFields = toFieldConsts(publicInputType, publicInput);
+
+      let nonPureInputExists = publicInputType
+        .toAuxiliary(publicInput)
+        .some((aux) => aux.length !== 0);
+
+      let publicInputFields, publicInputAux;
+      if (nonPureInputExists) {
+        // serialize publicInput into pure provable field elements and auxilary data
```
Trivo25, line 408: typo `auxiliary` (i fckd this up so many times myself :D )

src/lib/proof-system/zkprogram.ts
```diff
@@ -391,10 +398,30 @@ function ZkProgram<
...
         );
       }
-      let publicInputFields = toFieldConsts(publicInputType, publicInput);
+
+      let nonPureInputExists = publicInputType
+        .toAuxiliary(publicInput)
+        .some((aux) => aux.length !== 0);
+
+      let publicInputFields, publicInputAux;
+      if (nonPureInputExists) {
```
Trivo25, line 407: what about unifying both branches here and only using one function `toFieldAndAuxConsts` instead of two?

src/lib/proof-system/zkprogram.ts
```diff
@@ -415,8 +442,21 @@ function ZkProgram<
         programState.reset(methodIntfs[i].methodName);
       }
 
+      let publicOutput;
       let [publicOutputFields, proof] = MlPair.from(result);
-      let publicOutput = fromFieldConsts(publicOutputType, publicOutputFields);
+      if (nonPureInputExists) {
```
Trivo25, line 447: this should be the case when non-pure output exists, and not input, no?

Trivo25: this should also fix the case where we have a constant non-pure output

src/lib/proof-system/zkprogram.ts
```diff
@@ -415,8 +442,21 @@ function ZkProgram<
...
+        let nonPureOutput = programState.getNonPureOutput();
+
+        publicOutput = fromFieldConsts(
+          publicOutputType,
+          publicOutputFields,
+          nonPureOutput
+        );
+
+        programState.reset('__nonPureOutput__');
+      } else {
```
Trivo25, line 457: nit: we could also simplify this logic here if we simply return `[]` or `undefined` from `getNonPureOutput`

ymekuria:
> looks good so far! There's one case missing tho, and that's the case where we return a constant non-pure data structure from the method, so something like this
> 
> ```ts
> // ..
>   baseCase: {
>       privateInputs: [],
>       async method() {
>         return {
>           publicOutput: new MyStruct({ label: 'in-circuit', value: Field(5) }),
>         };
>       },
>     },
> ```
> 
> this yields
> 
> `publicOutput.label = undefined` and not `publicOutput.label = "in-circuit"` as expected

This could appears to work for me and the original assertion this assertion holds true

assert(proof.publicOutput.label === 'in-circuit');

Here is a log of the proof from this code. Is there something I am missing?

```ts
proof ProgramProof {
  shouldVerify: Bool { value: [ 0, [Array] ] },
  publicInput: MyStruct { label: 'input', value: Field { value: [Array] } },
  publicOutput: MyStruct { label: 'in-circuit', value: Field { value: [Array] } },
  proof: [ 0, [ 0, [Array], [Array], [Array] ] ],
  maxProofsVerified: 0
}
```

Trivo25:
> > looks good so far! There's one case missing tho, and that's the case where we return a constant non-pure data structure from the method, so something like this
> > ```ts
> > // ..
> >   baseCase: {
> >       privateInputs: [],
> >       async method() {
> >         return {
> >           publicOutput: new MyStruct({ label: 'in-circuit', value: Field(5) }),
> >         };
> >       },
> >     },
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this yields
> > `publicOutput.label = undefined` and not `publicOutput.label = "in-circuit"` as expected
> 
> This code appears to work for me and this assertion holds true
> 
> ```ts
> assert(proof.publicOutput.label === 'in-circuit');
> ```
> 
> Here is a log of the proof from this code. Is there something I am missing?
> 
> ```ts
> proof ProgramProof {
>   shouldVerify: Bool { value: [ 0, [Array] ] },
>   publicInput: MyStruct { label: 'input', value: Field { value: [Array] } },
>   publicOutput: MyStruct { label: 'in-circuit', value: Field { value: [Array] } },
>   proof: [ 0, [ 0, [Array], [Array], [Array] ] ],
>   maxProofsVerified: 0
> }
> ```

Can you check this snippet for me?

<details>
<summary>Example</summary>

```ts
import { Field, Struct, ZkProgram } from 'o1js';

class MyStruct extends Struct({
  label: String,
  value: Field,
}) {}

let MyProgram = ZkProgram({
  name: 'example-with-non-pure-inputs',
  publicOutput: MyStruct,

  methods: {
    baseCase: {
      privateInputs: [],
      async method() {
        return {
          publicOutput: new MyStruct({ label: 'input', value: Field(5) }),
        };
      },
    },
  },
});

console.log('compiling MyProgram...');
await MyProgram.compile();
console.log('compile done');

let { proof } = await MyProgram.baseCase();
console.log(proof.publicOutput);

```

</details>

CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 
+-`ZkProgram` to support non-pure provable types as inputs and outputs https://github.com/o1-labs/o1js/pull/1828
```
querolita, line 28: I thought this should have to go inside the `Unreleased` section?

ymekuria: Good catch. Moved it.

Trivo25 (approved):
Looks good to me! FYI I have reverted the mina and bindings commit

src/examples/zkprogram/program-with-non-pure-input.ts
```diff
@@ -0,0 +1,83 @@
...
+      async method() {
+        return {
+          publicOutput: new MyStruct({ label: 'output', value: Field(5) }),
+        };
+      },
+    },
+  },
+});
+
+let NonPureInputProgram = ZkProgram({
```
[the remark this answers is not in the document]
Trivo25: ignore; leftover from a previous commit

mitschabaude (approved):
implementation-wise, looks good to me!

we just have a problem with converting proofs to/from JSON now: proofs are serialized by doing `toFields()` on `publicInput` and `publicOutput` and converting these fields to strings. this will work in `toJSON()`, but will fail in `fromJSON()` where the aux part in `fromFields()` will be missing.

to solve that we'd need a general way to serialize any provable type. I have some ideas on that that I might post in a separate issue

mitschabaude:
huge simplification for so many zkprograms! 🥳

Issue #1832, opened 2024-09-24
Decouple the compile-time and runtime artifacts of offchain state


45930:
## Summary

The way that offchain state works today is a developer will create their custom offchain state class and object in on call:

```ts
const offchainState = OffchainState(
  {
    someMap: OffchainState.Map(PublicKey, UInt64),
    someValue: OffchainState.Field(UInt64),
  },
  { logTotalCapacity: 10, maxActionsPerProof: 5 }
);
```

This object is then passed into the definition of the smart contract, e.g. in the `settle` method:

```ts
@method
async settle(proof: StateProof) {
  await offchainState.settle(proof);
}
```

Because the specific object called `offchainState` is referenced in the smart contract, the object and the class become linked.  _Every instance of the smart contract class_ will reference the same object `offchainState`.

## Where has this caused issues?

There is a community builder working on a project that has run into this issue: https://discord.com/channels/484437221055922177/1287933364007079967

They have duplicates of the same smart contract with which they want to use offchain state.

I also ran into this problem when trying to write isolated unit tests for an app that uses offchain state.

45930:
@mitschabaude 

`@state(MyOffchainState.Commitments) offchainState = MyOffchainState.initialCommitments();`

It is only the commitment that are referenced in the smart contract.  Then the `settle` method accepts a rollup proof that _must_ be the right kind of rollup proof.  I'm not sure if my proposal is enough to ensure that.

mitschabaude:
Somehow, for doing offchain state operations, the smart contract needs access to the Merkle map that the offchain state is kept in.

Currently, there is just a single Merkle map per offchain state "class", but I think we want to avoid that - because the Merkle map is created from actions fetched from a particular account, so there should never be two different accounts used with the same Merkle map.

A contract instance is basically a representation of zkapp account, so it makes sense to require a 1:1 correspondence between contract instances and offchain Merkle maps.
So I think, when instantiating a contract, we should store an "offchain state instance" on it, similar to how it's done with `reducer`

mitschabaude:
How about this @45930?

```ts
// not a class, because users don't need any features of classes like adding methods
const offchainState = OffchainState(
  {
    someMap: OffchainState.Map(PublicKey, UInt64),
    someValue: OffchainState.Field(UInt64),
  },
  { logTotalCapacity: 10, maxActionsPerProof: 5 }
);

class StateProof extends offchainState.Proof {}

class MySmartContract extends SmartContract {
[...]

PR #1834, opened 2024-09-24
Refactor offchain state

45930:
## Summary

The experimental offchain state API currently uses a singleton class to manage storage on behalf of a smart contract.  This blocks the use case where many instances of the same smart contract need to be instantiated at the same time, because they will all use the same offchain state under the hood, leading to data mismatches.

This PR splits the offchain state API into `OffchainState` and `OffchainStateInstance` such that the `OffchainState` can be compiled into a circuit definition, and the `OffchainStateInstance` stores the internal data like which contract instance it is associated with and the merkle trees of data.

## Note

I called it out in line, but please note that there is still an ongoing DevX problem where the smart contract _class definition_ has to reference a specific _offchain state instance_ in order to compile.  That leaves us in the same problem, where unless handled by the developer, multiple contracts will point to the same storage.  There is now a viable workaround, but it doesn't work as nicely as we'd like.  I'd appreciate any input to overcome this issue.

Closes: https://github.com/o1-labs/o1js/issues/1832


src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -20,80 +22,99 @@ const offchainState = OffchainState(
...
-    // setting `from` to `undefined` means that the account must not exist yet
-    offchainState.fields.accounts.update(address, {
-      from: undefined,
-      to: amountToMint,
-    });
+    // Since this line is executed over and over, every time a proof is generated, we can't rely on setting this value elsewhere
+    // We also can't rely on different instances of the contract having different offchain state instances, because this specific instance of
+    // offchain state is referenced by the class definition of the smart contract.
+    // By using a closure and exporting the offchain state and the contract together, we can always refer to the correct instance of the offchain state
+    offchainState = offchainStateInstance;
```
45930, line 37: This line is the source of a lot of frustration.  Here are some alternatives attempted:

- `offchainState = offchainState.init()`
    - Re-initialized the state every time the prover is invoked.  The smart contract is not capable of staying up to date with the contract state.
- `offchainState: typeof offchainStateInstance`
    - The contract can't compile because provable methods rely on offchainState being defined at compile time


That leaves the implementation as implemented here.  The instance `offchainStateInstance` is tied to the class definition of the smart contract.  That means 2 or more smart contract instances will all have the same offchain state instance.  Even if we reset the instance state after initialization, the prover will re-run this line and set the offchain state back to the orifinally-defined value.

I found the best/only way to manage this is to turn the whole smart contract class into a closure so that the definition of each smart contract can be paired with its offchain state instance.  This doesn't feel amazing, but it does unblock the use case.

src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -140,16 +159,161 @@ await testLocal(
...
+  offchainStateInstance: offchainStateInstanceA,
+  ExampleContract: ExampleContractA,
+} = ExampleContractWithState();
+
+const {
+  offchainStateInstance: offchainStateInstanceB,
+  ExampleContract: ExampleContractB,
+} = ExampleContractWithState();
+
+const Local = await Mina.LocalBlockchain({ proofsEnabled: true });
```
45930, line 180: `testLocal` accepts a single contract instance only, so I re-implemented the behavior here for the specific case I needed.

src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -140,16 +159,161 @@ await testLocal(
...
+const contractA = new ExampleContractA(contractAccountA);
+const contractB = new ExampleContractB(contractAccountB);
+
+console.time('compile offchain state program');
+await offchainState.compile();
+console.timeEnd('compile offchain state program');
+
+console.time('compile contract');
+await ExampleContractA.compile();
+await ExampleContractB.compile();
```
45930, line 195: Contracts have the same `_verificationKey.hash`, but `_provers` need to be recompiled, or else all the contracts will point to the same instance of offchain state again.

kadirchan:
Love it

Trivo25:
Given that these modifications affect both the public API and developer usage patterns, we should evaluate options for maintaining backwards compatibility. If backwards compatibility isn't feasible, we may have to consider targeting V2 for these changes

45930:
@Trivo25 this is in the experimental namespace, so do we need to be worried about compatibility?

I do think an accompanying PR in the docs is required to match this change before release.

Trivo25:
> @Trivo25 this is in the experimental namespace, so do we need to be worried about compatibility?

That's a good point, then probably not

mitschabaude:
> Attempted to store previously initialized instances in a map in global state to get around the re-init problem, but ran into issues trying to read variables in provable code. E.g. if we use the contract address as a key, the prover blows up.

@45930 sounds like that could be solved with `Provable.asProver()`

45930:
> sounds like that could be solved with Provable.asProver()

Ok, I think I have a working branch locally that works as expected with the memoization happening "behind the scenes".  I don't love that this is hidden from the developer, but I like it better than forcing them to use a closure.  WDYT?


```
// Closure ensures that different contract states never interfere
function ExampleContractWithState() {
  const offchainStateInstance = offchainState.init();
  class ExampleContract extends SmartContract { ]
  
  return { offchainStateInstance, ExampleContract }
}
```

OR

```
class ExampleContract extends SmartContract {
  @state(OffchainState.Commitments) offchainStateCommitments =
    offchainState.emptyCommitments();
  
  // init method handles the instance management and doesn't expose it to the user at all
  offchainState = offchainState.init(this);
}
```

45930:
@mitschabaude , bf0e09e61461e05c1fc6e8a617343d1623bdae7f is pretty clean!

I think this API is a lot less error-prone and requires less deep knowledge to use.  But it does add more secret global state, which tends to cause its own issues down the line.

src/lib/mina/actions/offchain-state.ts
```diff
@@ -190,271 +211,342 @@ function OffchainState<
...
+        });
+      }
+      let instance = memoizedInstances.get(key);
+      if (instance === undefined) {
+        instance = OffchainStateInstance();
+        instance.setContractClass(
+          contractInstance.constructor as OffchainStateContractClass<Config>
+        );
+        memoizedInstances.set(key, instance);
+      }
```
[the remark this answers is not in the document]
45930: That was my first instinct, but it continues to cause provable errors:

```
Error: variables only exist inside checked computations
This shouldn't have happened and indicates an internal bug.
    at Bug (file:///Users/coby/Projects/o1js/dist/node/lib/util/errors.js:226:12)
    at assert (file:///Users/coby/Projects/o1js/dist/node/lib/util/errors.js:233:15)
    at toConstantField (file:///Users/coby/Projects/o1js/dist/node/lib/provable/field.js:1083:5)
    at toConstant (file:///Users/coby/Projects/o1js/dist/node/lib/provable/field.js:1076:12)
    at Field.toConstant (file:///Users/coby/Projects/o1js/dist/node/lib/provable/field.js:118:16)
    at Object.fromPublicKey (file:///Users/coby/Projects/o1js/dist/node/lib/ml/conversion.js:64:24)
    at Object.getAccount (file:///Users/coby/Projects/o1js/dist/node/lib/mina/local-blockchain.js:77:52)
    at Module.getAccount (file:///Users/coby/Projects/o1js/dist/node/lib/mina/mina-instance.js:43:27)
    at Object.fetch (file:///Users/coby/Projects/o1js/dist/node/lib/mina/state.js:223:32)
    at onchainActionState (file:///Users/coby/Projects/o1js/dist/node/lib/mina/actions/offchain-state.js:77:81)
    at merkleMaps (file:///Users/coby/Projects/o1js/dist/node/lib/mina/actions/offchain-state.js:86:37)
    at Object.createSettlementProof (file:///Users/coby/Projects/o1js/dist/node/lib/mina/actions/offchain-state.js:222:43)
    at file:///Users/coby/Projects/o1js/dist/node/lib/mina/actions/offchain-contract.unit-test.js:114:50
    at runInstruction (file:///Users/coby/Projects/o1js/dist/node/lib/mina/test/test-contract.js:101:27)
    at execute (file:///Users/coby/Projects/o1js/dist/node/lib/mina/test/test-contract.js:77:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async testLocal (file:///Users/coby/Projects/o1js/dist/node/lib/mina/test/test-contract.js:93:9)
    at async file:///Users/coby/Projects/o1js/dist/node/lib/mina/actions/offchain-contract.unit-test.js:104:1
```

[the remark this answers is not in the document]
45930: This error throws after compiling and deploying successfully, when I try to call `createAccount` in the first transaction to the zkapp after deploying.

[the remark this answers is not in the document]
45930: I landed on the init method because it's the only place within the zkapp itself that will only be executed when I want it to be executed.

Setting the contract instance outside of the zkapp is still possible, e.g.

```
// already-deployed contract
const zkapp = Contract(deployedAddress);
zkapp.offchainState.setContractInstance(zkapp);
```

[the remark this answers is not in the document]
dfstio: Can you please export IndexedMerkleMapBase as part of this refactoring? It is needed for the serialization of IndexedMerkleMap of any height. Now, only the function IndexedMerkleMap is exported, and I need to get this type using syntax like 
```typescript
export function serializeIndexedMap(
  map: InstanceType<ReturnType<typeof IndexedMerkleMap>>
)
```
https://github.com/zkcloudworker/zkcloudworker-tests/blob/main/src/indexed-map.ts

45930:
@dfstio do you want to open a separate PR to expose that API?  This PR may take some more time to fully review, but something simpler may get through faster on its own.

dfstio:
> @dfstio do you want to open a separate PR to expose that API? This PR may take some more time to fully review, but something simpler may get through faster on its own.

Done: https://github.com/o1-labs/o1js/issues/1855

src/lib/mina/actions/offchain-state.ts
```diff
@@ -190,271 +211,342 @@ function OffchainState<
...
+      merkleMap: IndexedMerkleMapN;
+      valueMap: Map<bigint, Field[]>;
+    };
 
-  function maybeContract() {
-    try {
-      return getContract();
-    } catch {
-      return internal.contract;
+    type InitializedInternalState = {
```
boray, line 232: This type seems to be unused. Do you plan to use it?

45930: No, it can be removed, I'll do that in my next commit

45930:
@Trivo25 just to add some visual context and reasoning behind this change:

image

This diagram shows how the feature works today.  The instance of the offchain state, and the definition of the smart contract are directly linked.  They need to be linked somehow because offchain state described a zk circuit which is a dependency of the smart contract.  Unfortunately, this way doesn't work because a 2nd contract initiialized in the same js session will reference the same offchain state, and when the states differ, you won't be able to generate a valid proof for at least one of them.

image

My changes instantiate a separate instance of the offchain state for each smart contract, but they each reference the same one zk circuit described by the offchain state definition.  This leaves 2 "fake classes": `OffchainState` which has the methods `compile`, `emptyCommitments` and `init`, which compile the circuit, return the commitments of an empty copy of the state, and intantiate a new copy of the state respectively.  The other is `OffchainStateInstance` which has all the methods and properties related to an actual copy of the offchain state, e.g. `settle`, and `commitments`.

45930:
Trying to find the right API that a) works and b) is simple to use has been the challenging piece, and the last thing Gregor and I were haggling over.  What makes it tricky is that when a prover generates a proof, it re-runs the smart contract definition, so if the definition of the smart contract includes an empty state to start with, that's what will be used at proof time.  If the definition of the contract doesn't reference offchain state at all to avoid this problem, then we can't use the data in provable methods.

2 plausible patterns are:

1. closure

```js
function MyContract() {
  const offchainState = OffchainState.init();

  class MyContract {
    // ... reference offchainState here
    }
    
    return {
        state: offchainState,
        contract: MyContract
    }
```

2. Magic global state

```js
// in offchain state definition
// manages a memoization map behind the scenes so that a new instance is only initialized the first time

  init(contractInstance: OffchainStateContract<Config>) {
    let key = 'COMPILE_TIME';
    let contractAddress = contractInstance.address;
    if (contractAddress.isConstant()) {
      key = contractAddress.toBase58();
    } else {
      Provable.asProver(() => {
        key = contractAddress.toBase58();
      });
    }
    let instance = memoizedInstances.get(key);
    if (instance === undefined) {
      instance = OffchainStateInstance();
      instance.setContractClass(
        contractInstance.constructor as OffchainStateContractClass<Config>
      );
      memoizedInstances.set(key, instance);
    }
    return instance;
}
```

```js
// in contract file

class MyContract {
  // ... reference offchainState here
  
  offchainState = OffchainState.init(this);
}
```

Currently, option 2 is implemented, which I think is ok.  All things equal, I don't like the magic global state, but I don't have an alternative that I like better.

src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -140,16 +146,145 @@ await testLocal(
...
+  (await contractA.offchainState.fields.accounts.get(receiver)).value.toBigInt()
+);
+console.log(
+  'Contract B, Receiver: ',
+  (await contractB.offchainState.fields.accounts.get(receiver)).value.toBigInt()
+);
+assert((await contractA.getBalance(sender)).toBigInt() == 900n);
+assert((await contractB.getBalance(sender)).toBigInt() == 1300n);
+assert((await contractA.getBalance(receiver)).toBigInt() == 100n);
+assert((await contractB.getBalance(receiver)).toBigInt() == 200n);
```
Trivo25, line 290: imo it would be good to add a test where the state of both contracts actually diverge more aggressively (eg contract A doesn't settle for 10 actions, whereas contract B settles every action)

45930: Ok, I will add this kind of test, rm the extra type I added, noted in Boray's comment, and add to the changelog.  Anything else you want updated?

Trivo25: I need to get a bit familiar with the changes first, will do another pass

src/lib/mina/actions/offchain-contract.unit-test.ts
```diff
@@ -21,71 +23,77 @@ const offchainState = OffchainState(
...
+  @state(OffchainState.Commitments) offchainStateCommitments =
+    offchainState.emptyCommitments();
+
+  // o1js memoizes the offchain state by contract address so that this pattern works
+  offchainState = offchainState.init(this);
+
+  init(): void {
+    super.init();
+    this.offchainState.setContractInstance(this);
+  }
```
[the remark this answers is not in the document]
Trivo25: cc @45930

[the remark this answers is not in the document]
45930: @Trivo25 

> can't we move this into offchainState.init()?

No, we can't.

We can just remove the init method here to make it more clear that `offchainState.setContractInstance` needs to be called outside of the circuit.  The way it works in my example is:

```ts
// first time deploying the contract
const c = Contract(address);

Mina.transaction({ c.deploy }); // contract instance is set automatically
```

```ts
// contract is already deployed
const c = Contract(address);
c.offchainState.setActiveInstance(c); // set contract instance manually

Mina.transaction({ c.method(...params) });
```

Trivo25 (approved):

kadirchan:
wen release 👀

boray:
> wen release 👀

It will be included in the next release but you can use before that with `npm i https://pkg.pr.new/o1-labs/o1js@<FIRST 7 DIGITS OF COMMIT HASH>`. Commit hash should be the hash of merge commit in main branch.

src/lib/mina/actions/offchain-contract-tests/ExampleContract.ts
```diff
@@ -0,0 +1,89 @@
...
+
+class StateProof extends offchainState.Proof {}
+
+// example contract that interacts with offchain state
+class ExampleContract extends SmartContract {
+  @state(OffchainState.Commitments) offchainStateCommitments =
+    offchainState.emptyCommitments();
+
+  // o1js memoizes the offchain state by contract address so that this pattern works
+  offchainState: any = offchainState.init(this);
```
45930, line 30: `any` is required here because the type of `offchainState` is actually

```
OffchainStateInstance<{
    readonly accounts: { kind: "offchain-map"; keyType: typeof PublicKey; valueType: typeof UInt64; };
    readonly totalSupply: { kind: "offchain-field"; type: typeof UInt64; };
  }>
```

But that type is not exported, resulting in this error: `Public property 'offchainState' of exported class has or is using name 'OffchainStateInstance' from external module "/Users/coby/Projects/o1js/src/lib/mina/actions/offchain-state" but cannot be named.`

45930: Seems fine to me to handle in a different PR eventually, as this API remains experimental.

src/lib/mina/actions/offchain-contract-tests/multi-contract-instance.unit-test.ts
```diff
@@ -0,0 +1,184 @@
...
+  state,
+  PublicKey,
+  UInt64,
+} from '../../../../index.js';
+import * as Mina from '../../mina.js';
+import assert from 'assert';
+import { ExampleContract } from './ExampleContract.js';
+import { settle, transfer } from './utils.js';
+
+const Local = await Mina.LocalBlockchain({ proofsEnabled: false });
```
45930, line 13: proofs not enabled on the multi-contract test since it takes much longer.  This also exercises both the proofs-enabled and proofs-disabled paths.

src/lib/mina/actions/offchain-contract-tests/single-contract-instance.unit-test.ts
```diff
@@ -0,0 +1,62 @@
...
+import * as Mina from '../../mina.js';
+import assert from 'assert';
+
+import {
+  ExampleContract,
+  offchainState as exampleOffchainState,
+} from './ExampleContract.js';
+import { settle, transfer } from './utils.js';
+
+const Local = await Mina.LocalBlockchain({ proofsEnabled: true });
```
45930, line 11: Proofs enabled on the shorter test to make sure the artifacts compile correctly and proofs can generate.

src/lib/mina/actions/offchain-contract-tests/ExampleContract.ts
Trivo25: as a rule of thumb: If we can do it now, might as well do it now :D

src/lib/mina/actions/offchain-contract-tests/ExampleContract.ts
```diff
@@ -0,0 +1,89 @@
+import {
```
Trivo25, line 1: I am okay with leaving it here for now but next time we add larger tests we should add them to the example folder and run them from there - makes it a bit easier to keep the internal code clean yet test it and provide a good example to the community

45930: I'd be happy to follow convention, but it doesn't seem like there is one.  This unit-test file was already here and I added onto it.  It seemed like it was getting too big so I broke it out into different files.  I'll follow up with you about how you see o1js being organized and we can work towards that.

45930: I'd be happy to follow convention, but it doesn't seem like there is one.  This unit-test file was already here and I added onto it.  It seemed like it was getting too big so I broke it out into different files.  I'll follow up with you about how you see o1js being organized and we can work towards that.

Trivo25: I agree, definitely not your fault - just something I noticed with the larger test

src/lib/mina/actions/offchain-contract-tests/ExampleContract.ts
mitschabaude: woah doesn't omitting this export completely break the type-safety of interacting with offchain state then?

45930: Turns out to be a pretty simple fix

https://github.com/o1-labs/o1js/pull/1871

src/lib/mina/actions/offchain-contract.unit-test.ts
mitschabaude, line 1: I'm not a fan that this test got deleted and replaced with a more verbose test that has less coverage of the implementation

PR #1839, opened 2024-09-26
Merge main into V2 to fix CI

Trivo25:
mostly to make sure the CI fixes also land in V2


src/lib/proof-system/proof-system.unit-test.ts
```diff
@@ -35,7 +35,9 @@ it('pickles rule creation', async () => {
   // a rule that verifies a proof conditionally, and returns the proof's input as output
   function main(proof: EmptyProof, shouldVerify: Bool) {
     proof.verifyIf(shouldVerify);
-    return proof.publicInput;
+    return {
```
Trivo25, line 38: left over fix from the auxiliary output PR which wasn't caught in CI

querolita (approved):

PR #1841, opened 2024-09-27
Feature/unit tests

jmikedupont2:


mitschabaude:
Hey @jmikedupont2 before you put much more work into this, I want to point out that we don't want to move unit tests to Jest 😅 

Actually it'd be nice to move the Jest tests to use [node's native test runner](https://nodejs.org/api/test.html#test-runner)

jmikedupont2:
ok, well how can we run these tests? I am struggling to get them running and recording.

jmikedupont2:
@mitschabaude thank you for the prompt feedback, I appreciate it. Will go in that direction. Right now i just want to get all the unit tests to run in perf and clinic, I got the jest tests running already so dont want to touch them just yet.

mitschabaude:
Some context about the `.unit-test.ts` files:
* they can be run with `./run /path/to/unit-test`
* under the hood, the `./run` script transpiles that single file from TS to JS, and expects that all JS files it depends on already exist (so make sure to do `npm run build` before running tests)
* then, the transpiled file is just executed with Node.js
* the process just described works for any file in the code base, and there's nothing special about `.unit-test.ts` files (except that some of our scripts, like `npm run test:unit`, pick up the `.unit-test` file pattern)
* if you want to make tests like that that use "jest-style" `it()` and `describe()` directives, then you can import those methods from `node:test` (but still run them normally as described)

jmikedupont2:
Thank you, I did not want to use jest, but was just starting somewhere, and found jest. I have now the first steps of the perf recordings starting for unit tests but clinic seems to be failing, still have much to do for a perfect reporting system. I think soon i will work on more reports and integrating the building of the docker images as well.

jmikedupont2:
It seems that the mina js is a compound of many different things that are out of sync and not updated.

Trivo25:
Do you plan to propose to add these changes (incl. CI changes, benchmarking, docker builds, ..) directly to o1js or do you plan to provide them as a 3rd party standalone tool/repo?

mitschabaude:
> It seems that the mina js is a compound of many different things that are out of sync and not updated.

Note that all of these tests run successfully in our own CI on every PR 😅 You might want to check `run-ci-tests.sh` as a reference

jmikedupont2:
@mitschabaude @Trivo25 https://github.com/meta-introspector/o1js/blob/feature/unit-test-clean/README-dev.md#goals-and-progress here is a restatement of my goals and also a cleaned up branch with some noise removed. The idea is to have a transparent environment using docker compose and github actions and other build systems like gitlab later where we can test locally the same way as github. This increases stability. Also we can publish docker images from this and distribute performance tools and reports in a unified manner.

jmikedupont2:
In the latest branches have implemented better creation tools for docker images and am now working on using hugging face to publish the results so we can have multiple outputs

jmikedupont2:
> > It seems that the mina js is a compound of many different things that are out of sync and not updated.
> 
> Note that all of these tests run successfully in our own CI on every PR 😅 You might want to check `run-ci-tests.sh` as a reference

Yes, I took those as a starting point for the docker files, and struggled to get them running in a reproducible manner on my machine. I will review everything again as I attempt to bring our threads back together. please be patient with me I am learning a lot here.

Issue #1039, opened 2023-07-20
Make `Proof` a normal `Provable` to compose within Structs


mitschabaude:
**Why**: Right now, we can only pass in proofs to provable methods as a top-level method argument. But for building abstractions, some users want to pass arrays of proofs and similar nested structures containing proofs.

**Notes on implementation**

This is more subtle than it sounds. Currently, the code that processes private inputs in `proof_systems.ts` has a lot of special cases for handling proofs, because these have to be explicitly passed on to Pickles to be recursively verified.

If we allow proofs arbitrarily nested inside other Structs, it means in some places we have to find and pull out all these proofs in order to perform that special logic.

mitschabaude:
I'm picking this up, for https://github.com/zksecurity/mina-attestations/issues/4

PR #1846, opened 2024-10-02
Proof file restructure

mitschabaude:
- to make zkprogram / proof stuff easier to work with, split `zkprogram.ts` into 3 files
  - I want to work on #1039 and having these changes in the same PR would make it unreviewable
  - no changes to any code here, just shifting around code
- unrelatedly, export some utility types from o1js that were missing


Trivo25 (approved):

mitschabaude:
pushed another small change I forgot, prune imports in the new proof file

PR #1847, opened 2024-10-02
Proof as Provable

mitschabaude:
work towards #1039 (but doesn't close it because we can't put proofs inside structs yet)

this adds a `.provable` property on proofs so that dealing with proofs becomes the same as dealing with any other method arguments.

there are still certain exceptions for proofs left because we have to declare them to pickles. however, many steps like witnessing a proof are the same now as for any other inputs.

things left that I'll do in follow-up PRs:
* automatically parse inputs for nested proofs, so that we can pass in proofs as part of larger structs and not only as top-level method arguments
* enable "declaring" proofs from anywhere, not just at the start of `picklesRuleFromFunction()` > `main()`


Trivo25 (approved):
This makes everything so much cleaner! lets make sure that we capture this in the changelog in either this or the follow up PR

PR #1851, opened 2024-10-03
Extract Proofs from any Provable

### LINKED ISSUE #1039: Make `Proof` a normal `Provable` to compose within Structs

mitschabaude:
closes #1039 

Removes restriction that proof parameters can't be nested.

Instead, (private) inputs are deeply searched for any proofs they possibly contain, and those proofs are all declared to Pickles

**Caveat**: `SelfProof` can't be nested yet

### DIFF
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -54,7 +54,7 @@ import {
   featureFlagsFromGates,
   featureFlagsToMlOption,
 } from './feature-flags.js';
-import { emptyValue, emptyWitness } from '../provable/types/util.js';
+import { emptyWitness } from '../provable/types/util.js';
 import { InferValue } from '../../bindings/lib/provable-generic.js';
 
 // public API
@@ -78,7 +78,6 @@ export {
   picklesRuleFromFunction,
   compileProgram,
   analyzeMethod,
-  synthesizeMethodArguments,
   methodArgumentsToConstant,
   methodArgumentTypesAndValues,
   Prover,
@@ -679,7 +678,7 @@ function analyzeMethod(
   method: (...args: any) => unknown
 ) {
   return Provable.constraintSystem(() => {
-    let args = synthesizeMethodArguments(methodIntf, true);
+    let args = methodIntf.args.map(emptyWitness);
     let publicInput = emptyWitness(publicInputType);
     // note: returning the method result here makes this handle async methods
     if (publicInputType === Undefined || publicInputType === Void)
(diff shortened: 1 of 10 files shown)


src/lib/proof-system/zkprogram.ts
```diff
@@ -544,21 +543,14 @@ function isDynamicProof(
...
+  return methodArgs.flatMap(extractProofs).map((proof) => proof.proof);
 }
 
 type MethodInterface = {
   methodName: string;
-  args: { type: ProvableType<unknown>; isProof: boolean }[];
-  proofs: Subclass<typeof ProofBase>[];
+  args: ProvableType<unknown>[];
+  numberOfProofs: number;
   returnType?: Provable<any>;
```
mitschabaude, line 554: simplified the `MethodInterface` type, there was no real use for storing the proofs in the updated code (they have to be deeply extracted from input values anyway)

mitschabaude:
@Trivo25 or @ymekuria this is ready to be reviewed

src/lib/proof-system/zkprogram.ts
```diff
@@ -729,21 +721,21 @@ function picklesRuleFromFunction(
...
         let value = Provable.witness(type, () => {
-          return argsWithoutPublicInput?.[i] ?? emptyValue(type);
+          return argsWithoutPublicInput?.[i] ?? ProvableType.synthesize(type);
         });
         finalArgs[i] = value;
-        if (isProof) {
-          let Proof = type as Subclass<typeof ProofBase<any, any>>;
-          let proof = value as ProofBase<any, any>;
-          proofs.push({ proofInstance: proof, classReference: Proof });
+
```
ymekuria, line 735: I'm just curious why we don't check `if(isProof){` here anymore?

ymekuria (approved):
Nice!! I just had a minor comment/question.

src/lib/proof-system/zkprogram.ts
mitschabaude: now, every method argument can contain multiple proofs, so I iterate over all the proofs we get from `extractProofs()`, rather than just test if the top-level method argument is a proof

PR #1853, opened 2024-10-07
Fixed state.fetch() behaviour in regards to custom token contracts

rpanic:
When using contract.state.fetch(), the method uses the default tokens, even though contracts might be deployed on a custom token account. That leads to that account not being found.


mitschabaude (approved):
Thanks!

Trivo25 (approved):
Thanks! :)

Issue #1855, opened 2024-10-07
Export IndexedMerkleMapBase


dfstio:
**Proposal:**
Export IndexedMerkleMapBase. It is needed for the serialization of IndexedMerkleMap of any height. Now, only the function IndexedMerkleMap is exported, and I need to get this type using syntax like

```typescript
export function serializeIndexedMap(
  map: InstanceType<ReturnType<typeof IndexedMerkleMap>>
)
```
https://github.com/zkcloudworker/zkcloudworker-tests/blob/main/src/indexed-map.ts

45930:
Hey do you want to try to implement this yourself in a PR?  It sounds like you know what is required.  I think exporting that base type sounds ok.

dfstio:
> Hey do you want to try to implement this yourself in a PR? It sounds like you know what is required. I think exporting that base type sounds ok.

Added https://github.com/o1-labs/o1js/pull/1856

PR #1856, opened 2024-10-07
Export IndexedMerkleMapBase

dfstio:
Closes https://github.com/o1-labs/o1js/issues/1855

Exports IndexedMerkleMapBase that is useful for the serialization of IndexedMerkleMap of any height.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -156,6 +156,8 @@ const Experimental_ = {
   IndexedMerkleMap,
 };
 
+type ExperimentalIndexedMerkleMapBase = IndexedMerkleMapBase;
+
 /**
  * This module exposes APIs that are unstable, in the sense that the API surface is expected to change.
  * (Not unstable in the sense that they are less functional or tested than other parts.)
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -166,6 +168,7 @@ namespace Experimental {
   // indexed merkle map
   export let IndexedMerkleMap = Experimental_.IndexedMerkleMap;
   export type IndexedMerkleMap = IndexedMerkleMapBase;
+  export type IndexedMerkleMapBase = ExperimentalIndexedMerkleMapBase;
```
Trivo25, line 171: We already expose `IndexedMerkleMapBase` (see the line above) via
```ts
export type IndexedMerkleMap = IndexedMerkleMapBase;
```

does this not work for you?

dfstio: > does this not work for you?

Yes, it works if I import `IndexedMerkleMap` two times - as the function and as the IndexedMerkleMapBase type:

```typescript
const { IndexedMerkleMap } = Experimental;
type IndexedMerkleMap = Experimental.IndexedMerkleMap;
```

Then, I use both the function and the type in the serializing code, which is a little bit confusing, as the same name is used for the type that contains the data that I use and the function, but it works and I can even rename the IndexedMerkleMap back to IndexedMerkleMapBase while importing:
```typescript
const { IndexedMerkleMap } = Experimental;
type IndexedMerkleMapBase = Experimental.IndexedMerkleMap;
```

dfstio: > I would be in favor of not exporting another type, just for the sake of simplicity. This will be fixed indirectly when we move `IndexedMerkleMap` out of the experimental namespace

I agree. Let's wait for a move of `IndexedMerkleMap` out of the experimental namespace. I hope it will be soon, as I plan to use it on the mainnet!

Issue #1411, opened 2024-02-05
process hangs when verifying proof read&deserialized from file


Yoga2015:
Original issue at Discord: https://discord.com/channels/484437221055922177/1204071024191082496

All the code at https://github.com/TokeniZK/zkapp-test/tree/main/src
tried with o1js version:  0.15.2,  0.15.4, & 0.16.0

**Reproduce steps:**

I have a demo zkProgram(https://github.com/TokeniZK/zkapp-test/blob/main/src/zkprogram.ts),  

**step1**:  `npm run step1`
(seen at /src/step1.ts) I exec one method & gen proof, then serialize the `proof` into file: `proof.json `. and then read&deserialize it from file and verify it and smoothly get the result: `True`. 

However, 
**step2**: (seen at /src/step2.ts)      `npm run step2`
I start another process to read the proof from the file on step1(`proof.json`), deserialize it and try to verify the proof, But the process keep await at `verify(*)` line !!

Yoga2015:
hi, @mitschabaude, could u please suggest me? thx

dfstio:
Using compile({ forceRecompile: true }) fixes it

Yoga2015:
thx！

dfstio:
Yes, it is correct. I've created a small repo with my usual settings that work for me in MinaNFT:
https://github.com/dfstio/zkapp-test/tree/dfst (branch dfst) 
based on @Yoga2015 code

@mitschabaude, you can use it for debugging

SpaceManiac:
Running in to the same problem with the dynamic `verify` method. It seems the **workaround** is to either `forceRecompile: true` or `cache: Cache.None` *any program at all* before attempting to verify proofs. It doesn't have to be the same program you're verifying proofs for.

Single-file repro. Will succeed on first run, hang on second run, and can be fixed by changing `false` to `true` on line 4.

```js
import { Bool, Cache, verify, ZkProgram } from 'o1js';

//  v-- Change to `true` to fix the hang, even though NoopProgram != VerifyMeProgram.
if (false) {
  const NoopProgram = ZkProgram({
    name: 'NoopProgram',
    methods: {
      chaff: {
        privateInputs: [],
        async method(b) {}
      }
    }
  });
  await NoopProgram.compile({ cache: Cache.None });
}

const inputs = await (typeof getPreviousProof !== 'undefined' ? getPreviousProof : createProof)();
console.log('verifying...');
[...]

PR #1857, opened 2024-10-08
Update proof-systems for safe lagrange-basis tracking

mrmr1993:
This PR fixes https://github.com/o1-labs/o1js/issues/1411.

This PR also includes some auxiliary changes in a separate commit, without which I could not make `main` compile.

The bulk of the fix is [here](https://github.com/o1-labs/proof-systems/pull/2672), forcing an API which forces us to generate a lagrange basis if one is not present, and as such removing all opportunity for the panics that halt the WASM interpreter.

o1js-bindings PR is up! https://github.com/o1-labs/o1js-bindings/pull/307


mitschabaude:
Nice!

I think the auxiliary changes might not be necessary with the TS version that o1js uses (there were significant improvements to type inference from conditionals in recent versions).
To install the version that matches CI, you can just do 
```
npm i
```

mitschabaude:
@45930 you should revert all the unrelated type "fixes" that @mrmr1993 did, they were only necessary because of an outdated TS version

Trivo25 (approved):

PR #1858, opened 2024-10-08
Looking up values in runtime tables

querolita:
This PR adds a function inside `Gadgets` in order to look up values in a runtime table. It also adds the functionality to configure a runtime table. ~However, it still does not have the ability to _create_ a runtime table. When that's done, I will include unit tests with the new feature.~


querolita:
This commit [7553ad4](https://github.com/o1-labs/o1js/pull/1858/commits/7553ad44d52caaef9fbb7e2be4c118ba8dd736e0) fixed the feature flags configuration that was preventing us from being able to use runtime tables from o1js programs. 

Right now it is solved by adding an optional flag called `withRuntimeTables` that is passed as `true` inside `.compile()` and `.fromZkProgram()` whenever there are runtime tables configured (this is to be backwards compatible).

Nonetheless, a future improvement will consist on making this automatically by being able to access a global state which keeps track of whether the runtime table configuration was performed or not, so that the developer does not suffer this extra annoyance.

src/lib/proof-system/feature-flags.ts
```diff
@@ -108,7 +111,7 @@ async function featureFlagsfromFlatMethodIntfs(
     xor: false,
     rot: false,
     lookup: false,
-    runtimeTables: false,
+    runtimeTables: false, //
```
Trivo25, line 114: can we remove `//` here?

src/lib/provable/gates.ts
```diff
@@ -284,6 +286,13 @@ function raw(kind: KimchiGateType, values: Field[], coefficients: bigint[]) {
   );
 }
 
+function addRuntimeTableConfig(id: number, first_column: bigint[]) {
```
Trivo25, line 289: nit `firstColumn`

src/lib/provable/gates.ts
```diff
@@ -284,6 +286,13 @@ function raw(kind: KimchiGateType, values: Field[], coefficients: bigint[]) {
...
+  Snarky.gates.addRuntimeTableConfig(
```
Trivo25, line 290: if i remember correctly, the `id` of the runtime tables must be anything but 1 since that's reserved. I think we should make the developer aware (I am not sure what would happen right now if we specify a runtime table with id 1) - I think in the doc comment is fine for now

Trivo25 (approved):
Thanks, this is great!

I will open up a follow up issue so that we can build a simple API around that at some point :)

src/lib/provable/gates.ts
querolita: Yeah actually 0 and 1 iirc... What about asserting that id > 1? I haven't seen any of those functions performing checks...

hattyhattington17 (approved):

PR #1860, opened 2024-10-10
Expose some types and gadgets

mitschabaude:
Several additions that were handy when implementing dynamic-length SHA2 and general-purpose array types in https://github.com/zksecurity/mina-attestations/pull/25

* Exposes a few utility types like `InferValue`, `From`, `TupleN`
* Exposes a few gadgets like `arrayGet()` and UInt32 to/from bytes
* make the `Packed` class much better typed

CI run here: https://github.com/zksecurity/o1js/pull/2

### DIFF
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -1,9 +1,8 @@
 import { EmptyUndefined, EmptyVoid } from '../../bindings/lib/generic.js';
 import { Snarky, initializeBindings, withThreadPool } from '../../snarky.js';
 import { Pickles, Gate } from '../../snarky.js';
-import { Field, Bool } from '../provable/wrapped.js';
+import { Field } from '../provable/wrapped.js';
 import {
-  FlexibleProvable,
   FlexibleProvablePure,
   InferProvable,
   ProvablePureExtended,
@@ -12,7 +11,6 @@ import {
 import {
   InferProvableType,
   provable,
-  provablePure,
 } from '../provable/types/provable-derivers.js';
 import { Provable } from '../provable/provable.js';
 import { assert, prettifyStacktracePromise } from '../util/errors.js';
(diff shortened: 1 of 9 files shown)


src/lib/proof-system/zkprogram.ts
mitschabaude, line 1: some unused imports and an unused method purged here

src/lib/provable/int.ts
mitschabaude, line 1: exposing the uint32 / bytes converters that we already use internally for sha2

src/lib/provable/gadgets/sha256.ts
```diff
@@ -239,7 +234,7 @@ function sigma(u: UInt32, bits: TupleN<number, 3>, firstShifted = false) {
  *
  * @returns The updated intermediate hash values after compression.
  */
-function sha256Compression(H: UInt32[], W: UInt32[]) {
+function sha256Compression([...H]: UInt32[], W: UInt32[]) {
```
mitschabaude, line 237: this fixes a bug, the compression function mutated its input, which bit me because in dynamic sha2 you not necessarily want to use the output

src/lib/provable/types/provable-derivers.ts
```diff
@@ -169,3 +170,44 @@ function provableExtends<
     },
   } satisfies ProvableHashable<S, InferValue<A>>;
 }
+
+function mapValue<
+  A extends ProvableHashable<any>,
+  V extends InferValue<A>,
+  W,
+  T extends InferProvable<A>
+>(
```
mitschabaude, line 179: method to take one provable and modify its "value" type

src/lib/provable/wrapped-classes.ts
```diff
@@ -19,3 +19,6 @@ function Bytes(size: number) {
 Bytes.from = InternalBytes.from;
 Bytes.fromHex = InternalBytes.fromHex;
 Bytes.fromString = InternalBytes.fromString;
+
+// expore base class so that we can detect Bytes with `instanceof`
+Bytes.Base = InternalBytes;
```
mitschabaude, line 24: we should _always_ expose base classes somewhere for type factories, so that users can detect what class they have using `instanceof`.
putting base classes on `.Base` of the factory function seems like a nice, non-intrusive pattern to me.

mitschabaude:
@Trivo25 or @ymekuria this is ready for review

PR #1864, opened 2024-10-14
Setup codeowners file with new team structure

45930:
This PR cleans up the `CODEOWNERS` file and lays the groundwork for a new round-robin code review process that will give the exploration team and future zkapps team exposure to o1js PR reviews.

Ultimately, it is the `o1js-admins` job to manage what code gets into o1js.  This process aims to get more people to the level that they can be added to the list of admins.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,9 +5,3 @@
 # We are relying on good faith to wait for admin approval on these PRs
 # Dangerous PRs that could release a new version, new consensus rules, or anything like that will necessarily touch more than `/src/lib`, so this is minimally risky
 /src/lib @o1-labs/o1js-admins @o1-labs/exploration-team @o1-labs/future-zkapps
-
-# Changes to the following paths can be safely owned by the current and future zkapps teams (correct use of codeowners)
-/src/examples @o1-labs/exploration-team @o1-labs/future-zkapps
-/src/lib/util @o1-labs/exploration-team @o1-labs/future-zkapps
-
-
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,6 +1,13 @@
-# All files require a review from @o1-labs/o1js-admins
+# By default, review is required by o1js-admins
 * @o1-labs/o1js-admins
 
-# Changes to source code should be reviewed by exploration team and future zkapps team 
-/src/examples @o1-labs/o1js-admins @o1-labs/exploration-team @o1-labs/future-zkapps
+# Changes to the following paths should be reviewed by o1js-admins, but we will tag current and future zkapps teams as well (incorrect use of codeowners)
+# We are relying on good faith to wait for admin approval on these PRs
+# Dangerous PRs that could release a new version, new consensus rules, or anything like that will necessarily touch more than `/src/lib`, so this is minimally risky
 /src/lib @o1-labs/o1js-admins @o1-labs/exploration-team @o1-labs/future-zkapps
+
+# Changes to the following paths can be safely owned by the current and future zkapps teams (correct use of codeowners)
+/src/examples @o1-labs/exploration-team @o1-labs/future-zkapps
+/src/lib/util @o1-labs/exploration-team @o1-labs/future-zkapps
+
+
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -1 +1,9 @@
-/src/lib/gadgets @o1-labs/crypto-eng-reviewers @mitschabaude
+# All files require a review from @o1-labs/o1js-admins
+* @o1-labs/o1js-admins
+
+# Changes to source code should be reviewed by exploration team and future zkapps team 
+/src/examples @o1-labs/exploration-team @o1-labs/future-zkapps
+/src/lib @o1-labs/exploration-team @o1-labs/future-zkapps
+
+# Provable code should be reviewed by the crypto team
+/src/lib/provable @o1-labs/crypto-eng-reviewers
```
45930, line 9: This is a legacy rule.  I'm happy to keep it, but I'm not sure if `crypto-eng-reviewers` is a valid target anymore.

Trivo25: Let's remove this. `/provable` has a lot of stuff that won't require crypto engineer code reviews (eg the Merkle tree api and many other things)

CODEOWNERS
```diff
@@ -1 +1,9 @@
-/src/lib/gadgets @o1-labs/crypto-eng-reviewers @mitschabaude
+# All files require a review from @o1-labs/o1js-admins
+* @o1-labs/o1js-admins
+
+# Changes to source code should be reviewed by exploration team and future zkapps team
```
mrmr1993, line 4: I think these rules operate as an 'or', not an 'and'. That is to say, any time someone from one team approves, the other team's review will no longer be required.

CODEOWNERS
45930: Is there any specific part of `/provable` that you would want a crypto review on?

CODEOWNERS
```diff
@@ -1 +1,13 @@
-/src/lib/gadgets @o1-labs/crypto-eng-reviewers @mitschabaude
+# By default, review is required by o1js-admins
+* @o1-labs/o1js-admins
+
+# Changes to the following paths should be reviewed by o1js-admins, but we will tag current and future zkapps teams as well (incorrect use of codeowners)
+# We are relying on good faith to wait for admin approval on these PRs
+# Dangerous PRs that could release a new version, new consensus rules, or anything like that will necessarily touch more than `/src/lib`, so this is minimally risky
+/src/lib @o1-labs/o1js-admins @o1-labs/exploration-team @o1-labs/future-zkapps
```
45930, line 7: This incorrect use of codeowners is still a more secure subset of our current rules AFAICT.  Right now any o1 eng can approve o1js PRs.  This will require a review from a subset of that group.

Reviews from crypto, protocol, or node specialists may also be required from time to time, which can be requested manually.  We probably also want to ensure that a github admin can override and merge an emergency PR if needed.

45930:
@mrmr1993 With the updated comments, does this procedure work for you?

CODEOWNERS
```diff
@@ -1 +1,13 @@
...
+
+# Changes to the following paths can be safely owned by the current and future zkapps teams (correct use of codeowners)
+/src/examples @o1-labs/exploration-team @o1-labs/future-zkapps
+/src/lib/util @o1-labs/exploration-team @o1-labs/future-zkapps
```
Trivo25, line 11: I would be in favor of leaving these specifications out of the file until we see an actual need for it - right now it seems too arbitrary (other than the `examples` folder maybe). If we decide to make exceptions like these, we would have to add more than just this particular folder. On the other hands, it's hard to differentiate between changes that might seem trivial but are actually non-trivial

45930: Ok, I will rm this for now.

Just so you know what I'm going for, codeowners can be used to require review of specific files by specific teams.  In this case (and thus reduce the load of review from other teams).  If a big PR includes API changes, new data types, tests, docs, and an example, maybe we only need admin review on the API changes and data types.  The other changes still need a review, but that review can come from one of the teams instead.

I will go with the simplest version to start, which is just `o1js-admin` is required by default and current and future zkapps teams will be required on changed to `/src/lib` since that's basically the current model.

Trivo25 (approved):
anything else we need to land this?

45930:
@Trivo25 I'll merge but we may still need admins to set up the round robin assignment.

PR #1871, opened 2024-10-17
Include the type OffchainStateInstance in the internal API

45930:
Fixes the typing of #1834 , so shouldn't need a changelog if it goes out in the same version

### DIFF
--- a/src/lib/mina/actions/offchain-state.ts
+++ b/src/lib/mina/actions/offchain-state.ts
@@ -512,10 +512,7 @@ function OffchainState<
   const memoizedInstances = new Map<String, OffchainStateInstance<Config>>();
 
   return {
-    init(contractInstance: OffchainStateContract<Config> | null) {
-      if (contractInstance === null) {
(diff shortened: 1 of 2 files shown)


src/lib/mina/actions/offchain-state.ts
```diff
@@ -33,7 +33,7 @@ import { ProvableType } from '../../provable/types/provable-intf.js';
 export { OffchainState, OffchainStateCommitments };
 
 // internal API
-export { OffchainField, OffchainMap };
+export { OffchainField, OffchainMap, OffchainStateInstance };
```
Trivo25, line 36: This only exports the type internally, do you want to export it to users too?

45930: Nope, users don't need this, but exporting it internally lets TS imply the type when used in a smart contract.  Thus, we don't have to use `any`.

Trivo25 (approved):

Issue #1724, opened 2024-07-05
Any easy way to verify secp256r1 (NIST P256) ECDSA signature?


xomexh:
The custom curve params for the NIST curve does not seem to work. Anyway/any work around as to how a P256 ECDSA signature? 
Using curve params as:
 ```
const secp256r1Params: CurveParams = {
    name: 'secp256r1',
    modulus: 0xFFFFFFFF00000001000000000000000000000000FFFFFFFFFFFFFFFFFFFFFFFFn,
    order: 0xFFFFFFFF00000000FFFFFFFFFFFFFFFFBCE6FAADA7179E84F3B9CAC2FC632551n,
    a: -3n,
    b: 0x5AC635D8AA3A93E7B3EBBD55769886BC651D06B0CC53B0F63BCE3C3E27D2604Bn,
    generator: {
      x: 0x6b17d1f2e12c4247f8bce6e563a440f277037d812deb33a0f4a13945d898c296n,
      y: 0x4fe342e2fe1a7f9b8ee7eb4a7c0f9e162bce33576b315ececbb6406837bf51f5n,
    },
};```

Trivo25:
Hey! Unfortunately, we currently only support foreign curves with curve parameter `a !== 0`, and here `a = -3`

https://github.com/o1-labs/o1js-bindings/blob/3fca906694f102dc60bf64745324de234cd6812d/crypto/elliptic-curve.ts#L531-L532

xomexh:
Thank you for the reply, I would see I'd see for myself if I could do something about this. 

Also another completely unrelated issue/doubt, is there an easy way to pass a composite struct having string data inside to a zkProgram? 

Example:
'''class P256_Signature_Field extends Struct({
    sig: String,
    msg: String
}){
    constructor(value:{sig:string, msg:string}){
        super(value)
    }
}'''

This inherently doesn't give any errors, but the values of sig and msg are not accessible inside prove-able code (inside the methods of zkProgram) i.e I get no value from these fields of sig and msg.

Trivo25:
> Also another completely unrelated issue/doubt, is there an easy way to pass a composite struct having string data inside to a zkProgram?

You can't work with strings inside provable code, you would have to convert your strings into Field elements using an encoding

mitschabaude:
I think this was a recent feature request and should remain open

PR #1885, opened 2024-10-24
Support Secp256r1

mitschabaude:
closes https://github.com/o1-labs/o1js/issues/1724

this adds ECDSA (and general) support for the curve **secp256r1**, also known as **P-256**, which is the standard curve used in ECDSA outside the crypto world, due to its recommendation by NIST; see [P-256 in this document](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-186.pdf)

Most changes are in https://github.com/o1-labs/o1js-bindings/pull/310

changes here:
* properly treat and pass on curve parameter `a` in EC addition and doubling methods
  * no circuit changes! the parameter was just missing in witness generation
* add secp256r1 to the curve and ECDSA unit tests

## Efficiency

Sadly, secp256r1 doesn't support the endomorphism we have implemented, because that endomorphism crucially relies on the `a = 0` property. Therefore, in our ECDSA circuit the GLV method is not used, which means significant constraint savings get lost:
* ECDSA on secp256k1: **32k** constraints
* ECDSA on secp256r1: **40k** constraints

### DIFF
--- a/src/lib/provable/gadgets/elliptic-curve.ts
+++ b/src/lib/provable/gadgets/elliptic-curve.ts
@@ -20,7 +20,6 @@ import { provable } from '../types/provable-derivers.js';
 import { assertPositiveInteger } from '../../../bindings/crypto/non-negative.js';
 import { arrayGet, assertNotVectorEquals } from './basic.js';
 import { sliceField3 } from './bit-slices.js';
-import { Hashed } from '../packed.js';
 import { exists } from '../core/exists.js';
 import { ProvableType } from '../types/provable-intf.js';
 
@@ -54,7 +53,7 @@ namespace Ecdsa {
   export type signature = { r: bigint; s: bigint };
 }
 
-function add(p1: Point, p2: Point, Curve: { modulus: bigint }) {
+function add(p1: Point, p2: Point, Curve: { modulus: bigint; a: bigint }) {
   let { x: x1, y: y1 } = p1;
   let { x: x2, y: y2 } = p2;
   let f = Curve.modulus;
@@ -63,7 +62,7 @@ function add(p1: Point, p2: Point, Curve: { modulus: bigint }) {
 
   // constant case
   if (Point.isConstant(p1) && Point.isConstant(p2)) {
-    let p3 = affineAdd(Point.toBigint(p1), Point.toBigint(p2), f);
+    let p3 = affineAdd(Point.toBigint(p1), Point.toBigint(p2), f, Curve.a);
     return Point.from(p3);
   }
 
@@ -120,7 +119,7 @@ function double(p1: Point, Curve: { modulus: bigint; a: bigint }) {
 
   // constant case
   if (Point.isConstant(p1)) {
-    let p3 = affineDouble(Point.toBigint(p1), f);
+    let p3 = affineDouble(Point.toBigint(p1), f, Curve.a);
     return Point.from(p3);
   }
 
--- a/src/lib/provable/gadgets/elliptic-curve.ts
+++ b/src/lib/provable/gadgets/elliptic-curve.ts
@@ -128,7 +128,7 @@ function double(p1: Point, Curve: { modulus: bigint; a: bigint }) {
     let [x1_, y1_] = Field3.toBigints(x1, y1);
     let denom = inverse(mod(2n * y1_, f), f) ?? 0n;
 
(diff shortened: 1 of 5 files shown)


mitschabaude:
@Trivo25 can you get me the necessary reviewers for this?

Geometer1729 (commented):
LGTM

mitschabaude:
Thanks for the review @Geometer1729! I also need one [here](https://github.com/o1-labs/o1js-bindings/pull/310) and then I need you to merge both PRs since I don't have permission to

mitschabaude:
Oh and @Trivo25 I also need admin review for changing the bindings commit

Geometer1729:
@mitschabaude Can you run `npm run build:update-bindings` and push the changes that creates in `o1js-bindings`? I think the bindings not having been recompiled may be why CI is failing.

mitschabaude:
> @mitschabaude Can you run `npm run build:update-bindings` and push the changes that creates in `o1js-bindings`? I think the bindings not having been recompiled may be why CI is failing.

no @Geometer1729, I didn't change anything that needs compiling. also, CI isn't failing, just the benchmark job is, and that one _always_ fails on forks

mitschabaude:
updating the bindings is just needed when you need to recompile ocaml or rust sources

Geometer1729:
Oh, I thought it was also affected by some of the content of o1js-bindings. Fair enough sorry about that.

src/lib/provable/test/ecdsa.unit-test.ts
```diff
@@ -189,3 +190,36 @@ console.timeEnd('ecdsa verify (prove)');
...
+            config
+          ),
+        };
+      },
+    },
+  },
+});
+
+console.time('ecdsa verify, no endomorphism (build constraint system)');
+let csNoEndo = (await programNoEndo.analyzeMethods()).ecdsa;
```
45930, line 222: Could we add some test vectors to verify this implementation?

mitschabaude: yes! I'll do that

45930:
Amazing work!  Florian has final approval, but it LGTM

mitschabaude:
the test is failing because we're still running node 18 in CI, where the web crypto API is missing 😬 

do you mind if I change it to node 20?

mitschabaude:
need workflow approval again 😢 @Trivo25

mitschabaude:
and again @Trivo25!

mitschabaude:
@nicc @mrmr1993 that's why I need write access to the repo ^^ https://github.com/o1-labs/o1js/pull/1885#issuecomment-2436878827

Trivo25:
@hattyhattington17 can we get your eyes on this too?

Trivo25:
@mitschabaude fyi I will get this merged before the next release

Trivo25 (approved):

src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -129,7 +128,7 @@ function double(p1: Point, Curve: { modulus: bigint; a: bigint }) {
     let [x1_, y1_] = Field3.toBigints(x1, y1);
     let denom = inverse(mod(2n * y1_, f), f) ?? 0n;
 
-    let m = mod(3n * mod(x1_ ** 2n, f) * denom, f);
+    let m = mod((3n * mod(x1_ ** 2n, f) + Curve.a) * denom, f);
```
querolita, line 131: In [PR310](https://github.com/zksecurity/o1js-bindings/blob/e081466d5e435fe3fbd32fa83ef5c0ef3f030dec/crypto/elliptic-curve.ts#L578) this `m` is computed with a single modular operation, meaning: `let m = mod((3n * x * x + a) * d, p);`. Cannot we remove one `mod` from this line?

querolita (approved):

hattyhattington17 (approved):

src/lib/provable/gadgets/elliptic-curve.ts
mitschabaude: good question! I think using `mod()` twice is better actually, because both x and d are large field elements. so x^2 is already a bigint of 2x the length, and it's more efficient to first reduce it (with `mod()`) back to 1x the length before doing another bigint multiplication with d. In the code you mention, we get an intermediate result of 2x * 1x = 3x the length

PR #1886, opened 2024-10-24
Tiny array gadget fix

mitschabaude:
Single-letter fix :D For a bug that hit me when doing https://github.com/zksecurity/mina-credentials/pull/46

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/b04520d...HEAD)
 
+### Fixed
+
+- Witness generation error in `Gadgets.arrayGet()` when accessing out-of-bounds indices https://github.com/o1-labs/o1js/pull/1886
+
 ## [2.0.0](https://github.com/o1-labs/o1js/compare/7e9394...b04520d)
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


boray (approved):

mitschabaude:
thanks @boray - can you run CI please?

mitschabaude:
I don't think this needs a changelog - wdyt @boray? you can add the no-changelog label if you agree

boray:
> I don't think this needs a changelog - wdyt @boray? you can add the no-changelog label if you agree

I don't think there is any user-facing change so we can label no-changelog. Do you have any idea why benchmark fails?

Trivo25:
If it fixes a bug that a user can encounter it warrants a changelog imo!

45930:
@mitschabaude what is the error case that this fixes?

mitschabaude:
> @mitschabaude what is the error case that this fixes?

it throws an error in a place no error is wanted (witness generator). On an out of bounds array access, the gadget is supposed to fail gracefully and return an unconstrained value.

(If you ask yourself why this is useful: It can be useful when you're in a "dummy" situation already and don't care about the return value. This sort of stuff often happens when you do dynamic reasoning in a static-sized circuit: You'll end up doing superfluous operations, but those aren't supposed to crash. For me it happened here: https://github.com/zksecurity/mina-credentials/pull/46/files#diff-b368b5466f745e5f81526d49eaf016583efcf8c48efefd8276db224cdd9108d6R381-R385)

mitschabaude:
Pushed a changelog!

mitschabaude:
@Trivo25 @45930 now with the changelog I need o1js-admin review!

mitschabaude:
> Do you have any idea why benchmark fails?

Btw, benchmark always fails on PRs from forks. I don't know why (but it's also not a required CI job and doesn't matter)

PR #1894, opened 2024-11-14
Check mina commit in ci

Geometer1729:
Adds a check in ci that the bindings were build with the same version of mina as exists in the submodule

https://github.com/o1-labs/o1js-bindings/pull/311


Trivo25 (approved):
Cool! I think this is a good way of forcing ourselves to show more due diligences when working with the bindings :)

PR #1905, opened 2024-11-18
API chunking support

querolita:
This PR adds all the needed changes to the ZkProgram API to support chunking as soon as the memory in WASM is optimized to fit it.

### DIFF
--- a/src/examples/zkprogram/program-with-chunking.ts
+++ b/src/examples/zkprogram/program-with-chunking.ts
@@ -1,0 +1,27 @@
+import { Field, ZkProgram, verify } from 'o1js';
+
+let MyProgram = ZkProgram({
+  chunks: 1,
+  overrideWrapDomain: 0,
+  name: 'example-with-chunking',
+  publicOutput: Field,
+
+  methods: {
+    baseCase: {
+      privateInputs: [],
+      async method() {
+        return Field(0);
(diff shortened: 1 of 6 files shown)


querolita:
Closing in favor of https://github.com/o1-labs/o1js/pull/2215

src/examples/zkprogram/program-with-chunking.ts
```diff
@@ -0,0 +1,35 @@
+import { Field, Cache, Gadgets, ZkProgram } from 'o1js';
+
+let MyProgram = ZkProgram({
+  chunks: 2,
```
45930, line 4: Is the expectation on developers that they will specify number of chunks explicitly like this?  How will we recommend choosing a size?  Is there any cost to selecting a chunk size of 2 for a proof that is only 5k rows?

querolita: Ideally, it should be performed automatically when in compile time we could count the number of rows and choose the chunks accordingly. But because chunking is still not working until we solve the memory limitation issue, this PR was just aiming at pinpointing the minimal changes required. There is definitely a cost implied with chunking, both in terms of performance and proof length. So you don't want to have more chunks than you really need.

Trivo25 (approved):

Issue #1891, opened 2024-11-07
Compiling ZkProgram in the browser never ends


berkay1532:
I noticed that a simple zkprogram I created did not compile in the browser. Then I tried to compile the zkprogram in the tutorial to see if the problem was with my program, but it did not compile either. Finally, when I removed the "**SelfProof**" parts in the zkprogram and compiled it, I saw that the process was completed without any problems. That's why I decided to open an issue about this.

Tutorial ZkProgram
`
const Add = ZkProgram({
  name: 'add-example',
  publicInput: Field,

  methods: {
    init: {
      privateInputs: [],

      async method(state: Field) {
        state.assertEquals(Field(0));
      },
    },

    addNumber: {
      privateInputs: [SelfProof, Field],

      async method(
        newState: Field,
        earlierProof: SelfProof<Field, void>,
        numberToAdd: Field
      ) {
        earlierProof.verify();
        newState.assertEquals(earlierProof.publicInput.add(numberToAdd));
      },
    },

    add: {
      privateInputs: [SelfProof, SelfProof],

      async method(
        newState: Field,
        earlierProof1: SelfProof<Field, void>,
        earlierProof2: SelfProof<Field, void>
      ) {
        earlierProof1.verify();
        earlierProof2.verify();
        newState.assertEquals(
          earlierProof1.publicInput.add(earlierProof2.publicInput)
        );
      },
    },
  },
});
`

YofiY:
I have been having the same issue.

45930:
What version of o1js are you using?  Could you make sure it's either 1.9.1 or 2.0.0?

berkay1532:
Yes i tried both.Unfortunately neither worked.

kadirchan:
https://github.com/o1-labs/o1js/issues/1890#issuecomment-2469029108
can you try `"o1js": "https://pkg.pr.new/o1-labs/o1js@fc16ce3"`

berkay1532:
Finally, it's working! This problem took me about 10 days. Thank u @kadirchan.

45930:
@berkay1532 , I don't think we can close this issue if the fix is to use a non-released version of o1js.  We still need to release a fixed version.

berkay1532:
Yep , i think so. Let’s not close till it is solved

mitschabaude:
It sounds like the fix is to use an older o1js version, so there was a recent regression?

45930:
[...]

Issue #1890, opened 2024-11-07
zkProgram stalling on proving


kadirchan:
I made a sequencer that collects and recursively bundles them, also this sequencer runs in a docker container. But when I first build and up this container it compiles contracts and zkprograms but when it comes to proving a method of zkprogram in [like this](https://github.com/DRM-Mina/drm-mina-sequencer/blob/a11a0e29b810debef3e6fcb4bb3b688b61d16379/src/sequencer/bundler.ts#L131), whole app stuck in this line without throwing any error and utilizing 0% cpu and 6 GB ram (total 32 GB available).
Only way to solve this I need to restart container, it takes much lesser time to compiling again probably due to caching and this time it does not stuck in proving. Also ram usage of app reduces almost 20-30% after restarting.

kadirchan:
I'am not sure if it is related or not to starting of snarkjs backend, as I said I didn't receive any errors it just acts like #1857

mitschabaude:
> I'am not sure if it is related or not to starting of snarkjs backend

It's not! "Starting" that backend just means doing a dynamic `import()` of a big file. This is not an error-prone process, and it already must have happened before compiling. Similarly, starting the worker thread pool is also done when compiling

kadirchan:
Also found that compiling of offchainState on browser stuck too, when I was using `"o1js": "https://pkg.pr.new/o1-labs/o1js@fc16ce3"` I was able to compile and use them but in 1.9.1 this problems occur.
Degraded my version to commit fc16ce3 for now

PR #1906, opened 2024-11-19
Fix Lagrange bases caching on the web

mitschabaude:
fixes https://github.com/o1-labs/o1js/issues/1891

changes are in the bindings PR: https://github.com/o1-labs/o1js-bindings/pull/313

EDIT:
a frustrating aspect of this fix is that the problem with compiling recursive programs in the browser must have been present for at least a year, if I'm not mistaken -- ever since Lagrange basis caching was introduced.

Ideally, following up this PR we would add a test that catches such a regression. We do currently have web tests but they don't use recursion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/e1bac02...HEAD)
 
+### Fixed
+
+- Compiling stuck in the browser for recursive zkprograms https://github.com/o1-labs/o1js/pull/1906
+
 ## [2.1.0](https://github.com/o1-labs/o1js/compare/b04520d...e1bac02) - 2024-11-13
 
 ### Added


45930:
@mitschabaude you're a hero

> a frustrating aspect of this fix is that the problem with compiling recursive programs in the browser must have been present for at least a year, if I'm not mistaken -- ever since Lagrange basis caching was introduced.

Yeah, caching never worked until 1.9.0, which introduced the performance bug.  Then in 1.9.1 it worked _well_ in node but still not in web.  Going forward we need to be more diligent about testing our API in every way a user might actually use it.  Since you saved me a lot of time in trying to track down this bug, I can re-allocate it to adding more tests :)

mitschabaude:
> Yeah, caching never worked until 1.9.0, which introduced the performance bug. Then in 1.9.1 it worked well in node but still not in web.

This is not true AFAICT, caching always worked in node.

In node, we only had the edge case of failing verification when lagrange bases hadn't been computed before. This didn't involve caching but was a problem _exposed_ by caching, since caching made o1js skip computation of the thing `verify()` had relied on.

kadirchan:
In node (o1js 1.9.1), I faced recursive proving stuck without using any cpu or giving any error on proving phase and only workaround I found is restarting app as I said #1890. On second run programs compile faster and I can run recursive proofs.

mitschabaude:
> In node (o1js 1.9.1), I faced recursive proving stuck without using any cpu or giving any error on proving phase and only workaround I found is restarting app as I said #1890. On second run programs compile faster and I can run recursive proofs.

I'm not sure why it gets stuck, but it could be memory related. My idea for why it works after restart is that by the restart, you're already done compiling and compilation artifacts are used from the cache. This uses much less memory than full compiling, so you have more memory available for proving on the second try

kadirchan:
Tried on 16 and 32 GB ram vps and it stuck on both. In first run docker container uses around 5-6 GB of ram and stuck on proving phase, gives no error and almost 0% ram usage. After restarting containers ram usage reduced to around 4 GB and they able to run proving.

mitschabaude:
> Tried on 16 and 32 GB ram vps and it stuck on both. In first run docker container uses around 5-6 GB of ram and stuck on proving phase, gives no error and almost 0% ram usage. After restarting containers ram usage reduced to around 4 GB and they able to run proving.

ok! no idea what's happening there. but if it's on nodejs it can't be related to this PR, since here we only change the web version

kadirchan:
Maybe it's something that's always been there, I just notice it.

45930 (approved):
I confirmed the fix locally.  The tradeoff with the writable cache case seems fine to me since the current behavior is silently failing in all cases, the 80/20 version is a strict upgrade.

45930:
@boray will follow up with the test cases PR soon!

45930:
(Leaving this unmerged to give Florian an opportunity to veto my approval, but we will target the next o1js release)

PR #1908, opened 2024-11-19
Replace isomorphic-fetch

### LINKED ISSUE #1904: Remove isomorphic-fetch dependency
Isomorphic-fetch is no longer being maintained and is causing a deprecation warning on node v22.11.0 by pulling in punycode.

From the zkapp-cli:
```
> node --trace-deprecation src/bin/index.js project test
? Create an accompanying UI project too? …  (node:2145) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
    at node:punycode:3:9
    at BuiltinModule.compileForInternalLoader (node:internal/bootstrap/realm:399:7)
    at BuiltinModule.compileForPublicLoader (node:internal/bootstrap/realm:338:10)
    at loadBuiltinModule (node:internal/modules/helpers:114:7)
    at Function._load (node:internal/modules/cjs/loader:1100:17)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)
    at Module.require (node:internal/modules/cjs/loader:1340:12)
    at require (node:internal/modules/helpers:141:16)
    at Object.<anonymous> (zkapp-cli/node_modules/whatwg-url/lib/url-state-machine.js:2:18)

> npm ls whatwg-url
zkapp-cli@0.22.0
└─┬ o1js@2.0.0
  └─┬ isomorphic-fetch@3.0.0
    └─┬ node-fetch@2.7.0
      └── whatwg-url@5.0.0
```

When this is done, remove the whatwg-url version override from zkapp-cli's package.json https://github.com/o1-labs/zkapp-cli/pull/707

boray:
Fixes #1904 

This PR replaces [isomorphic fetch](https://github.com/matthew-andrews/isomorphic-fetch) with it's isomorphic alternative [cross-fetch](https://github.com/lquixada/cross-fetch) and updates fetch examples to devnet from berkeley.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "blakejs": "1.2.1",
         "cachedir": "^2.4.0",
-        "isomorphic-fetch": "^3.0.0",
+        "cross-fetch": "^4.0.0",
         "js-sha256": "^0.9.0",
         "reflect-metadata": "^0.1.13",
         "tslib": "^2.3.0"
@@ -23,7 +23,6 @@
         "@influxdata/influxdb-client": "^1.33.2",
         "@noble/hashes": "^1.3.2",
         "@playwright/test": "^1.48.0",
-        "@types/isomorphic-fetch": "^0.0.36",
         "@types/jest": "^27.0.0",
         "@types/node": "^18.14.2",
         "@typescript-eslint/eslint-plugin": "^5.0.0",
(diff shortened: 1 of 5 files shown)


mitschabaude:
Why do we need this instead of using built-in fetch?

ymekuria:
> Why do we need this instead of using built-in fetch?

I agree with @mitschabaude here. We should use the built-in fetch instead of adding a dependancy. We moved to node fetch in the `zkApp-CLI` as well.

boray:
> Why do we need this instead of using built-in fetch?

The main reason is to not drop support for node versions <18.

mitschabaude:
> > Why do we need this instead of using built-in fetch?
> 
> The main reason is to not drop support for node versions <18.

I think we can do us a favor and drop that support!

45930 (commented):
I think we should include a changelog entry for this, but otherwise LGTM!

boray:
> I think we should include a changelog entry for this, but otherwise LGTM!

You're right, we're dropping support for node versions <18 with the last commit, so this needs to be included in the changelog. Also, this breaks backward compatibility. Should we wait for v3 to release this PR?

mitschabaude:
I'd consider it not breaking, and would release it under a patch version, because we have been explicit about not supporting node < 18 before:
* Our [docs state that node >= 18 is required](https://docs.minaprotocol.com/zkapps/writing-a-zkapp/introduction-to-zkapps/install-zkapp-cli#dependencies)
* the `"engine"` field in o1js `package.json` has `"node": ">=18.14.0"`

45930:
I suppose we need to be more explicit about our definition of _breaking_.  I'm inclined to agree with Gregor that we are very explicit already about node 18+ being required.  As we release major versions in the future, we may want to aggressively bump that supported node version so we can continue to make changes like these.

I'm ok to merge this in a minor version.  Even though I see the argument that we're "dropping support", I don't think that we actually had previously "supported" those versions.  They just happened to work.

45930 (approved):

Issue #1872, opened 2024-10-17
Failed to derive correct actions hash


aii23:
This issue was previously identified and fixed in [this pull request](https://github.com/o1-labs/o1js/pull/1784). However, it still appears to persist.

o1js version: 1.8.0

The error can be reproduced using the following script with any of the provided addresses:
```javascript
const network = Mina.Network({
  mina: 'https://api.minascan.io/node/devnet/v1/graphql',
  archive: 'https://api.minascan.io/archive/devnet/v1/graphql',
});
Mina.setActiveInstance(network);

const actions = await Mina.activeInstance.fetchActions(
  PublicKey.fromBase58(
    'B62qrHzs8Sn6rJW3Jkd8xvUkPKb4RBC4xsTgKnBd7KvVWnjhkXV7YKJ'
    // 'B62qmASMoUYbRA2TwbB6gDTcdaS9QxEQYghV67i8oMeqA5tsbfvkJ6P'
    // 'B62qio1AyjVw6tBqYCuEJqDz3ej3qVCyJjcQfTTbt3VgE6MZmtruiMJ'
  )
);
```
The order of actions within a single block remains problematic. We have updated the `o1js` fetch script to attempt different permutations of actions. This approach successfully identifies the correct order of actions, but for blocks with 7 or more actions, it has to check too many permutations.

Additionally, we have not found a logical way to sort the array of actions, and their order seems random for blocks with a broken sequence. The order of actions does not match the order of events, even though they are dispatched within the same function.

45930:
Sorry this is happening again.  Can you share the order that o1js is returning and the order that works when you try to re-order them?

Last time, the fix was to order them by account update id.  Seeing the out-of-order list and the data in it should help us pinpoint the sorting issue.

aii23:
Sure.
'B62qrHzs8Sn6rJW3Jkd8xvUkPKb4RBC4xsTgKnBd7KvVWnjhkXV7YKJ': 
Error for state  11553781833537412187998791170094246504179739906426596650283148421062134031747
Failed at permutation:  [ '40612', '40613', '40614', '40615' ]
Right order:  [ '40613', '40612', '40614', '40615' ]

'B62qmASMoUYbRA2TwbB6gDTcdaS9QxEQYghV67i8oMeqA5tsbfvkJ6P':
This account appears to be unaffected by the bug

[...]

PR #1917, opened 2024-11-22
Sort actions by tx sequence when available

45930:
Relates to https://github.com/o1-labs/o1js/issues/1872

The archive node currently in production does not expose enough information to accurately sort actions and produce an actions hash.

https://github.com/o1-labs/Archive-Node-API/pull/107 is the PR to fix the archive node API, and this is the PR to update the query and sorting in o1js.


src/lib/mina/fetch.ts
```diff
@@ -710,7 +710,15 @@ async function fetchActions(
     graphqlEndpoint,
     networkConfig.archiveFallbackEndpoints
   );
-  if (error) throw Error(error.statusText);
+  if (error) {
+    // retry once without querying transaction info
+    [response, error] = await makeGraphqlRequest<ActionQueryResponse>(
```
45930, line 715: Rather than retry the query, perhaps we can set the archive node API version in the Mina.Network settings somewhere.  Then we could make a different query based on which version we believe we're talking to.

For this specific example, after we get Minascan to update their version, the new query will work in nearly every case, but we ought to prepare for the case that different versions of the API could be running.

45930:
This PR is not to be merged until after Minascan deploys the fixes in https://github.com/o1-labs/Archive-Node-API/pull/107

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixed
 
 - Compiling stuck in the browser for recursive zkprograms https://github.com/o1-labs/o1js/pull/1906
+- Sort order for actions now includes the transaction sequence number and the exact account id sequence
```
Trivo25, line 23: nit: link to PR

src/lib/mina/fetch.unit-test.ts
```diff
@@ -136,96 +137,1181 @@ describe('Fetch', async (t) => {
...
+            hash: '1589729766029695153975344283092689798747741638003354620355672853210932754595',
+          },
+          {
+            actions: [
+              [
+                '17137397755795687855356639427474789131368991089558570411893673365904353943290',
+                '1',
+              ],
+            ],
+            hash: '10964420428484427410756859799314206378989718180435238943573393516522086219419',
```
Trivo25, line 227: given our history with the action sequence, I think we should seriously consider some advanced testing here - similar to the property testing we do in other places of the codebase

src/lib/mina/fetch.unit-test.ts
```diff
@@ -136,96 +137,1181 @@ describe('Fetch', async (t) => {
...
+                '7',
+                '5706376253943510402705579272467768662733130021399469094173930489992680696927',
+                '1',
+                '1',
+              ],
+            ],
+            hash: '3432500319723012006583645985859849032413742507994268064744704124629760504866',
+          },
+        ]);
+      });
```
Trivo25, line 1314: this is very long and makes it very hard to read, can we clean this up a little?

src/lib/mina/fixtures/fetch-actions-response-with-transaction-info.ts
```diff
@@ -0,0 +1,1864 @@
+export { mockFetchActionsResponse };
```
Trivo25, line 1: see what i wrote above; this could save us a few thousand lines here too

src/lib/mina/graphql.ts
```diff
@@ -317,7 +321,8 @@ const getActionsQuery = (
   publicKey: string,
   actionStates: ActionStatesStringified,
   tokenId: string,
-  _filterOptions?: EventActionFilterOptions
+  _filterOptions?: EventActionFilterOptions,
+  retryWithoutTxInfo: boolean = false
```
Trivo25, line 325: nit: we could generalize this into the options parameter above

src/lib/mina/fetch.unit-test.ts
45930: Can you describe what cases you'd like to see?  The existing tests here are simply real payloads from the graphQL node, and the real output of `fetchActions` in a case where the actions are derived successfully.  In particular, it is one of the edge cases discovered by the original reporter of the issue: https://github.com/o1-labs/o1js/issues/1872#issue-2595537724

src/lib/mina/fetch.unit-test.ts
45930: I don't know what you mean by clean it up?  Do you mean move the payload to a JSON file and read it in?

I would push back on removing this test or trying to prune it to its most useful bits.  Since it is a raw input-output from a real edge case, I think it's worth leaving in full.  It's not really meant to be readable, but it will flag regressions.

src/lib/mina/fixtures/fetch-actions-response-with-transaction-info.ts
45930: FWIW, this is the block in which the actions are out of order by the currently-live algorithm.  So if we did trim it, we should at least keep this.

```
      {
        blockInfo: {
          distanceFromMaxBlockHeight: 10037,
        },
        actionState: {
          actionStateOne:
            '11553781833537412187998791170094246504179739906426596650283148421062134031747',
          actionStateTwo:
            '2657110063538905863496230550574434206907731812034404344231638673775478011856',
        },
        actionData: [
          {
            accountUpdateId: '40618',
            data: [
              '7',
              '7',
              '7',
              '7',
              '7',
              '7',
              '15661593045172372274349484602975176160403678593916071611247487060653084577990',
              '0',
              '1',
            ],
            transactionInfo: {
              sequenceNumber: 4,
              zkappAccountUpdateIds: [40618, 6396, 33066],
            },
          },
          {
            accountUpdateId: '40617',
            data: [
              '3',
              '4',
              '5',
              '7',
              '6',
              '2',
              '6453147199940115032546911372596674482366995380739961315278593325733504245402',
              '0',
              '1',
            ],
            transactionInfo: {
              sequenceNumber: 5,
              zkappAccountUpdateIds: [40617, 40372, 40373],
            },
          },
          {
            accountUpdateId: '40619',
            data: [
              '3',
              '4',
              '5',
              '7',
              '6',
              '3',
              '21594276700301685638451290563121567472719745179788000956082286147863718718533',
              '1',
              '1',
            ],
            transactionInfo: {
              sequenceNumber: 6,
              zkappAccountUpdateIds: [40619, 33539, 33540],
            },
          },
          {
            accountUpdateId: '40620',
            data: [
              '3',
              '4',
              '5',
              '7',
              '6',
              '3',
              '21594276700301685638451290563121567472719745179788000956082286147863718718533',
              '1',
              '1',
            ],
            transactionInfo: {
              sequenceNumber: 7,
              zkappAccountUpdateIds: [40620, 33539, 33540],
            },
          },
        ],
      },
      ```

File: src/lib/mina/fetch.unit-test.ts (hunk 4 above)
Comment (Trivo25) [reply src/lib/mina/fetch.unit-test.ts:1314]: I don't want to remove it but I would like to make it more readable (dedicated JSON file for the payloads, for example)

File: src/lib/mina/fixtures/fetch-actions-response-with-transaction-info.ts (hunk 5 above)
Comment (Trivo25) [reply src/lib/mina/fixtures/fetch-actions-response-with-transaction-info.ts:1]: Sorry, I meant the property testing would be benefical here too

File: src/lib/mina/fetch.unit-test.ts (hunk 3 above)
Comment (Trivo25) [reply src/lib/mina/fetch.unit-test.ts:227]: Take a look at this https://github.com/o1-labs/o1js/blob/00eb12df2b35cd7bc6f020570e20c5e28eee8ed5/src/lib/testing/property.ts

File: src/lib/mina/fetch.unit-test.ts (hunk 4 above)
Comment (45930) [reply src/lib/mina/fetch.unit-test.ts:183]: I updated the pattern here, instead of doing a strict `bigObject.equals(anotherObject)`, I just check the hashes, which takes up less room and still inherently tests that the data is in the right order.

File: src/lib/mina/graphql.ts (hunk 6 above)
Comment (45930) [reply src/lib/mina/graphql.ts:325]: `EventActionFilterOptions` directly maps to the graphQL schema, but `retryWithoutTxInfo` is strictly o1js logic, so I think it's ok as is.

I wonder what you think about just skipping this retry logic though.  Minascan has updated their endpoint already, and if anyone else is using a self-hosted archive node API, they will eventually realize they need to upgrade.

I think we need to keep this retry step for a few minor versions at least, just in case Minascan needs to roll back the changes, we would rather o1js continue to work.  But eventually we can just remove this option.

File: src/lib/mina/fetch.ts [hunk 7]
```diff
@@ -757,9 +796,33 @@ export function createActionsList(
         `No action data was found for the account ${publicKey} with the latest action state ${actionState}`
       );
 
-    actionData = actionData.sort((a1, a2) => {
-      return Number(a1.accountUpdateId) < Number(a2.accountUpdateId) ? -1 : 1;
-    });
+    // DEPRECATED: In case the archive node is running an out-of-date version, best guess is to sort by the account update id
+    // As of 2025-01-07, minascan is running a version of the node which supports `sequenceNumber` and `zkappAccountUpdateIds` fields
+    // We could consider removing this fallback since no other nodes are widely used
```
Trivo25, line 801: > We could consider removing this fallback since no other nodes are widely used

I would be careful with that and I don't think we should. App developers are currently incentivized to run their own services for that type of stuff, so we will never know what users are exactly using. Imo we should just keep it in there, it doesn't hurt

Trivo25 (approved):

src/lib/mina/fetch.ts
```diff
@@ -691,10 +691,30 @@ async function fetchEvents(
...
+ * @param accountInfo - An {@link ActionsQueryInputs} containing the public key, and optional query parameters for the actions query
+ * @param graphqlEndpoint - The GraphQL endpoint to fetch from. Defaults to the configured Mina endpoint.
+ *
+ * @returns A promise that resolves to an object containing the final actions hash for the account, and a list of actions
+ * @throws Will throw an error if the GraphQL endpoint is invalid or if the fetch request fails.
+ *
+ * @example
+ * const accountInfo = { publicKey: 'B62qiwmXrWn7Cok5VhhB3KvCwyZ7NHHstFGbiU5n7m8s2RqqNW1p1wF' };
+ * const acitonsList = await fetchAccount(accountInfo);
+ * console.log(acitonsList);
```
boray, line 706:
```suggestion
 * const actionsList = await fetchAccount(accountInfo);
 * console.log(actionsList);
```
nit: typo

boray (approved):

src/lib/mina/fetch.unit-test.ts
mitschabaude: yeah I think the problem with property testing (or any unit testing) here is that the e2e lifecycle of actions doesn't live in this repo -- most of it lives in the archive node & API. so a unit test of correct behavior given arbitrary inputs would make sense there, but not here

src/lib/mina/fetch.ts
45930: > Imo we should just keep it in there, it doesn't hurt

It's a little sloppy because we will retry on _any_ error.  The archive node API responds with `Error: Unknown Error: {}` when this situation is hit.  So we should also improve the error handling on the API side.

But I'll leave it as-is for now.  The most common case will be that that the first request succeeds and this case is never hit.

PR #1919, opened 2024-11-25
Unlimited account update support 

### LINKED ISSUE #1910: Remove account update limit for zeko
Hello,
Currently o1js has been designated for mina and relies on the current mina configuration when generating a proof to know if it will work on Mina, if the transaction is too expensive an error is returned:
Error: The transaction is too expensive, try reducing the number of AccountUpdates that are attached to the transaction.

But on zeko you don't have this kind of limit, if you want to create a proof that exceeds the number of account updates allowed by Mina, you have to modify this file:
[O1js constants](https://github.com/o1-labs/o1js/blob/0ccf4a6fe2d0ffa3800a65006d122c9dd9de5289/src/lib/mina/constants.ts#L19)
and put a bigger value to COST_LIMIT.

It would be nice to have a parameter in o1js to remove this limit when you want to generate a proof for zeko.

youtpout:
Some blockchain like zeko are not limited by account update, to enable transaction creation for these blockchains, I've added a new parameter to the network property `enforceTransactionLimits` as for local blockchain.

Resolve #1910

### DIFF
--- a/src/lib/mina/mina.ts
+++ b/src/lib/mina/mina.ts
@@ -162,7 +162,8 @@ function Network(
       Fetch.setLightnetAccountManagerEndpoint(lightnetAccountManagerEndpoint);
     }
 
-    if (options.enforceTransactionLimits) {
+    if (options.enforceTransactionLimits !== undefined &&
+      typeof options.enforceTransactionLimits === 'boolean') {
       enforceTransactionLimits = options.enforceTransactionLimits;
     }
   } else {
--- a/src/lib/mina/mina.ts
+++ b/src/lib/mina/mina.ts
@@ -104,7 +104,7 @@ function Network(options: {
   mina: string | string[];
   archive?: string | string[];
   lightnetAccountManager?: string;
-  enforceTransactionLimits?: boolean;
+  bypassTransactionLimits?: boolean;
 }): Mina;
 function Network(
   options:
@@ -113,7 +113,7 @@ function Network(
       mina: string | string[];
       archive?: string | string[];
       lightnetAccountManager?: string;
-      enforceTransactionLimits?: boolean;
+      bypassTransactionLimits?: boolean;
     }
     | string
 ): Mina {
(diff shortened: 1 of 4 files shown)


src/lib/mina/mina.ts
```diff
@@ -158,6 +161,11 @@ function Network(
       lightnetAccountManagerEndpoint = options.lightnetAccountManager;
       Fetch.setLightnetAccountManagerEndpoint(lightnetAccountManagerEndpoint);
     }
+
+    if (options.enforceTransactionLimits !== undefined &&
```
45930, line 165: I think it would be cleaner to name the option `bypassTransactionLimits`.  That way `undefined` and `false` both resolve to the default behavior.  Then I think this could be 

`enforceTransactionLimits = options.bypassTransactionLimits ? false : true;`

45930 (commented):
Thanks for the PR!

I will do a more thorough review before approving, but I think this looks good.  The one consideration is that it doesn't allow for a dev to set their own custom transaction limits.  It's only an on/off.  Does zeko not have transaction limits at all?

src/lib/mina/mina.ts
youtpout: I tried to reuse the same name than LocalBlockchain :
`async function LocalBlockchain({
  proofsEnabled = true,
  enforceTransactionLimits = true,
} = {}) {`

https://github.com/o1-labs/o1js/blob/a0f0f349811fd812b27cb18bac74e8edecd48f9a/src/lib/mina/local-blockchain.ts#L74

src/lib/mina/transaction.test.ts
45930, line 1: How did you actually run this test?  It doesn't import from jest or from node:test.  Our pattern for unit tests is to name the file `*.unit-test.ts` and use node:test as the testing library.  I don't mind porting this over since our open source guidelines are not very good.

But on a bigger level, this isn't really a unit test of `transaction`, but rather `Mina.Network`.  So maybe this file ought to be called `mina.network.unit-test.ts`.  Does that make sense to you?

45930 (changes requested):

src/lib/mina/transaction.test.ts
youtpout: I renamed the test file and rename the parameter byPassTransactionLimits

shimkiv:
Hey @youtpout, can you please merge the `main` branch changes into your branch.

youtpout:
> Hey @youtpout, can you please merge the `main` branch changes into your branch.

Done

45930:
@shimkiv, do you know what's wrong with the benchmark o1js job?  It looks like it build o1js then immediately fails.

shimkiv:
@45930 it is related to self-hosted runners, repo CI config and external contributors. Back then we decided that it is not a big issue and didn't provide a fix (if it exists).

45930 (approved):

shimkiv:
@45930 for example, this particular job relies on repo secrets that external contributors simply don't have, hence the job is failing.

45930:
Thanks for explaining @shimkiv 

@youtpout thanks again for the contribution!

PR #1848, opened 2024-10-02
Dynamic arrays in o1js

querolita:
Addressing https://github.com/o1-labs/o1js/issues/1086, based on https://github.com/o1-labs/o1js/issues/633

~Making this public as WIP to receive early feedback~

Closes https://github.com/o1-labs/o1js/issues/2102

PR #1921, opened 2024-11-25
[easy] Added `zip` and `pad` functionalities for arrays

querolita:
Introduced in https://github.com/o1-labs/o1js/pull/1848, creating a separate PR to facilitate the change earlier.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Support secp256r1 in elliptic curve and ECDSA gadgets https://github.com/o1-labs/o1js/pull/1885
 
+- Provide `zip` and `pad` for the array utility https://github.com/o1-labs/o1js/pull/1921
+
 ### Fixed
 
 - Witness generation error in `Gadgets.arrayGet()` when accessing out-of-bounds indices https://github.com/o1-labs/o1js/pull/1886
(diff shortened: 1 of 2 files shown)


src/lib/util/arrays.ts
```diff
@@ -12,3 +15,16 @@ function chunk<T>(array: T[], size: number): T[][] {
 function chunkString(str: string, size: number): string[] {
   return chunk([...str], size).map((c) => c.join(''));
 }
+
+function zip<T, S>(a: T[], b: S[]) {
+  assert(a.length === b.length, 'zip(): arrays of unequal length');
```
Geometer1729, line 20: AFAIK zip is usually implemented to take the minimum length when the lists are unequal. I think we should default to that unless there's a reason that wouldn't make sense here.

Geometer1729 (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 - Compiling stuck in the browser for recursive zkprograms https://github.com/o1-labs/o1js/pull/1906
 
+### Added
+
+- Provide `zip` and `pad` for the array utility https://github.com/o1-labs/o1js/pull/1921
```
Trivo25, line 26: nit: we actually dont expose the array utilities to the outside world directly, we use them as proxy most of the time - so these functions are just internal for now. I would remove the changelog entry but still merge the additions!

src/lib/util/arrays.ts
```diff
@@ -12,3 +15,19 @@ function chunk<T>(array: T[], size: number): T[][] {
 function chunkString(str: string, size: number): string[] {
   return chunk([...str], size).map((c) => c.join(''));
 }
+
+function zip<T, S>(a: T[], b: S[]) {
+  assert(
+    a.length <= b.length,
+    'zip(): first array must be at least as long as the second array'
```
boray, line 22: message is not consistent with assertion.

querolita: 🤦🏻

PR #1922, opened 2024-11-25
[easy] Add assertNone method for options

querolita:
Added in https://github.com/o1-labs/o1js/pull/1848, creating a separate PR to facilitate the change earlier

### DIFF
--- a/src/lib/provable/option.ts
+++ b/src/lib/provable/option.ts
@@ -9,7 +9,7 @@ export { Option, OptionOrValue };
 
 type Option<T, V = any> = { isSome: Bool; value: T } & {
   assertSome(message?: string): T;
-  assertNone(): void;
+  assertNone(message?: string): void;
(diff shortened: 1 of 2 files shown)


Geometer1729 (approved):
LGTM

src/lib/provable/option.ts
```diff
@@ -104,6 +105,10 @@ function Option<A extends ProvableType>(
       return this.value;
     }
 
+    assertNone(): void {
```
Trivo25, line 108: nit: we could pass in an optional `message?: string` argument to the function and pass it to `this.isSome.assertFalse(message)` so that a user can specify a optional error message on faliure (see `assertSome` above)

Trivo25 (approved):

ymekuria (approved):

hattyhattington17 (approved):

PR #1928, opened 2024-11-27
Expose proof base64 conversion

mitschabaude:
low-level method exposed to have more flexibility for how to serialize proofs

I need this in https://github.com/zksecurity/mina-credentials/pull/82 to get a synchronous version of `Proof.fromJSON()`

### DIFF
--- a/src/lib/proof-system/proof.ts
+++ b/src/lib/proof-system/proof.ts
@@ -92,6 +92,13 @@ class ProofBase<Input = any, Output = any> {
   publicFields() {
     return (this.constructor as typeof ProofBase).publicFields(this);
   }
+
+  static _proofFromBase64(proofString: string, maxProofsVerified: 0 | 1 | 2) {
+    return Pickles.proofOfBase64(proofString, maxProofsVerified)[1];
+  }
+  static _proofToBase64(proof: Pickles.Proof, maxProofsVerified: 0 | 1 | 2) {
+    return Pickles.proofToBase64([maxProofsVerified, proof]);
+  }
 }
 
 class Proof<Input, Output> extends ProofBase<Input, Output> {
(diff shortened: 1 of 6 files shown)


Geometer1729 (approved):
LGTM

mitschabaude:
that changelog CI job is effective :D

src/lib/proof-system/proof.ts
```diff
@@ -92,6 +92,13 @@ class ProofBase<Input = any, Output = any> {
   publicFields() {
     return (this.constructor as typeof ProofBase).publicFields(this);
   }
+
+  static _proofFromBase64(proofString: string, maxProofsVerified: 0 | 1 | 2) {
```
45930, line 96: In the existing `fromJSON` implementation, we await `initializeBindings`.  What are the implications of skipping that step before entering pickles?

Shigoto-dev19 (commented):
looks good!

src/lib/proof-system/proof.ts
```diff
@@ -92,6 +92,13 @@ class ProofBase<Input = any, Output = any> {
...
+    return Pickles.proofOfBase64(proofString, maxProofsVerified)[1];
```
45930, line 97:
```suggestion
    if (!areBindingsInitialized) {
      throw Error(
        'Pickles bindings must be initialized before calling this method'
      );
    }
    return Pickles.proofOfBase64(proofString, maxProofsVerified)[1];
```

What if we did something like this, coupled with update the export in snarky.js

```
export {
  Snarky,
  Ledger,
  Pickles,
  Test,
  withThreadPool,
  wasm,
  initializeBindings,
  isInitialized as areBindingsInitialized, <-- add this
};
```

That way no one accidentally gets into trouble with this.

mitschabaude:
@Trivo25 I think this is ready for code owner approval

Shigoto-dev19 (approved):

Trivo25 (approved):

PR #1930, opened 2024-11-27
2024 11 automatically tag releases

45930:
After releasing a new version to NPM, tag that version on githb


Trivo25 (approved):
this is probably obsolete then and we can remove it 
https://github.com/o1-labs/o1js/blob/a0f0f349811fd812b27cb18bac74e8edecd48f9a/.github/workflows/release.yml#L63-L66

PR #1931, opened 2024-11-28
Introduce recursive proofs from within ZkPrograms

mitschabaude:
This PR enables to "declare" a recursively verified proof from _within a ZkProgram method_. Thus, recursive proofs no longer need to be inputs to the method, and therefore can be abstracted away much more easily.

bindings: https://github.com/o1-labs/o1js-bindings/pull/318

The new low-level way of introducing a proof that hasn't been passed as method argument is as follows:
```ts
let proof = await Provable.witnessAsync(MyProof, async () => { /* generate and return proof */  });
proof.declare();
// ... use the proof as if it had been a method argument ...
```

## New Recursion API

_see UPDATE below for final version of this API_

On top of this, I added an even more pleasant API, which you can use in the typical case where the ZkProgram that produces the inner proof is in scope. For this, assume the proof in question is supposed to come from a method `innerProgram.run()`. Then, the entire logic above is now encapsulated inside the new property `innerProgram.proveRecursively.run()`:

```ts
// inside outer zkprogram method
let output = await innerProgram.proveRecursively.run(...inputs);
```
note: this expects the inputs (public and private) as arguments and returns the public output -- which is the natural way to think of calling a method.

Internally, we make sure that the inner proof's _public_ input matches what the outer circuit passed in.

**Caveats**: of course we can't make any statement about the private inputs, because the inner proof inherently doesn't reveal those. so the private inputs must be understood only as hints to the recursive proving. Similarly, the inner proof doesn't guarantee the _particular method_ that was called on the inner program: as usual with recursive proofs, all of the methods can produce it.

An e2e example, that I added to CI and also manually ran in the browser, is at `src/tests/inductive-proofs-internal.ts`

**UPDATE**: after discussions, the API was changed to live separately from ZkProgram:

```ts
let inner = ZkProgram({ ..., methods: { run: ... });
let innerRecursive = Experimental.Recursive(program);

// inside another -- or even the same -- program:
let output = await innerRecursive.run(...inputs);
```

### DIFF
--- a/src/lib/mina/zkapp.ts
+++ b/src/lib/mina/zkapp.ts
@@ -46,7 +46,6 @@ import {
   compileProgram,
   computeMaxProofsVerified,
   Empty,
-  getPreviousProofsForProver,
   MethodInterface,
   sortMethodArguments,
   VerificationKey,
@@ -337,8 +336,6 @@ function wrapMethod(
               {
                 methodName: methodIntf.methodName,
                 args: clonedArgs,
-                // proofs actually don't have to be cloned
-                previousProofs: getPreviousProofsForProver(actualArgs),
                 ZkappClass,
                 memoized,
                 blindingValue,
@@ -429,7 +426,6 @@ function wrapMethod(
             {
               methodName: methodIntf.methodName,
               args: constantArgs,
-              previousProofs: getPreviousProofsForProver(constantArgs),
               ZkappClass,
               memoized,
               blindingValue: constantBlindingValue,
--- a/src/lib/mina/zkapp.ts
+++ b/src/lib/mina/zkapp.ts
@@ -44,13 +44,14 @@ import {
 import {
   analyzeMethod,
   compileProgram,
+  computeMaxProofsVerified,
   Empty,
   getPreviousProofsForProver,
   MethodInterface,
   sortMethodArguments,
   VerificationKey,
 } from '../proof-system/zkprogram.js';
-import { Proof } from '../proof-system/proof.js';
+import { Proof, ProofClass } from '../proof-system/proof.js';
 import { PublicKey } from '../provable/crypto/signature.js';
 import {
   InternalStateType,
@@ -154,11 +155,6 @@ function method<K extends string, T extends SmartContract>(
   // FIXME: overriding a method implies pushing a separate method entry here, yielding two entries with the same name
   // this should only be changed once we no longer share the _methods array with the parent class (otherwise a subclass declaration messes up the parent class)
   ZkappClass._methods.push(methodEntry);
-  ZkappClass._maxProofsVerified ??= 0;
-  ZkappClass._maxProofsVerified = Math.max(
-    ZkappClass._maxProofsVerified,
-    methodEntry.numberOfProofs
-  ) as 0 | 1 | 2;
   let func = descriptor.value as AsyncFunction;
   descriptor.value = wrapMethod(func, ZkappClass, internalMethodEntry);
 }
@@ -593,10 +589,10 @@ class SmartContract extends SmartContractBase {
       rows: number;
       digest: string;
       gates: Gate[];
+      proofs: ProofClass[];
     }
   >; // keyed by method name
   static _provers?: Pickles.Prover[];
-  static _maxProofsVerified?: 0 | 1 | 2;
   static _verificationKey?: { data: string; hash: Field };
 
   /**
@@ -644,6 +640,7 @@ class SmartContract extends SmartContractBase {
     forceRecompile = false,
   } = {}) {
     let methodIntfs = this._methods ?? [];
+    let methodKeys = methodIntfs.map(({ methodName }) => methodName);
     let methods = methodIntfs.map(({ methodName }) => {
       return async (
         publicInput: unknown,
@@ -657,13 +654,15 @@ class SmartContract extends SmartContractBase {
     });
     // run methods once to get information that we need already at compile time
     let methodsMeta = await this.analyzeMethods();
-    let gates = methodIntfs.map((intf) => methodsMeta[intf.methodName].gates);
+    let gates = methodKeys.map((k) => methodsMeta[k].gates);
+    let proofs = methodKeys.map((k) => methodsMeta[k].proofs);
     let { verificationKey, provers, verify } = await compileProgram({
       publicInputType: ZkappPublicInput,
       publicOutputType: Empty,
       methodIntfs,
       methods,
       gates,
+      proofs,
       proofSystemTag: this,
       cache,
       forceRecompile,
@@ -687,6 +686,17 @@ class SmartContract extends SmartContractBase {
       Object.values(methodData).map((d) => Field(BigInt('0x' + d.digest)))
     );
     return hash.toBigInt().toString(16);
+  }
+
+  /**
+   * The maximum number of proofs that are verified by any of the zkApp methods.
+   * This is an internal parameter needed by the proof system.
+   */
+  static async getMaxProofsVerified() {
+    let methodData = await this.analyzeMethods();
+    return computeMaxProofsVerified(
+      Object.values(methodData).map((d) => d.proofs.length)
+    );
   }
 
   /**
@@ -1189,7 +1199,7 @@ super.init();
       try {
         for (let methodIntf of methodIntfs) {
           let accountUpdate: AccountUpdate;
-          let { rows, digest, gates, summary } = await analyzeMethod(
+          let { rows, digest, gates, summary, proofs } = await analyzeMethod(
             ZkappPublicInput,
             methodIntf,
             async (publicInput, publicKey, tokenId, ...args) => {
@@ -1207,6 +1217,7 @@ super.init();
             rows,
             digest,
             gates,
+            proofs,
           };
           if (printSummary) console.log(methodIntf.methodName, summary());
         }
--- a/src/lib/proof-system/workers.ts
+++ b/src/lib/proof-system/workers.ts
@@ -1,4 +1,4 @@
-export { workers, setNumberOfWorkers };
+export { workers, setNumberOfWorkers, WithThreadPool };
 
 const workers = {
   numWorkers: undefined as number | undefined,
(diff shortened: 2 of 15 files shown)


mitschabaude:
@Trivo25 this is ready for review.

In a follow-up PR, I also want to add a conditional version: `proveRecursivelyIf()`, to cover the common use case of implementing base case and recursive case within the same method.

This, in turn, would enable us to simplify the offchain state API, by handling the recursive settling proof internally. @45930 do you think that should be done?

src/lib/proof-system/zkprogram.ts
```diff
@@ -471,6 +466,50 @@ function ZkProgram<
     return compileOutput.verify(statement, proof.proof);
   }
 
+  let regularRecursiveProvers = mapObject(regularProvers, (prover, key) => {
+    return async function proveRecursively_(
```
mitschabaude, line 470: this is the implementation of the new recursive proving API

src/lib/proof-system/zkprogram.ts
```diff
@@ -803,44 +845,53 @@ function picklesRuleFromFunction(
...
+    let proofs: DeclaredProof[];
+
+    try {
+      if (publicInputType === Undefined || publicInputType === Void) {
+        result = (await func(...finalArgs)) as any;
+      } else {
+        let input = fromFieldVars(publicInputType, publicInput);
+        result = (await func(input, ...finalArgs)) as any;
+      }
+      proofs = ZkProgramContext.getDeclaredProofs();
```
mitschabaude, line 879: instead of collecting all proofs inside this function as before, we get them from the new global context `ZkProgramContext`, which is also written to by `Proof.declare()`

src/lib/proof-system/zkprogram.ts
```diff
@@ -513,7 +555,7 @@ function ZkProgram<
     get: () => doProving,
   });
 
-  return program as any;
+  return program;
```
mitschabaude, line 558: I worked a bit on types in this file to reduce the amount of `as any`, in particular it's nice to have this final sanity check that the returned program matches the output type. for that, I had to put a boolean into `program.proofsEnabled` which we overwrite later with `Object.defineProperty()`

src/lib/proof-system/workers.ts
```diff
@@ -15,3 +15,68 @@ const workers = {
...
   workers.numWorkers = numWorkers;
 };
+
+type ThreadPoolState =
+  | { type: 'none' }
+  | { type: 'initializing'; initPromise: Promise<void> }
+  | { type: 'running' }
+  | { type: 'exiting'; exitPromise: Promise<void> };
+
+function WithThreadPool({
```
mitschabaude, line 25: the added logic here comes from the bindings, see https://github.com/o1-labs/o1js-bindings/pull/318

mitschabaude:
⚠️ **Update: This is not ready for review yet** ⚠️ 

I forgot to declare the internally declared proofs to pickles in the top level of `picklesRuleFromFunction()`. so, Pickles doesn't currently add any proof verification logic, and the "recursive proofs" work even if we use dummy proofs.

To fix this, I'll need to refactor a bit to find out the proofs prior to calling `pickledRuleFromFunction()`. Luckily this should be possible without extra overhead, because we're already calling `analyzeMethods()` up front to determine the gates, and in that step we'll also be able to count the declared proofs.

Because this caught me by surprise and seems like a dangerous footgun, I'm going to add runtime check that the number of dynamically returned proofs to be verified matches the number that was determined at compile time.

mitschabaude:
Ok it works now, ready for review!

mitschabaude (commented):
a few more comments

src/lib/proof-system/zkprogram-context.ts
```diff
@@ -0,0 +1,29 @@
+import { Context } from '../util/global-context.js';
+import type { Subclass } from '../util/types.js';
+import type { ProofBase } from './proof.js';
+
+export { ZkProgramContext, DeclaredProof };
+
+type DeclaredProof = {
+  Proof: Subclass<typeof ProofBase<any, any>>;
```
Trivo25, line 8: can we name this something like `ProofClass´ and `proofInstance` instead? This way its more clear which is what

src/lib/proof-system/zkprogram.ts
```diff
@@ -527,9 +577,12 @@ function ZkProgram<
       rawMethods: Object.fromEntries(
         methodKeys.map((key) => [key, methods[key].method])
       ) as any,
+      proofsEnabled: doProving,
       setProofsEnabled(proofsEnabled: boolean) {
         doProving = proofsEnabled;
       },
+
+      proveRecursively,
```
Trivo25, line 585: I will let the others be the judge if we want to express the API this way or if this is too confusing. I wonder if we can/should type this method in a certain way so its only available when its called in the correct context (within another program), and not in the plain native verison to avoid confusion

src/lib/proof-system/zkprogram-context.ts
mitschabaude: sure!

src/lib/proof-system/zkprogram.ts
mitschabaude: one note: `proveRecursively` doesn't return you a proof (just the proof output), and it's therefore hard to confuse with the plain proving API.

like, if you need a proof, you'll figure out quickly that this API doesn't give you what you want.

(if you don't care about a proof, then it's fine to use this API outside a circuit, because it does run the method)

src/lib/proof-system/zkprogram.ts
```diff
@@ -351,61 +376,49 @@ function ZkProgram<
     }
   }
 
-  function toProver<K extends keyof Methods & string>(
+  type RegularProver<K extends MethodKey> = (
```
45930, line 379: All the other types are at the bottom of this file.  Perhaps we ought to break out some of the type declarations since this file is so massive now.

What makes this type `Regular`?  Is it in opposition to `RecursiveProver`?  The new types could use documentary comments I think.

45930 (commented):
I'm not sure how I feel about the proposed API change.  It is quite clean, but also hides a lot of detail, and perhaps makes it easier to write programs with accidental or intentional exploits.  It seems to move the opposite direction of our new API design, which is more explicit and verbose.  Since it's optional for developers, I don't feel strongly that we _shouldn't_ release it.

I also don't love new APIs that are out ahead of the documentation.  That's the state of where we are now, so I wouldn't block this release on that.  But I intend to get to the point where I would block such a release for that reason in the future.

src/lib/proof-system/zkprogram.ts
45930: But for the example, I wouldn't understand how to use this.  `.recursiveProvers` seems slightly better, at least because it sounds like it may return an object.  It still sounds like `.recursiveProvers.baseCase()` will return a proof, so potentially `.recursiveProverFunctions` is the most natural sounding.

I think there's an asterisk to this which is that zkprogram is still a very opaque API that people don't have an intuitive sense for, so I think we can only be so intuitive with the naming anyways.

src/lib/proof-system/zkprogram.ts
mitschabaude: this type is not at the bottom of the file since it depends on the generic type params!

it's a `RegularProver` as opposed to the normal `Prover`, because it expects a public input argument, even if the public input type is `undefined`. In the final `Prover`s that are exposed on the program, we add some additional wrapping so that users don't have to pass in an `undefined` input, which mimics the way the methods themselves are also defined.

regular = method signature doesn't depend on the runtime publicInputType

for the internal logic, the "regular" prover signature is much easier to deal with, that's why I separated out the wrapping to remove the public input argument into a different step, so that I can use the regular prover for the new _recursive_ prover

src/lib/mina/zkapp.ts
```diff
@@ -657,13 +650,15 @@ class SmartContract extends SmartContractBase {
     });
     // run methods once to get information that we need already at compile time
     let methodsMeta = await this.analyzeMethods();
-    let gates = methodIntfs.map((intf) => methodsMeta[intf.methodName].gates);
+    let gates = methodKeys.map((k) => methodsMeta[k].gates);
+    let proofs = methodKeys.map((k) => methodsMeta[k].proofs);
```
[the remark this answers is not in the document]
mitschabaude: no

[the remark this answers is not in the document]
mitschabaude: proofs within a method are in the order of the `privateInputs`, same as before

src/lib/proof-system/zkprogram.ts
45930: > this type is not at the bottom of the file since it depends on the generic type params!

Ah right!

> it's a RegularProver as opposed to the normal Prover, because it expects a public input argument, even if the public input type is undefined

Could you add a comment to that effect in the code so in the future a reader will understand that at a glance?

mitschabaude:
@45930 @Trivo25 If you're really scared of introducing a different API out of the blue -- it would probably also work as a wrapper around ZkProgram. Something like this should work and would produce less of a irreversible change:

```ts
let program = ZkProgram(...);

let programRecursive = Experimental.Recursive(program);

// inside other program
let output = programRecursive.run(...inputs);
```

Personally think the API is awesome as is and would be easier to use and find on ZkProgram directly, but I don't want it to block the main addition of this PR, internal proof declaring, which is much more important.

mitschabaude:
Note: I pushed one more commit https://github.com/o1-labs/o1js/pull/1931/commits/f591f02816c0e29bcc6a5caa7db951329a6dccae which in emerged as a necessary fix in https://github.com/zksecurity/mina-credentials/pull/86 when running internal proofs with `proofsEnabled = false`.

Also, by experience on https://github.com/zksecurity/mina-credentials/pull/86, I can now say that proof declaring works well in a complicated, real-world circuit.
Also worth mentioning, there are scenarios where I couldn't use `.proveRecursively` but had to use manual proof witnessing: when I wanted to pick the particular recursive method based on the remaining size of proof inputs

Trivo25:
I think we should go for the wrapper and potentially move the functionality into zkProgram natively some time in the future

mitschabaude:
cool, I'll push a refactor sometime soon

mitschabaude:
@Trivo25 @45930 I moved `Recursive` outside `ZkProgram`!

mitschabaude:
All I Want For Christmas Is getting my PR merged 🎄 🎅🏻 🎄 🎁 🎄

45930 (approved):

Issue #1248, opened 2023-11-16
Move `Provable<T>` methods to a separate `.provable` namespace, on each provable type


mitschabaude:
**Why**: `Provable<T>` interface methods, like `.toFields()` and `.fromFields()`, are supposed to be an internal thing, used in circuit creation plumbing, or by advanced users to create general-purpose methods like `Provable.if()`. They work in very specific ways and circumstances. But, because of our interfaces right now, these functions have to be static methods on all our provable type classes, which are highly visible to users. So they are treated as public API. Sometimes they are used as conversion helpers, i.e.
* trying `PrivateKey.fromFields([x])` to convert a field element to a private key (which makes sense as a use case, but is not what `fromFields()` does and will create a `PrivateKey` with the wrong number of internal fields, that will cause breakage somewhere else)
* using `UInt64.fromFields([x])` to create a `UInt64` (dangerous, as it bypasses the 64-bit range constraint we would usually add when instantiating a `UInt64`)

I think we should move those internal methods to less visible namespaces where they are clearly marked as serving a special purpose: `PrivateKey.provable.{toFields, fromFields, ...}`, `UInt64.provable`, etc

**How**: 
* Step 1: Create a type `Provable<T> | { provable: Provable<T>}`, and accept that as input in all APIs that right now expect a `Provable<T>`. Same for variants like `FlexibleProvable<T>` and `ProvablePure<T>`
* Step 2: Deprecate provable methods on their current locations (hand-written provable classes, Struct, etc), and move them to `.provable`

**To discuss:** "provable" is quite an overloaded term here. We often use it to refer to code that becomes part of a proof. Here we use it for helper methods to define a type that can be used in provable code -- very different. What could be a less ambiguous namespace than `.provable`?

mitschabaude:
I think we should definitely still do this one, especially now that we got `.provable` types to work just as well as old ones everywhere

PR #1934, opened 2024-11-29
Allow JS values as ZkProgram inputs

mitschabaude:
We can already pass in normal js values when witnessing provable types. This PR enables passing normal values into ZkProgram provers (both normal and recursive provers).

Also improves the supported JS values for a few important types like `Signature` and `UIntX`


mitschabaude:
Ready for review!

src/lib/provable/int.ts
```diff
@@ -1842,6 +1842,15 @@ class UInt8 extends Struct({
     return new UInt8(x);
   }
 
+  static fromValue(
```
Shigoto-dev19, line 1845: In the case of `fromField(ff)`, isn't this similar to `new UInt8(ff.value)` or `UInt8.from(ff)`, which are not recommended to use instead of `UInt8.Unsafe.fromField(ff)` because they don't clearly indicate the potential unsafety of wrapping a Field into a UInt8, making the code less readable.

Shigoto-dev19 (commented):
Great work @mitschabaude! However, I’m concerned that this PR might make `ZkProgram` code less readable :thinking:

src/lib/provable/int.ts
mitschabaude: yeah absolutely, and this method isn't really intended for developer to use (that's what `.from()` was designed for).

`fromValue()` is part of the `Provable<T>` API and what we use internally to convert JS values into the type -- it's needed here for that reason.

Note that this PR doesn't add `UInt8.fromValue()` -- it was already there, we just override it to also handle `number` inputs (which is not unsafe), because `number` is the most natural type that you'd want to convert a UInt8 from

mitschabaude: there are two things to improve IMO, both of which aren't related to this PR though:
1. internal `Provable` methods should be on a separate namespace `.provable` instead of on the class itself, to mark then more clearly as not-developer facing https://github.com/o1-labs/o1js/issues/1248
2. `UInt8` shouldn't be a `Struct`, because while `Struct` is great for easily _composing_ types, it's not great for defining the base primitives. In this case, the problem is that the `fromValue()` type of `UInt8` is inferred from the `Struct({ value: Field })` definition to include `{ value: Field }`, which really should be `UInt8` (but that doesn't work because the Struct definition doesn't know anything about `UInt8`)

mitschabaude: >  isn't this similar to new UInt8(ff.value) or UInt8.from(ff), which are not recommended to use

side note: `UInt8.from(field)` is perfectly fine, because it adds a range-check

Shigoto-dev19: Thanks for the insight! 

> UInt8 shouldn't be a Struct, because while Struct is great for easily composing types, it's not great for defining the base primitives.

I recall encountering an issue a while ago where I couldn't use `Provable.witness` for a `UInt32` with `proofsEnabled=true`. Is this related to the same issue, or is it unrelated? I’ll need to revisit and check this again!

mitschabaude: `UInt32` is not implemented with `Struct` (but with the older `CircuitValue`) so that might be a different issue!

in the case of `UInt8`, one thing that bites me now and then is that `UInt8 satisfies Provable<UInt8>` is not true (and similar for other Structs), because e.g. the return type of `UInt8.fromFields()` is `{ value: Field }` and not `UInt8`

image

mitschabaude: While if you look at newer types like `Bytes`, you'll see that because of the way we made `.provable` aware of the class itself, this works:

```ts
Bytes(32) satisfies ProvableType<Bytes>;
```

mitschabaude:
with #1931 and #1933 approved, this is ripe for review!

45930:
@mitschabaude FYI we are discussing this change but with the holidays and some other urgent work going on at the moment, not everyone has had a chance to review.

Trivo25 (approved):
Thanks!

Issue #1935, opened 2024-11-29
offChain State does not update internal state


kadirchan:
For a few weeks I have been doing on chain testing in my application that uses offChain state api. I use services in my app that send and settle the proofs I receive from users to Mina and this services work consistently. 

However, after updating on offchain state, I realized that the api was not working correctly. When I tried to update again it after the updates were done, I was getting a root mismatch error and the only way to solve it was to restart the service.

I thought the problem might be that it was not settling and tested for that. After sending the update, I waited for it to settle completely, but the problem was the same. Then I console logged the states in some places on offchain-state.js and got the following results.

```sh
19127837625513839750015801518488884205754193410383643610887852957945537699566
first check false second check false
action state 18571608787403596221185729134339335640005846451227827507499404664500316550112
merkle map 4822366981302605685996191473523237511494367078661562577361822805082477390442
map root 4822366981302605685996191473523237511494367078661562577361822805082477390442 state root 4822366981302605685996191473523237511494367078661562577361822805082477390442
first check true second check true
returning internal
Current session 2009824360
2009824360  ->  2115950408
Generating proof
Transaction sent

## I waited for update and settlement here ##

19127837625513839750015801518488884205754193410383643610887852957945537699566
first check true second check true
returning internal
map root 4822366981302605685996191473523237511494367078661562577361822805082477390442 state root 19818629034789886366259936965517333313478084681883682142698168018525779052913
Error: root mismatch
Field.assertEquals(): 4822366981302605685996191473523237511494367078661562577361822805082477390442 != 19818629034789886366259936965517333313478084681883682142698168018525779052913
```

in the part
```ts
    let map = await Provable.witnessAsync(
      IndexedMerkleMapN,
      async () => (await merkleMaps()).merkleMap
    );
```
returns old map state and this causes root mismatch error, also I cannot figure out how to update this internal merkleMap.

the condition inside 
```ts
       if (
        internal.merkleMap.root.toString() !== emptyMerkleMapRoot.toString() ||
        internal.valueMap.size > 0
      )
```

gives true || true despite of actual tree is updated.

mitschabaude:
@45930 we might want a method to explicitly fetch and build the latest merkle map

45930:
@kadirchan do you want to try a PR that exposes the behavior you want?

I think we need a public function that fetches the latest state and sets the internal state on the offchain state instance.

```
let { merkleMap, valueMap } = await fetchMerkleMap(
    height,
    internal.contract,
[...]

PR #1936, opened 2024-11-30
remove condition temporarily

kadirchan:
I opened it to use this version with pkg.pr.new in my project. Tried to workaround #1935


Trivo25:
fyi pkg currently only works for commits that are merged into main

PR #1942, opened 2024-12-09
Use proof-systems master

dannywillems:
See https://github.com/MinaProtocol/mina/pull/16405 for additional context.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Method for optional types to assert none https://github.com/o1-labs/o1js/pull/1922
 - Increased maximum supported amount of methods in a `SmartContract` or `ZkProgram` to 30. https://github.com/o1-labs/o1js/pull/1918
 - Expose low-level conversion methods `Proof.{_proofToBase64,_proofFromBase64}` https://github.com/o1-labs/o1js/pull/1928
+- Improve usage of branches of proof-systems. Mina uses only the master branch
+  from now. https://github.com/o1-labs/o1js/pull/1942
 
 ### Fixed
 
--- a/README-dev.md
+++ b/README-dev.md
@@ -1,13 +1,20 @@
 # o1js README-dev
 
-o1js is a TypeScript framework designed for zk-SNARKs and zkApps on the Mina blockchain.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -25,6 +25,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Method for optional types to assert none https://github.com/o1-labs/o1js/pull/1922
 - Increased maximum supported amount of methods in a `SmartContract` or `ZkProgram` to 30. https://github.com/o1-labs/o1js/pull/1918
 - Expose low-level conversion methods `Proof.{_proofToBase64,_proofFromBase64}` https://github.com/o1-labs/o1js/pull/1928
+- Improve usage of branches of proof-systems. Mina uses only the master branch
+  from now. https://github.com/o1-labs/o1js/pull/1942
```
45930, line 29: I don't think we need this for a markdown-only change

README-dev.md
```diff
@@ -1,22 +1,31 @@
 # o1js README-dev
 
-o1js is a TypeScript framework designed for zk-SNARKs and zkApps on the Mina blockchain.
+o1js is a TypeScript framework designed for zk-SNARKs and zkApps on the Mina
```
45930, line 3: Are these line-length diffs intentional, or just your local IDE settings?

45930 (commented):
I'm not sure we need this PR at all if there's no functional changes.  Whoever next updates the bindings will include this history in their branch right?

README-dev.md
dannywillems: These are intentional, yes.
I added these changes in two independent commits that can be removed if there is a strong opinion on not enforcing a limit of 80 characters.

It is quite commonly accepted to keep 80 characters as a limit, not only for historical reasons, but also because it is easier to go through files in modern editors too when we have multiple buffers open on the same screen.

CHANGELOG.md
dannywillems: There is also an update of the bindings.

README-dev.md
dannywillems: (but also my editor is configured to make it relatively automatic when I open a file)

dannywillems:
Moving back in draft. It seems I did not update the bindings correctly.

dannywillems:
Closing

PR #1946, opened 2024-12-11
Configure GH before creating a tag

45930:
Fixes: https://github.com/o1-labs/o1js/actions/runs/12272522986/job/34242446940


45930:
@Trivo25 I think this is the fix.  I don't know how to confirm that this will work without waiting for it to rerun.  Do you?

Trivo25 (approved):

PR #1949, opened 2024-12-12
Twisted Edwards curves operations

querolita:
This PR is the twisted counterpart of the operations found in `/src/lib/provable/gadgets/elliptic-curve.ts`. It is a necessary step to support EdDSA, which uses the twisted curve Ed25519. Related to https://github.com/o1-labs/o1js-bindings/pull/317.


src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,462 @@
...
+  add,
+  double,
+  negate,
+  assertOnCurve,
+};
+
+function add(
+  p1: Point,
+  p2: Point,
+  Curve: { modulus: bigint; a: bigint; d: bigint }
```
dannywillems, line 35: Is `modulus` the order of the curve? or the modulus of the base field?

dannywillems: the order of the base foreign field I guess, based on the code below.

querolita: the base field yes

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,513 @@
...
+type point = { x: bigint; y: bigint };
+
+const Point = {
+  from({ x, y }: point): Point {
+    return { x: Field3.from(x), y: Field3.from(y) };
+  },
+  toBigint({ x, y }: Point) {
+    let x_ = Field3.toBigint(x);
+    let y_ = Field3.toBigint(y);
+    return { x: x_, y: y_, infinity: x_ === 0n && y_ === 1n };
```
querolita, line 51: I created an equivalent `Point` type (elliptic-curve also has it) with the difference that the infinity is not always set to `false` as it is the case in the other gadgets.

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,513 @@
...
+  assert(
+    Curve.modulus > l2Mask + 1n,
+    'Base field moduli smaller than 2^176 are not supported'
+  );
+
+  // witness and range-check denominators, x3, y3
+  let witnesses = exists(12, () => {
+    let [x1_, x2_, y1_, y2_] = Field3.toBigints(x1, x2, y1, y2);
+
+    // TODO: reuse code in twistedAdd to avoid recomputing these
```
querolita, line 91: This would be a nice feature and much cleaner code, but apparently this wouldn't be a trivial refactor at all. Ideas for the future: inspiration from the interpreter structure in o1vm?

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,513 @@
...
+ * we assert that the result is zero and just return the constant zero point.
+ *
+ * Implementation: We double all points together and leverage a precomputed table of size 2^c to avoid all but every cth addition.
+ *
+ * Note: this algorithm targets a small number of points
+ *
+ * TODO: could use lookups for picking precomputed multiples, instead of O(2^c) provable switch
+ * TODO: custom bit representation for the scalar that avoids 0, to get rid of the degenerate addition case
+ */
+function multiScalarMul(
```
querolita, line 337: basically the difference between this function and the counterpart in `elliptic-curve.ts` is that the Curve is a `CurveTwisted` instead of an affine curve. I don't know if making this more generic would be a good idea, maybe we want to have a specific implementation for each curve instead of parameterizing it. Open to suggestions (similar concerns arise in other functions of this interface as well).

querolita (commented):
left some comments for reviewers

mitschabaude:
Since twisted curve addition is complete, there's a lot of complexity in the scalar multiplication code around handling zeros that could be removed. E.g. you don't need the initial aggregator point, and you can just return the result instead of asserting it to be zero or non-zero

querolita:
@mitschabaude Oh that makes sense, I will look into it and clean the algorithm for twisted msm.

lib/provable/test/twisted-curve.unit-test.js
45930, line 1: Was this file mistakenly added to git?  There is no `lib/` directory at the top level.

src/lib/provable/test/twisted-curve.unit-test.ts
```diff
@@ -0,0 +1,83 @@
...
+const Edwards25519 = createCurveTwisted(TwistedCurveParams.Edwards25519);
+let curves = [Edwards25519];
+
+for (let Curve of curves) {
+  // prepare test inputs
+  let field = foreignField(Curve.Field);
+  let scalar = foreignField(Curve.Scalar);
+
+  // point shape, but with independently random components, which will never form a valid point
+  let badPoint = spec({
```
45930, line 30: I don't understand why we define `badPoint` as a point which cannot be valid, but then only use it to define a different point which is valid.

src/lib/provable/test/twisted-curve.unit-test.ts
```diff
@@ -0,0 +1,83 @@
...
+    back: Point.toBigint,
+    provable: Point.provable,
+  });
+
+  // valid random point
+  let point = map({ from: field, to: badPoint }, (x) =>
+    simpleMapToCurve(x, Curve)
+  );
+
+  // two random points that are not equal, so are a valid input to EC addition
```
45930, line 46: [nit] comments referencing EC in the twisted curve file

src/lib/provable/test/twisted-curve.unit-test.ts
```diff
@@ -0,0 +1,83 @@
...
+  let unequalPair = onlyIf(array(point, 2), ([p, q]) => !Curve.equal(p, q));
+
+  // test ec gadgets witness generation
```
45930, line 49: [nit] ec -> twisted curve

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  let message: string | undefined;
+  if (Point.isConstant(p)) {
+    message = `assertOnCurve(): (${x}, ${y}) is not on the curve.`;
+  }
+  ForeignField.assertMul(dTimesX2, y2, aTimesX2PlusY2Minus1, f, message);
+}
+
+/**
+ * EC scalar multiplication, `scalar*point`
```
45930, line 267: [nit] EC -> Twisted curve

45930 (commented):
Given how complex the witness and assertion logic is for these curve functions, I'd feel more comfortable with more extensive testing.

e.g. instead of testing each method on a random point, can we test a few different points?  Are there any special points that could be the source of an edge case error that we could test specifically?  Can we also( or at least) test cases where the witnesses will be invalid and confirm that the assertions trigger?

src/lib/provable/gadgets/twisted-curve.ts
ymekuria: Is the only difference between this function and the `elliptic-curve.ts` counterpart is that the Curve is a `CurveTwisted`? I'm curious to hear why you lean towards specific implementations of each curve vs a more generic function.

src/lib/provable/test/twisted-curve.unit-test.ts
```diff
@@ -0,0 +1,83 @@
...
+    (p) => Curve.isOnCurve(p) || throwError('expect on curve'),
+    (p) => CurveTwisted.assertOnCurve(p, Curve),
+    `${Curve.name} on curve`
+  );
+
+  equivalentProvable({ from: [unequalPair], to: point, verbose: true })(
+    ([p, q]) => Curve.add(p, q),
+    ([p, q]) => CurveTwisted.add(p, q, Curve),
+    `${Curve.name} add`
+  );
```
mitschabaude, line 61: do we have to test addition on unequal pairs of points only? it should work on any two points I think

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,451 @@
...
+
+  // constant case
+  if (Point.isConstant(p1) && Point.isConstant(p2)) {
+    let p3 = twistedAdd(Point.toBigint(p1), Point.toBigint(p2), f, a, d);
+    return Point.from(p3);
+  }
+
+  assert(
+    Curve.modulus > l2Mask + 1n,
+    'Base field moduli smaller than 2^176 are not supported'
```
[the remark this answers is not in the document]
querolita: sure, I'll change that

lib/provable/test/twisted-curve.unit-test.js
querolita: oh this is probably a git error yeah, it should be in /src/lib/provable/test/twisted-curve.unit-test.js instead

src/lib/provable/test/twisted-curve.unit-test.ts
querolita: I am not a fan either, I was copying the format we already had for elliptic curves unit tests

lib/provable/test/twisted-curve.unit-test.js
querolita: Looks like a compiled version of the unit test

src/lib/provable/gadgets/twisted-curve.ts
querolita: If the curve is an affine curve there are some special cases that need to be treated which do not happen in the twisted case. That's the reason that I removed the zeros cases after receiving @mitschabaude's feedback [here](https://github.com/o1-labs/o1js/pull/1949#issuecomment-2595381551), so the type is not the only difference. 

However, I also thought that having a more generic approach could be beneficial. But then I chatted a bit with @dannywillems, who believed that because elliptic curve operations are so sensitive, it can be a better idea to have almost duplicate implementations of certain functions, instead of maximizing generics. There's room for debate I suppose.

src/lib/provable/test/twisted-curve.unit-test.ts
querolita: Oh, I can also do that. Like, checking that adding the same point is the same result as the doubling for example. On it.

src/lib/provable/test/twisted-curve.unit-test.ts
mitschabaude: if you're not a fan, you can also define the valid point directly

```ts
  let point = spec({
    rng: Random.map(field.rng, (x) => simpleMapToCurve(x, Curve)),
    there: Point.from,
    back: Point.toBigint,
    provable: Point.provable,
  });
  ```

File: src/lib/provable/test/twisted-curve.unit-test.ts (hunk 9 above)
Comment (mitschabaude) [reply src/lib/provable/test/twisted-curve.unit-test.ts:61]: good idea!
nit: you still have the `unequalPair` in there which is not needed. you can test correctness of addition on any pair

File: src/lib/provable/crypto/crypto.ts [hunk 11]
```diff
@@ -12,11 +17,22 @@ const Crypto = {
   createCurve(params: Crypto.CurveParams): Crypto.Curve {
     return createCurveAffine(params);
   },
+  /**
+   * Create twisted elliptic curve arithmetic methods.
```
dannywillems, line 21: Nit: I would stress that is in affine coordinates.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+  /**
+   * Coerce the input to a {@link ForeignTwisted}.
+   */
+  static from(g: ForeignTwisted | FlexiblePoint) {
+    if (g instanceof this) return g;
+    return new this(g);
+  }
+
+  /**
+   * Parses a hexadecimal string representing an uncompressed elliptic curve point
```
dannywillems, line 67: Can you mention which endianness is expected?

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+   * The method extracts the x and y coordinates from the provided hex string and
+   * verifies that the resulting point lies on the curve.
+   *
+   * **Note:** This method only supports uncompressed elliptic curve points, which
+   * are 65 bytes in total (1-byte prefix + 32 bytes for x + 32 bytes for y).
+   *
+   * @param hex - The hexadecimal string representing the uncompressed elliptic curve point.
+   * @returns - A point on the foreign curve, parsed from the given hexadecimal string.
+   *
+   * @throws - Throws an error if the input is not a valid public key.
```
dannywillems, line 79: IIUC, we are talking about curve points in this case, not the signature scheme.
[Ed25519](https://en.wikipedia.org/wiki/EdDSA#Ed25519) is the signature scheme.
Curve25519 is the elliptic curve.

I guess in this case you want to talk about the elliptic curve.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+
+type FlexiblePoint = {
+  x: AlmostForeignField | Field3 | bigint | number;
+  y: AlmostForeignField | Field3 | bigint | number;
+};
+function toPoint({ x, y }: ForeignTwisted): Point {
+  return { x: x.value, y: y.value };
+}
+
+class ForeignTwisted {
```
dannywillems, line 30: Is this code supposed to be generic for any Edwards curve given in affine coordinates?

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+   * Create a new {@link ForeignTwisted} from an object representing the (twisted) x and y coordinates.
+   *
+   * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Please refer to the {@link ForeignField} constructor comments for more details.
+   *
+   * @example
+   * ```ts
+   * let x = new ForeignTwisted({ x: 1n, y: 1n });
+   * ```
+   *
+   * **Warning**: This fails for a constant input which does not represent an actual point on the curve.
```
dannywillems, line 44: If the code aims to be generic, we must also ensure it is in the prime subgroup.
If the code is strictly for Curve25519, we should mention it.

src/lib/provable/test/twisted-curve.unit-test.ts
```diff
@@ -0,0 +1,87 @@
...
+  equivalentProvable({ from: [point], to: point, verbose: true })(
+    (p) => Curve.add(p, p),
+    (p) => CurveTwisted.double(p, Curve),
+    `${Curve.name} adding same point equals doubling`
+  );
+
+  equivalentProvable({ from: [point], to: point, verbose: true })(
+    (p) => Curve.double(p),
+    (p) => CurveTwisted.add(p, p, Curve),
+    `${Curve.name} doubling equals adding same point`
```
dannywillems, line 85: I would add the following properties:
- scaling by the order gives the identity
- adding the identity gives the same point
- commutativity
- `r * (p + q) = r * p + r * q`
- `p + (q + s) = (p + q) + s`
- `(-(-p)) = p`
- `- identity` = `identity`
- test fromHex. In particular in the case that we don't have a point in the prime subgroup if we aim to be generic.

src/lib/provable/crypto/foreign-twisted.ts
dannywillems: My apologizes, you use correctly the name. It is named `Edwards25519`, and `Curve25519` is the one in Montgomery form.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+import { Bytes } from '../bytes.js';
+
+// external API
+export { createForeignTwisted, ForeignTwisted };
+
+// internal API
+export { toPoint, FlexiblePoint };
+
+type FlexiblePoint = {
+  x: AlmostForeignField | Field3 | bigint | number;
```
dannywillems, line 23: Side-note: wondering why number is accepted, as it is implemented as `double`?

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+   * class Edwards25519 extends createForeignCurveTwisted(Crypto.TwistedCurveParams.Edwards25519) {}
+   *
+   * // Example hex string for uncompressed point
+   * const publicKeyHex = '04f8b8db25c619d0c66b2dc9e97ecbafafae...';
+   * const point = Edwards25519.fromHex(publicKeyHex);
+   * ```
+   *
+   * **Important:** This method is only designed to handle uncompressed elliptic curve points in hex format.
+   */
+  static fromHex(hex: string) {
```
mitschabaude, line 92: why does this method exist? is this a common format used in applications? if yes that should be mentioned.

but in general I don't think packed hex strings are a great format that we should encourage people to use

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+ * const Curve = createForeignTwisted(Crypto.TwistedCurveParams.Edwards25519);
+ * ```
+ *
+ * `createForeignTwisted(params)` takes curve parameters {@link TwistedCurveParams} as input.
+ * We support `modulus` and `order` to be prime numbers up to 259 bits.
+ *
+ * The returned {@link ForeignCurveNotNeeded} class represents a _non-zero curve point_ and supports standard
+ * elliptic curve operations like point addition and scalar multiplication.
+ *
+ * {@link ForeignCurveNotNeeded} also includes to associated foreign fields: `ForeignCurve.Field` and `ForeignCurve.Scalar`, see {@link createForeignField}.
```
mitschabaude, line 284: "ForeignCurveNotNeeded"? :D

> represents a _non-zero curve point_

why only a non-zero point? I think all points can be represented?

src/lib/provable/gadgets/elliptic-curve.ts
```diff
@@ -68,7 +73,7 @@ function add(p1: Point, p2: Point, Curve: { modulus: bigint; a: bigint }) {
 
   assert(
     Curve.modulus > l2Mask + 1n,
-    'Base field moduli smaller than 2^176 are not supported'
+    `Base field moduli smaller than ${l2Mask + 1n} are not supported`
```
mitschabaude, line 76: can you revert this? a random-looking decimal string will be harder to read than `2^176`

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,451 @@
...
+  let y1y2 = ForeignField.mul(y1, y2, f);
+  let x1x2 = ForeignField.mul(x1, x2, f);
+  let ax1x2 = ForeignField.mul(Field3.from(a), x1x2, f);
+  let y3Num = ForeignField.sub(y1y2, ax1x2, f);
+
+  let x1x2y1y2 = ForeignField.mul(x1x2, y1y2, f);
+  let dx1x2y1y2 = ForeignField.mul(Field3.from(d), x1x2y1y2, f);
+  // check denominators are correctly computed:
+  // den = 1 / (1 +- d * x1^2 * y1^2)
+  Provable.equal(
```
mitschabaude, line 139: * in general, you should use `Provable.assertEqual()` instead of `Provable.equal(...).assertTrue()`. the latter unnecessarily uses constraints to compute the boolean, while assertEqual is free (just wiring)
* in this case though, why not use `ForeignField.assertMul()`? that should save the constraints for range-checking the addition result

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,451 @@
...
+    Field3,
+    one,
+    ForeignField.mul(x3Den, ForeignField.add(one, dx1x2y1y2, f), f)
+  ).assertTrue();
+  Provable.equal(
+    Field3,
+    one,
+    ForeignField.mul(y3Den, ForeignField.sub(one, dx1x2y1y2, f), f)
+  ).assertTrue();
```
mitschabaude, line 148: use `ForeignField.assertMul()`

src/lib/provable/gadgets/twisted-curve.ts
```diff
@@ -0,0 +1,451 @@
...
+    ForeignField.mul(x3Den, ForeignField.add(one, dx1x1y1y1, f), f)
+  ).assertTrue();
+  Provable.equal(
+    Field3,
+    one,
+    ForeignField.mul(y3Den, ForeignField.sub(one, dx1x1y1y1, f), f)
+  ).assertTrue();
+
+  ForeignField.assertMul(x3Num, x3Den, x3, f);
+  ForeignField.assertMul(y3Num, y3Den, y3, f);
```
mitschabaude, line 226: the method of first witnessing the denominators, then checking that they are computed correctly, and _then_ multiplying them with the numerators seems inefficient to me.

did you count the constraints created by these methods? and compare them with https://github.com/o1-labs/o1js/pull/1283?

my intuition says what I did in that PR is much more efficient, especially double but also add

mitschabaude (changes requested):
some comments attached here.

but I also have a general comment: I don't think the circuits are optimized enough yet. I think you should try to minimize the number of constraints before merging.

after releasing this, changing the constraints would be a breaking change, so better to get it right from the get-go

src/lib/provable/crypto/foreign-twisted.ts
45930: > "ForeignCurveNotNeeded"? :D

https://github.com/o1-labs/o1js/commit/10cfa8a8a59625b37515722d53630464b0ff4da3#diff-e2b15bfc1f6103679925c6ccf6697f5877d73e219b0433976dfba6111cf45febR306

👀 😁

src/lib/provable/gadgets/elliptic-curve.ts
45930: I agree about the readability.  Still think it's a shame to hard code a value like this that can go out of date, but I'm ok accepting readability as the dominant factor.

src/lib/provable/crypto/foreign-twisted.ts
mitschabaude: haha damn

querolita:
TODO: subgroup check

src/lib/provable/crypto/foreign-twisted.ts
querolita: The code in the bindings can be used for any curve, but in the provable side of things, we are currently only supporting the edwards25519 curve (but can be easily extended in the future). Thus, I can write a disclaimer on top explaining this.

src/lib/provable/crypto/foreign-twisted.ts
querolita: I thought it could be useful for testing purposes, just like in the weierstrass counterpart https://github.com/o1-labs/o1js/blob/c8240013102706c9113a0ccc4147a4a949a8cf29/src/lib/provable/crypto/foreign-curve.ts#L92

src/lib/provable/gadgets/twisted-curve.ts
querolita: I get your point, seems reasonable. I am now purging the circuit, comparing against your PR and the Weierstrass gadget counterpart.

mitschabaude: make sure to have a constraint counting script! that will help you build intuition more quickly

like this one:
https://github.com/o1-labs/o1js/blob/a910d2c614c704b289c97ca5017d4b93fe49a577/src/examples/crypto/twisted-edwards.ts#L96-L103

src/lib/provable/crypto/crypto.ts
```diff
@@ -12,11 +17,22 @@ const Crypto = {
   createCurve(params: Crypto.CurveParams): Crypto.Curve {
```
dannywillems, line 17: That would require a bigger change, but I would have used a more explicit name to understand which form of curve and which coordinates we're using.

`createCurve` -> `createWeierstrassAffineCurve`
`createCurveTwisted` -> `createTwistedEdwardsAffineCurve`

The space (i.e. affine vs projective) could be omitted as most of the time, in circuit, we use the affine coordinates.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+class ForeignTwisted {
+  x: AlmostForeignField;
+  y: AlmostForeignField;
+
+  /**
+   * Create a new {@link ForeignTwisted} from an object representing the (twisted) x and y coordinates.
+   *
+   * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants. Please refer to the {@link ForeignField} constructor comments for more details.
```
dannywillems, line 37: Nit: but it is in the entire o1js codebase - could we wrap at 80 characters please? It is painful when you want to split your window in different buffers.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+   *
+   * **Warning**: This fails for a constant input which does not represent an actual point on the curve.
+   */
+  constructor(g: {
+    x: AlmostForeignField | Field3 | bigint | number;
+    y: AlmostForeignField | Field3 | bigint | number;
+  }) {
+    this.x = new this.Constructor.Field(g.x);
+    this.y = new this.Constructor.Field(g.y);
+    // don't allow constants that aren't on the curve
```
dannywillems, line 52: As mentioned above, does it include being in the subgroup?

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+  static fromHex(hex: string) {
+    // trim the '0x' prefix if present
+    if (hex.startsWith('0x')) {
+      hex = hex.slice(2);
+    }
+
+    const bytes = Bytes.fromHex(hex).toBytes();
+    const sizeInBytes = Math.ceil(this.Bigint.Field.sizeInBits / 8);
+
+    // extract x and y coordinates from the byte array
```
dannywillems, line 101: I would check the input size first. Raising an example when the format is not the one expected.

src/lib/provable/crypto/foreign-twisted.ts
```diff
@@ -0,0 +1,309 @@
...
+    const tail = bytes.subarray(1); // skip the first byte (prefix)
```
dannywillems, line 102: What do you do with the prefix?

src/lib/provable/crypto/foreign-twisted.ts
dannywillems, line 79: Aren't we more explicit in o1js for the kind of exceptions we raise? Like the user should be able to do a pattern matching on the exception to understand more the kind of issues that happened.

src/lib/provable/crypto/crypto.ts
querolita: I am a big fan of this refactor, I have been getting confused as well while working on this project. Let's open an issue.

src/lib/provable/crypto/foreign-twisted.ts
querolita: I would agree to set a 100-char limit. However, my IDE automatically wraps adjusting to the tab width, so I never find this annoying personally.

src/lib/provable/crypto/foreign-twisted.ts
querolita: Right now there is no code to check the subgroup, will be adding it shortly.

src/lib/provable/crypto/foreign-twisted.ts
querolita: Actually this was extracted from the weierstrass counterpart, but I think there must be a typo. Note that some lines above the prefix is already removed if it starts with `0x`, so there's no need to crop it again here. I have pushed an update in the twisted case, but I haven't touched the other file.

src/lib/provable/test/twisted-curve.unit-test.ts
querolita: will leave this 
> test fromHex. In particular in the case that we don't have a point in the prime subgroup if we aim to be generic.
for follow up as it is not clear that it won't be removed

Trivo25:
This is finished and ported to its own [package](https://github.com/o1-labs-XT/o1js-eddsa) - great job! (keeping the branch alive just in case)

PR #1953, opened 2024-12-16
Expose method to fetch internal state on an `OffchainStateInstance`

hattyhattington17:
Created to address [1935](https://github.com/o1-labs/o1js/issues/1935).

Exposes a method on `OffchainStateInstance` `fetchInternalState` which fetches settled actions from the archive node and rebuilds the internal state trees.


45930:
I think it would be easiest to test on lightnet or devnet so you can have 2 different nodeJS threads running different instances of offchain state and talk to the same blockchain... that ought to work.  You could add it to the integration tests, or just run a test yourself to make sure this let's you sync back up with the chain after the chain diverges from your local state.

Issue #1941, opened 2024-12-05
sha2-512


Trivo25:

mitschabaude:
note: we already have sha2-512 in mina-credentials:
* static-length version: https://github.com/zksecurity/mina-credentials/blob/main/src/credentials/sha2.ts
  * this is directly forked from the o1js impl -- you could simply copy it back into o1js
* dynamic-length version: https://github.com/zksecurity/mina-credentials/blob/main/src/credentials/dynamic-sha2.ts

Trivo25:
Thanks!!

PR #1957, opened 2024-12-18
Supporting variants of SHA2

querolita:
Closes https://github.com/o1-labs/o1js/issues/1941


Geometer1729 (approved):
LGTM

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+    512: K384_512,
+  },
+};
+
+const SHA2 = {
+  /**
+   * Implementation of [NIST SHA-2](https://csrc.nist.gov/pubs/fips/180-4/upd1/final)
+   * hash Function. Supports output lengths of 224, 256, 384, or 512 bits.
+   *
+   * Applies the SHA-3 hash function to a list of big-endian byte-sized {@link Field}
```
dannywillems, line 189: SHA-2

src/lib/provable/gadgets/gadgets.ts
```diff
@@ -973,6 +974,27 @@ const Gadgets = {
    */
   SHA256: SHA256,
```
dannywillems, line 975: Should we have a follow-up PR where we get rid of SHA256.ts and use SHA2 instead? The interface could stay the same.

src/lib/provable/int.ts
```diff
@@ -521,6 +521,28 @@ class UInt64 extends CircuitValue {
...
+   */
+  static fromBytes(bytes: UInt8[]): UInt64 {
+    assert(bytes.length === 8, '8 bytes needed to create a uint64');
+    return UInt64.Unsafe.fromField(bytesToWord(bytes));
+  }
+
+  /**
+   * Combine 8 UInt8s into a UInt64, in big-endian order.
+   */
+  static fromBytesBE(bytes: UInt8[]): UInt64 {
```
dannywillems, line 543: Nit: don't you want to check the length as above?

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+    const { proof } = await Sha2_384Program.sha384(x);
+    await Sha2_384Program.verify(proof);
+    return proof.publicOutput;
+  });
+
+  for (let { preimage, hash } of testVectors()) {
+    let actual = Gadgets.SHA2.hash(384, Bytes.fromString(preimage));
+    expect(actual.toHex()).toEqual(hash);
+  }
+
```
dannywillems, line 215: Can you add a one-line command to mention where they come from?

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+    const { proof } = await Sha2_512Program.sha512(x);
+    await Sha2_512Program.verify(proof);
+    return proof.publicOutput;
+  });
+
+  for (let { preimage, hash } of testVectors()) {
+    let actual = Gadgets.SHA2.hash(512, Bytes.fromString(preimage));
+    expect(actual.toHex()).toEqual(hash);
+  }
+
```
dannywillems, line 287: Nit: Can you add a one-line command to mention where they come from?

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+    },
+    { runs: RUNS }
+  )(nobleSha384, async (x) => {
+    const { proof } = await Sha2_384Program.sha384(x);
+    await Sha2_384Program.verify(proof);
+    return proof.publicOutput;
+  });
+
+  for (let { preimage, hash } of testVectors()) {
+    let actual = Gadgets.SHA2.hash(384, Bytes.fromString(preimage));
```
dannywillems, line 212: Nit: not sure about the convention, but as a user, reviewing the PR and running the tests, I would like to see some logs here, as it seems it is taking a bit of time to run. I was looking at the stdout for some seconds and stop it as I didn't know what it was running.

(again: not blocking at all)

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+
+  await equivalentAsync(
+    {
+      from: [bytes(192)],
+      to: bytes(48),
+    },
+    { runs: RUNS }
+  )(nobleSha384, async (x) => {
+    const { proof } = await Sha2_384Program.sha384(x);
+    await Sha2_384Program.verify(proof);
```
dannywillems, line 207: Nit: is there a convention to check it also verifies? I added on my side.
```
const is_verified = await Sha2_384Program.verify(proof);
expect(is_verified);
```

I don't know if there is another place where we actually test the fact it verifies.

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+    // SHA256 and SHA512, because 8 words of 32 | 64 bits needs no truncation.
+    // Truncation is required for SHA224 and SHA384 though.
+    digest.bytes = digest.bytes.slice(0, length / 8);
+
+    return digest;
+  },
+  compression,
+  messageSchedule,
+  padding,
+  initialState<T extends UInt32 | UInt64>(length: Length): T[] {
```
dannywillems, line 244: Unrelated to the PR: is there a convention in o1js to order the methods? As an engineer not working often in TS, I was surprised to see the method `initialState` defined after a place it is used (`hash` above).

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+   */
+  hash<T extends Length>(length: T, data: FlexibleBytes): Bytes {
+    // Infer the type T based on the value of `length` (conditional type)
+    type Type = T extends 224 | 256 ? UInt32 : UInt64;
+
+    // preprocessing §6.2
+    // padding the message $5.1.1 into blocks that are a multiple of 512
+    let messageBlocks = padding<Type>(length, data);
+
+    let H = SHA2.initialState<Type>(length);
```
dannywillems, line 221: Nit: it seems we're doing the Uint32/Uint64 transformations every time we execute `hash`. A direct optimization would be to do it once and for all.

Not blocking.

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+  let k = Number(mod(paddingValue - (BigInt(l) + 1n), blockLength));
+
+  let lBinary = l.toString(2);
+
+  let paddingBits = (
+    '1' + // append 1 bit
+    '0'.repeat(k) + // append k zero bits
+    '0'.repeat(lengthChunk - lBinary.length) + // append 64|128-bit containing the length of the original message
+    lBinary
+  ).match(/.{1,8}/g)!; // this should always be divisible by 8
```
dannywillems, line 295: Oh. Is it a common practice in TS? :o

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+  // new intermediate hash value
+  intermediateHash(length, [a, b, c, d, e, f, g, h], H);
+
+  return H;
+}
+
+// Helper functions
+
+// Helper function to check if it is a short SHA2 (SHA2-224 or SHA2-256) or not
+function isShort(length: Length): boolean {
```
dannywillems, line 419: Nit: I would inline this check as `isShort` does not have a clear semantics for the primitive.

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+    await Sha2_256Program.verify(proof);
+    return proof.publicOutput;
+  });
+
+  for (let { preimage, hash } of testVectors()) {
+    let actual = Gadgets.SHA2.hash(256, Bytes.fromString(preimage));
+    expect(actual.toHex()).toEqual(hash);
+  }
+
+  function testVectors() {
```
dannywillems, line 144: Can you add a one-line comment to mention where they come from?

src/lib/provable/test/sha2.unit-test.ts
```diff
@@ -0,0 +1,309 @@
...
+    await Sha2_224Program.verify(proof);
+    return proof.publicOutput;
+  });
+
+  for (let { preimage, hash } of testVectors()) {
+    let actual = Gadgets.SHA2.hash(224, Bytes.fromString(preimage));
+    expect(actual.toHex()).toEqual(hash);
+  }
+
+  function testVectors() {
```
dannywillems, line 72: Can you add a one-line comment to mention where they come from?

dannywillems (commented):
Mostly comments at the moment.
I am not approving (sorry for being picky) because I would love to see :
- the origin of the test vectors
- a regression test for the number of constraints

I also think we should therefore get rid of sha256.ts if the primitive is implemented twice.

I have not fully verified that the circuits are not under-constrained.

src/lib/provable/gadgets/gadgets.ts
querolita: I hope so. But there are concerns that this is somehow a "breaking change" and existing zkapps would have to change the syntax to align with this one. Or maybe we can just be backwards compatible in that sense and give an alias for the old name. Either way, this looks like something we could do for a future version of the api.

src/lib/provable/gadgets/sha2.ts
querolita: Yeah I didn't like it either... At first it was is256 but then it was true also in 224. Any suggestion?

src/lib/provable/gadgets/gadgets.ts
mitschabaude: if you add an alias, you can also deprecate the old name with a `@deprecated` annotation

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+   * let digest224 = SHA2.hash(224, preimage);
+   * let digest256 = SHA2.hash(256, preimage);
+   * let digest384 = SHA2.hash(384, preimage);
+   * let digest512 = SHA2.hash(512, preimage);
+   * ```
+   *
+   */
+  hash<T extends Length>(length: T, data: FlexibleBytes): Bytes {
+    // Infer the type T based on the value of `length` (conditional type)
+    type Type = T extends 224 | 256 ? UInt32 : UInt64;
```
Trivo25, line 215: damn, some serious typescript! :D

src/lib/provable/gadgets/sha2.ts
Trivo25: Not really a convention, TS doesn't care - I personally try to order methods by factors such as importance or level of abstraction

src/lib/provable/gadgets/sha2.ts
Trivo25: Not common, but it's something I had to do in the other SHA implementation a while ago - cant remember why tho

src/lib/provable/gadgets/sha2.ts
```diff
@@ -0,0 +1,621 @@
...
+  } else {
+    for (let i = 0; i < 8; i++) {
+      H[i] = (variables[i] as UInt64).addMod64(H[i] as UInt64) as T;
+    }
+  }
+}
+
+// Subroutines for SHA2
+
+function Ch<T extends UInt32 | UInt64>(length: Length, x: T, y: T, z: T): T {
```
Trivo25, line 465: This works well, but if you are interested you could take a look at type guards that might make things in certain cases a bit cleaner

https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

src/lib/provable/test/sha2.unit-test.ts
Trivo25: Good point.. we are not doing that currently, anyway, but we probably should start doing that for completness altought I am not completely sure if Pickles or Kimchi checks if a proof is valid after generating it as a sanity check already

Trivo25 (approved):

src/lib/provable/test/sha2.unit-test.ts
querolita: It is already there, top line of the file 😅 

> // In-circuit tests of the SHA2 family of hash functions together with
// test vectors extracted from https://www.di-mgt.com.au/sha_testvectors.html

PR #1962, opened 2024-12-20
Fix correctness issue in ZkProgram typing

rpanic:
Currently, the ZkProgram method declares a lot of convenience types as generic parameters. However, this leads to wrong typings when explicitly typing those generic parameters as they become inconsistent with the original ZkProgram's Config and therefore the resulting ZkProgram.
Furthermore, this hugely increases friction when trying to type anything outside of o1js that does stuff with ZkPrograms in some way.


src/lib/proof-system/zkprogram.ts
```diff
@@ -265,10 +264,26 @@ function ZkProgram<
...
+    >;
+  };
+  // derived types for convenience
+  type MethodSignatures = Config['methods'];
+  type PrivateInputs = {
+    [I in keyof Config['methods']]: Config['methods'][I]['privateInputs'];
+  };
+  type AuxiliaryOutputs = {
+    [I in keyof MethodSignatures]: Get<MethodSignatures[I], 'auxiliaryOutput'>;
+  };
```
mitschabaude, line 285: you use the derived types above here as well?

mitschabaude (commented):
haven't tested intellisense yet, but this seems like a great idea to me!

src/lib/proof-system/zkprogram.ts
rpanic: You're right, i've fixed it to also use the global derived types here

src/lib/proof-system/zkprogram.ts
```diff
@@ -229,40 +258,13 @@ let SideloadedTag = {
...
-  } = {
-    [I in keyof Config['methods']]: Config['methods'][I]['privateInputs'];
-  },
-  AuxiliaryOutputs extends {
-    [I in keyof MethodSignatures]: Get<MethodSignatures[I], 'auxiliaryOutput'>;
-  } = {
-    [I in keyof MethodSignatures]: Get<MethodSignatures[I], 'auxiliaryOutput'>;
-  }
+  Config extends ConfigBaseType,
+  _ extends unknown = unknown // weird hack that makes methods infer correctly when their inputs are not annotated
```
mitschabaude, line 262: this extra line fixed the type error in `foreign-field-gadgets.unit-test.ts`

I also noticed that you could remove the errors when explicitly annotating method inputs in that file. Which would be good practice, since if you don't do it the inputs will be typed as `any` inside the method.

however, I also found that adding _any_ extra generic type argument makes the error disappear. no idea why!
since writing a zkprogram is already hard, because you tend to get confusing type errors on an incomplete ZkProgram definition, I think it's worth keeping this hack to remove one more confusing error.

mitschabaude (approved):
I merged main and fixed the type error, let's land this!

Trivo25 (approved):

PR #1968, opened 2025-01-10
Add setFee and setFeePerWU

### LINKED ISSUE #1626: Specify fee per weight unit
https://discord.com/channels/484437221055922177/1234429008427618387/1234765711298007071

> can we add to Mina.transaction the possibility to specify the fee per weight unit instead of fee per transaction? According to the explanation above, the market fee will be fee/wu, and to get the tx fee right, we need to know the number of account updates, which is finally known only after the tx is created.

Geometer1729:
Closes #1626


querolita (approved):

dfstio:
This PR clears the proof and signature authorizations for all AccountUpdates. At the same time, for many txs, only the signature of the feePayer should be changed. Is it possible to differentiate between the signatures and proofs that should be changed in case of fee change and those that can stay the same?

mitschabaude:
> At the same time, for many txs, only the signature of the feePayer should be changed. Is it possible to differentiate between the signatures and proofs that should be changed in case of fee change and those that can stay the same?

I would go even further: It's currently not possibly that _any_ proofs are invalidated by a fee change, because they can't refer to the fee payer.

Signatures are affected *if and only if* the have `useFullCommitment = true`.

So you can actually be fully precise with invalidation.

src/lib/mina/transaction.ts
```diff
@@ -544,6 +572,26 @@ function newTransaction(transaction: ZkappCommand, proofsEnabled?: boolean) {
       }
       return pendingTransaction;
     },
+    setFeePerWU(newFeePerWU:UInt64) {
+      //Currently WU is just the number of accountUpdates + 1
+      //https://github.com/MinaProtocol/mina/blob/a0a2adf6b1dce7af889250ff469a35ae4afa512f/src/lib/mina_base/zkapp_command.ml#L803-L823
```
45930, line 577: This is the weight of a zkapp_command, which ends up being different than the weight of a signed command: 

Also, there is this other formula for "cost", which I think is the most relevant formula to use for this context: https://github.com/MinaProtocol/mina/blob/a0a2adf6b1dce7af889250ff469a35ae4afa512f/src/lib/mina_base/zkapp_command.ml#L1317, ( duplicated in o1js here: https://github.com/o1-labs/o1js/blob/2f4d4379d866edb5375472c6e46a6163e73ce8b2/src/lib/mina/transaction-validation.ts#L91)

`(*10.26*np + 10.08*n2 + 9.14*n1 < 69.45*)`

The way I understand it, this formula estimates how long it will take to snark a transaction, which is turn is what makes snark workers want more payment.  Number of account updates +1 doesn't count proofs at all.  I think we should disambiguate "cost" and "weight" to make sure we are capturing the right thing with this new method.

dfstio:
`au.body.useFullCommitment` is `Bool`, therefore the condition `if (au.body.useFullCommitment)` will always be true

src/lib/mina/transaction.ts
```diff
@@ -114,6 +115,25 @@ type Transaction<
...
+   * // Waits for some time and decide to resend with a higher fee
+   *
+   * tx.setFee(newFee);
+   * await tx.sign([feePayerKey]));
+   * await tx.send();
+   * ```
+   */
+  setFee(newFee:UInt64) : TransactionPromise<Proven,false>;
+  /**
+   * setFeePerSnarkCost behaves identically to {@link setFee} but the fee is given per Account Update in the transaction. This is useful because nodes prioritize transactions by fee per weight unit.
```
45930, line 134: > but the fee is given per Account Update in the transaction. This is useful because nodes prioritize transactions by fee per weight unit.

This seems to describe the previous behavior.  Should be updated to mention snark cost instead of weight unit, and mention proofs as well as AccountUpdates.  Might be nice to `{@link getTotalTimeRequired}` as well, but I'm not sure that's a public function.

src/lib/mina/transaction.ts
```diff
@@ -250,6 +270,15 @@ type PendingTransaction = Pick<
    * ```
    */
   errors: string[];
+
+  /**
+   * setFee is the same as {@link Transaction.setFee(newFee)} but for a {@link PendingTransaction}.
+   */
+  setFee(newFee:UInt64):TransactionPromise<boolean,false>;
+  /**
+   * setFeePerSnarkCost is the same as {@link Transaction.setFeeWU(newFeePerWU)} but for a {@link PendingTransaction}.
```
45930, line 279: Update here as well

45930 (approved):

Shigoto-dev19 (approved):

src/lib/mina/transaction.ts
Geometer1729: The link at least works in my LSP

Geometer1729:
Thanks everyone, for bearing with me and catching all those mistakes.

PR #1972, opened 2025-01-17
Add Provable to documentation

45930:
## Summary

[Provable<T>](https://github.com/o1-labs/o1js/blob/82db28b6cb491801f42f5b3d4b89c8aa7b5d177d/src/lib/provable/provable.ts#L55) and [Provable](https://github.com/o1-labs/o1js/blob/82db28b6cb491801f42f5b3d4b89c8aa7b5d177d/src/lib/provable/provable.ts#L57) are both exported as `Provable` from the same file.  There are good documentation comments on the constant variable `Provable` that we want to expose in the [o1js-reference documentation](https://docs.minaprotocol.com/zkapps/o1js-reference), but it is tricky to disambiguate the two symbols.

### Approach

This PR creates a dummy type called `ProvableNamespace` and exports it such that documentation about the type generates under `type-aliases/ProvableNamespace`.  This isn't ideal, but it does quickly and without fuss get the documentation out there.

#### Modifying the export

```diff

- export { Provable };

+ export type Provable<t> = ...

+ export const Provabe = ...
```

We can abandon our style of declaring module exports at the top of the file, but the result is that many of the comments are missing in the ultimate documentation.  For instance, this documentation is what we get for `witness`:

> ### witness()
> 
> ```ts
> witness: <A, T>(type: A, compute: () => T) => InferProvable<ToProvable<A>>;
> ```
> 
> #### Type parameters
> 
> • **A** *extends* [`ProvableType`](../type-aliases/ProvableType.mdx)\<`any`, `any`\>
> 
> • **T** *extends* `any` = `From`\<[`ToProvable`](../type-aliases/ToProvable.mdx)\<`A`\>\>
> 
> #### Parameters
> 
> • **type**: `A`
> 
> • **compute**
> 
> #### Returns
> 
> [`InferProvable`](../type-aliases/InferProvable.mdx)\<[`ToProvable`](../type-aliases/ToProvable.mdx)\<`A`\>\>

And this is the documentation we get using the fake type workaround:

> ### witness()
> 
> ```ts
> witness: <A, T>(type: A, compute: () => T) => InferProvable<ToProvable<A>>;
> ```
> 
> Create a new witness. A witness, or variable, is a value that is provided as input
> by the prover. This provides a flexible way to introduce values from outside into the circuit.
> However, note that nothing about how the value was created is part of the proof - `Provable.witness`
> behaves exactly like user input. So, make sure that after receiving the witness you make any assertions
> that you want to associate with it.
> 
> #### Example
> 
> Example for re-implementing `Field.inv` with the help of `witness`:
> ```ts
> let invX = Provable.witness(Field, () => {
>   // compute the inverse of `x` outside the circuit, however you like!
>   return Field.inv(x);
> }
> // prove that `invX` is really the inverse of `x`:
> invX.mul(x).assertEquals(1);
> ```
> 
> #### Type parameters
> 
> • **A** *extends* [`ProvableType`](ProvableType.mdx)\<`any`, `any`\>
> 
> • **T** *extends* `From`\<[`ToProvable`](ToProvable.mdx)\<`A`\>\> = `From`\<[`ToProvable`](ToProvable.mdx)\<`A`\>\>
> 
> #### Parameters
> 
> • **type**: `A`
> 
> • **compute**
> 
> #### Returns
> 
> [`InferProvable`](InferProvable.mdx)\<[`ToProvable`](ToProvable.mdx)\<`A`\>\>

The second version is what we want to show to developers. 

#### Changing the name of one or the other `Provable`

```diff

- export { Provable };
+ export { ProvableT, ProvableV };

- type Provable<T> = ...
+ type ProvableT<T> = ...

- const Provable = ...
+ const ProvableV = ...
```

This approach seems like the best long-term solution, but involves fixing a lot of imports across the code-base and a potential for regression that my lightweight approach does not risk.  I'd like to take this approach eventually, but first learn more about this code, why the names were chosen the way they were, and which footguns to look out for in extricating them.

### DIFF
--- a/src/lib/provable/provable.ts
+++ b/src/lib/provable/provable.ts
@@ -29,6 +29,7 @@ import {
 import { witness, witnessAsync, witnessFields } from './types/witness.js';
 import { InferValue } from '../../bindings/lib/provable-generic.js';
 import { ToProvable } from '../../lib/provable/types/provable-intf.js';
+import { TupleN } from '../util/types.js';
 
 // external API
 export { Provable, ProvableNamespace };
@@ -81,19 +82,19 @@ type ProvableNamespace = {
    * invX.mul(x).assertEquals(1);
    * ```
    */
-  witness: Function;
+  witness: <A, T>(type: A, compute: () => Promise<T>) => Promise<T>;
   /**
    * Witness a tuple of field elements. This works just like {@link Provable.witness},
    * but optimized for witnessing plain field elements, which is especially common
    * in low-level provable code.
    */
-  witnessFields: Function;
+  witnessFields: <C>(size: number, compute: C) => TupleN<Field, number>;
   /**
    * Create a new witness from an async callback.
    *
    * See {@link Provable.witness} for more information.
    */
-  witnessAsync: Function;
+  witnessAsync: <A, T>(type: A, compute: () => Promise<T>) => Promise<T>;
   /**
    * Proof-compatible if-statement.
    * This behaves like a ternary conditional statement in JS.
@@ -108,7 +109,7 @@ type ProvableNamespace = {
    * const result = Provable.if(condition, Field(1), Field(2)); // returns Field(1)
    * ```
    */
-  if: Function;
+  if: <T>(condition: Bool, type: FlexibleProvableType<T>, x: T, y: T) => T;
   /**
    * Generalization of {@link Provable.if} for choosing between more than two different cases.
    * It takes a "mask", which is an array of `Bool`s that contains only one `true` element, a type/constructor, and an array of values of that type.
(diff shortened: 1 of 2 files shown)


mitschabaude:
removing the docstrings from `Provable` seems to prioritize the online references over inline docstrings.

does this priority really reflect how developers work in practice? do they look up online docs more than just reading the docstrings from intellisense / by going to definition in the editor?
honest question, personally I almost never do the former and do the latter all the time when interacting with a library

mitschabaude:
> tricky to disambiguate the two symbols

Btw, TypeScript itself has no problems with that. docstrings in the editor are simply combined from both the type and the value definition.
Only typedoc seems to be confused by it 😕

45930:
> removing the docstrings from Provable seems to prioritize the online references over inline docstrings.

I don't intend the side effect of removing intellisense from `Provable`, that's a good catch.  I can duplicate the comments so the show up in both places, and eventually I'd like to get rid of the dummy type and return to just having the site show the correct documentation.

I don't want to prioritize the online references over the inline docstrings, but I _do_ want to prioritize the online references 😉

45930:
> Only typedoc seems to be confused by it 😕

This gave me the bright idea of just updating the typedoc version, and that seems promising.  I will leave this PR in draft while I explore this option.

PR #1973, opened 2025-01-17
Update typedoc version

45930:
The newest version of [typedoc](https://typedoc.org/documents/Overview.html) is maintained and fully supports TS 5.

This is a cleaner way of accomplishing #1972 - Running locally, I can see documentation!

image


Trivo25 (approved):
❤️

Trivo25:
Looks like the nix flake needs to be updated to consider the new dependencies (i think?)

PR #1974, opened 2025-01-18
Add doc comments to ZkProgram et al

45930:
Adding some comments for intellisense and the o1js reference page as part of an ongoing effort to better document the best parts of the code.


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Added
 - `setFee` and `setFeePerSnarkCost` for `Transaction` and `PendingTransaction` https://github.com/o1-labs/o1js/pull/1968
+- Doc comments for ZkProgram, verify, SelfProof...https://github.com/o1-labs/o1js/pull/1974
```
Trivo25, line 22: nit `Doc comments for various ZkProgram methods` instead of `...`

src/lib/proof-system/zkprogram.ts
```diff
@@ -118,6 +123,13 @@ function createProgramState() {
   };
 }
 
+/**
+ * Initializes {@link Pickles} bindings, serializes the input proof and VK for use in Ocaml, then calls into {@link Pickles.verify} to verify the proof.
```
Trivo25, line 127: nit: `OCaml`

Besides that, I wonder if we should go this deep into the stack with the doc comments, because if we decide to do that, we might go down a rabbit hole. 

Woulndn't a generic description be easier for us as well as easier to understand for the developer (that has no context of what's going on deeper in our stack)?

src/lib/proof-system/zkprogram.ts
```diff
@@ -118,6 +123,13 @@ function createProgramState() {
...
+ *
+ * @param proof Either a Proof instance or a JSON proof which gets converted into an {@link MlPair} of {@link FieldConst} arrays for use in the bindings.
```
Trivo25, line 129: same here with the OCaml conversion types, I think ideally these things shouldn't be exposed to the user since they might just cause more confusion

src/lib/proof-system/zkprogram.ts
```diff
@@ -155,11 +167,16 @@ async function verify(
   );
 }
 
+/**
+ * Serializeable representation of a Pickles proof, useful for caching compiled proofs.
```
hattyhattington17, line 171: typo!

src/lib/proof-system/zkprogram.ts
45930: > Besides that, I wonder if we should go this deep into the stack with the doc comments, because if we decide to do that, we might go down a rabbit hole.

Are you concerned about this showing up on the website, or in intellisense, or both?  I have one priority, which is to fill out the website and make it kind of usable as a source of information, and I have another priority, which is make it feasible for open source contributors to get involved with the project.

My goal with this one is more about the OSS devs who have no way at all to discover how o1js _actually_ works.  This is a very small step obviously, but it's in the spirit of something I want to do more of.  I'd like the Pickles API to be exposed to o1js devs so they can tell when they've arrived at the edge of the library.  Otherwise we have these functions which come from nowhere and have no implementation in JS and are like black magic.

If you think it's too much detail for the site, I am very open to reorganizing this so that somehow it does not link to anything on the site, but it remains linked in code (this might be the behavior already, since `Pickles` is not exposed on the site, there's nothing to link to).  But I'd like to work on exposing the real stuff to developers who know where to look.  WDYT?

Trivo25: My concern is that it will confuse normal developers that are only consuming the API and using o1js. For open source contributions, I don't think we should rely on doc comments - oss devs will end up reading the code and the implementation, which means we will have to comment the internal code and their implementations better, rather than doc comments (which are public facing). I think it's too much to link to detailed lower-level functions, but I don't mind still explaining what's happening in a more generic way

45930: Ok, I'll remove links for now, but leave the summary

src/lib/proof-system/zkprogram.ts
45930: Removed the links here and skipped the reference to `MLFieldArray` - should be more generic now

Trivo25 (approved):

PR #1976, opened 2025-01-21
Fix devshell for macos

Geometer1729:
`nix develop '.?submodules=1#default'` should now work on macos.
Flake registries seem to be unreliable so this is recommended over `nix develop o1js`
This is useful because the previous work around of using the mina devshell relies on the `PLONK_WASM_NODEJS` and `PLONK_WASM_WEB` variables which skip some rust rebuilds unless the devshell itself is rebuilt.

`nix run .?submodules=1#update-bindings` still doesn't work on macos.

There are also nix documentation improvements from @hattyhattington17 

Sorry for the large flake.nix diff, I also formatted it to improve readability.


hattyhattington17 (approved):

Trivo25 (approved):
😍

PR #1981, opened 2025-01-24
Do not mutate the ciphertext in decryption

### LINKED ISSUE #1980: Encryption.decrypt should not mutate the ciphertext
Current `Encryption.decrypt` implemented here does `cipherText.pop()`, which mutates the passed in ciphertext unnecessarily:

https://github.com/o1-labs/o1js/blob/6ebbc23710f6de023fea6d83dc93c5a914c571f2/src/lib/provable/crypto/encryption.ts#L28-L59

And it will make decryption with the same ciphertext work only the first time.

maple3142:
Fix #1980

Just a one line change to avoid mutating the input ciphertext.

### DIFF
--- a/src/lib/provable/crypto/encryption.ts
+++ b/src/lib/provable/crypto/encryption.ts
@@ -33,7 +33,7 @@ function decrypt(
   const sharedSecret = publicKey.scale(privateKey.s);
   const sponge = new Poseidon.Sponge();
   sponge.absorb(sharedSecret.x);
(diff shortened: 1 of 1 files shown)


Geometer1729 (approved):
LGTM

Shigoto-dev19 (approved):
Nice catch! LGTM

maple3142:
Found a bug causing decryption error due to the code after is expecting `cipherText` array to not contain the auth tag, so pushed a new commit to fix it.

PR #1993, opened 2025-01-28
Resolve documentation signature mismatches

45930:
This PR resolves two classes of validation warnings from the typedoc build: signature mismatches, and not exported warnings.

Not exported warnings occur in a case like this:

```ts

type NotExported = {
  a: string;
  b: number;
}

// Warning: type NotExported, defined in ..., is referenced by Exported but is not included in the documentation
export class Exported extends NotExported {
   ...
 }
```

We make liberal use of this pattern, and I don't see a problem with continuing to do so.  We would only respect this warning if we were committed to a 100% complete autogenerated o1js reference.

### DIFF
--- a/src/lib/mina/zkapp.ts
+++ b/src/lib/mina/zkapp.ts
@@ -1039,7 +1039,6 @@ super.init();
 
   /**
    * Asynchronously fetches events emitted by this {@link SmartContract} and returns an array of events with their corresponding types.
-   * @async
    * @param [start=UInt32.from(0)] - The start height of the events to fetch.
    * @param end - The end height of the events to fetch. If not provided, fetches events up to the latest height.
    * @returns A promise that resolves to an array of objects, each containing the event type and event data for the specified range.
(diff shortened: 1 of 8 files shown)


src/lib/mina/zkapp.ts
```diff
@@ -1039,7 +1039,6 @@ super.init();
 
   /**
    * Asynchronously fetches events emitted by this {@link SmartContract} and returns an array of events with their corresponding types.
-   * @async
```
ymekuria, line 1042: Nice catch.

src/lib/provable/field.ts
```diff
@@ -836,7 +839,7 @@ class Field {
    *
    * **Important**: If the given `bytes` array is an array of `booleans` or {@link Bool} elements that all are `constant`, the resulting {@link Field} element will be a constant as well. Or else, if the given array is a mixture of constants and variables of {@link Bool} type, the resulting {@link Field} will be a variable as well.
    *
-   * @param bytes - An array of {@link Bool} or `boolean` type.
+   * @param bits - An array of {@link Bool} or `boolean` type.
```
ymekuria, line 842: I see what you did here, but the doc comment above says "If the given `bytes` array is an array of ...". Should we make the comment and @param name consistent?

ymekuria (approved):
Nice! I just had a minor comment.

PR #1994, opened 2025-01-28
Resolve link warnings

45930:
This PR resolves warnings related to `{@link}` tags linking to things that are not documented.

This problem requires some finesse, as we use link tags for intellisense on symbols that aren't linked in the documentation site.  We also use links on symbols that _are_ on the site, but link to things that _aren't_.

An easy fix for the internal-only use case is to tag the doc comment with `@internal`, so the intellisense will work and the site generator will know to skip it.

The trickier case is when a public comment references an internal API.  For these, we can either make the whole chain public, make the whole chain private, or remove the link tag, and leave the reference to a symbol that is internal unlinked.  I will call out these cases in inline comments.

This PR also resolves all warnings from executing `npx typedoc`, so we could conceivably add CI to keep it that way.


src/lib/proof-system/proof.ts
```diff
@@ -231,7 +231,7 @@ let sideloadedKeysCounter = 0;
  * NOTE: In the case of `DynamicProof`s, the circuit makes no assertions about the verificationKey used on its own.
  * This is the responsibility of the application developer and should always implement appropriate checks.
  * This pattern differs a lot from the usage of normal `Proof`, where the verification key is baked into the compiled circuit.
- * @see {@link src/examples/zkprogram/dynamic-keys-merkletree.ts} for an example of how this can be done using merkle trees
+ * @see `src/examples/zkprogram/dynamic-keys-merkletree.ts` for an example of how this can be done using merkle trees
```
45930, line 234: removing this link, which was not functional in any case.

src/lib/provable/crypto/foreign-curve.ts
```diff
@@ -18,7 +18,7 @@ import { Bytes } from '../bytes.js';
 export { createForeignCurve, ForeignCurve };
 
 // internal API
-export { toPoint, FlexiblePoint };
+export { toPoint, FlexiblePoint, ForeignCurveNotNeeded };
```
45930, line 21: By exporting `ForeignCurveNotNeeded`, we make it linkable in other comments, where it is already in use.  This makes the whole chain public.

`ForeignCurveNotNeeded` in particular seems weird.  I'm not sure if we should just delete it in favor of `ForeignCurve`.  See: [this PR](https://github.com/o1-labs/o1js/pull/1757/files#diff-e2b15bfc1f6103679925c6ccf6697f5877d73e219b0433976dfba6111cf45febR280) which introduces it.

src/lib/provable/crypto/foreign-curve.ts
```diff
@@ -421,10 +425,10 @@ class ForeignCurveNotNeeded extends ForeignCurve {
  * const Curve = createForeignCurve(Crypto.CurveParams.Secp256k1);
  * ```
  *
- * `createForeignCurve(params)` takes curve parameters {@link CurveParams} as input.
+ * `createForeignCurve(params)` takes curve parameters `CurveParams` as input.
```
45930, line 428: Removing the link to `CurveParams` here so that `createForeignCurve` can remain public, and we don't also have to make `CurveParams` public.  This is my opinionated way of resolving, since the type `CurveParams` is linked in the function signature, it is quite a similar experience for the developer.

src/lib/provable/foreign-field.ts
```diff
@@ -91,11 +91,11 @@ class ForeignField {
...
    * let x = new ForeignField(5);
    * ```
-   * 
+   *
    * Note: Inputs must be range checked if they originate from a different field with a different modulus or if they are not constants.
-   * 
+   *
    * - When constructing from another {@link ForeignField} instance, ensure the modulus matches. If not, check the modulus using `Gadgets.ForeignField.assertLessThan()` and handle appropriately.
-   * - When constructing from a {@link Field3} array, ensure all elements are valid Field elements and range checked.
+   * - When constructing from a `Field3` array, ensure all elements are valid Field elements and range checked.
```
45930, line 98: Removing the link to `Field3` here.  The type is used in the function signature.  The outcome here is that `new ForeignField()` remains public, and `Field3` is private.

typedoc.json
```diff
@@ -19,7 +19,10 @@
   },
   "entryPoints": [
     "src/index.ts",
+    "src/lib/mina/account-update.ts",
+    "src/lib/provable/crypto/foreign-curve.ts",
```
45930, line 23: Foreign curves are already public: https://docs.minaprotocol.com/zkapps/o1js-reference/classes/ForeignCurve

This change makes all the doc comments exposed in the file available for other files to reference in `@link`s.

src/lib/provable/int.ts
```diff
@@ -77,7 +77,7 @@ class UInt64 extends CircuitValue {
     return this.value.toString();
   }
   /**
-   * Turns the {@link UInt64} into a {@link BigInt}.
+   * Turns the {@link UInt64} into a BigInt.
```
ymekuria, line 80: I'm curious why you removed the `BigInt` link?

ymekuria (approved):

src/lib/provable/int.ts
45930: It would also be valid to turn this rule off like we did with `notExported` in #1993.  I think it is worth knowing if there is a typo in a link or if a public API is referencing private APIs, and resolving those, rather than ignoring.  But if it becomes too much effort, or too much value is lost by working around this rule, we can just turn it off too.

Trivo25 (approved):

boray (approved):

PR #1997, opened 2025-01-29
Add important MerkleList method

mitschabaude:
Somehow, `MerkleList` was missing a `popOption()` method and didn't have a good way to emulate it with the existing API.
That's pretty stupid (of me) because this is the most important method, if not the only method that a `MerkleList` really needs :D


mitschabaude:
Ready for review @45930

45930 (approved):

45930:
I don't have permission to force merge.

PR #1998, opened 2025-01-30
ZkProgram type fix

mitschabaude:
fixes a left-over of the "pure" input type requirement in the `ZkProgram` type


mitschabaude:
@Trivo25 this should be a quick one

Trivo25 (approved):

PR #2005, opened 2025-02-04
Make ECDSA easier to use with alternative hash functions

mitschabaude:
By default, ECDSA uses the `Keccak.ethereum()` hash function, e.g. in `EcdsaSignature.verify()` and `EcdsaSignature.verifyEthers()`.

In principle, the code was intended to also be usable with other hash functions. That's why there are `verifySignedHash()` and `signHash()` methods which let you pass in the curve scalar directly, not making any assumptions how that scalar was obtained.

However, this is still kinda hard to use with alternative hash functions because the caller has to convert the hash function output, usually `Bytes`, to a `ForeignField` element (the curve scalar). Internally, o1js does this with the `keccakOutputToScalar` method.

To not make it necessary that callers reimplement `keccakOutputToScalar()`, this PR makes `verifySignedHash()` and `signHash()` optionally accept the `Bytes` output directly, and internally call `keccakOutputToScalar()` on it.

Thus, it's now very easy to use ECDSA with a different hash implementation!

I'll be using this for making ECDSA work for dynamic length inputs https://github.com/zksecurity/mina-attestations/pull/95

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,13 +18,16 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/b857516...HEAD)
 
 ### Added
+
 - `setFee` and `setFeePerSnarkCost` for `Transaction` and `PendingTransaction` https://github.com/o1-labs/o1js/pull/1968
 - Doc comments for various ZkProgram methods https://github.com/o1-labs/o1js/pull/1974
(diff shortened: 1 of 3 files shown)


45930:
@mitschabaude could you add a quick example to `src/examples/crypto/ecdsa/ecdsa.ts` demonstrating the new option?

mitschabaude:
@45930 ok done!

Geometer1729 (approved):

45930 (approved):

PR #2008, opened 2025-02-06
Provable BigInt 

boray:
This PR introduces `createProvableBigInt()` class factory function that generates a provable bigint class which has a hardcoded modulus and can hold a integer that is in range [0,modulus). This primitive is intended to be a building block for foreign fields with moduli bigger than `ForeignField` supports. The class holds the bigint as limbs. Size and number of limbs are chosen by the factory function from presets by default. The factory function has an optional parameter for custom limb configurations.


mitschabaude:
I recommend to add (basic) in-circuit tests with `Provable.runAndCheck()` as soon as possible, to see if that actually works.

Right now you're using `Unconstrained` in provable code in a way that doesn't work:
```ts
Unconstrained.from((this.value.get() + a.value.get()) % this.Constructor.modulus.toBigint())
```
You can't `.get()` directly in provable code! It's designed to only be called in `witness()` blocks.
So it should be rewritten to:
```ts
Unconstrained.witness(() => (this.value.get() + a.value.get()) % this.Constructor.modulus.toBigint())
```

45930:
I prefer the name bignum to bigint because `toBigint()` is a common method our provable classes implement, and `BigInt` in that context is not this new class we're adding.  Bignum is not currently meaningful at all in our namespace, and it is used in the noir ecosystem in a similar context.

boray:
> I prefer the name bignum to bigint because `toBigint()` is a common method our provable classes implement, and `BigInt` in that context is not this new class we're adding. Bignum is not currently meaningful at all in our namespace, and it is used in the noir ecosystem in a similar context.

I've named ProvableBigInt to distinguish from BigInt but as you mentioned in cases such as `toBigint()` and `fromBigint()` it's still confusing. BigNum sounds nice!

boray:
I chose a design where the modulus is hardcoded into the class and every operation returns a reduced value. This limits its use for RSA, which requires a dynamic modulus. I chose this approach since this project was initially focused on fields, and it's relatively easy to implement RSA without a primitive from o1js, as it requires fewer mathematical operations than fields.

This design decision is open to discussion. I believe this primitive should also support RSA. This can be done by adding another class factory with a follow-up PR.

src/examples/crypto/bigint.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  @method async myMethod(y: SmallField) {
+    let x = y.mul(SmallField.fromBigint(3n));
+    Provable.log(x);
+    this.x.set(x);
+  }
+}
+await MyContract.analyzeMethods(); // works
+
+// btw - we support any finite field up to 259 bits. for example, the seqp256k1 base field:
```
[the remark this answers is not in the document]
boray: yes, this was committed as a placeholder and copied from ForeignField. Given that we have unit tests, we might not need such an example

src/examples/crypto/bigint.ts
```diff
@@ -0,0 +1,36 @@
...
+import { Experimental, method, Provable, Scalar, SmartContract, state, State } from 'o1js';
+
+const { createProvableBigInt } = Experimental;
+
+// Let's create a small finite field: F_17
+
+class SmallField extends createProvableBigInt(97n) {}
+
+let x = SmallField.fromBigint(16n);
+x.assertEquals(SmallField.fromBigint(-1n)); // 16 = -1 (mod 17)
```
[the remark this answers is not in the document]
boray: Should we have an example for bigint like foreign-field?

src/examples/crypto/bigint.ts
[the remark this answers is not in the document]
boray: Should we have an example for bigint like foreign-field?

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,881 @@
...
+     * @returns A Bool indicating if a is less than or equal to b
+     */
+    assertEquals(a: ProvableBigInt_) {
+      this.equals(a).assertTrue('ProvableBigInts are not equal');
+    }
+  }
+  return ProvableBigInt_;
+}
+
+abstract class ProvableBigInt<T extends ProvableBigInt<T>> {
```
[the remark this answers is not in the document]
boray: no, we can remove it.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+    limb_num: 36,
+    limb_size: 116n,
+    mask: (1n << 116n) - 1n,
+    MAX: (1n << 4176n) - 1n,
+  },
+};
+
+/**
+ * Creates a class representing a ProvableBigInt with modular arithmetic capabilities.
+ * This is particularly useful for implementing prime fields that doesn't fit into the native field.
```
querolita, line 60:
```suggestion
 * This is particularly useful for implementing prime fields that don't fit into the native field.
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+ * const power = a.pow(b);
+ * const sqrt = c.sqrt();
+ *
+ * ```
+ *
+ * The class automatically handles modular reduction after
+ * arithmetic operations to maintain valid representations. All operations are
+ * designed to be provable and optimised for less constraints.
+ *
+ * @param modulus The modulus for the big integer arithmetic (must be prime)
```
querolita, line 118: I pointed this out earlier, but: either assert it is a prime, or specify that if it is not a prime then some operations are not possible.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      let value = x;
+
+      if (value < 0n) {
+        value = ((x % modulus) + modulus) % modulus;
+      }
+      if (value >= ProvableBigInt_.modulus.toBigInt()) {
+        value = value % modulus;
+      }
+      for (let i = 0; i < ProvableBigInt_.config.limb_num; i++) {
+        fields.push(Field.from(value & ProvableBigInt_.config.mask)); // fields[i] = x & 2^64 - 1
```
querolita, line 199: seems like the comment with 2^64 is off

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      let fields = [];
+      let value = x;
+
+      if (value < 0n) {
+        value = ((x % modulus) + modulus) % modulus;
+      }
+      if (value >= ProvableBigInt_.modulus.toBigInt()) {
+        value = value % modulus;
+      }
+      for (let i = 0; i < ProvableBigInt_.config.limb_num; i++) {
```
querolita, line 198: Seems like this is quite similar to what's on line [142](https://github.com/o1-labs/o1js/blob/5b87a1aa69bafde33b7e2139c3d6b4348b23a38c/src/lib/provable/bigint.ts#L142-L145), I wonder if it makes sense to make this modular to avoid code duplication.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      for (let i = 0; i < this.Constructor.config.limb_num; i++) {
+        delta[i] = delta[i].sub(R[i]).seal();
+      }
+
+      let carry = Field.from(0);
+
+      for (let i = 0; i < this.Constructor.config.limb_num - 1; i++) {
+        let deltaPlusCarry = delta[i].add(carry).seal();
+
+        carry = Provable.witness(Field, () => deltaPlusCarry.div(1n << 116n));
```
querolita, line 321: This assumes limbs will always be 116 bits long? Then, I would use a constant somewhere, otherwise looks error prone.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        rangeCheck(carry, 128, true);
```
querolita, line 322: Why this is the range check needed? I was expecting carry to be much smaller than that (no pen and paper claim, just hand waving)

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      return new ProvableBigInt_(this.fields, this.value);
+    }
+
+    /**
+     * Adds two ProvableBigInt instances
+     * @param a The ProvableBigInt to add
+     * @returns The sum as a ProvableBigInt
+     */
+    add(a: ProvableBigInt_, isDouble = false): ProvableBigInt_ {
+      if (isDouble) a = this;
```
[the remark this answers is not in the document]
querolita: Oh I see, you actually have it. Forget about it.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        this.fields,
+        a.fields,
+        q.fields,
+        r.fields,
+        this.Constructor.modulus.fields,
+      ];
+
+      // compute X + Y limb-by-limb
+      for (let i = 0; i < this.Constructor.config.limb_num; i++) {
+        if (isDouble) delta[i] = X[i].mul(2);
```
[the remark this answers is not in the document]
querolita: Never mind, I saw that there is indeed `double` and `square` methods, which call these under the wood with the flag = true.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        delta[i] = delta[i].sub(R[i]).seal();
+      }
+
+      let carry = new Field(0);
+
+      for (let i = 0; i < 2 * this.Constructor.config.limb_num - 2; i++) {
+        let deltaPlusCarry = delta[i].add(carry).seal();
+
+        carry = Provable.witness(Field, () =>
+          deltaPlusCarry.div(1n << BigInt(this.Constructor.config.limb_size))
```
querolita, line 403: Better like this instead of hardcoded 116

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        );
+        rangeCheck(carry, 128, true);
```
querolita, line 405: Again, not sure where the 128 comes from and if we should be more accurate. Otherwise, I wonder if there could be multiple and fake candidates. Like, I see the completeness, but haven't analyzed  the soundness.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+          // If t is smaller than zero, add modulus
+          if (t < 0n) {
+            t = t + p;
+          }
+
+          return { res: ProvableBigInt_.fromBigInt(t) };
+        }
+      );
+
+      res.mul(this).assertEquals(ProvableBigInt_.one());
```
querolita, line 472: Nice, "expensive" witness computation block but then in circuit all we want is this assertion.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        }
+      };
+
+      let result = ProvableBigInt_.one();
+      let base = this.clone();
+
+      for (const chunk of processChunk(exponentBits, 100)) {
+        for (const bit of chunk) {
+          result = Provable.if(bit, ProvableBigInt_, result.mul(base), result);
+          base = base.mul(base);
```
querolita, line 518: Just a heads up. With pow, one can easily overflow the limbs capacity. I just wanted to raise a concern in case it was not checked: 
What happens when after exponentiating, you get a number that is larger than the prime but smaller than the MAX of the limbs? Like, larger than a prime of 512 bits, but smaller than the 580 being 5*116. 

Or the other way round, do you allow "overflown" field elements or is it always expected to fit in <p?

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      }
+
+      return result;
+    }
+
+    /**
+     * Computes the square root of a Provable BigInt
+     * @returns The square root as a ProvableBigInt
+     */
+    sqrt(): ProvableBigInt_ {
```
querolita, line 529: I suppose this is the most expensive operation? I think we should include in the docs of each function how many rows (or a range of them) are expected for each of these operations. That helps developers understand how complex they can afford their circuit to become.

querolita (commented):
So far, my felicitations for this PR, very interesting gadgets live inside 😄 

I have dropped a few comments so we can discuss openly, before this gets accepted.

src/lib/provable/bigint.ts
boray: Yes, it shouldn't be constant. I missed it.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        result |= this.fields[i].toBigInt() << (this.Constructor.config.limb_size * BigInt(i)); // result = result | fields[i] << 64 * i
+      }
+      return result;
+    }
+
+    toFields(): Field[] {
+      return this.fields.slice(0, this.Constructor.config.limb_num);
+    }
+
+    static fromFields(fields: Field[]): ProvableBigInt_ {
```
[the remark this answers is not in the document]
boray: It is limited `limb_size`-bit length. It's also an unsafe method. This needs to be doccommented though.

src/examples/crypto/bigint.ts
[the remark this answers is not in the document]
boray: Update: I added an example!

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+
+type BigIntParameter = {
+  limb_num: number;
+  limb_size: bigint;
+  mask: bigint;
+  MAX: bigint;
+};
+
+const BigIntParams: { [key: string]: BigIntParameter } = {
+  348: {
```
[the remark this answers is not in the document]
boray: 116-bit is the most optimal limb size(I'll add a comment about this). Therefore, it's more performant to choose 464-bit limbs for 381-bit prime fields as opposed to using 384-bit limbs (64x6).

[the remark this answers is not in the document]
boray: The parameter choice is not exposed to the user by default. The config with the least bit length that is greater than prime field's bit length is chosen by the class factory.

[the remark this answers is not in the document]
boray: Actually, we don't need to have a list of this parameters. `findConfig()` can choose the appropriate parameters with the rule I explained above.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+        value >>= ProvableBigInt_.config.limb_size; // x = x >> limb_size
+      }
+      return new ProvableBigInt_(fields, Unconstrained.from(value));
+    }
+
+    get Constructor() {
+      return this.constructor as typeof ProvableBigInt;
+    }
+
+    toBigInt(): bigint {
```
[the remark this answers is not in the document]
boray: I added a comment to clarify it is transforming a provable bigint to JS bigint.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
boray: I added a comment!

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+  348: {
+    limb_num: 3,
+    limb_size: 116n,
+    mask: (1n << 116n) - 1n,
+    MAX: (1n << 348n) - 1n,
+  },
+  464: {
+    limb_num: 4,
+    limb_size: 116n,
+    mask: (1n << 116n) - 1n,
```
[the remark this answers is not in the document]
boray: These values are being chosen by `findConfig()` now.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
boray: fixed!

src/lib/provable/bigint.ts
boray: `mul()` ensures the result is reduced. Therefore, every step of the exponentiation is also reduced. We don't allow "overflown" elements.

src/lib/provable/bigint.ts
boray: I added cost expectations to JSDoc comments. I didn't put exact numbers since they change depending on the limb configuration.

src/examples/crypto/bigint.ts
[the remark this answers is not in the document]
boray: Numbers are now correct in the updated example.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+          deltaPlusCarry.div(1n << BigInt(this.Constructor.config.limb_size))
+        );
+        rangeCheck(carry, 128, true);
+
+        deltaPlusCarry.assertEquals(carry.mul(1n << BigInt(this.Constructor.config.limb_size)));
+      }
+
+      delta[2 * this.Constructor.config.limb_num - 2].add(carry).assertEquals(0n);
+
+      return r;
```
Trivo25, line 402: we witness `r` and return it as the result but don't constraint it anywhere else, is this correct?

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+  abstract mod(): T;
+  abstract greaterThan(a: T): Bool;
+  abstract greaterThanOrEqual(a: T): Bool;
+  abstract lessThan(a: T): Bool;
+  abstract lessThanOrEqual(a: T): Bool;
+  abstract equals(a: T): Bool;
+  abstract assertEquals(a: T): void;
+}
+
+function rangeCheck(x: Field, bits: number, signed?: boolean) {
```
Trivo25, line 796: imo we should move these into `range-check.ts`

src/lib/provable/test/bigint.test.ts
```diff
@@ -0,0 +1,2503 @@
+import { Experimental, Field } from 'o1js';
```
Trivo25, line 1: This is a huge test file! Can we not simplify this and merge it with the other test file? somehow

src/examples/crypto/bigint.ts
```diff
@@ -0,0 +1,52 @@
+import { Experimental, method, Provable, SmartContract, state, State } from 'o1js';
```
Trivo25, line 1: Unused imports

Trivo25 (commented):
first pass done, great job!

a few general comments: I think we should try to stay as close as possible to camelCase (given that we do that in the rest of the codebase too)
I also think we should think about the cost comments as well, since they are relative and depend on a few external factors

src/examples/crypto/bigint.ts
boray: Can you point out which ones are unused? All of them are used as far as I see.

Trivo25: oops sorry, i was looking at something else - you are right

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+    /**
+     * Multiplies two ProvableBigInt instances
+     * Cost: Cheap
+     * @param a The ProvableBigInt to multiply
+     * @returns The product as a ProvableBigInt
+     */
+    mul(a: ProvableBigInt_, isSquare = false): ProvableBigInt_ {
+      if (isSquare) a = this;
+
+      let { q, r } = Provable.witness(Struct({ q: ProvableBigInt_, r: ProvableBigInt_ }), () => {
```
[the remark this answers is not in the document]
boray: I didn't know this. Now, structs are replaced by` provable()`.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+import { ProvablePureExtended, Struct } from './types/struct.js';
+import { Gadgets } from './gadgets/gadgets.js';
+import { assert } from './gadgets/common.js';
+import { provableFromClass } from './types/provable-derivers.js';
+import { Unconstrained } from './types/unconstrained.js';
+
+export { createProvableBigInt, ProvableBigInt };
+
+type BigIntParameter = {
+  limb_num: number;
```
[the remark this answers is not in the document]
boray: now, all different cases are replaced with camelCase!

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+        // For the last limb, only include significant bits
+        return bits.slice(0, totalBits - index * limbBits);
+      });
+
+      // Ensure the total length doesn't exceed what's needed
+      return allBits.slice(0, totalBits);
+    }
+
+    /**
+     * UNSAFE
```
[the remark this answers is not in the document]
boray: The reason is input is not checked. I added more explanation and packaged unsafe method together.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+     * Clones a ProvableBigInt instance
+     * @returns A new ProvableBigInt instance with the same value
+     */
+    clone(): ProvableBigInt_ {
+      return new ProvableBigInt_(this.fields, this.value);
+    }
+
+    /**
+     * Adds two ProvableBigInt instances
+     * Cost: Cheap
```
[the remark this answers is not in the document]
boray: The cost depends on `limbNum`, `limbSize`, `limbNumber % 3 == 0`. Therefore, I chose cheap, moderate and expensive scale instead of numbers. I may add benchmarks of bls and brainpool prime fields to top as a comment. Would it work?

src/lib/provable/test/bigint.test.ts
boray: This test file has a greater surface area than the unit test as this one test all methods while the unit-test only tests provable methods. Coby proposed to have one generic test which we can use with different primes. But, I am inclined to having them as separate tests since we use hardcoded values at some cases and imo it's better to have them explicitly.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+ *
+ * ```ts
+ * const BigInt521 = createProvableBigInt(2n ** 521n - 1n); // creates a class for 521-bit integers
+ * ```
+ *
+ * `createProvableBigInt(modulus, config?)` takes two parameters:
+ * - `modulus`: The modulus of the field (must be a prime)
+ * - `config`: Optional configuration for custom limb size and numbers
+ *
+ * The returned class supports comprehensive arithmetic operations including:
```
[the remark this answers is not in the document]
boray: The follow-up PR to support dynamic RSA will change this comments significantly. So, we don't need to worry about missing to update these later.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
Trivo25: I don't think we need numbers, but letting the developer know that `limbNum, limbSize, limbNumber % 3 == 0` impacts constraint cost should be important

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,957 @@
...
+    }
+
+    /**
+     * Converts a ProvableBigInt instance to a JS bigint
+     * @returns JS bigint representation of the ProvableBigInt
+     */
+    toBigInt(): bigint {
+      let result = 0n;
+      for (let i = 0; i < this.Constructor.config.limb_num; i++) {
+        result |= this.fields[i].toBigInt() << (this.Constructor.config.limb_size * BigInt(i)); // result = result | fields[i] << limb_size * i
```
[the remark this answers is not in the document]
Trivo25: what for?

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      const totalBits = this.Constructor.modulus.toBigInt().toString(2).length;
+      const limbBits = Number(this.Constructor.config.limb_size);
+
+      // Convert each field to bits with proper size
+      let allBits = this.fields.flatMap((field, index) => {
+        const bits = field.toBits();
+        // For all limbs except the last one, ensure proper bit length
+        if (index < this.Constructor.config.limb_num - 1) {
+          while (bits.length < limbBits) {
+            bits.push(new Bool(false));
```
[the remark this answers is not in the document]
boray: Reason of existence for the `toBits()` and `fromBits()` methods are basically their use in `pow()` (`fromBits() `is not directly used). With the latest commit, I simplified the `toBits()` by simply utilising the `field.toBits()` better with the length argument. Both unit and jest tests succeeds with the refactored method.  I didn't get how side-channelling would be risk. I would be happy to hear more about that.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,898 @@
...
+      // Calculate total needed bits based on modulus
+      const totalBits = this.Constructor.modulus.toBigInt().toString(2).length;
+      const limbBits = Number(this.Constructor.config.limb_size);
+
+      // Convert each field to bits with proper size
+      let allBits = this.fields.flatMap((field, index) => {
+        const bits = field.toBits();
+        // For all limbs except the last one, ensure proper bit length
+        if (index < this.Constructor.config.limb_num - 1) {
+          while (bits.length < limbBits) {
```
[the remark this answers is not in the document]
boray: This is not being done in the latest version!

src/lib/provable/bigint.ts
boray: no, we constraint it in the lines above.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
boray: Are you okay with using `cheap-medium-expensive` grading in addition to explaining the factors that affect constraint cost, `limbNum, limbSize, and limbNumber % 3 == 0`?

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
boray: It's the operation in plain English. I think it's relevant but may be redundant. Should I remove it?

src/lib/provable/bigint.ts
Shigoto-dev19: The `sqrt()` method is described as “cheap” because it only witnesses a single quadratic residue and constrains it with one `square` and one `assertEquals`. 

However, while its constraint cost is low, computing the actual square root via Tonelli-Shanks remains computationally expensive. For large `ProvableBigInt`s, witness generation in TS can take significant time; this runtime cost should not be confused with the method’s minimal constraint overhead or proof‑generation cost. So I think it would be helpful to add a note about this in the documentation.

Shigoto-dev19: Additionally, every non-zero quadratic residue modulo `p` has exactly two reduced square‑roots; `R` and `p − R`. We should document that this function always returns the canonical (smaller) square‑root.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+      let r = Provable.witness(ProvableBigInt_, () => {
+        const p = this.Constructor.modulus.toBigInt();
+        // Special cases
+        // Case 1: If input is 0, square root is 0
+        if (this.toBigInt() === 0n) return ProvableBigInt_.fromBigInt(0n);
+
+        // Case 2: If p ≡ 3 (mod 4), we can use a simpler formula
+        // In this case, the square root is a^((p+1)/4) mod p
+        if (p % 4n === 3n) {
+          const pplusonedivfour = (p + 1n) / 4n;
```
Shigoto-dev19, line 524: Nit: rename following camelCase naming conventions.
```suggestion
          const pPlusOneDivFour = (p + 1n) / 4n;
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+          const pplusonedivfour_reduced = pplusonedivfour % p;
```
Shigoto-dev19, line 525: Nit: rename following camelCase naming conventions.
```suggestion
          const pPlusOneDivFourReduced = pPlusOneDivFour % p;
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+          const sqrt = modularExponentiation(this.toBigInt(), pplusonedivfour_reduced, p);
```
Shigoto-dev19, line 526: Nit: rename following camelCase naming conventions.
```suggestion
          const sqrt = modularExponentiation(this.toBigInt(), pPlusOneDivFourReduced, p);
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+        .or(this.equals(a));
+    }
+
+    /**
+     * Checks if one ProvableBigInt is equal to another
+     * Cost: Cheap
+     * @param a The ProvableBigInt to compare
+     * @returns A Bool indicating if a is equal to b
+     */
+    equals(a: ProvableBigInt_): Bool {
```
Shigoto-dev19, line 661: Consider using `Provable.equal` instead.
```ts
return Provable.equal(ProvableBigInt_, this, a)
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+        .reduce((result, isEqual) => result.and(isEqual), new Bool(true));
+    }
+
+    /**
+     * Checks if one ProvableBigInt is less than or equal to another
+     * Cost: Cheap
+     * @param a The ProvableBigInt to compare
+     * @returns A Bool indicating if a is less than or equal to b
+     */
+    assertEquals(a: ProvableBigInt_) {
```
Shigoto-dev19, line 673: Consider using `Provable.assertEqual` instead.
```ts
Provable.assertEqual(ProvableBigInt_, this, a)
```

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+  return result;
+}
+
+/**
+ * Miller-Rabin primality test
+ * @param n The number to test for primality
+ * @param k Number of iterations (higher = more accurate)
+ * @returns true if n is probably prime, false if n is definitely composite
+ */
+function isPrime(n: bigint, k: number = 10): boolean {
```
Shigoto-dev19, line 838: The Miller-Rabin primality test already exists [here](https://github.com/o1-labs/o1js/blob/main/src/examples/crypto/rsa/utils.ts#L72) in the RSA example.

I think we should find a way to remove this redundancy from the codebase.

We can import it from the RSA utils but it would be ugly! maybe we can extract it into a shared utility (e.g. src/lib/util) that both modules can import :thinking: 

Same comment is applicable for the `randomBigintInRange` function.

src/lib/provable/test/bigint.test.ts
Shigoto-dev19, line 1: I recommend testing against prime fields larger than 256 bits **once** as it will better reflect `ProvableBigInt`’s utility (multi‑limb arithmetic) and provides stronger coverage of its algorithms. 

The only downside is that the tests execution time will be slower!

Shigoto-dev19: Moreover, it's possible to refactor the tests by extracting a reusable helper function for each method and invoking it in each test block with different **JS bigint** inputs.

src/lib/provable/test/bigint.unit-test.ts
Shigoto-dev19, line 1: Same here. I think it would be better to test against prime fields larger than 256 bits.

Shigoto-dev19 (commented):
Great work @boray :clap: 

I’ve left inline review comments covering mostly naming conventions, documentation clarifications, and a few refactoring suggestions.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
Trivo25: I still think it's misleading but as long as we give more information it should be okay

src/lib/provable/bigint.ts
Trivo25: We only use it to calculate `delta`, `delta[i] = delta[i].sub(R[i]).seal();`. are we sure there's no other `r` that still satisfies the assertions on `delta` and can be arbitrarily chosen? just making sure - we can take this to Slack too if you wish

Trivo25 (commented):
Please give me a quick ping when you had a chance to adress all the outstanding comments and before you're going to merge this

src/lib/provable/bigint.ts
Trivo25: what about `rangeCheck48` and co that come beneath this function? For this particular function, maybe a better name would be `rangeCheckSelector` or `rangeCheckHelper`

mitschabaude: IMHO this method is a classical example of premature abstraction and doesn't belong in `range-check.ts`
* makes the code less readable, e.g. `rangeCheck(x, 128, true)` is less clear than `rangeCheck128Signed(x)`
* adds one additional layer to go from where it's used to where the logic is => harder to navigate
* doesn't save any lines of code (actually it adds code, namely this function)

I would remove it and inline the different ways of doing range checks in this file!

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+      if (signed) {
+        rangeCheck128Signed(x);
+      } else {
+        throw new Error('128-bit unsigned range check not implemented');
+      }
+      break;
+  }
+}
+
+function rangeCheck48(x: Field) {
```
mitschabaude, line 765: this is less efficient than `rangeCheckN(48)`, would remove

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+       * **WARNING**: This method is UNSAFE and lacks checks on the fields. Use with caution as it can lead to incorrect results or security vulnerabilities.
+       * @param fields The limbs of the ProvableBigInt. Must be of the correct length.
+       * @returns A ProvableBigInt instance from the fields
+       */
+      fromFields(fields: Field[]): ProvableBigInt_ {
+        let value = 0n;
+        for (let i = 0; i < ProvableBigInt_.config.limbNum; i++) {
+          value |= BigInt(fields[i].toBigInt()) << BigInt(ProvableBigInt_.config.limbSize * i);
+        }
+        return new ProvableBigInt_(fields, Unconstrained.from(value));
```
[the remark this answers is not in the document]
Shigoto-dev19: Oh yeah, you’re right; the requested change wasn’t accurate.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
Shigoto-dev19: Great catch @mitschabaude !
Thank you for the information. It’s new to me and very helpful :)

45930:
@Trivo25 WDYT about merging this (within the experimental namespace) with the known issues commented, or otherwise pointed out?

It sounds like there is 95% buy-in on this feature with some call outs about clarity and performance, and one issue of soundness.  I think getting this feature to people with the caveat not to use it in a production zkapp until we promote it out of experimental is appropriate.  It will be a lot easier to address issues 1 by 1 in follow-up PRs as well.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+import { Provable } from './provable.js';
+import { ProvablePureExtended } from './types/struct.js';
+import { Gadgets } from './gadgets/gadgets.js';
+import { assert } from './gadgets/common.js';
+import { provable, provableFromClass } from './types/provable-derivers.js';
+import { Unconstrained } from './types/unconstrained.js';
+
+export { createProvableBigInt, ProvableBigInt };
+
+type bigIntParameter = {
```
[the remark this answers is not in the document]
boray: fixed!

src/examples/crypto/bigint.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const { createProvableBigInt } = Experimental;
+
+// p = 13
+class BigInt13 extends createProvableBigInt(13n) {}
```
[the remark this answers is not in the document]
boray: This is the toy example. We have such examples below.

src/lib/provable/test/bigint.test.ts
boray: We already test fields larger than 259 bits below. Coby proposed having a dynamic test function before. But, there are some hardcoded values in some test cases similar to existing tests in o1js and it's better to have them explicitly.

45930 (approved):
LGTM with the understanding that it's shipping to the experimental namespace and the remaining review comments are addressed in later PRs targeting the May release!

src/examples/crypto/bigint.ts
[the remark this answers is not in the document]
boray: `foreign-fields.ts` and `bigint.ts` are similar in that they are both primitives that other primitives can build upon. Similar to the `ecdsa` example, there should be an example for `bigint`. However, it should be a separate project since it’s a significant undertaking.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+      }
+      return result;
+    }
+
+    /**
+     * Converts a ProvableBigInt instance to field array representation of the limbs
+     * @returns Limbs of the ProvableBigInt
+     */
+    toFields(): Field[] {
+      return this.fields.slice(0, this.Constructor.config.limbNum);
```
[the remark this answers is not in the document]
boray: You're right, you can access limbs via `.fields` property. But if look closely, these fields are generated from JS bigint at instantiation with `bigIntToLimbs()` function. The same logic is also used at `fromBigInt()`, thus this function exists comply with DRY rule.

src/lib/provable/bigint.ts
```diff
@@ -0,0 +1,897 @@
...
+     * @param exp The exponent
+     * @returns The result as a ProvableBigInt
+     */
+    pow(exp: ProvableBigInt_): ProvableBigInt_ {
+      const exponentBits = exp.toBits();
+      const processChunk = function* (bits: Bool[], chunkSize: number) {
+        for (let i = 0; i < bits.length; i += chunkSize) {
+          yield bits.slice(i, i + chunkSize);
+        }
+      };
```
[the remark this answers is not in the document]
boray: I can't use that function since it requires the array to be a multiple of chunk size.

src/lib/provable/bigint.ts
boray: carry will be exceed 116 bits when summing/substracting, therefore we range check the carry to ensure no funny thing is going on.

src/lib/provable/bigint.ts
boray: The carry can be 116 bits so we can't use 116 bit range check. Thus, we use the next range check which is 128 bits as it's a safe ceiling.

PR #2015, opened 2025-02-07
Bytes: use FlexibleBytes alias directly

dannywillems:
Simply to unify the use of the alias.


querolita (approved):

src/lib/provable/bytes.ts
```diff
@@ -44,7 +44,7 @@ class Bytes {
    *
    * Inputs smaller than `this.size` are padded with zero bytes.
    */
-  static from(data: (UInt8 | bigint | number)[] | Uint8Array | Bytes): Bytes {
+  static from(data: FlexibleBytes): Bytes {
```
mitschabaude, line 47: FYI, I often avoided putting types in method signatures that are not exported, because for users the DX is better when they see an explicit type `(UInt8 | bigint | number)[] | Uint8Array | Bytes` instead of an opaque one.

Trivo25 (commented):
We should probably also export the `FlexibleBytes` type

Shigoto-dev19 (approved):

Trivo25 (approved):

Issue #1819, opened 2024-09-14
dev checkout failed `GIT_LFS_SKIP_SMUDGE=1` needed


jmikedupont2:
following the dev instructions you get this error
```Downloading rfcs/res/hard-fork-package-generation-buildkite-pipeline.dot.png (30 KB)
Error downloading object: rfcs/res/hard-fork-package-generation-buildkite-pipeline.dot.png (646e541): Smudge error: Er\
ror downloading rfcs/res/hard-fork-package-generation-buildkite-pipeline.dot.png (646e541b20428322a85dac4f07aa67ad2e66\
b732ca3b3ba1932410c7b64d2c0d): batch response: This repository is over its data quota. Account responsible for LFS ban\
dwidth should purchase more data packs to restore access.```

PR #2019, opened 2025-02-10
set `GIT_LFS_SKIP_SMUDGE` to 1 in .envrc

glyh:
This is more of a follow up on https://github.com/o1-labs/o1js/issues/1819. 

To help direnv users to prevent LFS from panicking, we should set `GIT_LFS_SKIP_SMUDGE` to 1.


Trivo25 (approved):
thanks!

PR #2024, opened 2025-02-10
Optimize bigint add method

Shigoto-dev19:
## Changes

- Replaced the add method with a new optimized one.

## Description 

- Supports unreduced bigint entries where  `this`, `a` > `p`
- More efficient than the vanilla `addMod` implementation
  - Total rows: `483` compared to `1895`
- Returns the **remainder** modulo **p**. 

## TODO

- Add a PR for more efficient version for a `double` method.

### DIFF
--- a/src/lib/provable/bigint.ts
+++ b/src/lib/provable/bigint.ts
@@ -176,53 +176,75 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     /**
      * Adds two ProvableBigInt instances
      * @param a The ProvableBigInt to add
-     * @returns The sum as a ProvableBigInt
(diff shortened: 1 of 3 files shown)


src/lib/provable/bigint.ts
```diff
@@ -176,53 +176,75 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     /**
      * Adds two ProvableBigInt instances
      * @param a The ProvableBigInt to add
-     * @returns The sum as a ProvableBigInt
+     * @returns The remainder of the sum as a ProvableBigInt
```
boray, line 179:
```suggestion
     * @returns The sum as a ProvableBigInt
```

PR #2025, opened 2025-02-11
Optimize bigint sub method

Shigoto-dev19:
## Changes

- Replaced the sub method with a new optimized one.

## Description 

- Supports unreduced bigint entries where  `this`, `a` > `p`
- More efficient than the vanilla `subMod` implementation
  - Total rows: `483` compared to `558`
- Returns the **remainder** of the difference modulo **p**.

### DIFF
--- a/src/lib/provable/bigint.ts
+++ b/src/lib/provable/bigint.ts
@@ -228,30 +228,74 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     /**
      * Subtracts one ProvableBigInt from another
      * @param a The ProvableBigInt to substract
-     * @returns The difference as a ProvableBigInt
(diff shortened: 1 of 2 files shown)


src/lib/provable/bigint.ts
```diff
@@ -228,30 +228,74 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     /**
      * Subtracts one ProvableBigInt from another
      * @param a The ProvableBigInt to substract
-     * @returns The difference as a ProvableBigInt
+     * @returns The remainder of the difference as a ProvableBigInt
```
boray, line 231:
```suggestion
     * @returns The difference as a ProvableBigInt
```

boray (approved):
LGTM!

PR #2027, opened 2025-02-11
Fix the bigint div method

Shigoto-dev19:
## Changes

- Updated the `div` method for modular arithmetic.
- Previously, the method returned the quotient and remainder from bigint division, which is incorrect for the following reasons:
  - For bigints `x` and `y` with modulus `p`, the `div` method should compute `x * (y^(-1) mod p)` rather than performing integer division as in JavaScript's bigint division.
  - Witnessing a quotient `q` and remainder `r` with the constraint `q.mul(a).add(r).assertEquals(this);` is insecure. Any pair `(q, r)` satisfying the constraint would pass, allowing a malicious prover to manipulate these values.
  - Witnessing the unique multiplicative inverse and asserting that its multiplication yields `1 mod p` is secure. This approach prevents any alternative valid values that could break the constraint.

## Description

- Only supports reduced bigint entries where both `this` and `a` are less than `p`.  
   - Note: This requirement exists because the `mul` method only supports reduced bigints._

- Total rows for Bigint2048: `1912`  
  _Note: The implementation is relatively expensive due to **two** provable bigint `multiplications` and one bigint `assertEquals`._

- Returns the **quotient** modulo **p**.

## TODO

- Adapt tests to align with the new multiplicative inverse logic.  
  (This might require implementing a separate bigint inverse function as a testing utility.)

### DIFF
--- a/src/lib/provable/bigint.ts
+++ b/src/lib/provable/bigint.ts
@@ -247,6 +247,10 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
       return r;
     }
 
+    double(): ProvableBigInt_ {
+      return this.add(this, true);
+    }
+
     /**
      * Subtracts one ProvableBigInt from another
      * @param a The ProvableBigInt to substract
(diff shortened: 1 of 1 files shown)


src/lib/provable/bigint.ts
```diff
@@ -336,35 +336,16 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     }
 
     /**
-     * Divides one ProvableBigInt by another and returns the quotient and remainder
+     * Divides one ProvableBigInt by another.
      * @param a The divisor as a ProvableBigInt
-     * @returns An object containing the quotient and remainder as ProvableBigInt
+     * @returns The remainder as ProvableBigInt
```
boray, line 341:
```suggestion
     * @returns The quotient as ProvableBigInt
```

src/lib/provable/test/bigint.test.ts
```diff
@@ -160,25 +160,22 @@ describe('BigInt17', () => {
...
       const b = BigInt17.fromBigint(3n);
       const result = a.div(b);
-      expect(result.quotient.toBigint()).toStrictEqual(
-        a.toBigint() / b.toBigint()
-      );
-      expect(result.remainder.toBigint()).toStrictEqual(
-        a.toBigint() % b.toBigint()
-      );
+      expect(result.mul(b).toBigint()).toStrictEqual(a.toBigint());
     });
```
boray, line 173: We need to add a case to make sure division by zero throws an error.

boray (changes requested):

boray (approved):
LGTM!

PR #2028, opened 2025-02-11
Add a new bigint mod method

Shigoto-dev19:
## Changes

- Add a new `mod` method.

## Description

- Supports unreduced bigint entries where `this` >`p`.  
  - Note: This method is only applicable to a bigint library, not to a ForeignField-like implementation, which assumes unreduced entries.
- Total rows: `429`
-  Returns the **remainder** modulo **p**.

## TODO

- Add unit tests.


boray (approved):
LGTM!

PR #2029, opened 2025-02-11
Add a new bigint double method

Shigoto-dev19:
## Changes

- Add a new bigint `double` method.

## Description 

- Supports unreduced bigint entries where  `this` > `p`
- More efficient than the `add` method if adding a bigint to itself is needed.
  - Total rows (Bigint2048): `429` compared to `483` for the add method.
- Returns the **remainder** modulo **p**. 

## TODO

- Add unit tests.


boray (approved):
LGTM!

PR #2033, opened 2025-02-13
Force-format all the files

Trivo25:
runs `npm run format` on all the top level o1js files, otherwise we'll indefinitely get formatting diffs in new PRs, which makes it hard to review them


45930:
Sorry, I will revert the bindings changes.  Also looks like some of the remaining warnings we from the examples dir, and we can probably ignore those.  We may define variables there just for demonstration purposes.

45930:
@Trivo25 I reverted all my changes, feel free to force a rebase back to `e4beadc8cd30f668b7e656678f80a75be78d255a`.  It turns out `--fix-dangerously` applies dangerous fixes!  I spotted some bad changes and it doesn't seem worth the risk.  I think the bulk of the changes from the safe formatting should be sufficient to create fewer diffs in the future.

45930 (approved):

mitschabaude:
so this happens when you change the prettier print width from 80 to 100? Love it
image

PR #2036, opened 2025-02-13
Seperate API versions

Trivo25:
Just a little "chore", moves the current Mina API into a designated v1 folder. I am doing this in a seperate PR to keep the diff in follow up PRs small and still reviewable. This will make it easier for internal and external contributors to follow the implementation of the new v2 API

bindings https://github.com/o1-labs/o1js-bindings/pull/338


ymekuria (approved):

45930 (approved):
o1js-mina seems like a good target for a separate package.  It would make this whole rewrite a lot easier IMO.  Do you want to consider building the v2 API as a standalone package for interacting with Mina?

Trivo25:
> o1js-mina seems like a good target for a separate package. It would make this whole rewrite a lot easier IMO. Do you want to consider building the v2 API as a standalone package for interacting with Mina?

I think we could aim to seperate that eventually, but right now I feel like it would slow everything down a bit

PR #2042, opened 2025-02-19
Transaction construction

Trivo25:


src/lib/proof-system/zkprogram.ts
```diff
@@ -62,6 +59,7 @@ export {
   RegularProver,
   TupleToInstances,
   PrivateInput,
+  Proof,
```
mitschabaude, line 62: seems unnecessary to re-export?

45930 (approved):
I think this is ok to start getting feedback on.  As I scan through the code, I wish we just had a `mina.d.ts` file that could be "the spec" of all the types in Mina and how they interact.  The behavior of accounts and permissions and transactions should all clearly be in their own files, but it might improve readability if we centralized just the types.  WDYT?

If you're open to the idea, I could take a stab on Friday to see if it feels natural.  I'm not trying to add busywork to your plate, but I think this could be a feed 2 birds with one scone situation where we get an o1js source of truth for all the Mina types, and the readability of all the other files is improved.

src/examples/v2/fund-transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+    publicKey: feePayerAddress,
+    fee: new UInt64(fee),
+    nonce: new UInt32(4),
+  }),
+  accountUpdates: [
+    new AccountUpdate('GenericState', GenericData, GenericData, {
+      authorizationKind: AccountUpdateAuthorizationKind.Signature(),
+      accountId: new AccountId(feePayerAddress, TokenId.MINA),
+      balanceChange: Int64.from(totalAmount + accountCreationFee).neg(),
+      incrementNonce: Bool(true),
```
mitschabaude, line 38: It would be nice if you could just pass in a plain `number` or `bigint` for the `balanceChange`, and a plain `boolean` for `incrementNonce`.

if the main API is going to be specifying every account update explicitly, I think it's worth making that as intuitive and non-verbose as possible!

src/examples/v2/fund-transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+const paymentCommand = await new ZkappCommand({
+  feePayment: new ZkappFeePayment({
+    publicKey: feePayerAddress,
+    fee: new UInt64(fee),
+    nonce: new UInt32(4),
+  }),
+  accountUpdates: [
+    new AccountUpdate('GenericState', GenericData, GenericData, {
+      authorizationKind: AccountUpdateAuthorizationKind.Signature(),
+      accountId: new AccountId(feePayerAddress, TokenId.MINA),
```
mitschabaude, line 36: I like the educational value of combining public key + token id into a single `accountId` field!
Also `TokenId.MINA` feels obvious and should've been added much earlier.

I'm not sure I like the approach of having a class (`AccountId`) for something that works well as a plain object. But if the class is not required i.e. the input type is `{ publicKey, tokenId }` it's fine

src/examples/v2/fund-transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+      authorizationKind: AccountUpdateAuthorizationKind.None(),
+      accountId: new AccountId(receiverAddress, TokenId.MINA),
+      balanceChange: Int64.from(amount),
+    }),
+  ],
+}).authorize({
+  networkId: 'testnet',
+  async getPrivateKey(pk: PublicKey): Promise<PrivateKey> {
+    if (pk === feePayerAddress) return feePayerKey;
+    throw new Error("Don't have the private key for this public key");
```
mitschabaude, line 50: might be worth making the return type `PrivateKey | undefined` so that the user doesn't have to manually throw an error

src/examples/v2/fund-transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+    }),
+    new AccountUpdate('GenericState', GenericData, GenericData, {
+      authorizationKind: AccountUpdateAuthorizationKind.None(),
+      accountId: new AccountId(receiverAddress, TokenId.MINA),
+      balanceChange: Int64.from(amount),
+    }),
+  ],
+}).authorize({
+  networkId: 'testnet',
+  async getPrivateKey(pk: PublicKey): Promise<PrivateKey> {
```
mitschabaude, line 48: curious why this was made a callback! just passing in the private keys seemed like an obvious API

src/examples/v2/fund-transfer.ts
```diff
@@ -0,0 +1,58 @@
...
+
+// build a manual zkapp command that transfers funds from the fee payer to the receiver address
+const paymentCommand = await new ZkappCommand({
+  feePayment: new ZkappFeePayment({
+    publicKey: feePayerAddress,
+    fee: new UInt64(fee),
+    nonce: new UInt32(4),
+  }),
+  accountUpdates: [
+    new AccountUpdate('GenericState', GenericData, GenericData, {
```
mitschabaude, line 34: feels tough to require a state + events + action layout on _every_ account update. this example shows why: often you'll create AUs where state/events/actions don't matter.

maybe there should be two different constructors? `AccountUpdate` and `ZkappAccountUpdate` or similar? and only the latter requires the layouts?

src/lib/provable/types/unconstrained.ts
```diff
@@ -39,7 +39,7 @@ export { Unconstrained };
 class Unconstrained<T> {
   private option: { isSome: true; value: T } | { isSome: false; value: undefined };
 
-  private constructor(isSome: boolean, value?: T) {
+  constructor(isSome: boolean, value?: T) {
```
mitschabaude, line 42: why was this necessary? breaks the API contract of `Unconstrained` :/

mitschabaude (commented):
@Trivo25 a bit of random feedback on the API! (didn't look at the code)

Trivo25:
> @Trivo25 a bit of random feedback on the API! (didn't look at the code)

Much appreicated! There's still a lot of gaps and holes in this implementation, i'll start collecting feedback from everywhere and make sure that we have a plan to address it

PR #2061, opened 2025-02-23
Update issue templates

Trivo25:


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,16 @@
...
+
+---
+
+**Describe the feature or improvement**
+
+_A clear and concise description of the proposed feature or change. Eg. It's important for zkApp developers to prove circuits in <20s_  
+
+**Describe the solution**
+
+_A clear _technical_ description of the implementation._
```
45930, line 16: The solution may not be known for many features.  Should we have a different template for "wishlist" features and feature requests with a specific solution in mind?

45930 (approved):
I like the direction of having these templates.  These seem like a good start and we can build from here.

PR #2063, opened 2025-02-24
o1js: bump up Rust version to nightly-1.79.0

dannywillems:
Related to https://github.com/MinaProtocol/mina/pull/16652.

Katie reported some issues while she was working on Kimchi and while trying to update wasm-bindgen - which is only available for more recent Rust versions.
In DM, we decided to bump up Rust in Mina. An easy path is to use Rust 1.79 as it is the latest version that is currently supported and tested in o1-labs/proofs-systems. It doesn't require other changes that simply bumping up the Rust toolchain in Mina.
More recent version will be supported right after, if requested.


Trivo25 (approved):

hattyhattington17 (approved):

dannywillems:
Merging even though "Build o1js bindings / build-bindings-ubuntu (pull_request)" is red.

PR #2065, opened 2025-02-26
Export Internal Testing Tools

boray:
This PR exposes existing internal testing tools for use of third party primitive developers.
### Changes:
* Created a`Testing` namespace in `src/lib/testing/testing.ts` that organizes the testing tools.
* `Testing` namespace is exported from `Experimental`.
* Added an example in `src/examples/testing.ts` to demonstrate the use of `Testing` utilities.

### DIFF
--- a/src/build/build-web.js
+++ b/src/build/build-web.js
@@ -89,6 +89,7 @@ async function buildWeb({ production }) {
     allowOverwrite: true,
     logLevel: 'error',
     minify,
+    external: ['node:assert/strict'],
   });
 }
 
--- a/src/build/build-web.js
+++ b/src/build/build-web.js
@@ -84,12 +84,11 @@ async function buildWeb({ production }) {
     resolveExtensions: ['.js', '.ts'],
     plugins: [wasmPlugin(), srcStringPlugin()],
     dropLabels: ['CJS'],
-    external: ['*.bc.js'],
+    external: ['*.bc.js', 'node:assert/strict'],
     target,
     allowOverwrite: true,
     logLevel: 'error',
     minify,
-    external: ['node:assert/strict'],
   });
 }
 
@@ -182,7 +181,7 @@ function srcStringPlugin() {
   };
 }
 
-function deferExecutionPlugin() {
+function _deferExecutionPlugin() {
   return {
     name: 'defer-execution-plugin',
     setup(build) {
--- a/src/lib/testing/constraint-system.ts
+++ b/src/lib/testing/constraint-system.ts
@@ -37,7 +37,10 @@ export {
 };
 
 // TODO get rid of this top-level await by making `test` support async functions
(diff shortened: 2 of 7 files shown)


src/build/build-web.js
```diff
@@ -84,7 +84,7 @@ async function buildWeb({ production }) {
     resolveExtensions: ['.js', '.ts'],
     plugins: [wasmPlugin(), srcStringPlugin()],
     dropLabels: ['CJS'],
-    external: ['*.bc.js'],
+    external: ['*.bc.js', 'node:assert/strict'],
```
mitschabaude, line 87: this breaks the web version doesn't it?

src/lib/testing/constraint-system.ts
```diff
@@ -37,7 +37,10 @@ export {
 };
 
 // TODO get rid of this top-level await by making `test` support async functions
-let { constraintSystemSync } = await synchronousRunners();
+let constraintSystemSync: any;
+(async () => {
+  constraintSystemSync = await synchronousRunners();
+})();
```
mitschabaude, line 43: how is this supposed to work? to me it seems to impose an ugly implicit assumption that bindings are initiated before using the testing framework. this should be made explicit

src/lib/testing/testing.ts
```diff
@@ -0,0 +1,51 @@
+import * as Property from './property.js';
+import * as Equivalent from './equivalent.js';
+import * as ConstraintSystem from './constraint-system.js';
+
+export { Testing };
+
+namespace Testing {
```
mitschabaude, line 7: I don't like how you pack unqualified names like `or` and `bool`, that have nothing to do with each other, into a flat namespace. why not export three namespaces?

mitschabaude (changes requested):
So a problem seems to be that the testing framework relies on Nodejs imports, and this isn't compatible with the web version.

On the other hand, `o1js` is designed to work in all JS runtimes.

I think for this reason it would be cleaner to export the testing framework as a separate package or package entry-point, like we already do with mina-signer.

This is a general direction we always wanted to move in: Have separate entry-points like `o1js/light`, `o1js/prover` and `o1js/testing` with different levels of functionality

src/lib/testing/constraint-system.ts
boray: this doesn't work. I was testing this to prevent having a top-level await.

src/lib/testing/testing.ts
boray: I don't like either.  I am just experimenting for now.

src/build/build-web.js
boray: yes, I wasn't sure how acceptable is this since testing tools won't be used in web but it still breaks the web version.

PR #2076, opened 2025-03-11
replace src/bindings submodule with contents

### LINKED ISSUE #2080: get rid of o1js-bindings as submodule
move all the files into o1js directly

### LINKED ISSUE #2133: artefacts should be built on-demand and no longer committed
- [x] the o1js-bindings submodule is removed from o1js
- [ ] The o1js-bindings repository is archived
- [x] All code from o1js-bindings is moved into the main o1js repository
- [x] WASM and jsoo artefacts are no longer committed to any repository
- [x] Artefacts can be generated locally by developers when needed
- [x] CI builds artefacts on tagged releases and publishes them to npm
- [x] CI can generate artefacts for non-release branches if needed (e.g. via manual trigger)
- [x] Prebuilt artefacts are uploaded to a temporary GCP Artifact Registry bucket
- [x] Developer tooling downloads artefacts automatically if no local build is available
- [x] Artefacts stored in GCP Artifact Registry are set to expire after a defined period (e.g., 7 days)
- [x] Developers without a working Nix or native build can still contribute to o1js
- [x] Clear steps for how to contribute to o1js as an external developer are outlined in a README
- [x] Build bindings on every PR that goes into main, pull bindings from latest main commit

-----
(speak to Luis for the GCP Artifact Registry bucket)

Bindings are not checked into the repository anymore, CI builds them for releases. We will outline clear steps for external contributors. If they dont want to or cant set up Nix/native builds, they will be able to use a manual CI workflow that builds the bindings off of a specific commit or branch and a script that automatically downloads these artifacts from a temporary GCP Artifact Registry bucket (artifacts expire after a few days to avoid bloat). If this becomes too much, we can also rate limit the job.

```sh
gh workflow run build-artefacts.yml \
  --ref my-branch \
  -f custom_ref=commit
```

Ideally the script to build the bindings is as easy and straightforward as possible. The script is executed, it triggered a build job, the artifacts are stored in a bucket, the ci job returns the bucket download link, the link automatically downloads the artifacts

Geometer1729:
- Replaces the o1js-bindings submodule with the same files but not as a submodle
- changes linting rules to be consistent with previous behavior
- Gitignore precompiled code
- remove `MINA_COMMIT` file entirely
- `npm run build:bindings-download` downloads compiled artifacts from github
- `npm run build:bindings-remote` triggers a github workflow to build the compiled artifacts then downloads them
- Remove the `Building the o1js bindings` section from `src/bindings/README.md` (it's outdated and redundant to README-dev anyway)

- replace the github suggestion for npmDepsHash with the auto-commit action
I tried forking the action to allow changes in files not in the pr but it turns out github doesn't support that.

`npm run:build` also now downloads the bindings if `src/bindings/compiled` does not exist. This should make contributing fairly painless, but if you do commit before running  `npm run:build` you get:
```
bindings have not been built for this commit
you may want to:
- switch to a commit where they have been built
- trigger a remote build with npm run build:bindings-remote
- open a pr and run ci
- if you are running remote-build you may not have pushed your latest commit
```
You do also now need the `gh` cli tool, but I confirmed that for `npm run:build` and `npm run build:bindings-download` you do not need to be in the o1labs org.

I apologize for the lengthy commit history (lots of troubleshooting ci) and for the large number of files.
The majority of the file changes are just adding files from o1js-bindings. Here are some diff command outputs to make that more reviewable: 
```
$diff -rq src/bindings ../o1js-bindings
Only in ../o1js-bindings: CODEOWNERS
Only in ../o1js-bindings/compiled: .gitignore
Only in ../o1js-bindings: .git
Only in ../o1js-bindings: .gitattributes
Only in ../o1js-bindings: .gitconfig
Only in ../o1js-bindings: .github
Only in ../o1js-bindings: LICENSE
Only in ../o1js-bindings: MINA_COMMIT
Only in ../o1js-bindings: .prettierrc.cjs
Files src/bindings/README.md and ../o1js-bindings/README.md differ
Only in ../o1js-bindings/scripts: check-commit.sh
Only in src/bindings/scripts: download-bindings.sh
Only in src/bindings/scripts: remote-build-bindings.sh
Files src/bindings/scripts/update-o1js-bindings.sh and ../o1js-bindings/scripts/update-o1js-bindings.sh differ

```
```
$diff src/bindings/scripts/update-o1js-bindings.sh ../o1js-bindings/scripts/update-o1js-bindings.sh
81a82,90
>
> # 3. update MINA_COMMIT file in o1js
>
> if [ -z "${SKIP_MINA_COMMIT}" ]
> then
> MINA_COMMIT=$(git -C src/mina rev-parse HEAD)
> echo "The mina commit used to generate the backends for node and web is" "$MINA_COMMIT" \
>   > src/bindings/MINA_COMMIT
> fi

```

Closes #2080 
Closes #2133


.github/workflows/build-bindings.yml
```diff
@@ -57,32 +51,40 @@ jobs:
...
         with:
-          name: bindings.patch
-          path: bindings.patch
\ No newline at end of file
+          name: bindings.tar.gz
+          path: bindings.tar.gz
+      - name: add build to gc-root if on main
+        if: github.ref == 'refs/heads/main'
+        run: |
+          nix build o1js#o1js-bindings --out-link /home/app/actions-runner/nix-cache/main-bindings-gcroot
```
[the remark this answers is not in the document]
Geometer1729: This won't actually build them cause the build should always be cached here. It just keeps a gcroot pointing to the latest commit on main so the nix garbage collection removes fewer derivations which are likely to be useful.

.github/workflows/build-bindings.yml
```diff
@@ -57,32 +51,40 @@ jobs:
...
+      - name: cache bindings
+        uses: actions/cache@v4
+        id: bindings-cache
+        with:
+          path: |
+            src/bindings/compiled
+            src/bindings/mina-transaction/gen
+          key: bindings-${{ github.event.pull_request.head.sha || github.sha }}
+
+  # seperate job because the rest of ci has to wait on the above job
```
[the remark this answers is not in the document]
Geometer1729: What I meant with this comment is that it would be more natural for this to all be one job but I split it here because the other workflows waiting for the cache have to wait for a job not a step so the split lets them wait a bit less. I'll update the comment to be more clear.

.github/workflows/pkg-pr-new-publish.yml
```diff
@@ -24,7 +24,11 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: ${{ matrix.node }}
+
       - name: Build o1js and mina-signer
+        env:
+          # needed to download bindings
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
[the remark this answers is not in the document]
Geometer1729: Only for `npm prepublishOnly`, it's required for `gh` when it downloads the bindings.
I tried to use the cache here but I couldn't get it to work. Apparently there are rules about when caches are visible more complex than just having the right key. The pull request trigger scopes them with `refs/pull/...` and then `workflow_run` only looks in `refs/heads/..` I think.

I hadn't thought about the security implications of having that in the environment. I'll just run the download bindings script directly in a separate step. I think that makes it hardest to have a subtle malicious pr.

README-dev.md
```diff
@@ -30,10 +30,19 @@ For most users, building o1js is as simple as running:
...
+This command downloads the artifacts from github if they are missing and compiles the TypeScript source files, making them ready for use.
+The compiled OCaml and WebAssembly artifacts are cached for each commit where ci is run or where they are manually built ie by `npm run build:bindings-remote`.
+These artifacts are stored under `src/bindings/compiled` and `src/bindings/mina-transaction/gen` and contain the artifacts needed for both node and web builds.
+These files only have to be regenerated if there are changes to the OCaml or Rust source files.
+Keep in mind merging a newer version of o1js may include Ocaml and rust changes so you may need to rerun `npm run build:bindings-download`.
+If this fails because the artifacts haven't been built for that commit.
+You can also run `REV=<some_git_rev> ./src/bindings/scripts/download-bindings.sh` to download artifacts for a different commit.
```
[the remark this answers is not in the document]
Geometer1729: You can. I'll suggest that in the docs instead.

src/bindings/README.md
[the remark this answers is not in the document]
Geometer1729: I'll just delete the whole building the o1js bindings section. There's a bunch of outdated stuff and `README-dev` covers it better anyway.

.github/workflows/nix-clean.yml
```diff
@@ -0,0 +1,25 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  nix-cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up Nix
+        runs-on: [sdk-self-hosted-linux-amd64-build-system]
```
boray, line 19: Is this line valid? Is this job supposed to run on a GitHub hosted runner or self-hosted?

boray (commented):
I am genuinely excited for this PR. I add a few comments.

src/bindings/scripts/download-bindings.sh
```diff
@@ -0,0 +1,24 @@
...
+  exit 1
+fi
+
+gh run watch "$RUN_ID" --exit-status || echo "Warning: ci failed on this job, trying to download bindings anyway"
+#remove if it exists but don't fail
+if [ -f ".bindings_download/bindings.tar.gz" ]; then
+  rm .bindings_download/bindings.tar.gz
+fi
+gh run download "$RUN_ID" --dir .bindings_download --name=bindings.tar.gz
+tar xzf .bindings_download/bindings.tar.gz -C src/bindings --overwrite
```
boray, line 24: I am on macOS, this line didn't work for me. It seems not all implementations of `tar` have the `--overwrite` flag. I replaced this line with those below and it worked for me. It may not be the most elegant solution. I leave it to you to decide the best approach.

```bash
rimraf src/bindings/mina-transaction/gen/*.js src/bindings/mina-transaction/gen/*.ts src/bindings/compiled 
tar -xf .bindings_download/bindings.tar.gz -C src/bindings
```

README-dev.md
```diff
@@ -30,10 +30,19 @@ For most users, building o1js is as simple as running:
 
 ```sh
 npm install
+gh auth login #gh is used to download the compiled artifacts
 npm run build
 ```
 
-This command compiles the TypeScript source files, making them ready for use. The compiled OCaml and WebAssembly artifacts are version-controlled to simplify the build process for end users. These artifacts are stored under `src/bindings/compiled` and contain the artifacts needed for both node and web builds. These files only have to be regenerated if there are changes to the OCaml or Rust source files.
+This command downloads the artifacts from github if they are missing and compiles the TypeScript source files, making them ready for use.
```
Trivo25, line 37: Can we add a section explicitly directed at external contributors that is easy to find for new developers that want to contribute to o1js?

.github/workflows/nix-clean.yml
Geometer1729: It should run on our self hosted machine, good catch thanks.

src/bindings/scripts/download-bindings.sh
Geometer1729: While fixing this I noticed my `.gitignore` is wrong there are now `v1` and `v2` directories in `src/bindings/mina-transaction/gen/` I'll remove the old bindings with `git clean -fdx src/bindings/compiled src/bindings/mina-transaction/gen` to try to avoid duplicating that shape everywhere.

README-dev.md
Geometer1729: I added subsections for external and internal contributors, I think it reads better now.

.github/workflows/build-action.yml
```diff
@@ -32,7 +32,20 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Restore bindings-cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            src/bindings/compiled
+            src/bindings/mina-transaction/gen
+          key: bindings-${{ github.event.pull_request.head.sha || github.sha }}
+          fail-on-cache-miss: true
```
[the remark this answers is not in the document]
Geometer1729: This makes a lot of sense. Like you mentioned the `wait for action` has a bit of a code smell. I definitely favor doing that in a new PR.

.github/workflows/nix-clean.yml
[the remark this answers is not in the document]
Geometer1729: It's only coupled in that working on this pr really highlighted how painful running the store cleanup every commit was and while I was refactoring the workflow anyway it felt like a good time to fix it.

:+1: on the name change.

README-dev.md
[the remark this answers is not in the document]
Trivo25: Ooops thats right, didn't scroll up!

README-dev.md
Trivo25, line 1: We might have to add `gh` to the list of required tools on line 14

README-dev.md
```diff
@@ -30,10 +30,19 @@ For most users, building o1js is as simple as running:
...
+The compiled OCaml and WebAssembly artifacts are cached for each commit where ci is run or where they are manually built ie by `npm run build:bindings-remote`.
+These artifacts are stored under `src/bindings/compiled` and `src/bindings/mina-transaction/gen` and contain the artifacts needed for both node and web builds.
+These files only have to be regenerated if there are changes to the OCaml or Rust source files.
+Keep in mind merging a newer version of o1js may include Ocaml and rust changes so you may need to rerun `npm run build:bindings-download`.
```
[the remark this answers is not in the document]
Geometer1729: That's probably a good idea. It's not quite perfect but hashing `src/bindings` and the mina commit would give a pretty good indication of rather the bindings are the same.

We could give a warning when the bindings seem out of data but our builds already have so many warnings it's hard to imagine anyone noticing. We could give an error but that feels annoying for developers. 

What do you think makes sense?

.github/workflows/live-tests.yml
```diff
@@ -37,6 +37,20 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Wait for bindings-cache
+        uses: fountainhead/action-wait-for-check@v1.2.0
```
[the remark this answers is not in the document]
Geometer1729: I haven't audited them, but I've at least pinned every action that uses a secret to a commit now.

Geometer1729:
@richardpringle 
> Is it also possible that we use `curl` as the default when fetching instead of `gh`? If I want to run the tests on main with a fresh clone, I shouldn't need to install gh, right?

Unfortunately github requires credentials to download artifacts. I think gh is the easiest way to do that. To me it doesn't seem worth hosting the artifacts separately to avoid. Maybe there's a better option I'm not seeing.

package.json
```diff
@@ -46,15 +46,18 @@
   },
   "scripts": {
     "dev": "npx tsc -p tsconfig.test.json && node src/build/copy-to-dist.js",
-    "build": "node src/build/copy-artifacts.js && rimraf ./dist/node && npm run dev && node src/build/build-node.js",
+    "build": "npm run checkForBindings && node src/build/copy-artifacts.js && rimraf ./dist/node && npm run dev && node src/build/build-node.js",
     "build:bindings": "./src/bindings/scripts/build-o1js-node.sh",
     "build:update-bindings": "./src/bindings/scripts/update-o1js-bindings.sh",
+    "build:bindings-remote": "./src/bindings/scripts/remote-build-bindings.sh && npm run:build",
```
Trivo25, line 52: `npm run:build` should be `npm run build`

Trivo25 (approved):
Great job, this feels insanely nice - especially compared to the native build experience!

mitschabaude (approved):
it's amazing to see all this code come home to the o1js repo 🎉

.gitignore
```diff
@@ -13,3 +13,9 @@ src/config.mlh
 .direnv
 .rustup
 result
+# precompiled bindings
+src/bindings/compiled
+# allows dune and README.md
+src/bindings/mina-transaction/gen/*/*.js
+src/bindings/mina-transaction/gen/*/*.ts
```
mitschabaude, line 20: is there a reason to ignore these files? they're small and the diff lets you easily see type changes, which is convenient when reviewing

Geometer1729: I just want to stop tracking generated files in git entirely.

I could write a script to download them for a pair of commits and show the diff if that would help.

mitschabaude: personally I would track them. they are much more "part of the codebase" than the bindings files. for example, the TS build won't even work without those files, while bindings are only needed once you compile circuits/create proofs

Geometer1729: Fair enough :+1:

PR #2083, opened 2025-03-17
Export more utility and bindings functions

### LINKED ISSUE #2082: export parts of the bindings through o1js
export some of the bindings in o1js directly - this will give developers more access to the underlying bindings logic (mina interface, crypto and math stuff, etc) - will be exported under its own namespace (eg Bindings) - this will also make it easy to modularize o1js

Trivo25:
see https://github.com/o1-labs/o1js-bindings/pull/345/

closes #2082


src/index.ts
```diff
@@ -188,3 +188,6 @@ namespace Experimental {
 }
 
 Error.stackTraceLimit = 100000;
+
+// export parts of the low-level bindings interface for advanced users
+export * as Bindings from './bindings/index.js';
```
mitschabaude, line 193: IMO "bindings" is not the best name/concept for what you're exporting here. So far, I see finite field and elliptic curve primitives. how about exporting them as part of the `Crypto` namespace instead? (currently used for elliptic curve parameters)

The fact that this code resides in the `/bindings` folder is more of a historical artifact. Arguably the crypto stuff (and some of the other TS code in bindings) should live somewhere else, and `/bindings` just contain the various layers of Ocaml and Rust bindings

Trivo25: > IMO "bindings" is not the best name/concept for what you're exporting here

Yep, we are still looking for a better name. The math and crypto exports are just the beginning, I imagine that we could export other things in the future too

src/lib/provable/gadgets/basic.ts
```diff
@@ -363,3 +365,25 @@ function getLinear(x: ScaledVar | Constant): [[bigint, VarFieldVar], bigint] {
 }
 
 const ScaledVar = { isVar, getVar, isConst, getConst };
+
+/**
+ * Get value from array in O(n) constraints.
+ *
+ * Assumes that index is in [0, n), returns an unconstrained result otherwise.
+ */
+function arrayGetGeneric<T>(type: ProvableType<T>, array: T[], index: Field) {
```
Trivo25, line 374: cherry picked from https://github.com/o1-labs/o1js/pull/1949

src/lib/provable/gadgets/foreign-field.ts
```diff
@@ -406,6 +407,19 @@ function equals(x: Field3, c: bigint, f: bigint) {
   }
 }
 
+// Field3 equality
+function assertEquals(x: Field3, y: Field3) {
```
Trivo25, line 411: cherry picked from https://github.com/o1-labs/o1js/pull/1949

src/lib/util/octets.ts
```diff
@@ -0,0 +1,123 @@
+import { Field, checkBitLength } from '../provable/field.js';
```
Trivo25, line 1: cherry picked from https://github.com/o1-labs/o1js/pull/1949

mitschabaude: this is weirdly placed, it's the only circuit code not located in `src/provable` or `src/mina`

also maybe it'd be more accessible as `field.toBytes()` and `Field.fromBytes()`?

src/lib/util/octets.ts
```diff
@@ -0,0 +1,123 @@
...
+      return Array.from({ length: bytelength }, (_, k) => new UInt8((x >> BigInt(8 * k)) & 0xffn));
+    }
+    let bytes = Provable.witness(Provable.Array(UInt8, bytelength), () => {
+      let x = input.toBigInt();
+      return Array.from({ length: bytelength }, (_, k) => new UInt8((x >> BigInt(8 * k)) & 0xffn));
+    });
+    let field = bytes
+      .reverse()
+      .map((x) => x.value)
+      .reduce((acc, byte) => acc.mul(256).add(byte));
```
mitschabaude, line 47: the default version of this method with 32 bytes is unsound! since there are several representations of a given field element with 256 bits

mitschabaude: compare Field to/from bits, where we only allow 254 bits or less

querolita: You're right, I was using this function only for smaller inputs (limb-length fields actually), but did not add any disclaimer about the soundness for large values when the outcomes are not rangechecked anywhere else. Would you suggest to fix that by adding those checks, limiting it to values which are smaller than the field length, or just remove the `Octets` utility entirely?

querolita (approved):
I limited the use of `toProvableBytes` to prevent the issue pointed out by @mitschabaude . Also added some comments and extended docs.

src/lib/util/octets.ts
querolita: Actually the reason that it went to its own util file was that I was getting circular dependencies preventing me from calling it from `Field` directly, which was my first attempt. Seems like that has been solved here.

src/lib/provable/gadgets/foreign-field.ts
```diff
@@ -471,8 +487,91 @@ const Field3 = {
...
+ *
+ * @param input - the field element to convert to bytes.
+ * @param bytelength - the number of bytes to fit the element. If the element
+ *                     does not fit in `length` bits, the functions throws an
+ *                     error.
+ *
+ * @return An array of {@link UInt8} element representing this {@link Field} in
+ *         little endian encoding.
+ */
+function toProvableBytes(input: Field, byteLength: number = 32) {
```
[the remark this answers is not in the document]
mitschabaude: I see that the method didn't end up getting merged, but just want to point out that limiting to 254 bits and then rounding up to 32 bytes does **not** fix the soundness issue. Limiting to 31 bytes does.

All of this is no concern within a foreign field gadget, because there each limb has at most 88 bits = 11 bytes. It's just a concern for to/from bytes methods on _general_ field elements.

PR #2085, opened 2025-03-19
Remove benchmarks from CI

richardpringle:
The benchmarks were flaking and they were being used for a regular check before merging PR. This PR removes said benchmarks.


Trivo25 (approved):
yaay!

PR #2091, opened 2025-03-20
Expose static contract method for setting key

### LINKED ISSUE #1843: Add SmartContract.setVerificationKeyUnsafe(vk: VerificationKey)
When building the SmartContract deploy transaction without calling `compile()`, o1js refuses to build, asking for the verification key to be present. In frontend, it does not make sense to compile the SmartContract if the proving is being made on backend. The workaround now is to directly set the `_verificationKey` variable using pre-calculated verification keys for different networks:

```typescript
import { verificationKeys } from "./vk";

const adminContractVerificationKey = verificationKeys[chain]?.admin;
const tokenContractVerificationKey = verificationKeys[chain]?.token;

FungibleTokenAdmin._verificationKey = {
      hash: Field(adminContractVerificationKey.hash),
      data: adminContractVerificationKey.data,
    };
FungibleToken._verificationKey = {
      hash: Field(tokenContractVerificationKey.hash),
      data: tokenContractVerificationKey.data,
    };
```
The variable  `_verificationKey` is an internal one and can be changed in future o1js versions.

### Proposal:
Add SmartContract.setVerificationKeyUnsafe(vk: VerificationKey) to set verification key of the SmartContract without compiling it.

richardpringle:
Expose a static method for manually setting the verification key

closes #1843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
   - New transaction construction API `new ZkappCommand()`. https://github.com/o1-labs/o1js/pull/2042
 - Bump up Rust version to 1.79.0. Bindings now depends on nightly-2024-06-13.
   https://github.com/o1-labs/o1js/pull/2063
+- `setVerificationKeyUnsafe` static method to `SmartContract` [#2091](httpts://github.com/o1-labs/o1js/pull/2091)
 
 ## [2.3.0](https://github.com/o1-labs/o1js/compare/b857516...fb625f)
 
--- a/src/lib/mina/v1/zkapp.ts
+++ b/src/lib/mina/v1/zkapp.ts
@@ -635,6 +635,16 @@ class SmartContract extends SmartContractBase {
   }
 
   /**
+   * Manually set the verificaiton key.
+   */
+  static setVerificationKeyUnsafe(verificationKey: { data: string; hash: Field | string }) {
+    SmartContract._verificationKey = {
+      data: verificationKey.data,
+      hash: Field.from(verificationKey.hash),
+    };
+  }
+
+  /**
    * Deploys a {@link SmartContract}.
    *
    * ```ts


CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/fb625f...HEAD)
 
+### Added
+
+- `setVerificationKeyUnsafe` static method to `SmartContract`
```
Trivo25, line 22: fyi we usually try to link to the PR as well https://github.com/o1-labs/o1js/pull/2091/

Trivo25 (approved):

Geometer1729 (approved):

src/lib/mina/v1/zkapp.ts
```diff
@@ -634,6 +634,16 @@ class SmartContract extends SmartContractBase {
     return computeMaxProofsVerified(Object.values(methodData).map((d) => d.proofs.length));
   }
 
+  /**
+   * Manually set the verificaiton key.
+   */
+  static setVerificationKeyUnsafe(verificationKey: { data: string; hash: Field | string }) {
```
45930, line 640: I don't think the type `string` belongs in our source code, but I don't need to die on the hill right now.

There is a type `Base64VerificationKeyString` which is an alias of string at the moment, but I prefer to use it in cases where that's what we actually mean.  If we ever scope down what is allowed in this type, it is nicer to already have a semantic alias for what we mean.

Also `static _verificationKey?: { data: string; hash: Field };` is already used in this file.  Why do we allow a Field or string in this method param (I noticed that there is no type error by just removing `| string`, so is there a downstream issue)?  I'd rather see a proper `VerificationKey` type used in both places.

45930: What is the intended usage of `SmartContract.setVerificationKeyUnsafe`?  It seems like it should be an instance method, not a class method right?  We want to set the vk on an instance of a smart contract.

There is a relevant test at `src/examples/zkapps/dex/upgradability.ts` - could you add a case there maybe to show the usage of this?

45930: Oh, I see the intended use from the issue!

45930: It's "fine" but it's not taking full advantage of typescript to telegraph how to use the API.  When I say we shouldn't use `string`, I don't mean we shouldn't accept string inputs into the method.  I mean that it's an overly broad type annotation.  This method does not actually accept a string.  It accepts a string representation of a base64 encoded ocaml struct, and string representation of a Poseidon hash.  We have an opportunity to communicate that to the developer explicitly with types, and we're not taking advantage of it.

Like I said, I don't need to die on this hill for this PR, but as an on-boarding exercise, I do think there could be room for improving the clarity argument types.

45930 (approved):

PR #2093, opened 2025-03-24
@o1js/testing prototype

boray:


src/testing/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "types": "./dist/node/testing/index.d.ts",
+    "node": "./dist/node/testing/index.js",
+    "default": "./dist/node/testing/index.js"
+  },
+  "files": [
+    "dist",
+    "README.md"
+  ],
+  "peerDependencies": {
+    "o1js": "^2.0.0"
```
45930, line 36: Should we target `>=2.0.0` instead of `^2.0.0`?

45930 (commented):
Overall this looks reasonable!  Can I try using it locally with `npm link`?  I tried from the `src/lib/testing` folder and it didn't work.

src/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+```bash
+npm install --save-dev @o1js/testing
+# or with yarn:
+yarn add -D @o1js/testing
+```
+
+Note that this package requires `o1js` as a peer dependency, so make sure you have it installed:
+
+```bash
+npm install --save o1js
```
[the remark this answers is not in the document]
boray: I think it's customary to put that, at least in most of the npm packages.

boray:
> Overall this looks reasonable! Can I try using it locally with `npm link`? I tried from the `src/lib/testing` folder and it didn't work.

I used [yalc](https://github.com/wclr/yalc) to test the package so far. It should work with `npm link` too. Can you explain how it failed?

45930:
For example, the react testing library doesn’t have a code block to install react.  It just mentions that it’s a peer dependency. https://github.com/testing-library/react-testing-library
I don’t think it’s a big deal either way.  But it just seems like anyone installing this package will know how to install o1js too.

  
On Thu, Apr 3, 2025 at 4:50 PM, Boray Saygılıer ***@***.***&gt; wrote:       Overall this looks reasonable! Can I try using it locally with npm link? I tried from the src/lib/testing folder and it didn't work.   I used yalc to test the package so far. It should work with npm link too. Can you explain how it failed?—
Reply to this email directly, view it on GitHub, or unsubscribe.
You are receiving this because you commented.Message ID: ***@***.***&gt;  
  boray left a comment  (o1-labs/o1js#2093)
   Overall this looks reasonable! Can I try using it locally with npm link? I tried from the src/lib/testing folder and it didn't work.  I used yalc to test the package so far. It should work with npm link too. Can you explain how it failed? —

boray:
> I don’t think it’s a big deal either way. But it just seems like anyone installing this package will know how to install o1js too.

oops, I missed it's for o1js not @o1js/testing. Indeed, it's an overkill. I changed this while ago but not commited. It just has a note about peer dependency. The current version of the readme is incomplete. I'll be pushing updates tomorrow.

src/testing/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@o1js/testing",
+  "description": "o1js testing utilities",
+  "version": "0.0.1",
+  "type": "module",
+  "scripts": {
+    "build": "tsc -p ../../tsconfig.testing.json &&  node copyBindings.js",
```
Trivo25, line 7: right now this builds everything in the project, incl. mina-signer, and copies all artefacts into `**/testing/dist/**` - I think we should be a bit more selective here

src/testing/copyBindings.js
mitschabaude, line 1: file name should be `copy-bindings.js` to follow our convention

src/testing/src/testing.ts
```diff
@@ -0,0 +1,59 @@
...
+  fulfills,
+  equals,
+  contains,
+  allConstant,
+  ifNotAllConstant,
+  isEmpty,
+  withoutGenerics,
+  print,
+  repeat,
+} = ConstraintSystem_;
```
mitschabaude, line 47: I would put the constraint system stuff in a separate name space.
most people won't need it, and it's confusing to pollute the name space with many simple functions like `and`, `equals` etc

mitschabaude: similar how `Random` is exported as its own namespace!

src/testing/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+This is package contains testing utils used internally in o1js to test primitives.
+
+## Installation
+
+```bash
+npm install --save-dev @o1js/testing
+# or with yarn:
+yarn add -D @o1js/testing
+```
```
mitschabaude, line 11: I would remove `yarn`, its v1 version has been deprecated for years now and v2+ never took off

src/testing/package.json
boray: Only mina-signer is not used, but I didn't try to prevent it from building since `npm run build` in o1js also builds mina-signer.

src/testing/src/testing.ts
boray: Agreed but then existing tests would be incompatible* (directly) with the package. What do you think about a new PR that creates such namespace and updates all tests?

mitschabaude: wdym they "would be incompatible"? I'm asking for a change in this file, `testing.ts`, not in the existing tests

mitschabaude:
```suggestion
export { ConstraintSystem_ as ConstraintSystem }
```

boray: It was a deliberate choice to not create a namespace to not change the usage from `constraintSystem.fromZkProgram()` to `ConstraintSystem.constraintSystem.fromZkProgram()` to make the package "plug and play" compatible with the existing tests.

mitschabaude: Wrong choice IMO. People won't use this, it pollutes the namespace, and it's still plug and play compatible if you just unpack the CobstraintSystem object at the beginning

boray: "it pollutes the namespace" I agree with this but I don't think the latter is a good solution. Why not we fix the pollution in its roots at `constraint-system.ts`? Similar fix would be needed for `equivalent.ts` too.

mitschabaude: Sure, fix it where you think it fits best!

Trivo25:
How do you want to release this to npm?

boray:
> How do you want to release this to npm, if at all?

We can release as a scoped package under o1js on demand similar to mina-signer.

Trivo25:
> > How do you want to release this to npm, if at all?
> 
> We can release as a scoped package under o1js on demand similar to mina-signer.

Right, but manually or through CI?

boray:
> > > How do you want to release this to npm, if at all?
> > 
> > 
> > We can release as a scoped package under o1js on demand similar to mina-signer.
> 
> Right, but manually or through CI?

Manually, I don't think we should release this package in a strict cadence.

Trivo25:
> > > > How do you want to release this to npm, if at all?
> > > 
> > > 
> > > We can release as a scoped package under o1js on demand similar to mina-signer.
> > 
> > 
> > Right, but manually or through CI?
> 
> Manually, I don't think we should release this package in a strict cadence.

Okay, but this means the person releasing it will need access to our npm details. If we add a CI workflow we could trigger it manually, given we already have access to the credentials in CI

boray:
> > > > > How do you want to release this to npm, if at all?
> > > > 
> > > > 
> > > > We can release as a scoped package under o1js on demand similar to mina-signer.
> > > 
> > > 
> > > Right, but manually or through CI?
> > 
> > 
> > Manually, I don't think we should release this package in a strict cadence.
> 
> Okay, but this means the person releasing it will need access to our npm details. If we add a CI workflow we could trigger it manually, given we already have access to the credentials in CI

I assumed you have the credentials for npm for o1js and you can release on my behalf. No worries, there won't be releases often :)

src/testing/index.ts
```diff
@@ -0,0 +1 @@
+export * from './src/testing.js';
```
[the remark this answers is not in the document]
boray: I needed to do this at some point of the process. Then, I didn't thought much on this.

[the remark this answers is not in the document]
boray: Let's get rid of it.

src/testing/README.md
```diff
@@ -0,0 +1,486 @@
...
+await Bitwise.compile();
+
+constraintSystem.fromZkProgram(Bitwise, 'notUnchecked', ifNotAllConstant(contains('Generic')));
+```
+
+## Examples
+
+Check out [`src/lib/provable/test`](https://github.com/o1-labs/o1js/tree/main/src/lib/provable/test) directory for more examples.
+
+## API Reference
```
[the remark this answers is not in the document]
boray: I think that also a good path. But, the tsdocs should be at relevant files in `src/lib/testing`, not in this package. I think it's acceptable to have this reference in the README. I don't think it would be hard to keep this up to date as the testing files wouldn't change much. And, if we decide to do a refactor, updating the docs must be part of the refactor.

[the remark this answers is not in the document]
boray: We can do a follow up PR once we figure out how we will present this package's docs in the new o1js documentation.

src/testing/CHANGELOG.md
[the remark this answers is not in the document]
boray: I am onboard with this idea!

src/testing/src/testing.ts
```diff
@@ -0,0 +1,59 @@
+import * as Equivalent_ from '../../lib/testing/equivalent.js';
+import * as Property_ from '../../lib/testing/property.js';
+import * as ConstraintSystem_ from '../../lib/testing/constraint-system.js';
```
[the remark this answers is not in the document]
boray: There is no index file at testing folder. How can this work at all? We can create an index.ts there and import only from that file. But, then the index would have these imports.

Trivo25:
> I assumed you have the credentials for npm for o1js and you can release on my behalf. No worries, there won't be releases often :)

We shouldn’t depend on a single person who may or may not have the credentials to release a package. For our other releases, we already have access to the credentials via CI, and we should follow the same approach here. Since we're introducing this new package to developers, we can expect feedback and requests for improvements - which means we'll likely need to release more frequently. Relying on one individual creates an unnecessary bottleneck and adds avoidable overhead to what could be a much simpler process.

src/testing/README.md
```diff
@@ -0,0 +1,486 @@
...
+    notUnchecked: {
+      privateInputs: [Field],
+      async method(a: Field) {
+        return { publicOutput: Gadgets.not(a, 240, false) };
+      },
+    },
+  },
+});
+
+await Bitwise.compile();
```
mitschabaude, line 122: nit: this line should be removed

src/testing/README.md
```diff
@@ -0,0 +1,486 @@
...
+  
+  // Verify that x + y = y + x
+  expect(xField.add(yField)).toEqual(yField.add(xField));
+});
+```
+
+```typescript
+import { test, Random } from '@o1js/testing';
+
+// Test Uint8.from() correcttly converts numbers to UInt8
```
0x471, line 50: nit: Uint -> UInt & correcttly -> correctly

0x471 (approved):
It seems good to merge for now, and the folder structure can be improved in follow-up PRs.

45930:
> We shouldn’t depend on a single person who may or may not have the credentials to release a package. For our other releases, we already have access to the credentials via CI, and we should follow the same approach here.

Yeah, releasing the package automatically if the version is incremented seems best to me.  How do we handle mina signer?

Trivo25:
> How do we handle mina signer?

Same way. We can simply copy the workflow from here https://github.com/o1-labs/o1js/blob/d6ae829578ffaa1254ea53216db50c92a3f6274e/.github/workflows/build-action.yml#L286-L371

src/testing/CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
+# Changelog
```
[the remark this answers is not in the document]
Trivo25: In a follow up PR we should also organize our github actions better. Right now everything is a bit chaotic

src/testing/package.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "bugs": "https://github.com/o1-labs/o1js/issues",
+  "main": "dist/node/testing/index.js",
+  "types": "dist/node/testing/index.d.ts",
+  "exports": {
+    "types": "./dist/node/testing/index.d.ts",
+    "node": "./dist/node/testing/index.js",
+    "default": "./dist/node/testing/index.js"
+  },
+  "files": [
```
Trivo25, line 30: We are including mina-signer and other internal dependencies in the package when we publish it to npm. While this might not be a big size increase in the overall package (given that the compile artefacts alone are about 45mb), it's still suboptiomal to include files that have nothing to do with the actual library in the released package

Trivo25: If or when we end up splitting everything into seperate `package` folders, we should definitely clean this up. So let's make sure we keep this in mind, I approve in hopes that we will follow up on this soon

Trivo25 (approved):

Issue #2098, opened 2025-03-26
Bits support for UInt32 and UInt64


kadirchan:
I think it would be good to have `toBits()` and `fromBits()` methods in `UInt32` and `UInt64` classes as it is in the Field class. It is especially convenient when compressing variables into a single `Field` element.

PR #2099, opened 2025-03-27
Add toBits and fromBits support for UInt32 and UInt64

kadirchan:
Corresponding issue: #2098 

Implemented `toBits()` and `fromBits()` methods for `UInt32` and `UInt64` classes and their tests.

### DIFF
--- a/src/lib/provable/int.ts
+++ b/src/lib/provable/int.ts
@@ -19,7 +19,6 @@ import {
 import { assert } from '../util/assert.js';
 import { TupleN } from '../util/types.js';
 import { bytesToWord, wordToBytes } from './gadgets/bit-slices.js';
-import { Fp } from '../../bindings/crypto/finite-field.js';
 import { BinableFp } from '../../mina-signer/src/field-bigint.js';
 
 // external API
--- a/src/lib/provable/int.ts
+++ b/src/lib/provable/int.ts
@@ -556,12 +556,7 @@ class UInt64 extends CircuitValue {
         throw Error(`UInt64.toBits(): ${this} does not fit in ${length} bits`);
       return bits.slice(0, length).map((b) => new Bool(b));
     }
-    let bits = Provable.witness(Provable.Array(Bool, length), () => {
-      let f = this.toBigInt();
-      return Array.from({ length }, (_, k) => new Bool(!!((f >> BigInt(k)) & 0x1n)));
-    });
-    UInt64.fromBits(bits).assertEquals(this);
-    return bits;
+    return this.value.toBits(length);
   }
 
   /**
@@ -580,21 +575,7 @@ class UInt64 extends CircuitValue {
   static fromBits(bits: (Bool | boolean)[]) {
     const length = bits.length;
     checkBitLength('UInt64.fromBits()', length, 64);
-    if (bits.every((b) => typeof b === 'boolean' || b.toField().isConstant())) {
-      let bits_ = bits
-        .map((b) => (typeof b === 'boolean' ? b : b.toBoolean()))
-        .concat(Array(Fp.sizeInBits - length).fill(false));
-      return new UInt64(BinableFp.fromBits(bits_));
-    }
-    return UInt64.Unsafe.fromField(
-      bits
-        .map((b) => new Bool(b))
-        .reduce((acc, bit, idx) => {
-          const shift = 1n << BigInt(idx);
-          return acc.add(bit.toField().mul(shift));
-        }, Field.from(0))
-        .seal()
-    );
+    return UInt64.Unsafe.fromField(Field.fromBits(bits));
   }
 }
 /**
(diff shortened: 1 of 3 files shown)


src/lib/provable/int.ts
```diff
@@ -532,6 +534,48 @@ class UInt64 extends CircuitValue {
   static fromBytesBE(bytes: UInt8[]): UInt64 {
     return UInt64.fromBytes([...bytes].reverse());
   }
+
+  /**
+   * Turns the {@link UInt64} into a {@link Bool} array. Default length is 64.
+   */
```
boray, line 540: clarify endiannes

src/lib/provable/int.ts
```diff
@@ -1023,6 +1067,48 @@ class UInt32 extends CircuitValue {
   static fromBytesBE(bytes: UInt8[]): UInt32 {
     return UInt32.fromBytes([...bytes].reverse());
   }
+
+  /**
+   * Turns the {@link UInt32} into a {@link Bool} array. Default length is 32.
+   */
```
boray, line 1073: same here

src/lib/provable/int.ts
```diff
@@ -532,6 +534,48 @@ class UInt64 extends CircuitValue {
...
+      let bits = BinableFp.toBits(this.toBigInt());
+      if (bits.slice(length).some((bit) => bit))
+        throw Error(`UInt64.toBits(): ${this} does not fit in ${length} bits`);
+      return bits.slice(0, length).map((b) => new Bool(b));
+    }
+    let bits = Provable.witness(Provable.Array(Bool, length), () => {
+      let f = this.toBigInt();
+      return Array.from({ length }, (_, k) => new Bool(!!((f >> BigInt(k)) & 0x1n)));
+    });
+    UInt64.fromBits(bits).assertEquals(this);
```
boray, line 553: this assertion should fail with a message

kadirchan: The `assertEquals` used by the current `UInt64` and `UInt32` classes do not support custom error messages, I can add this as a feature to `CircuitValue` or we can add a try catch block like [this](https://github.com/kadirchan/o1js/blob/0ba37fe6507e03acc5b19bbdcddcef4fe740141d/src/lib/provable/int.ts#L1764-L1765).

src/lib/provable/test/int.test.ts
```diff
@@ -625,6 +625,35 @@ describe('int', () => {
           });
         });
       });
+
+      describe('toBits() / fromBits()', () => {
+        it('should be the same as 12345678', async () => {
```
45930, line 630: Can you add a test for the `fromBits()` behavior without using `toBits()` first?

e.g.

```ts
it('fromBits from raw bits', () => {
  const bits = [Bool(false), Bool(true), Bool(false), Bool(true)];
  const uint = UInt64.fromBits(bits);
  expect(uint.toBits(4)).toEqual(bits);
});
```

kadirchan: Since these methods are very similar to the `Fields.fromBits` and `Fields.toBits` implementations, I kept the unit tests relatively minimal. Thank you for the feedback I added a few tests for the cases you mentioned.

src/lib/provable/int.ts
```diff
@@ -532,6 +534,68 @@ class UInt64 extends CircuitValue {
...
+    return bits;
```
mitschabaude, line 564: I don't get why you don't use the `field.toBits()` method here?

```suggestion
    checkBitLength('UInt64.toBits()', length, 64);
    return this.value.toBits(length);
```

src/lib/provable/int.ts
```diff
@@ -532,6 +534,68 @@ class UInt64 extends CircuitValue {
...
+    }
+    return UInt64.Unsafe.fromField(
+      bits
+        .map((b) => new Bool(b))
+        .reduce((acc, bit, idx) => {
+          const shift = 1n << BigInt(idx);
+          return acc.add(bit.toField().mul(shift));
+        }, Field.from(0))
+        .seal()
+    );
```
mitschabaude, line 597: just use `Field.fromBits()`
```suggestion
    return UInt64.Unsafe.fromField(Field.fromBits(bits));
```

mitschabaude (changes requested):
Nice addition! But this unnecessarily adds complexity by not reusing the well-tested `toBits()` and `fromBits()` methods on `Field`, even though the implementation is exactly the same

src/lib/provable/int.ts
kadirchan: You're right—I need to be more attentive and get more familiar with the library

src/lib/provable/int.ts
kadirchan: In this case, since we can't customize the part
```ts
if (this.isConstant()) {
  let bits = BinableFp.toBits(this.toBigInt());
  if (bits.slice(length).some((bit) => bit))
    throw Error(`Field.toBits(): ${this} does not fit in ${length} bits`); // this error message
  return bits.slice(0, length).map((b) => new Bool(b));
}
```
in the `toBits` method for the `UInt32` and `UInt64` classes, error messages will only be shown for the `Field` type, which might be confusing for developers. wdyt?

mitschabaude: yep that's a good point! but you could at least use it in the non-constant branch

kadirchan:
Thanks to @mitschabaude , I simplified the unnecessary parts of the code.  
I also added these methods for `UInt8`, along with their corresponding tests.

kadirchan:
I didn't notice the unused import before pushing requested changes by Gregor, should I remove line 22 and push again? @45930

kadirchan:
It should be fine now

PR #2107, opened 2025-03-27
Update provable docs

benSepanski:
Add documentation to `Provable.witness()` indicating that developers can rely on
1. Type checks to be enforced by `Provable.witness()`
2. Composite types to enforce type checks on each of their subcomponents.

### DIFF
--- a/src/lib/provable/provable.ts
+++ b/src/lib/provable/provable.ts
@@ -3,33 +3,33 @@
  * - a namespace with tools for writing provable code
  * - the main interface for types that can be used in provable code
  */
-import { InferValue } from '../../bindings/lib/provable-generic.js';
-import { ToProvable } from '../../lib/provable/types/provable-intf.js';
-import { Context } from '../util/global-context.js';
 import { Bool } from './bool.js';
-import {
-  asProver,
-  constraintSystem,
-  generateWitness,
-  inCheckedComputation,
-  inProver,
-} from './core/provable-context.js';
 import { Field } from './field.js';
+import { Provable as Provable_, ProvableType } from './types/provable-intf.js';
+import type { FlexibleProvable, FlexibleProvableType, ProvableExtended } from './types/struct.js';
+import { Context } from '../util/global-context.js';
 import {
   HashInput,
   InferJson,
   InferProvableType,
   InferredProvable,
 } from './types/provable-derivers.js';
-import { Provable as Provable_, ProvableType } from './types/provable-intf.js';
-import type { FlexibleProvable, FlexibleProvableType, ProvableExtended } from './types/struct.js';
+import {
+  inCheckedComputation,
+  inProver,
+  asProver,
+  constraintSystem,
+  generateWitness,
+} from './core/provable-context.js';
 import { witness, witnessAsync, witnessFields } from './types/witness.js';
+import { InferValue } from '../../bindings/lib/provable-generic.js';
+import { ToProvable } from '../../lib/provable/types/provable-intf.js';
 
 // external API
 export { Provable };
 
 // internal API
-  export { getBlindingValue, memoizationContext, MemoizationContext, memoizeWitness };
+export { memoizationContext, MemoizationContext, memoizeWitness, getBlindingValue };
 
 /**
  * `Provable<T>` is the general interface for provable types in o1js.
--- a/src/lib/provable/provable.ts
+++ b/src/lib/provable/provable.ts
@@ -87,7 +87,11 @@ const Provable = {
    *       .toUInt32();
    *   });
    *   // prove the 32-bit lower and upper limbs match the 64-bit value
-   *   value.assertEquals(new UInt64(lowerLimb).add(new UInt64(upperLimb).mul(2 ** 32)));
+   *   value.assertEquals(
+   *     limbs.lower
+   *       .toUInt64()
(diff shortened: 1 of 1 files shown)


benSepanski:
Ack, I just realized my vscode reformatted all the imports :(

Trivo25 (approved):
thank you so much!

Trivo25:
> Ack, I just realized my vscode reformatted all the imports :(

I'll take care of this :)

src/lib/provable/provable.ts
```diff
@@ -62,6 +71,44 @@ const Provable = {
    * // prove that `invX` is really the inverse of `x`:
    * invX.mul(x).assertEquals(1);
    * ```
+   *
+   * Example for decomposing a 64-bit integer into two 32-bit limbs. {@link Provable.witness} will
+   * prove that the two limbs are actually 32-bits, ensuring the decomposition is unique.
+   * ```ts
+   * function decompose(value: UInt64) {
```
45930, line 78: These examples red-squiggle because the UInt64 limbs are not the right type for a UInt32 return.

image
image

This function type-checks, but throws on the range check `upper: value.and(new UInt64(0xffffffff00000000n)).toUInt32(),`:

```ts
function decompose(value: UInt64) {
  // get two arbitrary 32-bit values from the prover
  let limbs = Provable.witness(Decomposition, () => {
    return new Decomposition({
      lower: value.and(new UInt64(0xffffffffn)).toUInt32(),
      upper: value.and(new UInt64(0xffffffff00000000n)).toUInt32(),
    });
  });
  // prove the 32-bit lower and upper limbs match the 64-bit value
  value.assertEquals(
    limbs.lower
      .toUInt64()
      .add(limbs.upper.toUInt64().mul(UInt64.from(2n * 32n)))
  );
}
```

45930: Can you clarify what you're trying to demonstrate with the example?

benSepanski: I'm trying to demonstrate that using a witness of type `Decomposition` will still prove that `limbs.lower` and `limbs.upper` are 32-bits. My goal is just to make explicit that `Struct`s call `check()` on each of the structure fields

src/lib/provable/provable.ts
```diff
@@ -62,6 +71,58 @@ const Provable = {
...
+   *     return value
+   *       .and(new UInt64(0xffffffff00000000n))
+   *       .div(2 ** 32)
+   *       .toUInt32();
+   *   });
+   *   // prove the 32-bit lower and upper limbs match the 64-bit value
+   *   value.assertEquals(
+   *     limbs.lower
+   *       .toUInt64()
+   *       .add(limbs.upper.toUInt64().mul(UInt64.from(2n * 32n)))
```
45930, line 93:
```suggestion
   *       .add(limbs.upper.toUInt64().mul(UInt64.from(2n ** 32n)))
```

src/lib/provable/provable.ts
```diff
@@ -62,6 +71,58 @@ const Provable = {
...
+   *         .and(new UInt64(0xffffffff00000000n))
+   *         .div(2 ** 32)
+   *         .toUInt32(),
+   *     });
+   *   });
+   *   // prove the 32-bit lower and upper limbs match the 64-bit value
+   *   value.assertEquals(
+   *     limbs.lower
+   *       .toUInt64()
+   *       .add(limbs.upper.toUInt64().mul(UInt64.from(2n * 32n)))
```
45930, line 122:
```suggestion
   *       .add(limbs.upper.toUInt64().mul(UInt64.from(2n ** 32n)))
```

src/lib/provable/provable.ts
```diff
@@ -62,6 +71,58 @@ const Provable = {
...
+   *   });
+   *   let upperLimb = Provable.witness(UInt32, () => {
+   *     return value
+   *       .and(new UInt64(0xffffffff00000000n))
+   *       .div(2 ** 32)
+   *       .toUInt32();
+   *   });
+   *   // prove the 32-bit lower and upper limbs match the 64-bit value
+   *   value.assertEquals(
+   *     limbs.lower
```
45930, line 91: limbs is not defined anymore

src/lib/provable/provable.ts
45930: Hah, well that's exactly what `upper: value.and(new UInt64(0xffffffff00000000n)).toUInt32()` proves!

If you wanted to leave in an explicit range check failure, that would be fine with me if it's clearly labeled as a failure case.

45930 (changes requested):
Nice, the new implemetations both work, making the superficial changes I called out.  Once the examples are runnable, I'm happy to merge!

src/lib/provable/provable.ts
benSepanski: `upper: value.and(new UInt64(0xffffffff00000000n)).toUInt32()`, so IIUC it doesn't add constraints

45930: Yes you're right.  It still fails, but only because of the javascript class.

45930 (approved):

45930:
Thanks again @benSepanski!

benSepanski:
Hooray! Thanks for all your help :)

PR #2111, opened 2025-03-28
Simplify transaction type

45930:
This PR removes `AuthorizedZkappCommand` and replaces it with an `isAuthorized` property on `ZkappCommand`.

I originally intended to create a `Transaction` interface that can be shared by all transaction types, but the only transaction type I found was `ZkAppCommand`, so maybe an interface is overkill?


src/lib/mina/v2/transaction.ts
```diff
@@ -1,78 +1,133 @@
...
+   * @returns An authorized {@link Transaction}
+   */
+  authorize(...args: any): T | Promise<T>;
+
+  /**
+   * Converts the transaction to a JSON object.
+   *
+   * @returns A JSON object representing the transaction
+   */
+  toJSON(): any;
```
45930, line 86: `any` could be replaced here by a `SerializedMinaTransaction` type that helps map to the raw transaction JSON that Mina nodes expect.

PR #2118, opened 2025-04-02
 Optimize Base58 encoding/decoding

VolodymyrBg:
- Use Object.fromEntries for inverse alphabet creation
- Replace let with const for immutable variables
- Add bounds checking to prevent potential errors
- Optimize toBase58 and fromBase58 algorithms
- Improve type safety with explicit type annotations
- Special case handling for empty inputs

### DIFF
--- a/src/lib/util/base58.ts
+++ b/src/lib/util/base58.ts
@@ -6,25 +6,22 @@ import { changeBase } from '../../bindings/crypto/bigint-helpers.js';
 export { toBase58Check, fromBase58Check, base58, withBase58, fieldEncodings, Base58, alphabet };
 
 const alphabet = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'.split('');
-let inverseAlphabet: Record<string, number> = {};
-alphabet.forEach((c, i) => {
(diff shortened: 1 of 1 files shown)


src/lib/util/base58.ts
```diff
@@ -6,25 +6,22 @@ import { changeBase } from '../../bindings/crypto/bigint-helpers.js';
 export { toBase58Check, fromBase58Check, base58, withBase58, fieldEncodings, Base58, alphabet };
 
 const alphabet = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'.split('');
-let inverseAlphabet: Record<string, number> = {};
-alphabet.forEach((c, i) => {
-  inverseAlphabet[c] = i;
-});
+const inverseAlphabet = Object.fromEntries(alphabet.map((c, i) => [c, i]));
```
45930, line 9: Is this actually an optimization?  It benchmarks as slower for me!

45930 (commented):
I'm not convinced that there's enough value here to merge.  Do you think this change actually improves o1js?

mitschabaude:
this makes the code harder to read, and I don't see any significant algorithmic improvements. wouldn't merge

PR #2123, opened 2025-04-04
Implement verification key hash validation

### LINKED ISSUE #2120: As a developer, I should be able to verify if a given data, hash pair are a valid verification key.
See: https://discord.com/channels/484437221055922177/1357003022307295405

It would be convenient if the SDK exposed a way to verify the validity of a given string verification key hash.  Ultimately, we would want to access this within a circuit, e.g. to upgrade a side-loaded verification key committed to on chain.

Deniz found that you can set any value for the verification key hash and the transaction will succeed on LocalBlockchain: https://github.com/0x471/o1js-vk-hash-validation/blob/main/README.md - This is a confusing DevX and we can improve it by validating in JS whether the pair is valid.

0x471:
## Description
This PR adds functionality to validate verification key data-hash pairs, addressing the need to verify the validity of verification key, particularly when upgrading side-loaded verification keys committed on chain.

## Changes
- Added `validateVkHash` function in `zkprogram.ts` to verify if a given verification key's data matches its hash: https://github.com/o1-labs/o1js/blob/06182ba70d5b84cca15a31b0a76c87eec386627b/src/lib/proof-system/zkprogram.ts#L142
- Added example zkApp demonstrating the validation functionality: 
https://github.com/o1-labs/o1js/blob/06182ba70d5b84cca15a31b0a76c87eec386627b/src/examples/zkapps/vk-hash-validation-zkapp.ts#L29

## Related Issue
Resolves #2120

## Additional Context
This feature is particularly useful for:
- Verifying verification keys before setting them in zkApps
- Ensuring data-hash pairs are valid when upgrading verification keys
- Preventing invalid verification keys from being committed on chain?
- https://discord.com/channels/484437221055922177/915745847692636181/1357003022307295405

## To-Do
- [ ] Add test cases for different verification key scenarios
- [ ] Investigate proper circuit context handling without `Provable.inCheckedComputation()`
- [ ] Update Changelog

### DIFF
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -42,7 +42,7 @@ import { mapObject, mapToObject, zip } from '../util/arrays.js';
 import { VerificationKey } from './verification-key.js';
 
 // public API
-export { SelfProof, JsonProof, ZkProgram, verify, Empty, Undefined, Void, Method };
+export { SelfProof, JsonProof, ZkProgram, verify, Empty, Undefined, Void, Method, validateVkHash };
 
 // internal API
 export {
@@ -130,6 +130,21 @@ async function verify(
   return prettifyStacktracePromise(
     withThreadPool(() => Pickles.verify(statement, picklesProof, vk))
   );
+}
+
+/**
+ * Validates that a verification key's hash matches the expected hash.
+ * This is useful for verifying that a verification key is valid in a circuit.
+ * 
+ * @param verificationKey - The verification key to validate
+ * @returns void, throws an error if the hash doesn't match
+ */
+async function validateVkHash(verificationKey: VerificationKey) {
+  const circuitVk = Pickles.sideLoaded.vkToCircuit(() => verificationKey.data);
+  
+  // Assert the validity of the auxiliary vk-data by comparing the witnessed and the hash as an input
+  const hash = inCircuitVkHash(circuitVk);
+  Field(hash).assertEquals(verificationKey.hash, 'Provided VerificationKey hash not correct');
 }
 
 /**
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -135,13 +135,13 @@ async function verify(
 /**
  * Validates that a verification key's hash matches the expected hash.
  * This is useful for verifying that a verification key is valid in a circuit.
- * 
+ *
  * @param verificationKey - The verification key to validate
  * @returns void, throws an error if the hash doesn't match
  */
 async function validateVkHash(verificationKey: VerificationKey) {
   const circuitVk = Pickles.sideLoaded.vkToCircuit(() => verificationKey.data);
-  
+
   // Assert the validity of the auxiliary vk-data by comparing the witnessed and the hash as an input
   const hash = inCircuitVkHash(circuitVk);
   Field(hash).assertEquals(verificationKey.hash, 'Provided VerificationKey hash not correct');
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -133,11 +133,11 @@ async function verify(
 }
 
 /**
- * Validates that a verification key's hash matches the expected hash.
- * This is useful for verifying that a verification key is valid in a circuit.
+ * Asserts that the hash computed from a verification key's data matches the provided hash.
+ * This is used to ensure that the verification key data and hash are consistent.
  *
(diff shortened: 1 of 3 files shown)


0x471:
When `validateVkHash()` is used in a zkApp without `Provable.inCheckedComputation()`, it throws the error: "This function can't be run outside of a checked computation". Any ideas?  cc @Trivo25

Trivo25:
> When `validateVkHash()` is used in a zkApp without `Provable.inCheckedComputation()`, it throws the error: "This function can't be run outside of a checked computation". Any ideas? cc @Trivo25

Probably because this function can't be run outside of a checked computation and we run smart contract code (transactions) outside of a checked computation

src/lib/proof-system/zkprogram.ts
```diff
@@ -132,6 +132,21 @@ async function verify(
...
 }
 
+/**
+ * Validates that a verification key's hash matches the expected hash.
+ * This is useful for verifying that a verification key is valid in a circuit.
+ *
+ * @param verificationKey - The verification key to validate
+ * @returns void, throws an error if the hash doesn't match
+ */
+async function validateVkHash(verificationKey: VerificationKey) {
```
mitschabaude, line 142: the docstring is misleading. this method just proves that the given hash belongs to _any_ verification key. not that it belongs to the given verification key. that's because the `VerificationKey` type doesn't represent the actual vk data as part of the circuit. To the circuit, it's just a hash. So there's actually no way to verify in a circuit that the `VerifcationKey` .data is valid

0x471: Thanks @mitschabaude. Could you please check the updated one?

mitschabaude: Still has the same problem I'd say

adamczykm:
This issue has already misled some important pieces of o1js code such as the new fungible token standard and almost an upcoming zkusd government contract. Maybe it is worth a warning somewhere in the docstrings?

0x471:
Hey @mitschabaude,

I just pushed a commit that improves the example for using a different zkApp's hash. However, as you can see from the output, it's throwing an error: "Error: Provided VerificationKey hash is not correct." Here's a more detailed output for reference: [Gist Link](https://gist.github.com/0x471/b1b11591a977f221738e0fc43f45780c).

I believe this aligns with the use case @45930 mentioned earlier:
> This is interesting. From the actual user story we're trying to solve, the goal is to confirm that a given Verification Key (VK) upgrade matches exactly the data they expect. This isn't in the context of upgrading a zkApp's VK, but in relation to upgrading a state field containing the hash of a sideloaded VK.

If we're unable to prove that a given key's data and hash match, does this imply we can't prove that a specific state field is the true hash of a given constraint system?

Could you elaborate further on why you believe this might be misleading?

mitschabaude:
> Could you elaborate further on why you believe this might be misleading?

@0x471 your test assumes an honest prover. However, a malicious prover can create a valid proof for your example zkApp, using a modifed version of o1js, without the error, and where the vk's data doesn't match the hash.

> So from the actual user story we want to solve, they want to confirm that a given VK upgrade corresponds to precisely the data they believe. Not in a context of upgrading the VK of a ZKApp, but in the context of upgrading a state field of the hash of a sideloaded VK.

> If we can't prove that a given key data and key hash match, does that mean we can't prove that a given state Field is the true hash of a specific constraint system?

@45930 thanks for the context. I would guess that, in principle, you can prove that a given vk hash belongs to some constraint system (but that it is infeasible in terms of circuit complexity). In any case,  the `VerificationKey` type is the wrong way to represent vks in that case, because it doesn't maintain the actual vk as part of the circuit (just its hash). For the user story, you'd need to represent all of the vk inside the circuit, as well as the entire constraint system. (However, as I said, that sounds infeasible to me, without knowing more)

0x471:
Closing this draft PR as it's not feasible to proceed at the current stage

src/lib/proof-system/zkprogram.ts
```diff
@@ -132,6 +132,21 @@ async function verify(
...
 }
 
+/**
+ * Asserts that the hash computed from a verification key's data matches the provided hash.
+ * This is used to ensure that the verification key data and hash are consistent.
+ *
+ * @param verificationKey - The verification key whose data and hash to check for consistency
+ * @returns void, throws an error if the computed hash doesn't match the provided hash
+ */
+async function validateVkHash(verificationKey: VerificationKey) {
```
mrmr1993, line 142: This is redundant, this logic already exists in the zkProgram. Instead, you should just constrain `hash` somehow in relation to your state.

PR #2128, opened 2025-04-04
Cargo: enforce some versions previously used by succesful builds on o1js

dannywillems:


dannywillems:
As briefly described to @Trivo25. I think the job "Build o1js bindings / build-bindings-ubuntu (pull_request)" does not pick the right versions of the dependencies, as constrained in [proof-systems/Cargo.toml](https://github.com/o1-labs/proof-systems/blob/ede9198eeb55114be9b2426dc3150437a2e70f0f/Cargo.toml).

The commit https://github.com/o1-labs/o1js/pull/2128/commits/b3aa7672fd7dd3abb147c7712fb38b528bcafaa8 has been done on my machine (modulo the diff [1]). We can see that the CI is unhappy, [here](https://github.com/o1-labs/o1js/actions/runs/14313553755) and [here](https://github.com/o1-labs/o1js/actions/runs/14313553738/). However, after regenerating without Nix, with the commands:
```
dune clean
find . -name Cargo.toml \
  -not -path "*/target/*" \
  | grep -vE "snarky-deriver|ethereum-optimism|kimchi-stubs-vendors" \
  | xargs -I {} dirname {} \
  | sort -u \
  | xargs -I {} bash -c '
      echo "Cleaning {}"
      cd "{}" && cargo clean
    '
npm run build:update-bindings
```
we get the changes introduced in https://github.com/o1-labs/o1js/pull/2128/commits/c0ae41f6255424a6fcede1d536515f3856dd6391 which seems to correctly select the dependencies I enforced in the file Cargo.toml (by looking at the logs on my machine while generating).


[1]
```
diff --git a/flake.nix b/flake.nix
index 5b42d5fdf..974593c22 100644
--- a/flake.nix
+++ b/flake.nix
@@ -107,7 +107,7 @@
           sha256 =
             if pkgs.stdenv.isDarwin
             then "sha256:1mv8skl4l2q782741r1yakbf0y4q6v9358fm91r45gj97j20il1y"
-            else "sha256:0x7h8nggyw3vsvgqv0s2hxzn393l4gzpm3vb2qvsnnm1rf0r86v2";
+            else "sha256:0hpczvjwn0c74jssc1l10f11gg5j790cxhnqr6g9pyb8z0g1vqby";
         };
         toolchain = pkgs.symlinkJoin {
           name = "toolchain";
```

Geometer1729:
@dannywillems CI gets the wasm through mina's nix

https://github.com/MinaProtocol/mina/blob/0dcd9837b9fc9889586c6889fdffb59a7d6e1697/nix/rust.nix#L176-L219

It looks like you made all the changes and that code should be using `proof_systems/Cargo.lock` 
and when I delete something from `proof_systems/Cargo.lock` it breaks `nix run o1js#update-bindings` which is what ci runs.

If you want to double check that, you could delete a package from the lockfile like `console_error_panic_hook` and make sure CI fails with ` error: no matching package named `console_error_panic_hook` found`

Am I misunderstanding the issue?

dannywillems:
> It looks like you made all the changes and that code should be using proof_systems/Cargo.lock

That is correct.

The problem is I don't understand why when I run the build with nix, I do not have all the tests passing, if they are supposed to use the versions that are strictly constrained [here](https://github.com/o1-labs/proof-systems/blob/master/Cargo.toml#L51-L73). My hypothesis was that it was using a different version of once_cell or wasm-bindgen than the one I get [here](https://github.com/o1-labs/proof-systems/pull/3144#issuecomment-2779229353)

dannywillems:
Merging as https://github.com/MinaProtocol/mina/pull/16887 has been merged too

Issue #2006, opened 2025-02-04
Memory boosting mode


querolita:
### Problem description

WASM bytecode allows web browsers to execute high-performance code written in lower-level languages (e.g. Rust) with near native speed. However, WASM has a heap limitation set at 4GB, what can be a problem when compiling memory demanding applications. In our case, o1js makes use of the Kimchi SNARK behind the hood, which has been designed to provide the fastest proving time across our code stack. This is thank to, among many others, efficient algorithms, optimized formatting, and precomputation of costly parameters. On this end, several polynomials that describe the circuit are stored ahead of time to avoid having the prover generate them whenever needed. This comes at the cost of an important part of the heap being consumed by these structures. This configuration is preventing us from using that precious space for other features that could require further memory resources, such as chunking circuits from zkApps. Moreover, there's a bound in the number of methods that can be defined, and some of them are currently being used by the selector polynomials (also limiting the complexity of the circuits that can be deployed). 

### Proposal

The idea is to provide a _memory boosting mode_ inside the o1js prover interface that would send a signal throughout the stack to trade memory for proving time. That means, by default o1js would still function as it does today. But upon request by the zkApp developer, our SNARK will be able to operate in a (slightly) slower pace in order to release sufficient heap space that could be used by the circuit for other concerns. If that flag is set, these heavy polynomials would only be computed when needed by the SNARK instead of being recovered from memory. 

A second step which would alleviate this prover overhead consists on cacheing these polynomials directly in the browser. This way the Typescript code would obtain their evaluations in a database fashion from the webpage.

### Concrete steps

The relation defining our circuit is given by a set of polynomials that play an essential role in the proving procedure. They are split into three main groups: gate selectors, lookup polynomials, and coefficient polynomials. These are fixed and known per circuit, as they define the operation that the SNARK is proving claims about, leaking no information at all about the underlying witness. They can be seen as switches throughout the execution trace, turning on and off certain features of the gates involved. In particular, each of them has degree in the order of the number of rows (as they are computed by interpolating their evaluations across the domain). However, in order to compute other necessary polynomials for the proof generation, we need further evaluations of them. In particular, 8 times more than the domain size (what is referred to as D8 in the code). These are the concrete evaluations that we are targeting in this proposal. Instead of loading them from memory at proof generation, the prover would compute them whenever needed as the degrees of the constraints require so. 

### Implementation details

[...]

PR #2143, opened 2025-04-09
Lazy mode for WASM memory

querolita:
This PR adds support for the lazy mode from o1js. It is the last PR of this issue https://github.com/o1-labs/o1js/issues/2006. 

Gains are not as clear as they are in Kimchi or Pickles. In particular, not sufficient to support 2 chunks unfortunately for now.
 
memory_comparison

cpu_comparison


src/lib/proof-system/lazy-mode.unit-test.ts
```diff
@@ -0,0 +1,87 @@
...
+  return fieldWithRng(Random.bignat((1n << BigInt(n)) - 1n));
+};
+
+function logWasmMemoryByteLength(label: string) {
+    let bytes = get_memory_byte_length();
+    console.log(`[WASM Memory] ${label}: ${bytes} bytes`);
+}
+
+function logWasmMemory(label: string) {
+    const memory = get_memory() as WebAssembly.Memory;
```
dannywillems, line 42: To keep a note: you might want to be sure you always get a new view on the memory.
For history, here some discussions we shared before.

--------------------


So you are mostly interested in getting a view on the memory field of the wasm-module, like Uint8Array(plonk_wasm.memory) (or smth like this, see the signature returned by the WebAssembly module given above) - also be **careful** that Uint8Array returns a view, and it would be corrupted if the WebAssembly memory grows, apparently. And you can use "basic" JavaScript method to analyze bytes by bytes the Uint8Array.
let mem = Uint8Array(plonk_wasm.memory);
mem.get(0) <- get the value of the first byte saved in the memory of the module.
mem.get(1) <- get the value of the second byte.
etc.
Note, you can also get by "machine word" using a different view, like Uint32Array.
From there, you can build a method printing regularly the memory addresses you are interested in.


Ah!
Memory instances can also grow, for example via the Memory.grow() method in JavaScript or memory.grow in the WebAssembly. Since ArrayBuffer objects can't change size, the current ArrayBuffer is detached and a new ArrayBuffer is created to point to the newer, bigger memory.
This explains my sentence here (link to "be careful about Uin8Array being a view")!
Source: [here](https://developer.mozilla.org/en-US/docs/WebAssembly/Guides/Understanding_the_text_format#webassembly_memory).

CHANGELOG.md
```diff
@@ -17,20 +17,22 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/3eef10d...HEAD)
 
+- [PR !2143]( https://github.com/o1-labs/o1js/pull/2143) Lazy mode for prover index computation
```
Trivo25, line 20: the changelog seems to be a bit out of date, make sure to bring it up to date with main

src/lib/proof-system/lazy-mode.unit-test.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    let isValid = await LazyMode.verify(proof);
+
+    console.log(`(${lazyMode ? 'Lazy' : 'Eager'}) Memory after verify`, getMemory());
+
+    assert(isValid);
+
+  }
+
+  await testLazyMode(false);
```
Trivo25, line 67: should we run this a second time with `true` as well? I guess right now there isn't too much to compare but maybe in the future

Trivo25 (approved):
Nice! I will think about a better way to benchmark o1js' performance in the coming weeks, this should include things like compile/prover performance as well as memory consumption - once we have a better way to analyze what's going on interally we can pick this back up and make o1js great again :D

PR #2144, opened 2025-04-09
Add bitwise methods to UInt8

boray:
This pull request includes several updates to the `src/lib/provable/int.ts` file. The changes include improvements to the documentation and the addition of new bitwise operation methods for the `UInt8` class.

Documentation improvements:
* Removed the unnecessary parameters in comments from the `and` method in the `UInt32` class and `not` method in the `UInt64` class. 

New bitwise operation methods for `UInt8` class:

* Added a new `xor` method for performing bitwise XOR operations.
* Added a new `not` method for performing bitwise NOT operations.
* Added a new `and` method for performing bitwise AND operations.
* Added a new `or` method for performing bitwise OR operations.

### DIFF
--- a/src/lib/provable/int.ts
+++ b/src/lib/provable/int.ts
@@ -928,7 +928,7 @@ class UInt32 extends CircuitValue {
    * let a = UInt32.from(3);    // ... 000011
    * let b = UInt32.from(5);    // ... 000101
    *
-   * let c = a.and(b, 2);    // ... 000001
+   * let c = a.and(b);    // ... 000001
    * c.assertEquals(1);
    * ```
    */
@@ -1745,7 +1745,7 @@ class UInt8 extends Struct({
    * let a = UInt8.from(3);    // ... 000011
    * let b = UInt8.from(5);    // ... 000101
    *
-   * let c = a.and(b, 2);    // ... 000001
+   * let c = a.and(b);    // ... 000001
    * c.assertEquals(1);
    * ```
    */
--- a/src/lib/provable/int.ts
+++ b/src/lib/provable/int.ts
@@ -284,7 +284,7 @@ class UInt64 extends CircuitValue {
    *
    * @example
    * ```ts
-   * // NOTing 4 bits
+   * // NOTing 4 bits with the unchecked version 
    * let a = UInt64.from(0b0101);
    * let b = a.not();
    *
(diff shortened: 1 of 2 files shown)


src/lib/provable/int.ts
```diff
@@ -284,9 +284,9 @@ class UInt64 extends CircuitValue {
    *
    * @example
    * ```ts
-   * // NOTing 4 bits with the unchecked version
+   * // NOTing 4 bits
```
boray, line 287: not() methods doesn't have `checked` parameter anymore.

Trivo25: While the method doesn't have the parameter, the implementation still uses the unchecked version. This should at least be noted

querolita: What do you mean? I can still see this in main 🤔  https://github.com/o1-labs/o1js/blob/ace28ce36072dc814313ee6da4731f89bf2f5716/src/lib/provable/gadgets/bitwise.ts#L12

boray: > What do you mean? I can still see this in main 🤔

The parameter exists in `Bitwise.not()` but it's hardcoded in UInt bitwise ops. 

e.g.
```js  
  not() {
    return new UInt32(Bitwise.not(this.value, UInt32.NUM_BITS, false).value);
  }
```

0x471 (approved):
It looks good to me. Mentioning that it's unchecked under the hood is a nice touch.

src/lib/provable/int.ts
```diff
@@ -1670,6 +1670,106 @@ class UInt8 extends Struct({
...
+   *
+   * You can find more details about the implementation in the [Mina book](https://o1-labs.github.io/proof-systems/specs/kimchi.html?highlight=gates#and)
+   *
+   *
+   * @example
+   * ```typescript
+   * let a = UInt8.from(3);    // ... 000011
+   * let b = UInt8.from(5);    // ... 000101
+   *
+   * let c = a.and(b, 2);    // ... 000001
```
45930, line 1748:
```suggestion
   * let c = a.and(b);    // ... 000001
```

45930 (approved):

45930:
Can you update the PR description to match the changes?  The current description does not match the behavior.

PR #2145, opened 2025-04-09
Implement TokenId.derive to construct from its parts

VolodymyrBg:
This commit implements the TODO in the TokenId class by adding proper functionality to construct a TokenId from its component parts instead of using raw Field values directly. Specifically:

- Added TokenId.derive method that creates a TokenId from a token owner's public key and a parent token ID
- Added TokenId.fromField helper method for converting raw Field values to TokenIds
- Fixed the necessary type assertions to ensure the code is type-safe

This change ensures TokenIds are properly derived from their components, making the code more descriptive and maintainable.

### DIFF
--- a/src/lib/mina/v2/core.ts
+++ b/src/lib/mina/v2/core.ts
@@ -7,6 +7,7 @@ import { GenericHashInput } from '../../../bindings/lib/generic.js';
 import { hashWithPrefix, packToFields } from '../../../lib/provable/crypto/poseidon.js';
 import { prefixes } from '../../../bindings/crypto/constants.js';
 import { Types } from '../../../bindings/mina-transaction/v1/types.js';
+import { PublicKey } from '../../provable/crypto/signature.js';
 
 export {
   Option,
@@ -83,7 +84,6 @@ class ZkappUri {
 }
 
 class TokenId {
-  // TODO: construct this from it's parts, don't pass in the raw Field directly
   constructor(public value: Field) {}
 
   equals(x: TokenId): Bool {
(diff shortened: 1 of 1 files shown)


src/lib/mina/v2/core.ts
```diff
@@ -95,6 +95,44 @@ class TokenId {
   }
 
   static MINA: TokenId = new TokenId(new Field(1));
+
+  /**
+   * Derives a TokenId from a token owner's public key and a parent token ID.
+   * This implementation exactly matches the OCaml code in Mina Protocol:
+   * https://github.com/MinaProtocol/mina/blob/2b0f60cf908bf8ed74de98fcc5442d3947f89d8e/src/lib/mina_base/account_id.ml#L133
```
Trivo25, line 102: I think we can leave out the link here

src/lib/mina/v2/core.ts
```diff
@@ -95,6 +95,44 @@ class TokenId {
...
+    
+    return new TokenId(hash);
+  }
+
+  /**
+   * Creates a TokenId from a Field value.
+   * @param field - The Field value.
+   * @returns A new TokenId.
+   */
+  static fromField(field: Field): TokenId {
```
Trivo25, line 133: I don't see additional value to the original constructor here `constructor(public value: Field) {}`

src/lib/mina/v2/core.ts
```diff
@@ -95,6 +96,33 @@ class TokenId {
...
+   * 
+   * @param tokenOwner - The public key of the token owner.
+   * @param parentTokenId - The parent token ID (defaults to MINA token ID).
+   * @returns A new TokenId.
+   */
+  static derive(tokenOwner: PublicKey, parentTokenId: Field = new Field(1)): TokenId {
+    // Use the provable function to create an AccountId type
+    const AccountId = provable({ tokenOwner: PublicKey, parentTokenId: Field });
+    
+    // Create the accountId using the provable type
```
Trivo25, line 112: technically this comment isn't correct, we are not using the provable type here

src/lib/mina/v2/core.ts
```diff
@@ -95,6 +96,33 @@ class TokenId {
...
+    const accountId = { tokenOwner, parentTokenId };
+    
+    // Use the toInput method from the provable type to get the hash input
+    const input = AccountId.toInput(accountId);
+    
+    // First pack the input (Random_oracle.pack_input), then hash with the prefix
+    const packed = packToFields(input);
+    // hashWithPrefix returns a value that needs to be cast to Field
```
Trivo25, line 120: this is not true, `hashWithPrefix` already returns a `Field`

45930:
The latest changes are not closer than the previous round.  Since this isn't an important issue and we're being distracted by reviews, I'm going to close it.

PR #2157, opened 2025-04-14
cleanup nix taking advantage of changes in mina

richardpringle:
This is actually cherry-picked from [here](https://github.com/o1-labs/o1js/pull/2131/commits/d3af820c1b4e77316ed6b3227373301f7bf07128)

Thanks @Geometer1729

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -272,8 +272,6 @@
               PLONK_WASM_WEB
               KIMCHI_STUBS
               KIMCHI_STUBS_STATIC_LIB
-              MARLIN_PLONK_STUBS
-              MARLIN_PLONK_STUBS_STATIC_LIB
               ;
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -284,6 +272,8 @@
               PLONK_WASM_WEB
               KIMCHI_STUBS
               KIMCHI_STUBS_STATIC_LIB
+              MARLIN_PLONK_STUBS
```
dannywillems, line 275: It should not be re-introduced. The variables have been renamed into kimchi_stubs

Geometer1729: In this pr the fact that they were missing was breaking ci:
https://github.com/o1-labs/o1js/pull/2131/commits/d3af820c1b4e77316ed6b3227373301f7bf07128

I guess updating mina would fix this if it didn't break other stuff.

PR #2167, opened 2025-04-16
Update wasm-bindgen

richardpringle:
I think this is probably behind at this point, but I want to see if this builds on CI


dannywillems:
Closing in favor of https://github.com/o1-labs/o1js/pull/2177

PR #2170, opened 2025-04-18
Add check to verification key struct

45930:
This PR adds a `static check()` method to the verification key struct.  The result is that we add assertions on the value of the VK such that the data and hash must match.

It's possible that adding these assertions are overkill, since the Protocol should not allow an updated VK that is not internally valid.  It does provide a more consistent API experience though.

### DIFF
--- a/src/examples/zkapps/zkapp-self-update.ts
+++ b/src/examples/zkapps/zkapp-self-update.ts
@@ -59,7 +59,29 @@ Provable.log('original verification key', fooVerificationKey);
 
 const { verificationKey: barVerificationKey } = await Bar.compile();
 
+try {
+  const illegalVerificationKey = new VerificationKey({
+    data: fooVerificationKey!.data,
+    hash: barVerificationKey.hash,
+  });
+
+  const tx2 = await Mina.transaction(deployer, async () => {
+    // VK with mismatched hash and data should throw
+    await contract.replaceVerificationKey(illegalVerificationKey);
+  });
+  await tx2.prove();
+} catch (error: any) {
+  if (
+    error.message.includes('The verification key hash is not consistent with the provided data')
+  ) {
+    console.log('correctly threw on illegal verification key');
+  } else {
+    throw error;
+  }
+}
+
 const tx2 = await Mina.transaction(deployer, async () => {
+  // This call will work because the vk is valid
   await contract.replaceVerificationKey(barVerificationKey);
 });
 await tx2.prove();
--- a/src/lib/proof-system/verification-key.ts
+++ b/src/lib/proof-system/verification-key.ts
@@ -2,6 +2,7 @@ import { initializeBindings, Pickles } from '../../snarky.js';
 import { provable } from '../provable/types/provable-derivers.js';
 import { Struct } from '../provable/types/struct.js';
 import { Field } from '../provable/wrapped.js';
+import { inCircuitVkHash } from './zkprogram.js';
 
 export { VerificationKey };
 
@@ -25,6 +26,21 @@ class VerificationKey extends Struct({
       ...RAW_VERIFICATION_KEY,
       hash: Field(RAW_VERIFICATION_KEY.hash),
     });
+  }
+
+  /**
+   * In-circuit check that data and hash are consistent.
+   */
+  static check(x: VerificationKey) {
+    if (x.hash.isConstant()) {
+      return;
+    }
+    const circuitVk = Pickles.sideLoaded.vkToCircuit(() => x.data);
+    const inCircuitHash = inCircuitVkHash(circuitVk);
+    Field(inCircuitHash).assertEquals(
+      x.hash,
+      'The verification key hash is not consistent with the provided data'
+    );
   }
 }
 
--- a/src/lib/proof-system/zkprogram.ts
+++ b/src/lib/proof-system/zkprogram.ts
@@ -60,6 +60,7 @@ export {
   TupleToInstances,
   PrivateInput,
   Proof,
+  inCircuitVkHash,
 };
 
 type Undefined = undefined;


45930:
Actually, this will be a breaking change, so I will create a new class with the check instead.

src/examples/zkapps/zkapp-self-update.ts
```diff
@@ -59,7 +59,29 @@ Provable.log('original verification key', fooVerificationKey);
...
+    await contract.replaceVerificationKey(illegalVerificationKey);
+  });
+  await tx2.prove();
+} catch (error: any) {
+  if (
+    error.message.includes('The verification key hash is not consistent with the provided data')
+  ) {
+    console.log('correctly threw on illegal verification key');
+  } else {
+    throw error;
```
boray, line 79:
```suggestion
    throw new Error(`Unexpected error: ${error.message}`);
```
This would only be thrown when an unexpected error occurs. It could be better to be more explicit about this.

src/lib/proof-system/verification-key.ts
```diff
@@ -26,6 +27,21 @@ class VerificationKey extends Struct({
       hash: Field(RAW_VERIFICATION_KEY.hash),
     });
   }
+
+  /**
+   * In-circuit check that data and hash are consistent.
+   */
+  static check(x: VerificationKey) {
+    if (x.hash.isConstant()) {
```
boray, line 35: Why do we assume that if the hash is constant, it is correct?

boray (commented):
I think it is cool being able to check VK at the SDK side too.

src/lib/proof-system/verification-key.ts
45930: If it's a constant, we are not in a circuit, so we can't calculate the in-circuit hash.  I don't think we have the tools available to us to confirm validity outside of a circuit.

45930:
> It's not clear to me why this happened. The underlying provable is not changed at all. Do we hash the class at circuit generation?

The new `VerificationKey` has new constraints.  We can't call `check` for free.  The downside to this change is it costs constraints, and doesn't actually do anything, because such illegal keys will be rejected by the network.  So really what we gain is DevX improvement.

0x471 (approved):

boray:
> > It's not clear to me why this happened. The underlying provable is not changed at all. Do we hash the class at circuit generation?
> 
> The new `VerificationKey` has new constraints. We can't call `check` for free. The downside to this change is it costs constraints, and doesn't actually do anything, because such illegal keys will be rejected by the network. So really what we gain is DevX improvement.

Why this needs to be constrained at all. Since the vk is already checked by network, it would be double checked. Isn't there way to not introduce new constraints. Maybe, we can export is as a function instead?

mitschabaude:
What is the intention of this change? It seems pointless to me, because the vk.data is not part of the circuit, so it's impossible to make a proof statement about it.

What the check() method now does is witness an in-circuit VK, checks that it hashes to the hash, and then immediately forget that VK. It feels like a waste of constraints since you're making a statement about something which is not connected to anything else in the circuit

45930:
> Why this needs to be constrained at all. Since the vk is already checked by network, it would be double checked. Isn't there way to not introduce new constraints. Maybe, we can export it as a function instead?

The problem with the Mina Protocol approach is that currently, the transaction will not "fail" with an error message clearly explaining why.  It just won't show up on the block explorer.  We also can't point to a simple line of code that clearly guards against this case.  In `LocalBlockchain` testing, the illegal case works fine, so it gives the impression that this attack is possible.  I think at the cost of a few hundred constraints, some devs will opt into this much clearer check that happens on the SDK side and they can rely on it.

You can also imagine hypotheticals where this would be useful: a custom Mina network that skips this check, an offchain constraint system with sideloaded proof upgradability.

Exporting the function `inCircuitVkHash` was my first thought as well, but you can't call it in JS.  You get the "Cannot call outside of a checked function issue".

src/lib/proof-system/verification-key.ts
```diff
@@ -26,10 +27,63 @@ class VerificationKey extends Struct({
       hash: Field(RAW_VERIFICATION_KEY.hash),
     });
   }
+
+  /**
+   * In-circuit check that data and hash are consistent.
+   */
+  static check(x: VerificationKey) {
```
Trivo25, line 34: This is a breaking change too

src/lib/proof-system/verification-key.ts
```diff
@@ -26,10 +27,63 @@ class VerificationKey extends Struct({
...
 };
+
+class VerificationKeyNext extends Struct({
```
Trivo25, line 53: If we end up introducing a change like this I would be in favor of naming the new class `VerificationKeyV2` and depracating `VerificationKey` - this is how we have been doing it in the past - I feel like `ClassNameNext` can be a bit misleading

Trivo25:
> In `LocalBlockchain` testing, the illegal case works fine, so it gives the impression that this attack is possible.

We can just fix `LocalBlockchain` instead of introducing a breaking change that adds a bunch of constraints (if that's a viable solution) and fixing the Mina node error message in the upcoming HF or providing solutions that dont require circuit changes

src/lib/proof-system/verification-key.ts
45930: I made a new class so it's not breaking

45930: Oh I see I forgot to rm it from the original!  My bad

src/lib/proof-system/verification-key.ts
45930: Yeah, it needs a better name.  Since it's not objectively better, I wondered if `Checked` would work, without deprecating either one.

Trivo25: I feel like two seperate classes for the same thing is confusing. When should you use what class? Instead, why not just add a method `checkHash` or something to the old one - this would achive the same effect without causing confusion and not introducing more overhead

Trivo25: There's an argument to be made for encoding behavior more explicitly via types - for example, introducing `Checked` and `Unchecked` variants like Mina does. But since o1js hasn't followed that pattern so far, introducing it now might cause more confusion, as it wouldn't align with the rest of our type conventions

45930:
> We can just fix LocalBlockchain instead of introducing a breaking change that adds a bunch of constraints (if that's a viable solution) and fixing the Mina node error message in the upcoming HF or providing solutions that dont require circuit changes

Yeah we could "just" do this, but fixing the experience for developers in several months doesn't appeal to me.

`LocalBlockchain` calls into [Transaction_logic](https://github.com/o1-labs/o1js-bindings/blob/f946c48997b8c4a0c105ee0e633c715286584f90/ocaml/lib/local_ledger.ml#L233C5-L233C22), which does not seem like a viable path to do a quick devx fix.  Maybe a more viable option is to carve out an edge case in [updateSubclass](https://github.com/o1-labs/o1js/blob/4ac507dd3037bc7c42267bab14b9333e9d2e5510/src/lib/mina/v1/precondition.ts#L216), but that doesn't solve the case where a verification key hash is set as a state variable, so it leaves edge cases and makes things even more complicated.  The solution I came up with is, I think, the simplest way to consistently block this case within JS and raise a coherent error message.

45930:
> What is the intention of this change? It seems pointless to me, because the vk.data is not part of the circuit, so it's impossible to make a proof statement about it.
> 
> What the check() method now does is witness an in-circuit VK, checks that it hashes to the hash, and then immediately forget that VK. It feels like a waste of constraints since you're making a statement about something which is not connected to anything else in the circuit

The intention is to make the javascript SDK throw errors for illegal inputs.  Currently, `LocalBlockchain` will allow this illegal transaction to happen without checking the vk, and sending the transaction to devnet results in successfully proving the tx locally and sending it

```
{
  status: 'pending',
  data: { sendZkapp: { zkapp: [Object] } },
  errors: [],
  transaction: {
    feePayer: {
      body: [Object],
      authorization: '7mX32iUdcrGBb48MQXDq9zbx2TtK1dPzcUdXKE3EqniJazE45aYpjq744Ui3bYqic1dg15zCkMR8EL3DbWTs7Ebbh6PPb7iG'
    },
    accountUpdates: [ [AccountUpdate] ],
    memo: ''
  },
  setFee: [Function: setFee],
  setFeePerSnarkCost: [Function: setFeePerSnarkCost],
  hash: '5Ju1Pd3fJwuzZ7Q5kagXunpdxg3WfpjzekT9YtJwXSdKSAAByMwe',
  toJSON: [Function: toJSON],
  toPretty: [Function: toPretty],
  wait: [AsyncFunction: wait],
  safeWait: [AsyncFunction: safeWait]
}
```

But then the tx just never appears in the explorer: https://minascan.io/devnet/tx/5Ju1Pd3fJwuzZ7Q5kagXunpdxg3WfpjzekT9YtJwXSdKSAAByMwe

With my change, calling `.prove` on an illegal transaction will fail.  Even if this only affects honest provers, it still provides the correct devx, and it's hard to replicate this any other way.

Trivo25:
> > We can just fix LocalBlockchain instead of introducing a breaking change that adds a bunch of constraints (if that's a viable solution) and fixing the Mina node error message in the upcoming HF or providing solutions that dont require circuit changes
> 
> Yeah we could "just" do this, but fixing the experience for developers in several months doesn't appeal to me.
> 
> `LocalBlockchain` calls into [Transaction_logic](https://github.com/o1-labs/o1js-bindings/blob/f946c48997b8c4a0c105ee0e633c715286584f90/ocaml/lib/local_ledger.ml#L233C5-L233C22), which does not seem like a viable path to do a quick devx fix. Maybe a more viable option is to carve out an edge case in [updateSubclass](https://github.com/o1-labs/o1js/blob/4ac507dd3037bc7c42267bab14b9333e9d2e5510/src/lib/mina/v1/precondition.ts#L216), but that doesn't solve the case where a verification key hash is set as a state variable, so it leaves edge cases and makes things even more complicated. The solution I came up with is, I think, the simplest way to consistently block this case within JS and raise a coherent error message.

Happy to chat more about this next week, but we don't need to change the OCaml implementation _now_, we have been patching the local ledger behaviour via `sendTransaction` for a while now. Take a look at how we check permissions, for example

https://github.com/o1-labs/o1js/blob/3cc34d0e6ae5b913af4e2d3292e179cbb34a0fd8/src/lib/mina/v1/local-blockchain.ts#L164-L171

45930:
If there is another way to make `.prove` fail for the illegal transaction: 

```ts
const illegalVerificationKey = new VerificationKeyNext({
  data: fooVerificationKey!.data,
  hash: barVerificationKey.hash,
});

const tx2 = await Mina.transaction(deployer, async () => {
  // VK with mismatched hash and data should throw
  await contract.replaceVerificationKey(illegalVerificationKey);
});
await tx2.prove();
```

Then I'm open to that.  I agree adding constraints for devx is a pretty suspicious tradeoff.  But the current experience is so bad, that I think it's justified here as a quick fix.  If we could even update the Node software to process the transaction and report that the tx failed with an error text associated, that would be acceptable.

boray:
>  but we don't need to change the OCaml implementation _now_
I agree, I created a quick PR to add vk validation in vk updates. We may also validate vk when proofsEnabled is true.
https://github.com/o1-labs/o1js/pull/2171

45930:
Closing in favor of #2171

PR #2171, opened 2025-04-18
fix: validate vk in updates

boray:


45930:
This looks good, can you add the test case from my PR? `src/examples/zkapps/zkapp-self-update.ts`

45930:
Oh, fix CI first :)

boray:
> Oh, fix CI first :)

wow, I forgot to build locally but I wasn't expecting this  😅

src/lib/mina/v1/transaction-validation.ts
```diff
@@ -305,6 +307,17 @@ async function verifyAccountUpdate(
     }
   });
 
+  if (accountUpdate.update.verificationKey.isSome.toBoolean()) {
+    const vk = VerificationKey.fromValue(accountUpdate.update.verificationKey.value);
+    const circuitVk = Pickles.sideLoaded.vkToCircuit(() => vk.data);
+    const inCircuitHash = inCircuitVkHash(circuitVk);
+    if (inCircuitHash !== vk.hash) {
```
boray, line 314: this is wrong, the condition will always be false 😄

boray:
> This looks good, can you add the test case from my PR? `src/examples/zkapps/zkapp-self-update.ts`

Which test file should have this case? Do you have an opinion?

45930:
> Which test file should have this case? Do you have an opinion?

The file `src/examples/zkapps/zkapp-self-update.ts` already exists.  You can copy the same diff from my PR.  I just want to confirm that this branch has the same behavior as mine.

src/lib/mina/v1/transaction-validation.ts
```diff
@@ -331,6 +344,13 @@ type AuthorizationKind = { isProved: boolean; isSigned: boolean };
 
 const isPair = (a: AuthorizationKind, b: AuthorizationKind) => !a.isProved && !b.isProved;
 
+const isValidVk = (vk: VerificationKey) => {
+  const verificationKey = VerificationKey.fromValue(vk);
+  const circuitVk = Pickles.sideLoaded.vkToCircuit(() => verificationKey.data);
+  const inCircuitHash = inCircuitVkHash(circuitVk);
```
mitschabaude, line 350: nit: would be cleaner to export an out-of-circuit hashing method from ocaml for this purpose

src/lib/mina/v1/transaction-validation.ts
```diff
@@ -331,6 +349,20 @@ type AuthorizationKind = { isProved: boolean; isSigned: boolean };
 
 const isPair = (a: AuthorizationKind, b: AuthorizationKind) => !a.isProved && !b.isProved;
 
+const isValidVk = async (vk: VerificationKey) => {
+  let { runAndCheckSync } = await synchronousRunners();
+  const verificationKey = VerificationKey.fromValue(vk);
+  try {
```
45930, line 355: Do you think it makes sense to also add this method to the struc?  Not as `static check()`, but maybe an instance method called `isValid`?  That would give people more freedom to access it where they want, rather than only in this workflow.

boray: Which struct exactly?

45930: `VerificationKey`

boray: Wouldn't it be a breaking change?

boray: We can export this as a function in` src/lib/proof-system/verification-key.ts`. No more "This function can't be run outside of a checked computation" error.

45930: Why would it be a breaking change if it's just passing some auxiliary data into a function?

`static check()` is a breaking change because it's a secret special method that adds constraints to the circuit, but I don't see why adding `isValid` would be.

boray: I still don't have a full grasp on this. But, I tried adding as an instance method and it caused a type catastrophe. I think it better to keep the PR scoped.

mitschabaude: Yeah, instance methods on Structs are in general not working well, they make some of the Provable method type signatures incorrect.

I came to the conclusion that they should just be avoided; there are different (slightly more complicated) ways to have proper provable classes

mitschabaude: Structs are still great for simple composed types

src/lib/mina/v1/transaction-validation.ts
```diff
@@ -242,9 +243,15 @@ async function verifyAccountUpdate(
         publicOutput: [],
       };
 
-      let verificationKey = account.zkapp?.verificationKey?.data;
+      let verificationKeyRaw = account.zkapp?.verificationKey;
+      assert(verificationKeyRaw !== undefined, 'Account does not have a verification key');
+      let verificationKey = verificationKeyRaw.data;
       assert(verificationKey !== undefined, 'Account does not have a verification key');
```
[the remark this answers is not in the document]
boray: yes, that's right. I had to use non-null assertion operator instead of the assert() which was rejected by the linter somehow. It somehow doesn't bother linter now.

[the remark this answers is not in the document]
Trivo25: Hm, that's odd! Just a small nit: it would be cleaner to keep the `assert` you previously added directly on the verificationKeyRaw object. That way, we assert it's defined up front and can avoid using the non-null assertion operator (`!`) later, like in `await checkVkValidity(verificationKeyRaw!)`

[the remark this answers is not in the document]
boray: I added the `assert` back and updated the message for the second `assert` to make the differentiable.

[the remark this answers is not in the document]
Trivo25: Why two asserts? One `assert` on `verificationKeyRaw` aka `account.zkapp?.verificationKey` is enough to ensure that `verificationKeyRaw.data` exists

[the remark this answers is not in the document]
boray: yeah, that's right.

src/lib/proof-system/verification-key.ts
```diff
@@ -33,3 +35,26 @@ const RAW_VERIFICATION_KEY = {
...
 };
+
+/**
+ * Validates a VerificationKey by checking its hash against the hash computed from its data in-circuit.
```
Trivo25, line 40: `hash computed from its data in-circuit` is a little misleading, I think - we should explicitly mention that this is _not_ an in-circuit operation and it can only be used outside of provable code

src/lib/mina/v1/transaction-validation.ts
Trivo25: agree, would be nice to do this

src/examples/zkapps/zkapp-self-update.ts
```diff
@@ -59,7 +59,30 @@ Provable.log('original verification key', fooVerificationKey);
 
 const { verificationKey: barVerificationKey } = await Bar.compile();
 
+try {
+  const illegalVerificationKey = new VerificationKey({
```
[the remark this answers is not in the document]
boray: I copied the example from the predecessor PR #2170. I haven't thought much about the naming etc. I agree `invalid` sounds better

src/examples/zkapps/zkapp-self-update.ts
```diff
@@ -59,7 +59,30 @@ Provable.log('original verification key', fooVerificationKey);
...
+  await tx2x.sign([deployer.key]).send();
+} catch (error: any) {
+  if (
+    error.message.includes('The verification key hash is not consistent with the provided data')
+  ) {
+    console.log('correctly threw on illegal verification key');
+  } else {
+    throw error;
+  }
+}
```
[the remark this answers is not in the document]
boray: I see your point and I agree. We have lots of similar patterns in `/examples`. AFAIU, the reason we have such test cases in examples is since we use these examples in CI. I don't think naming these as examples is right. But, this is how it has been done so far.

src/lib/proof-system/verification-key.ts
```diff
@@ -26,6 +28,21 @@ class VerificationKey extends Struct({
...
   }
+
+  static async checkValidity(x: VerificationKey): Promise<boolean> {
+    try {
+      let { runAndCheckSync } = await synchronousRunners();
+
+      runAndCheckSync(() => {
+        let vk = Pickles.sideLoaded.vkToCircuit(() => x.data);
+        let inCircuitHash = inCircuitVkHash(vk);
+        inCircuitHash.assertEquals(x.hash);
```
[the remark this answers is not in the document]
boray: It's not clear what `keyDataCircuit` means to me. For `x`, let's change it.

45930:
@richardpringle I can own the original decision to use the example instead of a test.

I'm open to adding a test, but this is a particularly tricky thing to test.  In a sense we're "adding a feature" here, and we can test the behavior of that feature.  But the important thing we're doing with the example (and what the whole examples folder does) is check that the feature we're adding doesn't break other features that we didn't intend to change.  Our examples do double duty where they are run in CI as integration tests, and therefore we also get alerted to cases where the example has become out of date.

I'm open to writing separate integration tests and leaving the examples as examples _only_, but that's a different conversation.  I'm also not defending my original sloppiness that is ultimately what you're calling out here.  Your comments are valid.  But in testing different versions of this PR, running that example script has been a crucial tool to look for VK regressions, and as a CI integration test, it ensures that exactly the right error is thrown in this case, or the test will fail.

👍 to more tests and using _test_ files in general rather than overloading examples with sloppy code, but please don't throw out the regression test benefits of the example.

src/lib/proof-system/verification-key.ts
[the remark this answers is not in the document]
45930: `x` is a commonly-used param in such methods in o1js: https://github.com/search?q=repo%3Ao1-labs%2Fo1js%20(x%3A%20&type=code

Using `key` instead sounds fine, but I think it makes sense to follow convention by default.  If you don't like the convention, then change it, but if every other reviewer has a different style guide, we will never merge any code.  My preference for something like this would be to create linting rules that can be objectively resolved before review.

src/examples/zkapps/zkapp-self-update.ts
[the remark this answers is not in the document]
mitschabaude: @boray I agree with @richardpringle here. The added code here spams the example with something that wasn't its purpose. just because some of the other examples make the same mistake doesn't mean you should repeat it. just add a unit test!

src/lib/proof-system/verification-key.ts
```diff
@@ -26,6 +28,21 @@ class VerificationKey extends Struct({
...
+
+  static async checkValidity(key: VerificationKey): Promise<boolean> {
+    try {
+      let { runAndCheckSync } = await synchronousRunners();
+
+      runAndCheckSync(() => {
+        let vk = Pickles.sideLoaded.vkToCircuit(() => key.data);
+        let inCircuitHash = inCircuitVkHash(vk);
+        inCircuitHash.assertEquals(key.hash);
+      });
```
mitschabaude, line 40: why not just use the async `runAndCheck` here?

src/lib/proof-system/verification-key.unit-test.ts
```diff
@@ -7,3 +7,6 @@ let cached = VerificationKey.dummySync();
 
 assert(generated.data === cached.data, 'data equals');
 assert(generated.hash.equals(cached.hash).toBoolean(), 'hash equals');
+
+let vkIsValid = await VerificationKey.checkValidity(generated);
```
45930, line 11: Could you add the negative test case too?  I think this is nearly ready to approve!

boray: done!

src/lib/proof-system/verification-key.ts
boray: I tried using it but it didn't worked.

45930: It looks like the only difference here is whether we initialize bindings first and return a synchronous function from withing a closure or we await the bindings to be initialized when we call the function:

```ts
  await initializeBindings();
  let id = snarkContext.enter({ inCheckedComputation: true });
  try {
    let finish = Snarky.run.enterGenerateWitness();
    if (!checkConstraints) Snarky.run.setEvalConstraints(MlBool(false));
    await f();
    return finish();
  } catch (error) {
    throw prettifyStacktrace(error);
  } finally {
    if (!checkConstraints) Snarky.run.setEvalConstraints(MlBool(true));
    snarkContext.leave(id);
  }
```

```ts
  await initializeBindings();

  function runAndCheckSync(f: () => void) {
    let id = snarkContext.enter({ inCheckedComputation: true });
    try {
      let finish = Snarky.run.enterGenerateWitness();
      f();
      finish();
    } catch (error) {
      throw prettifyStacktrace(error);
    } finally {
      snarkContext.leave(id);
    }
  }
```

45930 (approved):

PR #2175, opened 2025-04-22
Bump up wasm-bindgen to 0.2.89

dannywillems:


dannywillems:
Mina + proof-systems merged. Updating now o1js with latest `mina/compatible`

PR #2187, opened 2025-04-29
Push to the nix cache in ci

Geometer1729:
- Push nix builds to the mina-nix-cache
- add `nixConfig` to encourage users to use the cache and other sensible nix settings


.github/actions/build/action.yml
```diff
@@ -30,11 +50,20 @@ runs:
...
+        AWS_SECRET_ACCESS_KEY: ${{ inputs.gcp_secret }}
       shell: bash
       run: |
         set -Eeu
         ./pin.sh
+        nix build o1js#bindings --accept-flake-config
+        temp_key=$(mktemp)
+        echo ${{ inputs.nar_secret }} > "$temp_key"
+        nix store sign --key-file "$temp_key" --recursive ./result
+        nix copy --to "s3://mina-nix-cache?endpoint=https://storage.googleapis.com" $(nix path-info ./result)
```
45930, line 64: What is the implication of cahing the nix build?  I assume that users on their local machine will be able to develop locally much faster by downloading freshly cached dependencies?

flake.nix
```diff
@@ -14,6 +14,18 @@
     dune-nix.inputs.flake-utils.follows = "flake-utils";
     flake-utils.url = "github:numtide/flake-utils";
   };
+  nixConfig = {
```
45930, line 17: Is `README-nix.md` up to date, accounting for this flow?  It looks like we're adding config here that is called out as suggestions in the readme.  Could you take a pass over that and prune any irrelevant content?

.github/actions/build/action.yml
```diff
@@ -1,16 +1,36 @@
...
-      run: echo "PATH=$PATH:/nix/var/nix/profiles/default/bin" >> $GITHUB_ENV
+      run: command -v unzstd || (sudo apt-get update && sudo apt-get install -y zstd)
+    - uses: nixbuild/nix-quick-install-action@v30
+      continue-on-error: true
+      with:
+        nix_conf: |
+          keep-env-derivations = true
+          keep-outputs = true
+          extra-substituters = https://storage.googleapis.com/mina-nix-cache
+          extra-trusted-public-keys = nix-cache.minaprotocol.org:fdcuDzmnM0Kbf7yU4yywBuUEJWClySc1WIF6t6Mm8h4= nix-cache.minaprotocol.org:D3B1W+V7ND1Fmfii8EhbAbF1JXoe2Ct4N34OKChwk2c= mina-nix-cache-1:djtioLfv2oxuK2lqPUgmZbf8bY8sK/BnYZCU2iU5Q10=
```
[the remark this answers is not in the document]
Geometer1729: They should have, but they are also redundant here now so I'll just remove them. My bad.

.github/actions/build/action.yml
Geometer1729: Yeah. We already sort of had this, the difference being made here is that we are adding things built in o1js CI to that cache so the cache should be much more useful than when it was only mina builds.

45930 (approved):

PR #2190, opened 2025-04-29
Run fetchGenesisConstants even if fetchLastBlock fails

MartinOndejka:
Genesis constants do not get fetched if `fetchLastBlock` fails. On [Zeko](https://github.com/zeko-labs/zeko), we don't have a last block query so it's fine if it fails, however we still want the o1js to query correct `accountCreationFee`.

### DIFF
--- a/src/lib/mina/v1/fetch.ts
+++ b/src/lib/mina/v1/fetch.ts
@@ -373,9 +373,13 @@ async function fetchMissingData(graphqlEndpoint: string, archiveEndpoint?: strin
       (async () => {
         try {
           await fetchLastBlock(graphqlEndpoint);
-          await fetchGenesisConstants(graphqlEndpoint);
-          delete networksToFetch[network[0]];
         } catch {}
+
+        try {
+          await fetchGenesisConstants(graphqlEndpoint);
+        } catch {}
+
+        delete networksToFetch[network[0]];
       })()
     );
   }
--- a/src/lib/mina/v1/fetch.ts
+++ b/src/lib/mina/v1/fetch.ts
@@ -371,15 +371,18 @@ async function fetchMissingData(graphqlEndpoint: string, archiveEndpoint?: strin
   if (network !== undefined) {
     promises.push(
       (async () => {
-        try {
-          await fetchLastBlock(graphqlEndpoint);
-        } catch {}
+        const [lastBlockOk, genesisOk] = await Promise.all([
+          fetchLastBlock(graphqlEndpoint)
+            .then(() => true)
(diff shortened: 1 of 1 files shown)


src/lib/mina/v1/fetch.ts
```diff
@@ -373,9 +373,13 @@ async function fetchMissingData(graphqlEndpoint: string, archiveEndpoint?: strin
...
         try {
           await fetchLastBlock(graphqlEndpoint);
+        } catch {}
+
+        try {
           await fetchGenesisConstants(graphqlEndpoint);
-          delete networksToFetch[network[0]];
         } catch {}
+
+        delete networksToFetch[network[0]];
```
45930, line 382: It looks like we're changing the behavior here from `delete the networksToFetch entry if fetch was successful` to `delete the networksToFetch entry no matter what`.  I don't think you need a second try/catch block for this, you could just move the `delete` out of the try/catch.

Can you be more surgical? Like only change behavior if the network == zeko?

Trivo25: I think we should avoid behaviour that's tied to specific custom networks and try to implement a solution that works for all networks that provide the interface otherwise we'll have to consider all possible custom networks ids and their custom behaviour in the future

MartinOndejka: I agree with both points, we don't want to change the logic of refetching and also I wouldn't want to hardcode any zeko specific edge cases.

This not only keeps both, but also optimizes for latency since the fetches are done in parallel.
```ts
const [lastBlockOk, genesisOk] = await Promise.all([
  fetchLastBlock(graphqlEndpoint)
    .then(() => true)
    .catch(() => false),
  fetchGenesisConstants(graphqlEndpoint)
    .then(() => true)
    .catch(() => false),
]);

if (lastBlockOk && genesisOk) {
  delete networksToFetch[network[0]];
}
```

boray (approved):

src/lib/mina/v1/fetch.ts
45930: This implementation looks good to me, but won't the `fetchLastBlock` promise still resolve to `false` for zeko?  If this is ok for you, it's ok for me!

45930 (approved):

src/lib/mina/v1/fetch.ts
MartinOndejka: That's okay since we don't support network preconditions at the moment. Now we mainly want to fix the `fundNewAccount` function to work properly with lower account creation fee.

PR #2206, opened 2025-05-03
Comment build scripts with Claude

dannywillems:


mitschabaude:
also it seems to have hallucinated `.sh` scripts in the README :D

dannywillems:
> also it seems to have hallucinated `.sh` scripts in the README :D

It did not hallucinate. There was a bash script on the machine.

dannywillems:
> If I'm being completely honest here, I think this documentation is more of a maintenance burden than it is helpful. The code is incredibly simple and anyone who can't understand it, probably shouldn't be editing it. 
> 
> None of the documentation explains *why* these build scripts exist, which is really the only valuable documentation that could exist here IMO. The words just explain what the code does, so the words are only really helpful to someone who can't read the code. 

This PR was not supposed to be open. It wasn't a real change.

PR #2210, opened 2025-05-06
Prepare Provable BigInt for Standard API

boray:


Trivo25 (commented):
please add a changelog entry!

querolita (approved):
Modulo @Trivo25 suggestions and some comments, looks good to me

src/lib/provable/bigint.ts
```diff
@@ -752,7 +753,7 @@ abstract class ProvableBigInt<T> {
   abstract assertEquals(a: T): void;
 }
 
-function rangeCheck(x: Field, bits: number, signed?: boolean) {
+function rangeCheckHelper(x: Field, bits: number, signed?: boolean) {
```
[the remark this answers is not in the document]
Trivo25: > Do you mean src/lib/provable/gadgets/range-check.ts 
Yes

`rangeCheckHelper` isn't the only range check related function in this file - the same goes for `rangeCheck116` and `rangeCheck128Signed`. Also, `rangeCheckHelper` doesn't seem too specific that would exclude it's use with any other generic input - what this really is, it's a range check factory - I don't see a reason why this can't be with the rest of them

src/lib/provable/bigint.ts
```diff
@@ -752,7 +753,7 @@ abstract class ProvableBigInt<T> {
...
   const supportedBits = new Set([32, 48, 64, 116, 128]);
```
Trivo25, line 757: Also, I just spotted this - why not simply enforce this on the type level? That would be much cleaner and easier given that we use typescript for that exact reason

Fizzixnerd (commented):
Regarding the naming of `_bigint` vs. `_bigInt`, there are still inconsistencies.  I suggest making an issue and following up in a separate PR.

Some examples I found are below.

src/bindings/crypto/bigint-helpers.ts
```diff
@@ -10,6 +10,9 @@ export {
   sign,
   bytesToBigint32,
   bigintToBytes32,
+  isPrime,
+  modularExponentiation,
+  randomBigintInRange
```
Fizzixnerd, line 15: here.

src/bindings/crypto/bigint-helpers.ts
```diff
@@ -234,3 +237,76 @@ function abs(x: bigint) {
...
+  return true;
+}
+
+/**
+ * Generates a random bigint in the range [min, max)
+ * @param min The minimum value (inclusive)
+ * @param max The maximum value (exclusive)
+ * @returns A random bigint in the range [min, max)
+ */
+function randomBigintInRange(min: bigint, max: bigint): bigint {
```
Fizzixnerd, line 292: I notice this one acts on non-provable `bigint`s.  However, the one you changed (`bigIntToLimbs`) also has that property.  So I think it is still an inconsistency.

src/lib/provable/bigint.ts
Fizzixnerd, line 756: Can this be named something better?  `mkRangeCheck` maybe?

src/lib/provable/bigint.ts
```diff
@@ -763,7 +764,7 @@ function rangeCheck(x: Field, bits: number, signed?: boolean) {
       Gadgets.rangeCheck32(x);
       break;
     case 48:
-      rangeCheck48(x);
+      Gadgets.rangeCheckN(48, x);
       break;
     case 64:
       Gadgets.rangeCheck64(x);
```
Fizzixnerd, line 770: If we are commiting to `rangeCheckN`, then I think the rest of them should be changed too.  This should not block this PR however, and should be documented in a new issue and prioritized accordingly.

src/lib/provable/bigint.ts
```diff
@@ -218,9 +225,11 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
      * @returns An array of bits representing the ProvableBigInt
      */
     toBits(): Bool[] {
-      return this.fields.flatMap((field) => {
+      const bits = this.fields.flatMap((field) => {
         return field.toBits(this.Constructor.config.limbSize);
       });
+      ProvableBigInt_.Unsafe.fromBits(bits).assertEquals(this);
```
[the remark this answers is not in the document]
boray: It's essentially a sanity check. If we remove it, it wouldn't change the expected behaviour. But, I would prefer to keep it.

src/lib/provable/bigint.ts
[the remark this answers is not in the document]
boray: I was able to get rid of `rangeCheckHelper ` and I moved `rangeCheck128Signed` to `range-check.ts`.

src/lib/provable/bigint.ts
```diff
@@ -37,8 +38,9 @@ type BigIntParameter = {
  * Implementation details:
  *
  * Internally, a ProvableBigInt is represented as an array of Field elements (limbs),
- * where each limb holds 116 bits as default. The total size is determined by the configuration,
- * with preset options supporting different bit lengths:
+ * where each limb holds 116 bits as default. The number 116 originates from the calculation of 116 x 2 = 232,
```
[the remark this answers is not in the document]
boray: We can't have 128 because of the calculation you mentioned. We don't choose 120 since then we will need a 56-bit range check (64+56) which we don't have. The highest range check we can have under 64-bit is 52-bit. From that, we get the number 116 (64+52)

see: https://github.com/o1-labs/o1js/blob/c39f24ee67f62229c82632c5c40c708e0c20d1ec/src/lib/provable/gadgets/range-check.ts#L43

src/bindings/crypto/bigint-helpers.ts
```diff
@@ -234,3 +237,76 @@ function abs(x: bigint) {
...
+    // Convert bytes to bigint
+    let x = BigInt(0);
+    for (const byte of bytes) {
+      x = (x << 8n) + BigInt(byte);
+    }
+
+    // Check if x is within the range
+    if (x < range) {
+      return min + x;
+    }
```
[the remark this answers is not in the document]
mitschabaude: That obviously wouldn't be uniformly random. Say you create values in the range [0, 255) this way, from a single byte.
When you do % 255 on the byte, then both 0 and 255 are mapped to 0, which means the result is 0 with twice the probability of any other value.

The method I wrote yields uniform distribution. I wouldn't worry about "unbounded time", this is fast in practice

src/lib/provable/bigint.ts
```diff
@@ -141,6 +151,10 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
       return ProvableBigInt_.fromBigInt(BigInt(x));
     }
 
+    static toCanonical(x: ProvableBigInt_): ProvableBigInt_ {
+      return x.mul(ProvableBigInt_.one);
+    }
```
mitschabaude, line 156: this doesn't prove that the result is canonical 🤦🏻‍♂️ it witnesses an arbitrary $q, r$ such that $x * 1 = q * p + r$. nowhere does it prove that $r < p$. the prover could literally witness $q = 1$ and $r = x$.

note that multiplication also _shouldn't_ prove canoncal results. this method should have a check for that!
see how `ForeignField` does this:
https://github.com/o1-labs/o1js/blob/a6b65037a7ac1f1dbe647bb3767d00882cc3b02f/src/lib/provable/gadgets/foreign-field.ts#L429-L437

mitschabaude (changes requested):

mitschabaude (approved):

src/lib/provable/bigint.ts
```diff
@@ -88,6 +94,10 @@ function createProvableBigInt(modulus: bigint, config?: BigIntParameter) {
     modulus < config_.max,
     `ProvableBigInt: modulus exceeds the max supported size of 2^${config_.max}`
   );
+  assert(
```
Trivo25, line 97: Let's enforce these checks on the type level, please. 

Given that these values cannot be dynamic, type level checks will make it easier for the developer to use this API in a type safe way without unexpected runtime errors - it's also less error prone and gives better readability imo

Trivo25 (commented):
I just noticed that typing seems to be off - can anyone else check? Not sure if it's only my environment 

grafik

Trivo25:
> I just noticed that typing seems to be off - can anyone else check? Not sure if it's only my environment
> grafik


The issue persists. Even though the API is experimental, we should still strive to offer users type-safe, fully functional interfaces - even within the experimental namespace. I should have caught the type-safety problems in the original PR; that oversight is on me. Sorry for letting it slip originally

Trivo25:
CI failures are fixed in the latest commit of `main`

PR #2216, opened 2025-05-08
Fixed error in download-bindings.sh for multiple workflow runs

rpanic:
If there existed multiple workflow runs for any given head, then the `RUN_ID` would contain all of the runs, delimited by a whitespace. This broke the script, this PR fixes that by extracting the latest build before downloading


dannywillems:
Related to https://github.com/o1-labs/o1js/pull/2199

rpanic:
@dannywillems Fwiw the fix in #2199 didn't fix it for me, because the limit flag applies to all workflows, not only the ones that we are interested in, which are filtered locally via jq in the next line

hattyhattington17 (approved):

PR #2217, opened 2025-05-08
Export analyzeMethod

rpanic:
For protokit, we want to extract some additional data statically from any given circuit method. This is difficult using `zkprogram.analyzeMethods()` as it runs through all methods at once.
Using `analyzeMethod` directly, this becomes possible

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -61,7 +61,14 @@ export { Reducer } from './lib/mina/v1/actions/reducer.js';
 export { state, State, declareState } from './lib/mina/v1/state.js';
 
 export type { JsonProof } from './lib/proof-system/zkprogram.js';
-export { SelfProof, verify, Empty, Undefined, Void } from './lib/proof-system/zkprogram.js';
+export {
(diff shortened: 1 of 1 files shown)


Trivo25:
I am happy to export this method, but `analyzeMethod` is relatively "low level" as it requires the user to define the method interfaces manually. What do you think about an optional argument to `ZkProgram.analyzeMethods(methodName?: string)` that only analyzes that specific method? Or maybe some an entirely different function dedicated to analyzing single methods?

rpanic:
I think for the sake of API-cleanness, the best would probably to extract most of the zkprogram-specific logic into a method `analyzeMethod(zkProgram, methodName)` that is exported and internally also used by ZkProgram.analyzeMethods()
This however means that we need to make the ZkProgram's methodIntfs somewhat accessible, as that is purely internal right now.
If that makes you uncomfortable, I'd add a method `ZkProgram.analyzeMethod(methodName: string)`

PR #2229, opened 2025-05-20
remove self-hosted runners

Geometer1729:
Removes the self hosted runners and just runs ci on `ubuntu-latest`


.github/actions/build/action.yml
```diff
@@ -45,24 +30,13 @@ runs:
...
     - name: Build the o1js bindings
       if: steps.bindings-cache.outputs.cache-hit != 'true'
-      env:
-        AWS_ACCESS_KEY_ID: ${{ inputs.cache_id }}
-        AWS_SECRET_ACCESS_KEY: ${{ inputs.gcp_secret }}
       shell: bash
       run: |
         set -Eeu
         ./pin.sh
-        nix build o1js#bindings --accept-flake-config
```
Trivo25, line 58: do we not have to build the bindings?

.github/actions/upload/action.yml
```diff
@@ -16,11 +16,22 @@ runs:
   steps:
     - name: Build
       uses: ./.github/actions/build
-      with:
-          cache_id: ${{ inputs.cache_id }}
-          nar_secret: ${{ inputs.nar_secret }}
-          gcp_secret: ${{ inputs.gcp_secret }}
-
+    - name: Build the o1js bindings
```
Trivo25, line 19: oh i guess it moved down here

Trivo25 (approved):
Thanks! This should be easy to revert too once we move back to self hosted runners

PR #2242, opened 2025-05-31
Remove unused bindings functions

Trivo25:


src/bindings/crypto/bindings/bindings.unit-test.ts
```diff
@@ -124,97 +122,8 @@ equivalentRecord(Bigint256Bindings, wasm, {
   caml_bigint_256_deep_copy: { from: [bigint256], to: bigint256 },
 });
 
-equivalentRecord(
```
Trivo25, line 127: none of these are compiled to WASM anymore, so there's nothing to check

PR #2245, opened 2025-06-03
Convert the Circuit API to ZkFunction with ZkProgram-style structure

Shigoto-dev19:
## Description

> In contrast to `ZkProgram`, `Circuit` does not pass through Pickles, but creates a proof with Kimchi directly. Therefore, it does not support recursion, but is also much faster.

The new `ZkFunction` API provides the same functionality as the legacy `Circuit` API, but introduces a more ergonomic, functional structure inspired by ZkProgram.

## Notes

- A `ZkFunction` supports only a single main method; it does not support multiple methods as in `ZkProgram.`
- `privateInputTypes` are declared at the top level (not per-method) because there is only one circuit method.

## TODO

- [ ] Support `undefined` for `publicInputType` (make it optional and handle types and `undefined` as the first argument correctly)
- [ ] Abstract `.provable` for composite provable types


45930:
> ZkFunction nicely abstracts the exact thing ZkFunction does - it’s a single function or circuit with a single responsibility, same a function in a programming language. zkPrograms (or programs) are collections of multiple functions (including recursion)

This works for me IFF you can use a `ZkFunction` as an input to a `ZkProgram`.  e.g.

```ts
const MyFuncA = ZkFunction({ ... });
const MyFuncB = ZkFunction({ ... });
const MyProg = ZkProgram({
  name: 'zzz',
  publicInput: Field,
  publicOutput: Field,
  methods: {
    MyFuncA,
    MyFuncB
  }
});
```

If it doesn't work like this, then it's exactly the reason I don't like the naming.  As you say, one is `a single function` and the other is a `collections of multiple functions`.  The naming implies that they are compatible.

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+     * @example
+     * ```ts
+     * const keypair = await zkf.generateKeypair();
+     * ```
+     */
+    async generateKeypair() {
+      const main = mainFromCircuitData(config);
+      const publicInputSize = config.publicInputType.sizeInFields();
+      await initializeBindings();
+      return prettifyStacktracePromise(
```
Trivo25, line 33: and store the keypair internally so that the user wont have to pass it around manually - we can just return the verification key

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+     */
+    async prove(privateInputs: W, publicInput: P, keypair: Keypair) {
+      const publicInputSize = publicInputType.sizeInFields();
+      const publicInputFields = config.publicInputType.toFields(publicInput);
+      const main = mainFromCircuitData(config, privateInputs);
+      await initializeBindings();
+      return withThreadPool(async () => {
+        const proof = Snarky.circuit.prove(
+          main,
+          publicInputSize,
```
Trivo25, line 57: To answer one of your questions:
> Support undefined for publicInputType (make it optional and handle types and undefined as the first argument correctly)

an 'undefined' public input is simply a public input of size `0` with an empty array of field elements.

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+import { ProvablePure } from '../provable/types/provable-intf.js';
+
+// external API
+export { ZkFunction, Keypair, Proof, VerificationKey };
+
+type ZkFunctionConfig<P, W extends any[]> = {
+  name: string;
+  publicInputType: ProvablePure<P>;
+  privateInputTypes: { [K in keyof W]: ProvablePure<W[K]> };
+  main: (publicInput: P, ...privateInputs: W) => void;
```
Trivo25, line 16: In ZkProgram, the layout of the `main` function (or methods) depends on the type of `publicInputType` - we can derive the shape of the method based on whether we have a provable type as part of the ZkFunction config and type or simply leave the property "empty" (undefined) 

```ts
type Method<PublicInput> = PublicInput extends undefined
  ? {
      method(...args: any): Promise<any>;
    }
  : {
      method(publicInput: PublicInput, ...args: any): Promise<any>;
    };

type MyMethod1 = Method<undefined>;
        // ^ type MyMethod1 = { method(...args: any): Promise<any>; }
type MyMethod2 = Method<Field>;
        // ^ type MyMethod2 = { method(publicInput: Field, ...args: any): Promise<any>; }
```

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * Proves a statement using the public input, private input, and the {@link Keypair} of the circuit(ZkFunction).
+     * @example
+     * ```ts
+     * const keypair = await zkf.generateKeypair();
+     * const proof = await zkf.prove([privateInput], publicInput, keypair);
+     * ```
+     */
+    async prove(privateInputs: W, publicInput: P, keypair: Keypair) {
```
Trivo25, line 49: right now, private inputs seem to be expected in the form of an array - i think it would be good to split them up as well as require the public input as first argument

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+  const { publicInputType } = config;
+  return {
+    /**
+     * Generates a proving key and a verification key for this circuit(ZkFunction).
+     * @example
+     * ```ts
+     * const keypair = await zkf.generateKeypair();
+     * ```
+     */
+    async generateKeypair() {
```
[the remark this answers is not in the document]
Shigoto-dev19: Makes sense, I wanted to keep the naming unchanged so that the `ZkFunction` code components remain directly comparable to the `Circuit` API, making it easier to assess the integrity of the first draft.

Shigoto-dev19:
> > Abstract .provable for composite provable types
> 
> What do you mean by that? Composite types (such as Structs) should already be valid inputs

Composite types is not accurate. I was referring to provable types that expose a `.provable` property, such as `Bytes` and `Ecdsa`.

Ideally, these provable types should be declared directly; just like in the other examples in the codebase.

src/lib/proof-system/zkFunction.ts
Shigoto-dev19: Change addressed in commit: bf71ac28faa4071b342a0da47cc7990063a1f08e

src/index.ts
```diff
@@ -35,7 +35,7 @@ export { provable, provablePure } from './lib/provable/types/provable-derivers.j
 export { Struct } from './lib/provable/types/struct.js';
 export { Unconstrained } from './lib/provable/types/unconstrained.js';
 export { Provable } from './lib/provable/provable.js';
-export { Circuit, Keypair, public_, circuitMain } from './lib/proof-system/circuit.js';
+export { ZkFunction, Keypair } from './lib/proof-system/zkfunction.js';
```
Trivo25, line 38: I don't think `Keypair` needds to be part of the public API anymore

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+     * @throws If `compile` has not been called.
+     *
+     * @example
+     * ```ts
+     * const verificationKey = await zkf.compile();
+     * const proof = await zkf.prove([privateInput], publicInput);
+     * const isValid = await zkf.verify(publicInput, proof, verificationKey);
+     * ```
+     */
+    async verify(publicInput: P, proof: Proof, verificationKey?: VerificationKey) {
```
Trivo25, line 97: I think we can make the `verificationKey` parameter required, that way you can send proofs to an entity that hasn't compiled the contract and they can still verify it

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+      const main = mainFromCircuitData(config);
+      const publicInputSize = publicInputType.sizeInFields();
+      await initializeBindings();
+      _keypair = await prettifyStacktracePromise(
+        withThreadPool(async () => {
+          let keypair = Snarky.circuit.compile(main, publicInputSize);
+          return new Keypair(keypair);
+        })
+      );
+      return _keypair.verificationKey();
```
Trivo25, line 46: nit: returning `{ verificationKey }` would be more in line with the rest of our API

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+     *
+     * @throws If `compile` has not been called.
+     *
+     * @example
+     * ```ts
+     * await zkf.compile();
+     * const proof = await zkf.prove([privateInput], publicInput);
+     * ```
+     */
+    async prove(privateInputs: W, publicInput: P) {
```
Trivo25, line 64: nit: `publicInput: P, ...privateInputs: W` would allow us to pass arguments to the prove function like this `await main.prove(hash, preimage);` without having to wrap them in an array

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+            proof.value,
+            verificationKey.value
+          )
+        )
+      );
+    },
+  };
+}
+
+class Keypair {
```
Trivo25, line 115: nit: I think we can simplify the Keypair abstraciton given that we don't export that to the user

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+    } catch (error) {
+      throw prettifyStacktrace(error);
+    }
+  }
+}
+
+/**
+ * Proofs can be verified using a {@link VerificationKey} and the public input.
+ */
+class Proof {
```
Trivo25, line 147: We should add a `.verify` method to the proof class too and we need to figure out how to differenciate this Proof class from the others with have (the recursion related ones)

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  constructor(value: Snarky.Proof) {
+    this.value = value;
+  }
+}
+
+/**
+ * Part of the circuit {@link Keypair}. A verification key can be used to verify a {@link Proof} when you provide the correct public input.
+ */
+class VerificationKey {
```
Trivo25, line 158: Same here - we need to make sure that somehow we can differenciate between the different types of verification keys

Trivo25 (commented):
Great job, this already looks great! Let's hop on slack and discuss the reamining TODOs before we can ship a first version 👀

src/lib/proof-system/zkFunction.ts
```diff
@@ -0,0 +1,157 @@
...
+export { ZkFunction, Keypair, Proof, VerificationKey };
+
+type ZkFunctionConfig<P, W extends any[]> = {
+  name: string;
+  publicInputType: ProvablePure<P>;
+  privateInputTypes: { [K in keyof W]: ProvablePure<W[K]> };
+  main: (publicInput: P, ...privateInputs: W) => void;
+};
+
+function ZkFunction<P, W extends any[]>(config: ZkFunctionConfig<P, W>) {
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in commit: [4c2890b](https://github.com/o1-labs/o1js/pull/2245/commits/4c2890bcb0cefe1473ae95b6578d640e9561a17d)

Note: The [Get](https://github.com/o1-labs/o1js/blob/main/src/lib/proof-system/zkprogram.ts#L1220) type is not used to infer the keys of `ZkFunctionConfig` as in `ZkProgram` because `publicInput` is still not handled automatically when undefined and hence we don't need more accurate inference for optional keys.

src/lib/proof-system/zkfunction.ts
Shigoto-dev19: Nice Catch!

Fixed in commit: https://github.com/o1-labs/o1js/pull/2245/commits/ea8965c39368b042cdabb7ef2ea5b2ea4053088c

src/lib/proof-system/zkfunction.ts
Shigoto-dev19: Fixed in commit: https://github.com/o1-labs/o1js/pull/2245/commits/e07118edea9ef65cffbe3e9c4f2a2b08c404d607

src/index.ts
Shigoto-dev19: I agree, but we still need the `Keypair` class to call the `constraintSystem` method.
Wdyt about removing it from the public API and adding `constraintSystem` directly into the `ZkFunction` API with a warning that `compile` should be called beforehand to have internal access to the keypair.

src/lib/proof-system/zkfunction.ts
Shigoto-dev19: Done in: https://github.com/o1-labs/o1js/pull/2245/commits/f509d868c62cf6e1d6f42efb48aba26321b1676f

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+): {
+  compile: () => Promise<{ verificationKey: VerificationKey }>;
+  constraintSystem: () => Promise<Gate[]>;
+  prove: ProveMethodType<Config>;
+  verify: VerifyMethodType<Config>;
+} {
+  const publicInputType = provablePure(config.publicInputType ?? Undefined);
+  const hasPublicInput = config.publicInputType !== undefined;
+
+  type Keypair = Snarky.Keypair;
```
Shigoto-dev19, line 68: Note: `Snarky.Keypair` type is `unknown` and could be removed here.

src/examples/benchmarks/mul-web.ts
```diff
@@ -76,19 +75,19 @@ if (withPickles) {
...
   toc();
 
   tic('compile 2');
-  kp = await circuit.generateKeypair();
+  verificationKey = (await circuit.compile()).verificationKey;
   toc();
 
   tic('prove');
-  let p = await circuit.prove([], [], kp);
+  let p = await circuit.prove(Undefined.empty());
```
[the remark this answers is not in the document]
Shigoto-dev19: Yes, you're right and that's what the latest changes aimed to improve.

Trivo25 (commented):
Great work! The public API is clean, intuitive, and fully functional :) 

That said, our internal types are still fragile, leading to a lot of type casting. Let’s dedicate the coming week to cleaning them up so the codebase remains stable both inside and out.

- fix type consistency and address the comments i left
- get rid of casts (as, any, etc.) where possible
- consolidate shared types to reduce duplication

src/lib/provable/test/primitives.unit-test.ts
```diff
@@ -1,22 +1,23 @@
...
-  @circuitMain
-  static main() {
+const primitives = ZkFunction({
+  name: 'Primitives',
+  privateInputTypes: [],
+  main: () => {
     // division
     let x64 = Provable.witness(UInt64, () => 10n);
     x64.div(2).assertEquals(UInt64.from(5));
     let x32 = Provable.witness(UInt32, () => 15n);
```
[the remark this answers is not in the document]
querolita: I am checking again. Looks like this works for both 15 and 12... Probably the `div` function is just checking the integer quotient, regardless of whether it is a division with zero remainder or not. Maybe let's use less confusing values for the unit test.

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+      verificationKey: VerificationKey
+    ) => Promise<boolean>;
+
+function ZkFunction<Config extends ZkFunctionConfig>(
+  config: Config & {
+    main: InferMainType<Config>;
+  }
+): {
+  compile: () => Promise<{ verificationKey: VerificationKey }>;
+  constraintSystem: () => Promise<Gate[]>;
```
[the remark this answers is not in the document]
Shigoto-dev19: Done in e1c4277c11967354afa58b6a610e92bda3e854ee

src/examples/benchmarks/mul-web.ts
```diff
@@ -76,19 +74,19 @@ if (withPickles) {
   let circuit = simpleKimchiCircuit(nMuls);
 
   tic('compile 1 (includes srs creation)');
-  let kp = await circuit.generateKeypair();
+  let { verificationKey } = await circuit.compile();
   toc();
 
   tic('compile 2');
-  kp = await circuit.generateKeypair();
+  verificationKey = (await circuit.compile()).verificationKey;
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in ae66a39946072fbea8cf9df53c4475b1249c2c0b

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
+import { Gate, Snarky, initializeBindings } from '../../snarky.js';
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in 85ad25179460069908f38fa02a1da8665d8b3f5f

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+}
+
+function mainFromCircuitData<Config extends ZkFunctionConfig>(
+  config: Config & { main: InferMainType<Config> },
+  privateInputs?: PrivateInputs<Config>
+): Snarky.Main {
+  return function main(publicInputFields: MlFieldArray) {
+    let id = snarkContext.enter({ inCheckedComputation: true });
+    try {
+      const publicInput = provablePure(config.publicInputType ?? Undefined).fromFields(
```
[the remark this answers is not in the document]
Shigoto-dev19: Done in a7c1fe217d4174df5172d7d4e8731a240862b0e5

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+     * @example
+     * ```ts
+     * const { verificationKey } = await zkf.compile();
+     * const proof = await zkf.prove(publicInput, privateInput1, privateInput2);
+     * ```
+     */
+    async prove(...args: any[]) {
+      if (!_keypair) throw new Error('Cannot find Keypair. Please call compile() first!');
+
+      const publicInput = hasPublicInput ? args[0] : Undefined.empty();
```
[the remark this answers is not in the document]
Shigoto-dev19: Updated in a7c1fe217d4174df5172d7d4e8731a240862b0e5

src/lib/proof-system/zkfunction.ts
```diff
@@ -83,10 +84,11 @@ function ZkFunction<Config extends ZkFunctionConfig>(
     async compile() {
       const main = mainFromCircuitData(config);
       const publicInputSize = publicInputType.sizeInFields();
+      const lazyMode = config.lazyMode ?? false;
       await initializeBindings();
       _keypair = await prettifyStacktracePromise(
         withThreadPool(async () => {
-          return Snarky.circuit.compile(main, publicInputSize, false);
+          return Snarky.circuit.compile(main, publicInputSize, lazyMode);
```
querolita, line 96: yes, that's the only change needed for the lazy mode

Trivo25 (commented):
Nice, great job! I left some comments but we are almost there, only some minor things remaining <3 

edit: Oh, and let's add this to the Experimental namespace until we've fixed the other issues we have identified

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,286 @@
...
+     *
+     * @throws If `compile` has not been called.
+     *
+     * @example
+     * ```ts
+     * const { verificationKey } = await zkf.compile();
+     * const proof = await zkf.prove(publicInput, privateInput1, privateInput2);
+     * ```
+     */
+    async prove(...args: Parameters<ProveMethodType<Config>>) {
```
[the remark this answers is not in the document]
mitschabaude: my 2c: public and private inputs should be separate fields in an object. the optional first public argument is painful to work with.

I liked the zkfunction-like API we used in mina-attestations here: https://github.com/zksecurity/mina-attestations/blob/80f777989e41a119926c8017c92ee216bb123a1a/src/credential-imported.ts#L300-L313

used here for example: https://github.com/zksecurity/mina-attestations/blob/80f777989e41a119926c8017c92ee216bb123a1a/src/imported/ecdsa-credential.ts#L69-L91

src/lib/proof-system/zkfunction.ts
Shigoto-dev19: Added in these commits: 1dfe14559b2fbb3e03523cce5c058f3b7138f44c & 0b71aa3085b65f6b28ead4d378ba6755d845d8df

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,286 @@
...
+class Proof {
+  value: Snarky.Proof;
+  publicInputFields: Field[];
+
+  constructor(value: Snarky.Proof, publicInputFields: Field[]) {
+    this.value = value;
+    this.publicInputFields = publicInputFields;
+  }
+
+  async verify(verificationKey: VerificationKey) {
```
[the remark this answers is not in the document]
Shigoto-dev19: Added here 4203ef7b59e83438ac1fc6ba7ce9c26256cc66a9

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,286 @@
...
+    /**
+     * Generates and stores a proving key and a verification key for this circuit(ZkFunction).
+     *
+     * @returns The generated verification key.
+     *
+     * @example
+     * ```ts
+     * const { verificationKey } = await zkf.compile();
+     * ```
+     * @warning Must be called before `prove` or `verify`.
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in 4203ef7b59e83438ac1fc6ba7ce9c26256cc66a9

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,286 @@
...
+
+    /**
+     * Returns a low-level JSON representation of the constraint system (gates)
+     *
+     * @throws If compile() has not been called yet.
+     *
+     * @example
+     * ```ts
+     * await zkf.compile();
+     * const cs = await zkf.constraintSystem();
```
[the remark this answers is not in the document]
Shigoto-dev19: Nice catch!
Fixed in 4203ef7b59e83438ac1fc6ba7ce9c26256cc66a9

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+import { prettifyStacktrace, prettifyStacktracePromise } from '../util/errors.js';
+import { ProvableTypePure } from '../provable/types/provable-intf.js';
+import { provablePure, InferProvable } from '../provable/types/provable-derivers.js';
+import { Tuple } from '../util/types.js';
+import { TupleToInstances, Undefined } from './zkprogram.js';
+
+// external API
+export { ZkFunction, Proof, VerificationKey };
+
+type Get<T, Key extends string> = T extends { [K in Key]: infer _Value } ? _Value : undefined;
```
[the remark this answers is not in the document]
Shigoto-dev19: Moved the `Get` generic type to `src/lib/util/types.ts`: e571099a87bbbd515832490aefa9348801d501bb

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @throws If `compile` has not been called.
+     *
+     * @example
+     * ```ts
+     * const { verificationKey } = await zkf.compile();
+     * const proof = await zkf.prove(publicInput, privateInput1, privateInput2);
+     * ```
+     */
+    async prove(...args: any[]) {
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in 1d2fc64717c41d6adbd0babdd2f2c31269d9d770

src/lib/proof-system/zkfunction.ts
```diff
@@ -0,0 +1,248 @@
...
+     * @throws If `compile` has not been called.
+     *
+     * @example
+     * ```ts
+     * const { verificationKey } = await zkf.compile();
+     * const proof = await zkf.prove(publicInput, privateInput1, privateInput2);
+     * const isValid = await zkf.verify(publicInput, proof, verificationKey);
+     * ```
+     */
+    async verify(...args: any[]) {
```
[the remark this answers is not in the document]
Shigoto-dev19: Fixed in 1d2fc64717c41d6adbd0babdd2f2c31269d9d770

CHANGELOG.md
```diff
@@ -18,6 +18,12 @@ This project adheres to
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/045b1ab...HEAD)
 
+### Changed
+
+- Replace the `Circuit` API with the `ZkFunction` API, which provides the same
```
Trivo25, line 23: Please add that it's currently only available under the `Experimental` namespace, explicitly mention that `Circuit` has been deprecated

Trivo25 (approved):
Some tiny comments but approving in advance, great job!

CHANGELOG.md
```diff
@@ -18,10 +18,11 @@ This project adheres to
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/045b1ab...HEAD)
 
-### Changed
+### Deprecated
```
Trivo25, line 21: We added `ZkFunction` so that deserves it only entry!

PR #2265, opened 2025-06-29
fix: Correct TypeScript export declarations for type-only exports

Fizzixnerd:
## Summary
This PR fixes build errors caused by exporting TypeScript types as values instead of using proper `export type` declarations.

## Fixed Exports
- **foreign-field.ts**: `ForeignField`, `AlmostForeignField`, `CanonicalForeignField`
- **poseidon.ts**: `ProvableHashable` 
- **bytes.ts**: `FlexibleBytes`
- **provable-generic.ts**: `From`, `InferValue`, `InferJson`, `IsPure`
- **cache.ts**: `CacheHeader`
- **sign-zkapp-command.ts**: `CallForest`

## Problem
These exports were causing esbuild errors like:
```
✘ [ERROR] No matching export in "src/lib/provable/foreign-field.ts" for import "ForeignField"
```

The issue occurred because these items are TypeScript type aliases or interfaces, not runtime values, but were being exported as values using `export { ... }` instead of `export type { ... }`.

## Solution
- Changed value exports to type exports where appropriate
- Updated corresponding import statements to use `import type { ... }`
- Maintains backward compatibility for type usage
- Fixes build system compatibility with bundlers like esbuild

## Test plan
- [x] Build system now works without export errors
- [x] Existing functionality unchanged (types still available for use)
- [x] No runtime behavior changes (only build-time type exports)

🤖 Generated with [Claude Code](https://claude.ai/code)


Fizzixnerd:
Did not mean to do that, sorry for the spam.

Issue #2266, opened 2025-06-30
Support chunking in zkFunction


querolita:
Explore how chunking could be enabled in the upcoming zkFunction API https://github.com/o1-labs/o1js/pull/2245/

querolita:
@Trivo25 @Shigoto-dev19 

Sharing our thread for traceability

**What is chunking**
A mechanism inside Kimchi to be able to fit more complex circuits (more than 2¹⁶ rows) with just one proof, except that it is a longer one. Behind the hood, what it is doing is splitting the long circuit into "chunks" of 2¹⁶ size and then the prover creates a random combination of the chunks with a verifier challenge to condense them as if it was originally at most 2¹⁶ rows. But this means there are more polynomials to commit to than with a single-chunk circuit. But details should remain as a black box for us in this task.

**What is the state of this**
 - It is supported on Kimchi. When the number of rows is longer than the domain size, the number of chunks is inferred (a division) accordingly.
- It is supported on Pickles. But here, the number of chunks needs to be passed manually as an argument `num_chunks` of certain functions involving compilation of the circuit etc. Why not automatically? Mainly because recursion adds some rows to the kimchi circuit alone, so maybe it would require just 1 chunk on the kimchi side but as soon as pickles is involved then you need 2 for example. Not perfect, but also I don't think there's any deterministic method to predict ahead of time the exact number of extra rows involved in step proofs to decide whether or not to +1 the naive number of chunks or not...
- On o1js, we recently [merged](https://github.com/o1-labs/o1js/pull/1905) some API changes and fixed some bugs ([1](https://github.com/MinaProtocol/mina/pull/16057) [2](https://github.com/MinaProtocol/mina/pull/16092) [3](https://github.com/o1-labs/proof-systems/pull/2590) [4](https://github.com/o1-labs/proof-systems/pull/3220) [5](https://github.com/o1-labs/proof-systems/pull/3222) [6](https://github.com/o1-labs/proof-systems/pull/2521)) to technically support chunking inside zkprograms (basically adding placeholders for the `numChunks` parameter, equivalent to `num_chunks` in the ocaml code). In practice though, the wasm memory limit is reached if you have more than 2^16 rows, so the program hangs forever trying to generate the proof.

**Future tasks**
[...]

PR #2280, opened 2025-07-07
[WIP] Enabling chunking in ZkFunction

querolita:
`./run src/examples/zkfunction/function-with-chunking.ts` keeps running forever in my local tests, let's see what CI does with it


querolita:
Closing after https://github.com/o1-labs/o1js/issues/2266

PR #2290, opened 2025-07-10
CI: configure dependabot

dannywillems:


.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    reviewers:
+      - "o1-labs/o1js-maintainers"
```
45930, line 10: Who is this?  The teams I know about are:

image

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
...
+    labels:
+      - "dependencies"
+      - "javascript"
+    commit-message:
+      prefix: "npm"
```
45930, line 15: Not sure we need this commit prefix, but 🤷

.github/dependabot.yml
```diff
@@ -0,0 +1,31 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
```
45930, line 4: Can we group either all deps, or production + development deps so that there are fewer total PRs?  Our CI is a nightmare, and opening 1 PR per dependency seems like it will not be adopted by the team.

45930: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#groups--

45930 (commented):
Willing to give this a try!

.github/dependabot.yml
dannywillems: See [5a325dd](https://github.com/o1-labs/o1js/pull/2290/commits/5a325ddb82d065d9a7ab294af5478bf4d1cd4294)

.github/dependabot.yml
dannywillems: It should give something like: `npm(dev): bump @types/node from 18.x to 20.x`.

45930 (approved):

PR #2322, opened 2025-07-23
Support Foreign Field Initialization from Field

45930:


src/lib/provable/foreign-field.ts
```diff
@@ -99,6 +103,19 @@ class ForeignField {
...
+    if (x instanceof Field) {
+      if (x.isConstant()) {
+        let value = x.toBigInt();
+        if (value >= p) {
+          throw new Error(
+            `Field value ${value} exceeds foreign field modulus ${p}. Please ensure the value is reduced to the modulus.`
+          );
+        }
+      }
+      this.value = fieldToField3(x);
```
45930, line 116: This function range checks each limb to < 2^88, so the resulting `Field3` _must_ be comprised of 3 limbs < 2^88.  It is further constrained that

```ts
// prove that x == x0 + x1*2^l + x2*2^2l
let x_ = x0.add(x1.mul(1n << l)).add(x2.mul(1n << l2));
x_.assertEquals(x);
```

Yet there is a warning in the comments:

```
**Warning:** the output is underconstrained up to a multiple of the modulus that could be added to the bigint.
```

I'm wondering if there is a relevant attack here that is not avoided by asserting the canonicalness of the result, or potentially asserting that the modulus of the foreign field must be > 2^88.

mitschabaude: I don't think so, so the only thing remaining to make this safe is checking that the result is canonical! Currently you only do that for constants

PR #2326, opened 2025-08-11
Update rust version in o1js nix flake to match proof systems

hattyhattington17:
Update rust version in o1js nix flake to match proof systems https://github.com/o1-labs/proof-systems/blob/master/Makefile#L33


bleepbloopsify:
this is done on `develop`, since we can't bump `mina` without breaking node

PR #2331, opened 2025-08-13
Add test values for base field overflows

boray:
This PR adds several new test private keys to the `privateKeys` array in `src/mina-signer/src/signature.unit-test.ts`. The new keys are specifically chosen to be larger than the modulus of the base field (`p`) but smaller than the modulus of the scalar field (`q`).

see [mina#17613](https://github.com/MinaProtocol/mina/pull/17613/files#diff-17949d726dc1da8a51017a79c1c081a6a475a24afb31f35a5baf254a584a4c9a)


src/mina-signer/src/signature.unit-test.ts
```diff
@@ -74,6 +74,17 @@ let privateKeys = [
   13507121957360975407754936805517309719995432919287716894090189496158972691353n,
   19152183577529784585604205414612388832924448123778079866321952213819564835523n,
   20617898261860919949586898353988281719931035849127590497832022423723702051831n,
+  28948022309329048855892746252171976963363056481941560715954676764349967630337n, // the keys below this line (including) are larger than p smaller than q
+  28948022309329048855892746252171976963363056481941560715954676764349967630337n, // where p is the modulus of the base field
```
dannywillems, line 78: This seems to be duplicated with the above.

PR #2388, opened 2025-08-22
Bump mina to `569e99`

Trivo25:


Shigoto-dev19 (approved):
I confirm that the performance regression is fixed thanks to this PR :partying_face:

Issue #2101, opened 2025-03-27
runtime table API


Trivo25:
**Describe the feature or improvement**

We can now sucessfully use runtime tables in o1js https://github.com/o1-labs/o1js/pull/1858! 
Right now, developers have to use raw gates and manage everything themselves. While it's good to give developers low level and powerful tools to design their own applications, I believe we should additionally wrap this in a simple API for the average developer (TBD)

Trivo25:
Something as simple as 

```ts
// Create a runtime table with predefined indices
const table = new RuntimeTable(2, [0n, 1n, 2n, 3n, 4n, 5n, 6n, 7n, 8n, 9n]);
// or maybe `new RuntimeTable(tableId, length)`

// Insert values into the table
table.insert([
  [5n, v0], 
  [6n, v1], 
  [7n, v2]
]);

// Check values at specific positions
table.check(2n, v0);
table.check(6n, v1);
table.check(6n, v1); // Can check multiple times
```

would do for now imo

PR #2402, opened 2025-08-28
Change runtime table API as a `RuntimeTable` class

querolita:
Closes https://github.com/o1-labs/o1js/issues/2101

~Note: I removed the old syntax assuming everyone would like to move to this API, but I can keep it to make it backwards compatible.~

TODO: get rid of the `withRuntimeTable` boolean flag to have o1js automatically obtain the `runtimeTable` feature flag from Snarky: https://github.com/o1-labs/o1js/issues/2405


src/lib/provable/gadgets/runtime-table.ts
```diff
@@ -0,0 +1,94 @@
...
+     *
+     * @param pairs Array of pairs [index, value] to insert into the runtime table.
+     */
+    insert(pairs: [bigint, Field][]) {
+        for (let i = 0; i < pairs.length; i += 3) {
+            const chunk = pairs.slice(i, i + 3);
+            const [idx0, value0] = chunk[0];
+            const [idx1, value1] = chunk[1] || [idx0, value0];
+            const [idx2, value2] = chunk[2] || [idx0, value0];
+            Gates.lookup(Field.from(this.id), Field.from(idx0), value0, Field.from(idx1), value1, Field.from(idx2), value2);
```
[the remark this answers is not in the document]
querolita: I thought the structure might be heavier in practice if we were storing the index as a `Field` directly. And also, back on the comment above, if we do add some "smart" shortcuts to prevent from looking up indices that were never initialized, it is cleaner if those indices are bigints vs fields.

src/lib/provable/test/runtime-table.unit-test.ts
```diff
@@ -0,0 +1,70 @@
...
+import { FeatureFlags } from '../../proof-system/feature-flags.js';
+import { Cache } from '../../proof-system/cache.js';
+
+let uint = (n: number | bigint): Spec<bigint, Field> => {
+  return fieldWithRng(Random.bignat((1n << BigInt(n)) - 1n));
+};
+
+// Runtime table tests
+{
+  let RuntimeTableZkProgram = ZkProgram({
```
Shigoto-dev19, line 17: Even though this is a unit test, I think it would be more valuable to highlight a practical use case for runtime tables. For example, the `ZkProgram` could take a private input as a preimage of a Poseidon hash. We could then hardcode specific digest bytes into the runtime table at given indices, and perform lookups to check whether certain byte sequences (e.g., the last 3-6 bytes) are zero; similar to a mining example. This would let us validate that the digest of a private preimage ends with, say, three zero bytes.

If implementing the zero-byte condition is too complex, a simpler approach would be to use any preimage, hardcode certain digest-derived pairs in the table, and include happy as well as unhappy tests to demonstrate how the API fails on invalid lookups.

Shigoto-dev19: This is just a proposition, the example as is; is fine as well!

hattyhattington17 (approved):

PR #2404, opened 2025-09-02
Enable zkProgram.compile() with cache in browsers

ymekuria:
## Summary

Enables `ZkProgram.compile({ cache })` to work in browsers. Previously, using a cache in web environments with a `ZkProgram` was disabled because the underlying lagrange basis functions were not compatible with the web worker API and would cause compilation to hang indefinitely.

## Background

The `ZkProgram.compile({ cache })` functionality was broken in browsers because:
1. Lagrange basis functions returned `WasmVector` types incompatible with web worker transfer
2. These functions were disabled in `worker-spec.js` to prevent hanging
3. Cache writing would fail silently or hang when attempting to compute lagrange basis

## Changes

- Added 2 new pointer-based function registrations to `worker-spec.js`
- Updated `srs.ts` with environment-aware `getLagrangeBasis` function
- Automatically detects Node.js vs Web environments
- Uses pointer functions in browsers, direct functions in Node.js

## Related PRs

This is part 3 of 3 for the complete fix:
1. [proof-systems#3325 (Rust pointer functions)](https://github.com/o1-labs/proof-systems/pull/3325)
2. [mina#17707 (submodule reference update)](https://github.com/MinaProtocol/mina/pull/17707)
3. **This PR**: o1js (worker spec + TypeScript integration)

## Testing

Manually tested in browser environment:
- Created test project with zkProgram.compile() using a cache in the browser
- Verified compilation completes successfully without hanging
- Confirmed cache files are written (lagrange-basis-fp-*, etc.)
- Pointer functions return valid memory addresses

### DIFF
--- a/src/bindings/crypto/bindings/srs.ts
+++ b/src/bindings/crypto/bindings/srs.ts
@@ -87,8 +87,8 @@ function srsPerField(f: 'fp' | 'fq', wasm: Wasm, conversion: RustConversion) {
 
     if (isWeb) {
       // Web: Use new pointer-based functions for worker compatibility
-      const ptr = wasm[`caml_${f}_srs_get_lagrange_basis_ptr`](srs, n);
-      return wasm[`caml_${f}_srs_get_lagrange_basis_read_from_ptr`](ptr);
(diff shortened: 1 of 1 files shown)


Trivo25:
continuing the discussion on slack

PR #2417, opened 2025-09-05
Allow signing zkapp command without fee payer private key

MartinOndejka:
Currently `mina-signer` doesn't allow signing the zkapp command account updates, that have different public key than fee payer. This limits use-cases for collaborative construction of zkapp command, where one user signs the account update, and the other pays the fee.

This change removes this restriction by signing account updates that have the corresponding public key to the supplied private key, and not the ones that are equal to the fee payer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,13 +21,9 @@ This project adheres to
 ## [2.9.0](https://github.com/o1-labs/o1js/compare/4b1dccdd...114acff) - 2025-09-02
 
 ### Added
+
 - Support for `ForeignField.Unsafe.fromField` as an alternative constructor
   https://github.com/o1-labs/o1js/pull/2322
-
-### Changed
-
-- Updated Rust nightly version from `2024-06-13` to `2024-09-05` to match the
-  version used in Mina proof-systems repository.
 
 ### Fixed
 


Hebilicious:
@45930 @querolita 
I can describe the scenario we're facing :
We have a transaction that is created server side, where the feePayer is the zeko sequencer. 
This transaction must be signed by the user, then send.
Currently this works with o1js, but not with `mina-signer`. It doesn't work with `mina-signer` because `mina-signer` doesn't sign the necessary account update due to the fact it assumes that the public key is the one of the feePayer, not the one of the user.
Here's a diff of what happens https://www.diffchecker.com/Pn7nIFvq/, on the left the transaction signed with o1js, on the right the transaction signed with `mina-signer` (using the same private key).

If you're worried about a breaking change, maybe this could accept a public key argument.

querolita:
My concern is whether or not this would be the only change that's necessary to support the desired behaviour. Meaning, for usual transactions, the fix still will allow the same workflow. But I am wondering if there could be any failing assumption somewhere else in the system if it encounters a zkapp command that's not been authorized (yet). We don't have any such unit-test/scenario, so it is hard to predict on the top of my head. Hence, I just wanted to raise awareness that this might not solve the issue alone, but maybe it could need further changes. Hopefully not.

I am on board with giving this flexibility to the mina signer, but I would feel more comfortable if we documented this change accordingly, for future reference in case we were reported with unexpected behaviours. Unfortunately, mina-signer has received very little attention when it comes to documentation, but it is time to change that trend 😄 . Feel free to add your notes as comments. Otherwise, I can push some to this branch. Would that be OK?

querolita (commented):
@Hebilicious Indeed, that's one of the places where the change could lead to unexpected behaviours:

```typescript
  // verify fee payer signature
  let signature = Signature.fromBase58(zkappCommand.feePayer.authorization);
  let ok = verifyFieldElement(signature, fullCommitment, publicKey, networkId);
  if (!ok) return false;
```

If one tries to verify the signature using the public key of a different party, this assertion will fail. Let me push a change accordingly, together with some comments. Let me revert my feedback to avoid merging unstable states, so it can all go together within the same PR.

Either way, all signatures need to be verified sooner or later. It could make sense to include a parameter including the target public key as you were suggesting to make it clearer in the code which one to use. I will work on this today.

mitschabaude:
let me just add that verifying signatures is not among the primary use cases of mina-signer, since it's not an operation that is usually done wallet-side. the core use case of mina-signer is to put signatures on transactions that will later be verified on the node side. the node will of course verify all the AU signatures!

personally I have no concerns about this PR, I think it's good!

MartinOndejka:
Another note is that this is not publicly visible function, so adding docs to it won't be super helpful. The public function that uses it is here https://github.com/o1-labs/o1js/blob/main/src/mina-signer/mina-signer.ts#L388-L391. I didn't want to introduce any breaking changes so I didn't touch it, however the interface of it is very questionable. It's weird that the arguments are `feePayer` and then the `zkappCommand` which also includes fee payer. The return type again, has duplicit fee payer and for some reason also returns public key of the fee payer?

PR #2436, opened 2025-09-16
update Transaction.fromJSON to accept serialized JSON string

45930:
## Summary

`tx.toJSON` returns a stringified JSON object, but `Transaction.fromJSON` does not accept a string input.  This makes is slightly awkward to work with serialized JSON transactions.

### Fix

This PR adds support for string as input to the `Transaction.fromJSON` function by adding a `ZkAppCommand.parse` function for handling string input.


Shigoto-dev19 (approved):
LGTM!

PR #2447, opened 2025-09-24
upgrade: bump mina

bleepbloopsify:
upgrade: bumping mina (after a long time) to `compatible
(8f958aac99f53f27ce34e62861775d7117821a2c)`

downgrades dune to 3.1 to be compatible
upgrades node to 22.0 for wasm support (across our CI)
nix rust toolchain upgrade to 2024-09-05
fixes API incompatibilities
adds `.gitattributes` to some improper files
fixes our `dune` files to have more standard flag usage (for better
inheritance)
adds a dependency debug line to a dune invocation in order to give
better tracing for objects when building (doesn't affect happy path)

Status: expected to work as soon as CI finishes

this is the root of a trunk.
children: #2502

### DIFF
--- a/src/bindings/crypto/.gitattributes
+++ b/src/bindings/crypto/.gitattributes
@@ -1,0 +1,1 @@
+constants.ts linguist-generated=true
--- a/src/bindings/ocaml/lib/local_ledger.ml
+++ b/src/bindings/ocaml/lib/local_ledger.ml
@@ -237,7 +237,7 @@ let apply_zkapp_command_transaction l (txn : Zkapp_command.Stable.Latest.t)
   check_account_update_signatures txn ;
   let ledger = l##.value in
   let application_result =
-    Transaction_logic.apply_zkapp_command_unchecked
+    Transaction_logic.apply_zkapp_command_unchecked ~signature_kind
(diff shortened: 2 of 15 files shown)


bleepbloopsify:
I don't have a nix setup, so if someone does have one, it would be nice if they could go format that `flake.nix` (if it isn't formatted)

src/bindings/crypto/.gitattributes
Shigoto-dev19: Good call!

src/bindings/ocaml/lib/local_ledger.ml
```diff
@@ -237,7 +237,7 @@ let apply_zkapp_command_transaction l (txn : Zkapp_command.Stable.Latest.t)
   check_account_update_signatures txn ;
   let ledger = l##.value in
   let application_result =
-    Transaction_logic.apply_zkapp_command_unchecked
+    Transaction_logic.apply_zkapp_command_unchecked ~signature_kind
```
Shigoto-dev19: This change seems different that what the PR is addressing.
Could you add more details about it in the PR description?

Trivo25: this is due to a change in the mina repo I'd imagine

PR #2449, opened 2025-09-25
Fix accidental regression of verification keys

Trivo25:
Main fix is in `src/lib/proof-system/feature-flags.ts`, the rest is mostly a quick and dirty example to test it doesn't happen again


src/examples/zkprogram/runtime-table/payroll.ts
```diff
@@ -0,0 +1,65 @@
+/**
```
Trivo25, line 1: maybe this is a little overkill but i didnt want to add something super boring :D

PR #2455, opened 2025-09-30
move changelog script

bleepbloopsify:
move scripts out-of-sight.

also flag off a piece of the script that probably shouldn't be run automatically every time


scripts/changelog/update-changelog.sh
```diff
@@ -37,11 +37,17 @@ current_date=$(date +%Y-%m-%d)
...
 # Step 6: Update the CHANGELOG
 sed -i "s/\[Unreleased\](.*\.\.\.HEAD)/\[Unreleased\](https:\/\/github.com\/o1-labs\/o1js\/compare\/$current_commit...HEAD)\n\n## \[$new_version\](https:\/\/github.com\/o1-labs\/o1js\/compare\/$previous_commit...$current_commit) - $current_date/" CHANGELOG.md
 
-# Step 7: Auto-fix compare URLs for all past versions
-node fix-changelog-links.cjs
+# Step 7: Auto-fix compare URLs for all past versions, if -r is set
+while getopts "r" opt; do
+  case $opt in
+    r)
+      node scripts/changelog/fix-changelog-links.cjs
```
querolita, line 50: It makes sense to run it only if it the optional is enabled

querolita (commented):
Nice

PR #2460, opened 2025-10-01
Store all cache artefacts 

Trivo25:
This fixes an issue where we didn't store all the neccesary cache artefacts and had to re-compute some even after loading the cache. The performance impact was neglecteble but the core issue was that this triggered the re-generation of this artefact after loading the cache, causing our worker logic to error with `caml_fp_srs_get_lagrange_basis is disabled on the web` and ideling indefinitely when browser caches had `canWrite` enable (note we don't yet support writeable browser caches anyway, but this was still a common problem devs encountered and it was relatively easy to fix).

Certain lagrange basis, even tho stored on the srs, weren't correctly written to the cache (or re-read from the restored srs(?)). I fixed it by adding a lazy cache read function, when the lagrange basis already exists, stores it if it didn't exist in the cache before.


Shigoto-dev19 (approved):

querolita (approved):
Makes sense to me

PR #2465, opened 2025-10-03
Format transaction layout definitions to clear the diff

querolita:
This PR adds one command run to the artifacts build of o1js to run prettier after the transaction layout definitions have been generated. The goal is to remove them from source control when simple white-space changes occur, so we don't have to take care of them manually before committing.


querolita:
!ci-bypass-changelog

Shigoto-dev19 (approved):
Nice catch @querolita and thanks for addressing it in this PR :clap:

PR #2469, opened 2025-10-03
chore: formatting zkprogram.ts, and organizing imports

bleepbloopsify:
I wanted to change `zkprogram.ts`, but my formatting really destroys the entire file (as you can probably see here)

We should do a pass across the entire repo at some point, but right now I want to format this file.

Also, I added `organize-imports` for another level of standardization (and also, my local config kinda explodes the imports when I "format" the file, so we should probably have this anyways).

Remember to set your default formatter in `vscode` to `prettier`, otherwise you get a lot of other errors :yikes:

Our CI only checks changed files afaict, so this will be a slow migration across the board for our files, but preferable to a large format blowout

### DIFF
--- a/.prettierrc.cjs
+++ b/.prettierrc.cjs
@@ -13,4 +13,5 @@ module.exports = {
       },
     },
   ],
+  plugins: [require.resolve('prettier-plugin-organize-imports')],
 };
--- a/npmDepsHash
+++ b/npmDepsHash
@@ -1,1 +1,1 @@
-sha256-hvRlb/ZgfEnSW+OJG4UibamdL5UzygQQthMpUqh2W0w=
+sha256-ww0EdkEWiciR6XLTu2/lfqtDMbvIDBLj+gyPH+lpLTE=
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,6 +41,7 @@
         "oxlint": "^1.0.0",
         "pkg-pr-new": "^0.0.9",
         "prettier": "^3.6.2",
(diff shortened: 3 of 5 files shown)


Trivo25:
We actually did this a while ago in a few different PRs and it's checked in CI - what changed? https://github.com/o1-labs/o1js/pull/2283

bleepbloopsify:
> We actually did this a while ago in a few different PRs and it's checked in CI - what changed? #2283 other than organised imports

I'm not sure?

This is the output of `npm run format ./src/lib/proof-system/zkprogram.ts` (after I stick the plugin in)

Perhaps we're not running the proper format checks in CI?

bleepbloopsify:
https://github.com/o1-labs/o1js/commit/e571099a87bbbd515832490aefa9348801d501bb#diff-2cd77c3d508cf1d738808d6a928504243b94d12537aa7b94738f5833a35d8c57

Looks like @Shigoto-dev19 isn't running `prettier` properly? (but the lint check passes?)

image

hmmm, looks like the lint check isn't working

added a ticket to fix: #2470

bleepbloopsify:
fixed the lint checks in #2471

Trivo25:
> Reproduced again by running:
> 
> ```
> git checkout main ./src/lib/proof-system/zkprogram.ts
> npm run format ./src/lib/proof-system/zkprogram.ts
> ```
> 
> on this branch

I can reproduce the imports being linted, but not the rest of the file

querolita (approved):

PR #2471, opened 2025-10-03
fix: pull request lint check

bleepbloopsify:
Moved this to `pull_requests` so we can use the proper branch diffing (pull request CI is also only triggered against `main`, so maybe not too much of a functional difference, but using the proper SHA here will help with the other error


bleepbloopsify:
image

it works!

bleepbloopsify:
codespell is working again as well
image

0x471 (approved):

PR #2548, opened 2025-10-14
Add performance regression tests to CI

Shigoto-dev19:
Closes https://github.com/o1-labs/o1js/issues/2545.
Depends on https://github.com/o1-labs/o1js/pull/2586.

### How it works

The performance regression workflow in CI ensures that compile, prove, and verify times remain consistent across commits. It is driven by two workflows with a single source-of-truth baseline stored in the repo at `tests/perf-regression/perf-regression.json`.

- **PR checks (`checks.yml`)**: the matrix includes **“Performance Regression”**. The shared script `run-ci-tests.sh` runs in **check mode** (defaults to `PERF_MODE=--check`) and **reads the baseline file directly from the repository**. It then runs the benchmarks and compares current results to the stored values to detect regressions. No artifacts are downloaded, and no labels are consulted.

- **Baseline updates (`dump-perf-baseline.yml`)**: this is a **separate, manual workflow** (`workflow_dispatch`) that runs on **the same branch**. It executes the same performance suite in **dump mode** (`PERF_MODE=--dump`), **regenerates `perf-regression.json` using GitHub runners**, and **commits the updated file to the same branch. This becomes the canonical baseline for subsequent PR checks.

---

### How to use

**1) Open a PR (check mode)**
- The **Checks** workflow runs the **Performance Regression** matrix entry.
- `run-ci-tests.sh` executes in `--check`, reading `tests/perf-regression/perf-regression.json` directly from the repo.
- **Outcome:** The PR is validated against the committed baseline; no file changes occur.

**2) Update the baseline (dump mode)**
- Trigger the manual workflow **“Dump Performance Regression Baseline”** (or run the script  `npm run  regression:ci:dump-perf`).
- CI runs the performance suite in `--dump` on stable GitHub runners and **commits the refreshed `perf-regression.json` to the same branch.
- **Outcome:** The baseline in the same branch is updated and used by all future PR checks.

---

### Notes

- The committed `tests/perf-regression/perf-regression.json` is the **canonical baseline**; PR checks read it directly.
- There are **no labels** and **no artifacts** in this flow.
- The **first baseline** must be seeded once (either via the dedicated dump workflow after merge, or via a one-off seed commit); See https://github.com/o1-labs/o1js/pull/2586.  
- Baselines should **not be committed manually** in normal operation; always update via the dump workflow to ensure consistent results from GitHub runners.

### DIFF
--- a/scripts/build/ts/build-node.sh
+++ b/scripts/build/ts/build-node.sh
@@ -1,28 +1,24 @@
 #!/usr/bin/env bash
-
-# - If bindings are missing, try `npm run check:bindings` (downloads CI artifacts).
-# - If that fails, fall back to `npm run build:bindings-node` (local build).
-# - Subsequent steps (copy/clean/builds) are critical: failures stop the script with a clear message.
-
 set -Eeuo pipefail
 
-# ---------- paths ----------
-ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
-BINDINGS_CHECK_PATH="${BINDINGS_CHECK_PATH:-$ROOT_DIR/src/bindings/compiled}"  # presence gate
-
-# ---------- shared libraries ----------
+# shared libraries
 source ./scripts/lib/ux.sh
+
+# paths
+ROOT_DIR="$(get_repo_root)"
+BINDINGS_CHECK_PATH="${BINDINGS_CHECK_PATH:-$ROOT_DIR/src/bindings/compiled}"
+
+# setup
 setup_script "${BASH_SOURCE[0]}" "Build"
 
-# ---------- steps ----------
-ensure_bindings() {
-  bold "[1/5] Ensuring compiled bindings"
+# steps
+bold "Starting build"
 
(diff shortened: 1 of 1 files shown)


tests/perf-regression/dump-perf-ci.sh
bleepbloopsify: This file is interesting, but is there a reason we need it?

My understanding of GH actions is like this:

We trigger the action -> it calls the bash script

not:
we run the bash code -> it calls the gh workflow

CI is meant to make sure that the scripts we can run locally stays running, hence "Continuous" of "Continuous Integration"

request: delete this file, since all it does it call `gh workflow run` (with preset flags, but `man gh` should take care of that if necessary).

if it's meant to be manual, just direct people to the web interface and have them call the action manually (since workflow_dispatch is available as a trigger)

bleepbloopsify (approved):
I like this much better! "It is CURRENT_REF" type 💩 

I dislike `gh` as a dependency, but since it's already here I guess we can let it hang out a bit longer.

Small nit about a script that only calls one command, at that point do we really need a script?

otherwise, 🚢 it

Trivo25 (approved):

PR #2550, opened 2025-10-14
feat: added test to check RSA usage of cache

bleepbloopsify:
formatted the old cache regression tests

added a new test: cache/rsa-regression that checks the RSA example so it works with a cache.

updated the cache regression artifact pinned to include RSA examples.

this is the root of a trunk.

children:
#2551

### DIFF
--- a/src/tests/cache/rsa-regression.ts
+++ b/src/tests/cache/rsa-regression.ts
@@ -1,0 +1,24 @@
+import minimist from 'minimist';
+import assert from 'node:assert';
+import { rsaZkProgram } from '../../examples/crypto/rsa/program.js';
+import { Bigint2048 } from '../../examples/crypto/rsa/rsa.js';
+import { generateRsaParams, rsaSign, sha256Bigint } from '../../examples/crypto/rsa/utils.js';
+import { CacheHarness } from './harness.js';
+
+const { mode, tarball } = minimist(process.argv.slice(2));
+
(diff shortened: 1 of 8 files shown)


src/tests/cache/rsa-regression.ts
```diff
@@ -0,0 +1,26 @@
...
+import { rsaZkProgram } from '../../examples/crypto/rsa/program.js';
+import { Bigint2048 } from '../../examples/crypto/rsa/rsa.js';
+import { generateRsaParams, rsaSign, sha256Bigint } from '../../examples/crypto/rsa/utils.js';
+import { CacheHarness } from './harness.js';
+
+const { mode, tarball } = minimist(process.argv.slice(2));
+
+const harness = await CacheHarness({ mode, tarball });
+
+const Program = rsaZkProgram;
```
Trivo25: necessary? `import { rsaZkProgram as Program } from` also works

src/tests/cache/rsa-regression.ts
```diff
@@ -0,0 +1,26 @@
...
+
+const { verificationKey: vk } = await Program.compile({ cache: harness.cache });
+harness.check(vk, 'vk');
+
+const input = await sha256Bigint('How are you!');
```
Trivo25: i am good x)

Trivo25 (approved):

PR #2553, opened 2025-10-14
feat: small and big zk program for cache regression

bleepbloopsify:
root: #2550 
parent: #2552 

uses the small and big zkprogram for the cache regression tests


src/examples/zkprogram/program-small-big.ts
Shigoto-dev19: When you import `SmallProgram` and `BigProgram`, I assumed and it's confirmed after checking that compile, prove, and verify steps are run twice. 

It might be better to have `program-small-big.ts` to include the ZkPrograms and a separate run.ts file for CI and/or performance regression checks.

PR #2566, opened 2025-10-16
chore: cspell dictionary and spelling mistakes

bleepbloopsify:
since the introduction of `cspell`, I have noticed that the linter seems to be quite loud on PRs.
this is "not good", and does not bode well for the longevity of using `cspell` or typochecks in general

I believe that this is due to the late introduction of a spell-checker, so I did the work to add the list of strange words we use to the `cspell.json` that I configured. This way only "new" words will be flagged, rather than words that already existed in our codebase.

This also comes with a slew (like 10) of spelling fixes, (and one in an actual string!)

hopefully the scope of this change is small enough, I can break it up more if necessary.


cspell.log
Shigoto-dev19: What does this file do?

Shigoto-dev19 (commented):
Many good spelling fixes :clap: 

I am just wondering about words like `Zabcdefghijkmnopqrstuvwxyz` :thinking:

cspell.log
bleepbloopsify: ah, this file is what I was using to log errors from the spell checker 🤦

bleepbloopsify:
> I am just wondering about words like Zabcdefghijkmnopqrstuvwxyz 🤔

Many of these dictionary "words" are fixtures that we specify in code. they have the same "look" and "feel" of words, so we could either write a regex to catch it (now we have a regex problem), or just add it to the dictionary

I have opted to add it to the dictionary here

Shigoto-dev19 (approved):

PR #2599, opened 2025-10-28
Only write during `dump` mode

Trivo25:
We only want to enable a writeable cache when we want to dump new artefacts, when we want to check against the fixtures we want to force the cache to read the artefacts instead of re-generating faulty ones

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,11 @@ This project adheres to
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/3453d1e53...HEAD)
 
+### Internal
+
+- Change cache harness to only allow writes when `dump` mode is active.
+  https://github.com/o1-labs/o1js/pull/2599
+
 ## [2.10.0](https://github.com/o1-labs/o1js/compare/114acff...3453d1e53) - 2025-09-27
 
 ### Internal


src/tests/cache/harness.ts
```diff
@@ -5,81 +5,94 @@ import { Cache, ProofBase, VerificationKey, verify as verifyProof } from 'o1js';
...
-  if (!isAllowedMode(mode)) { throw new Error(`mode should be one of ${allowedModes.join(' | ')}`) }
-
+export async function CacheHarness({
+  mode,
+  tarball,
+  debug = true,
+}: {
+  mode: string;
+  tarball: string;
+  debug?: boolean;
```
Trivo25: this is new

src/tests/cache/harness.ts
```diff
@@ -5,81 +5,94 @@ import { Cache, ProofBase, VerificationKey, verify as verifyProof } from 'o1js';
...
     } else {
       return tmpdir();
     }
-  })()
+  })();
 
-  const cacheDir = path.join(workingDir, "cache")
-  const cache = Cache.FileSystem(cacheDir)
+  const cacheDir = path.join(workingDir, 'cache');
+  const cache = { ...Cache.FileSystem(cacheDir, debug), canWrite: mode === 'dump' };
```
Trivo25: only write when we want to dump, don't allow writing when we only want to check

Shigoto-dev19: I am wondering if we should make `canWrite` stricter as it will only log(not throw) `Failed to write cache` in case of failure.

PR #2609, opened 2025-10-30
(native prover) verifier index

Trivo25:
https://github.com/o1-labs/o1js/pull/2609
https://github.com/o1-labs/proof-systems/pull/3360
https://github.com/MinaProtocol/mina/pull/18038


querolita (approved):
I think this still shows a wrong commit history due to the force pushing some weeks ago, we might have to solve conflicts on the build side when we merge. Maybe we should rebase before doing it to reduce the number of commits in the diff.

PR #2624, opened 2025-11-06
Bump proof-systems with cherry-picked `dd2f6ba`

Trivo25:


src/lib/proof-system/cache.ts
```diff
@@ -50,6 +50,10 @@ type Cache = {
    * but could just be a cache miss, or file system permissions incompatible with writing data.
    */
   debug?: boolean;
+  /**
+   * The directory where cache files are stored. Only applicable for file system caches.
+   */
+  cacheDirectory?: string;
```
Trivo25: I added this since I found it handy to have a way to retrieve the cache directory from the cach object

PR #2644, opened 2025-11-17
3.0 Trunk

bleepbloopsify:
Added features:
1. Bumps mina and adds macOS support
2. Drops nodejs 20 in favor of 22, 24
3. moving mina to lyh/mesa-testing to support hard fork
4. verification key JSON
5. tweaks dune to match mina's 3.1 version

Compatibility log:
1. Changed from time approximation of circuit to a transaction update cap
2. Updated events and actions limits to match `mesa`


bleepbloopsify:
update:
https://github.com/o1-labs/o1js/actions/runs/20108799382/job/57703022435?pr=2644

This failing test "compatible" I believe is failing because we are pulling the `mina#compatible` image, rather than the `mina#mesa` image. It should be ephemeral, so we can safely ignore.

bleepbloopsify:
> @bleepbloopsify what do you mean ephemeral? That as soon as `mina#mesa` is merged into `mina#compatible` then that CI check won't fail?

The CI check makes sure that if we're merging to main, we're compatible with the mina `compatible` docker image. That image is pinned to before the hard-fork, so we don't actually work against it. The test is failing properly at the moment.

once mina's compatible image becomes "up-to-date", the test will pass happily

PR #2689, opened 2025-12-10
enhancement: actions and events tests

### LINKED ISSUE #2670: Write unit tests for increased events/actions
Test description for the 1024 field (flat-length) limit for events and actions
note: we already have tests for a lot of this behaviour, we simply need to adjust the contants (which is arleady done by the time we implement all MIPs in o1js, but we should consider some additional edge cases if the current tests dont do that properly) 

goal: 
validate the 1024 field limit applies to the total flat length of events and actions

…and that it’s enforced correctly for:
- single AccountUpdate
- single Transaction across multiple AccountUpdates

> shape matters: items can be 1 field each or up to 1024 fields per event/action

some potential cases:

**PASS cases**
- proving succeeds, transaction applies.
- recorded events/actions match intended count + shape.
- (for reducer cases) state update proves reduction processed the intended # of actions

**FAIL cases**
- transaction is rejected deterministically 
- no state changes

Events: 

- Single AccountUpdate (one zkApp call producing one AccountUpdate)
- pass: `1024 events × 1 field` => total `1024`
- pass: `1 event × 1024 fields` => total `1024`
- fail: `1025 events × 1 field` => total `1025`
- fail: `1 event × 1024 fields` + `1 event × 1 field` => total `1025`

- single transaction, multiple AccountUpdates 
- pass: AU1 `512 events × 1 field` + AU2 `512 × 1` => total `1024`
- pass: AU1 `1 event × 512 fields` + AU2 `1 event × 512 fields` => total `1024`
- fail: AU1 `513 × 1` + AU2 `512 × 1` => total `1025`
- fail: AU1 `1×512` + AU2 `1×513` => total `1025`

---

actions
- Single AccountUpdate (dispatch + reduce in the same zkApp call / AccountUpdate)
- pass: dispatch `1024 actions × 1 field`; reduce `1024` pending actions
- pass dispatch `1 action × 1024 fields`; reduce `1024` pending actions
- fail: dispatch `1025 × 1`; (reduce not executed because tx fails)
- fail: dispatch `1×1024` + `1×1` => `1025`; (reduce not executed)

- Single transaction, multiple AccountUpdates (dispatch only, split across updates)
- pass: AU1 dispatch `512×1` + AU2 dispatch `512×1` => total `1024`
- pass: AU1 dispatch `1×512` + AU2 dispatch `1×512` => total `1024`
- fail: AU1 `513×1` + AU2 `512×1` => total `1025`
- fail: AU1 `1×512` + AU2 `1×513` => total `1025`

bleepbloopsify:
added tests to ensure we support this exact number of actions and events.

throwing tests exist to ensure the check is actually being checked

closes #2670


bleepbloopsify:
looking through mina's logic, it looks like action limits and events limits are separate, so I've also added tests that make sure they can co-exist in the same transaction happily (1024 events, 1024 actions in the same update)

PR #2693, opened 2025-12-12
enhancement: account state unit tests

### LINKED ISSUE #2669: Write unit tests for increased on chain state
we already have coverage for most cases, this issue is to ensure the boundaries are explicitly tested
 
- pass: tests that allocate < 32 fields 
- pass: tests that allocate 32 fields
- fail: tests that allocate > 32 fields

keep in mind that we already have a bunch of account update unit tests - we shouldn't duplicate the code but check what we have and then make sure we check these new edge cases properly

bleepbloopsify:
testing to see that we do support 32 account state fields, and reject anything above

closes #2669

### DIFF
--- a/src/lib/mina/v1/account-update-state.unit-test.ts
+++ b/src/lib/mina/v1/account-update-state.unit-test.ts
@@ -1,9 +1,19 @@
 import assert from 'node:assert';
 import { describe, it } from 'node:test';
-import { AccountUpdate, Field, Mina, SmartContract, declareState, method } from 'o1js';
+import {
+  AccountUpdate,
+  Field,
+  Mina,
+  PublicKey,
+  SmartContract,
+  State,
+  declareState,
+  method,
+} from 'o1js';
 
 type TestSpec = {
   nAppStateFields: number;
+  expectsToReject?: boolean;
 };
 
 function specToString(t: TestSpec) {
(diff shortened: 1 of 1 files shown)


bleepbloopsify:
```
o1js-small [develop-3.0] ♞  ./run src/lib/mina/v1/account-update-state.unit-test.ts
running /Users/leon/github/o1js-small/dist/node/lib/mina/v1/account-update-state.unit-test.js
▶ app state updates
  ✔ should succeed with spec: (fields: 1) (750.01975ms)
  ✔ should succeed with spec: (fields: 8) (430.418625ms)
  ✔ should succeed with spec: (fields: 32) (445.382541ms)
  ✔ should reject with spec: (fields: 33) (326.388041ms)
✔ app state updates (1952.968542ms)
ℹ tests 4
ℹ suites 1
ℹ pass 4
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 2491.93075
```
Looks like this

src/lib/mina/v1/account-update-state.unit-test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  return contract;
+}
+
+await describe('app state updates', async () => {
+  let Local = await Mina.LocalBlockchain({ proofsEnabled: true });
+  Mina.setActiveInstance(Local);
+  const [feePayer] = Local.testAccounts;
+
+  const tests: TestSpec[] = [
```
Trivo25: while this checks if we can declare `n` state fields, it doesn't check that we actually properly allocate them becuse they are 0 by default anyway. we can do this in a different test but we should make sure that when we declare `n` state fields, we should also try to modify them

Shigoto-dev19: I agree! you can also define a field array `Provable.Array(Field, 32)` and play around with the number of fields and state changes etc..

PR #2699, opened 2025-12-17
Correct `AccountUpdate` limit

Trivo25:
my bad - i thought it was 16 _in total_ but my memory scammed me

https://o1-labs.slack.com/archives/C08L3RGUCBY/p1765917354348219?thread_ts=1765916102.077259&cid=C08L3RGUCBY


bleepbloopsify (approved):
Ah I see, this makes more sense

Issue #2700, opened 2025-12-17
Field index support for runtime tables


querolita:
**Describe the feature or improvement**

Being able to lookup runtime table entries using field elements (aka variables) as indices.

**(Optional) Describe the solution**

Some preliminar checks have been performed and it's possible to change the API to accept `Field` for the `RuntimeTable` class, while keeping constant values at compile time for the table configuration.

Feature requested by @0x471

PR #2706, opened 2026-01-08
Add support for `Field` index lookups in runtime tables

querolita:
This simple PR modifies the API of `RuntimeTable` to support lookup indices that are `Field`, apart from the old `bigint` for backwards compatibility. This allows the index to be a provable type and link it to other parts of the witness. The same functionality could be obtained by replicating the behavior of the class by manually adding a lookup gate with the corresponding arguments, but that approach was hacky and we want our API to be gentler to use.

This was literally a 15min change that can make the difference for UX.

Closes https://github.com/o1-labs/o1js/issues/2700

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ This project adheres to
 
 ### Added
 
+- Added `fetchTimedAccountInfo()` to retrieve detailed balance information for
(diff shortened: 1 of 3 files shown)


Shigoto-dev19 (approved):
Just left a comment! otherwise lgtm :clap:

PR #2721, opened 2026-01-12
Add big state zkapp to ci tests

Trivo25:


Shigoto-dev19 (commented):
I can't see a `big-state-zkapp.ts` in the changes or codebase :thinking: 
nvm, I can see it in `leon/develop-3.0/account-state-unit-tests` branch :sweat_smile: 

After the hardfork is deployed, I think it would be better to increase the states of an existing zkApp example rather than adding a new one :bulb:

Shigoto-dev19 (approved):

Issue #2725, opened 2026-01-13
Fix the APIs in the o1js Mina module to fetch the current slot


Shigoto-dev19:
The o1js API to fetch the last block from the network:

- `Mina.currentSlot()` doesn't work for both MT and Devnet.
- `Mina.getNetworkState().globalSlotSinceGenesis` doesn't work for both MT and Devnet as well!

The API that works for both is `(await fetchLastBlock(MINA_NODE_ENDPOINT)).globalSlotSinceGenesis` where the `fetchLastBlock()` function is imported from `o1js`.

It seems we do have a functional API to fetch the network state including the current slot; however it was annoying not to have what appears to be the straightforward API from an active Mina instance.

If the APIs don't work we should either remove them or fix them by wrapping them around the `fetchLastBlock()` function.

PR #2736, opened 2026-01-15
Fix & Expose fetchCurrentSlot and implement currentSlot for remote blockchains

Shigoto-dev19:
Closes https://github.com/o1-labs/o1js/issues/2725.

### Changes

- Export `fetchCurrentSlot` publicly as the supported way to query the current slot from a remote blockchain.
- Clarify `currentSlot()` behavior
  - Works with `Mina.LocalBlockchain()`, but is not available for remote blockchains.
  - I considered wrapping it internally via `fetchCurrentSlot`, but that would require making `currentSlot()` async (and updating the local implementation too). This might be better suited for the `develop` branch; wdyt @Trivo25?
  - Added an error message that points users to `fetchCurrentSlot` (imported from `o1js`).
- Clarify `getNetworkState()` usage
  - It can be called freely with a local chain, but remote fetching is restricted outside a transaction.
  
- @Trivo25 `fetchCurrentSlot` used to incorrectly fetch the epoch slot instead of the global slot since genesis (which is what most users need). 
- I updated the function to return the global slot by default in addition to an optional `slotType` parameter (`'global'` | `'epoch'`) so users can explicitly choose which slot type to fetch. 
- The documentation is now clarified with examples explaining both slot types, making `fetchCurrentSlot` a correct and lightweight way to fetch either the global slot or the epoch slot as needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ This project adheres to
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/34caaedc9...HEAD)
 
+### Changed
+
+- `Mina.currentSlot()` now works for remote blockchains using cached network
+  state, instead of throwing an error about missing implementation.
+  https://github.com/o1-labs/o1js/pull/2736
+
 ### Added
 
 - Export `fetchCurrentSlot` publicly as the supported way to query the current


Trivo25:
> @Trivo25 I was playing around with the `fetchCurrentSlot` API and noticed it doesn’t seem to match the `globalSlotSinceGenesis` value we get from the last block.
> 
> I did read the code for both:
> 
>     * `fetchLastBlock(MINA_NODE_ENDPOINT)).globalSlotSinceGenesis`
> 
>     * `fetchCurrentSlot(MINA_NODE_ENDPOINT)`
> 
> 
> …but the difference is still not clear to me 🤔 Would appreciate a clarification 🙏
> image

With what network are you testing?

Trivo25:
> Clarify `currentSlot()` behavior
> 
>     * Works with `Mina.LocalBlockchain()`, but is not available for remote blockchains.
> 
>     * I considered wrapping it internally via `fetchCurrentSlot`, but that would require making `currentSlot()` async (and updating the local implementation too). This might be better suited for the `develop` branch; wdyt @Trivo25?
> 
>     * Added an error message that points users to `fetchCurrentSlot` (imported from `o1js`).

Why are you not using the cache?

Trivo25:
> > > Clarify `currentSlot()` behavior
> > > ```
> > > * Works with `Mina.LocalBlockchain()`, but is not available for remote blockchains.
> > > 
> > > * I considered wrapping it internally via `fetchCurrentSlot`, but that would require making `currentSlot()` async (and updating the local implementation too). This might be better suited for the `develop` branch; wdyt @Trivo25?
> > > 
> > > * Added an error message that points users to `fetchCurrentSlot` (imported from `o1js`).
> > > ```
> > 
> > 
> > Why are you not using the cache?
> 
> Cache?

We have a `networkCache` in `fetch.ts` but this stuff

Shigoto-dev19:
> > > > @Trivo25 I was playing around with the `fetchCurrentSlot` API and noticed it doesn’t seem to match the `globalSlotSinceGenesis` value we get from the last block.
> > > > I did read the code for both:
> > > > ```
> > > > * `fetchLastBlock(MINA_NODE_ENDPOINT)).globalSlotSinceGenesis`
> > > > 
> > > > * `fetchCurrentSlot(MINA_NODE_ENDPOINT)`
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > …but the difference is still not clear to me 🤔 Would appreciate a clarification 🙏
> > 
> > 
> > > With what network are you testing?
> > 
> > 
> > Remote blockchain, I values from the screenshot are from MT
> 
> Can you check devnet and mainnet again?

Devnet: 
image

Mainnet:
image

src/lib/mina/v1/mina-instance.ts
Shigoto-dev19: only alphabet-based reordering of imports and exports here

src/lib/mina/v1/fetch.ts
```diff
@@ -546,14 +553,36 @@ async function fetchLastBlock(graphqlEndpoint = networkConfig.minaEndpoint, head
...
+ * // Fetch epoch-relative slot
+ * const epochSlot = await fetchCurrentSlot(
+ *   'https://api.minascan.io/node/devnet/v1/graphql',
+ *   'epoch'
+ * );
+ * ```
  */
 async function fetchCurrentSlot(
   graphqlEndpoint = networkConfig.minaEndpoint,
+  slotType: 'global' | 'epoch' = 'global',
```
Shigoto-dev19: This is not a breaking change as the function was never exported as public API.

src/lib/mina/v1/mina.ts
```diff
@@ -193,7 +193,11 @@ function Network(
       return defaultNetworkConstants;
     },
     currentSlot() {
-      throw Error('currentSlot() is not implemented yet for remote blockchains.');
+      throw Error(
```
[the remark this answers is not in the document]
Shigoto-dev19: I'll give it a try 👀

Issue #2711, opened 2026-01-09
Polishing native prover and land it


querolita:
Similar to above, rust is a mess and we have to invest some time to clean that up and overall cleanly merge the PRs together and resolve mere conflicts

- [x] Remove prints before benchmarks
- [x] Clean up Rust
- [ ] Resolve conflicts and merge PRs

Issue #2710, opened 2026-01-09
Clean up the conversion core


querolita:
Right now JavaScript is a mess, we need to clean that up and have a clear separation of concerns. merge all PRs and resolve merge conflict

- [x] Remove logs before benchmarks
- [x] Implement conversion core (properly implement types)
- [ ] Resolve conflicts and merge PRs

PR #2741, opened 2026-01-23
Clean up native prover

querolita:
**CHANGES**
- plonk_wasm is now kimchi_wasm. This affects bindings files as well. 
- plonk_napi is now kimchi_napi.
- moved napi conversion files onto a `native` folder inside `bindings`
- created `napi-ffi.ts` with common helpers to improve readability of the conversion layer, and add type safety. Examples are `asArrayLike`, `arrayFrom` and constructors for classes expected on the rust side.
- whenever both `wasm` and `napi` types are involved in a file (such as `bindings.ts`), we reimport the types and alias them prefixing with either `wasm` or `napi` to improve understanding
- created `napi-wrappers.ts` as a workaround that redefines the types coming from Napi in a TS-friendly syntax to enable type safety at compile time. That is, this prevents us from using `as any` and `unknown`  in several locations of the conversion code, as opposed to what we had earlier. If there's a cleaner way to achieve this please do share in the comments, I would love to hear.
- moved specific types to `napi-wrappers.ts` to centralize where the conversion types are defined to avoid verbosity in the core/proof/verifier-index/oracles files.
- deleted unnecessary indirections that were used before while debugging other features in the PoC such as `toNodeBuffer`
- deleted try-catch blocks that were added while debugging, also unused `runInTryCatch`
- used `?` syntax more extensively to define or access optional fields
- unified `tsBindings` to contain a single `rustConversion` and `srs` elements that get instantiated differently at runtime using the function `getConversionBundle`
- for convenience, created a backend type to differentiate at runtime whether it is `wasm` or `napi` while keeping type safety at compilation time. This one relies on the global variables defined in `node_backend.js` on the mina repo.
- some functions in `bindings.ts` to facilitate the logic such as `getKimchiBackend`
- created a type `RustConversion` that depends on a generic `BackendKind`. This is used to determine what conversion type must be used
- unified output format of both `buildWasmRustConversion` and `buildNapiRustConversion` functions in case we want to further unify the wasm and napi conversion cores in the future
- The conversion bundle returns all the information that could be relevant for the conversion layer, adapted to a specific backend: that is, the actual backend kind, the whole rust namespace obtained from the bindings (whether wasm-styled or napi-styled at runtime), a specific conversion mechanism, and the specific srs type. 

Partially addresses https://github.com/o1-labs/o1js/issues/2710 and https://github.com/o1-labs/o1js/issues/2711

o1js: https://github.com/o1-labs/o1js/pull/2741
mina: https://github.com/MinaProtocol/mina/pull/18354
proof-systems: https://github.com/o1-labs/proof-systems/pull/3447

### DIFF
--- a/src/bindings/crypto/bindings/conversion-base.ts
+++ b/src/bindings/crypto/bindings/conversion-base.ts
@@ -87,18 +87,14 @@ function affineFromRust<A extends WasmAffine>(pt: A): OrInfinity {
   }
 }
 
-const tmpBytes = new Uint8Array(32);
-
 function affineToRust<A extends WasmAffine>(pt: OrInfinity, makeAffine: () => A) {
   let res = makeAffine();
   if (pt === Infinity) {
     res.infinity = true;
   } else {
     let [, [, x, y]] = pt;
-    // we can use the same bytes here every time,
-    // because x and y setters copy the bytes into wasm memory
-    res.x = fieldToRust(x, tmpBytes);
-    res.y = fieldToRust(y, tmpBytes);
+    res.x = fieldToRust(x);
+    res.y = fieldToRust(y);
   }
   return res;
 }
--- a/src/bindings/crypto/native/napi-conversion-proof.ts
+++ b/src/bindings/crypto/native/napi-conversion-proof.ts
@@ -28,7 +28,7 @@ import type {
 } from '../bindings/kimchi-types.js';
 import { ConversionCore, ConversionCores } from './napi-conversion-core.js';
 import {
-  arrayFrom,
+  asArrayLike,
   castCtor,
   type Ctor,
 } from './napi-ffi.js';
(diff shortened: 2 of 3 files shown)


src/bindings/crypto/napi-conversion-proof.ts
```diff
@@ -50,21 +59,48 @@ const proofEvaluationsFromRust = mapProofEvaluations(fieldFromRust);
...
+};
 
-type NapiProverCommitments = WasmFpProverCommitments | WasmFqProverCommitments;
-type NapiOpeningProof = WasmFpOpeningProof | WasmFqOpeningProof;
-type NapiProverProof = WasmFpProverProof | WasmFqProverProof;
-type NapiLookupCommitments = WasmFpLookupCommitments | WasmFqLookupCommitments;
+type NapiProverProof = {
+  commitments: NapiProverCommitments;
+  proof: NapiOpeningProof;
+  evals: unknown;
```
querolita: pending

Issue #2729, opened 2026-01-14
Determining finality of included L1 transactions


Trivo25:
this could be a easy addition in o1js, adding a API where you can query the “depth” of a transaction by abstracting over the mina node graphql API

Shigoto-dev19:
> this could be a easy addition in o1js, adding a API where you can query the “depth” of a transaction by abstracting over the mina node graphql API

## About the new Transaction Depth API

This PR is intended to implement an API to query the `depth` (confirmation count) of included transactions, abstracting over the Mina node GraphQL API.

### Overview

Transaction depth represents how many blocks have been built on top of the block containing a transaction. Higher depth = more confirmations = more secure.

Per [Mina documentation](https://docs.minaprotocol.com/mina-protocol/lifecycle-of-a-payment):
- 8 blocks = 98.67% confidence (~33 mins)
- **15 blocks = 99.9% confidence (~60 mins)** ← Default threshold
- 23 blocks = 99.9965% confidence (~1.5 hours)
- 30 blocks = 99.9998% confidence (~2 hours)

Here's what a transaction depth value represent in the intended API:
- Just sent: The transaction is in the mempool (pending), waiting to be picked up by a block producer. It hasn't been included in any block yet.

- Depth 0 (just included): The transaction has been included in a block (the latest block), but no subsequent blocks have been built on top of it yet.

- Depth 1: One block has been built on top of the block containing the transaction.

- Depth N: N blocks have been built on top of the block containing the transaction.
---

### New Types

```typescript
type TransactionDepthInfo = {
  depth: number;                 // Confirmations (0 = just included)
  inclusionBlockHeight: number;  // Block where tx was included
  currentBlockHeight: number;    // Current chain tip
  isFinalized: boolean;          // depth >= finalityThreshold
  finalityThreshold: number;     // Threshold used for calculation
};

type DepthOptions = {
  blockLength?: number;          // Blocks to search (default: 20)
  finalityThreshold?: number;    // Finality threshold (default: 15)
};
```

### New APIs

#### Methods on IncludedTransaction

```ts
// After waiting for inclusion
const included = await pendingTx.wait();

// Get depth (throws if not found in recent blocks)
const depthInfo = await included.getDepth();

// Safe variant (returns null instead of throwing)
const depthInfo = await included.safeGetDepth();

// With custom finality threshold
const depthInfo = await included.getDepth({ finalityThreshold: 10 });
```

#### Standalone Function

```ts
[...]

PR #2749, opened 2026-01-29
Add transaction depth APIs for tracking L1 inclusion finality

Shigoto-dev19:
Closes https://github.com/o1-labs/o1js/issues/2729.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,12 @@ This project adheres to
 
 ## [Unreleased](https://github.com/o1-labs/o1js/compare/34caaedc9...HEAD)
 
+### Changed
+
+- `Mina.currentSlot()` now works for remote blockchains using cached network
+  state, instead of throwing an error about missing implementation.
+  https://github.com/o1-labs/o1js/pull/2736
+
 ### Added
 
 - Added transaction depth and finality API to determine canonical finality of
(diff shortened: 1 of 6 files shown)


src/lib/mina/v1/transaction.ts
```diff
@@ -656,19 +757,59 @@ function createRejectedTransaction(
...
+    }
+    return result;
+  };
+
+  const waitForFinality = async (
+    options?: WaitForFinalityOptions
+  ): Promise<TransactionDepthInfo> => {
+    const finalityThreshold = options?.finalityThreshold ?? 15;
+    const interval = options?.interval ?? 60000;
+    const maxAttempts = options?.maxAttempts ?? 30;
```
Trivo25: with the interval being 60s, max attempts 30 and the expected finality threshold 15 we should expect a transaction to be final after at least 60min, likely even more - right now this stops any attempts after 30 * 60s = 30mins, so we already know this will never succeed because the math doesn't work out. we should at least set up the default parameters within a range where we expect the transaction to be final

src/lib/mina/v1/local-blockchain.ts
```diff
@@ -278,14 +279,19 @@ async function LocalBlockchain({ proofsEnabled = true, enforceTransactionLimits
           return pendingTransaction;
         };
 
+        // Track the block height at which this transaction was included
+        const inclusionBlockHeight = Number(networkState.blockchainLength.toBigint());
```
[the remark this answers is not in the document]
Shigoto-dev19: That true! But I think it would help testing the advancements of slots with local blockchain as described in the issue about `LocalBlockchain Support`.

```ts
const Local = await Mina.LocalBlockchain({ proofsEnabled: false });

// ... send transaction ...
const included = await pending.wait();

// Initially depth is 0
let info = await included.getDepth({ finalityThreshold: 3 });
console.log(info.depth); // 0

// Simulate blocks being added
Local.setBlockchainLength(Local.getNetworkState().blockchainLength.add(3));

// Now depth is 3
info = await included.getDepth({ finalityThreshold: 3 });
console.log(info.depth); // 3
console.log(info.isFinalized); // true
```

src/lib/mina/v1/transaction.ts
Shigoto-dev19: I see! I chose those default params to await for the highest probability for a canonical finality; also taking into consideration the network instability. Btw, even these params, with lower threshold and higher interval & attempts failed when testing Devnet.

What do you recommend setting as default params here?

PR #2751, opened 2026-02-02
Update examples to reference Mesa instead of Berkeley

Shigoto-dev19:
Closes https://github.com/o1-labs/o1js/issues/2673.

### Changes

- Update Mina node, archive, and explorer endpoints to Mesa
  - GraphQL endpoint: `https://plain-1-graphql.mina-mesa-network.gcp.o1test.net/graphql`
  - Archive endpoint: `http://mesa-archive-node-api.gcp.o1test.net`
  - Explorer URL: `https://mesa.minaexplorer.com` (replacing both `minascan.io/berkeley` and `berkeley.minaexplorer.com`)
  
- Default faucet network parameter changed from `devnet` to `mesa`
  - **Note**: we might need to change this again to `devnet` assuming that `mesa` would be the default devnet once the hardfork is applied.
  
- Rename example files referencing Berkeley to Mesa
  - `simple-zkapp-berkeley.ts` -> `simple-zkapp-mesa.ts`
  - `voting/run-berkeley.ts` -> `voting/run-mesa.ts `
  - **Note**: the replacement here is straightforward but renaming the examples with `live` or `devnet` suffix might be more meaningful.
   
- Replace `berkeley` hash option with network-agnostic `legacy` in mina-signer
  - `hashPayment` and `hashStakeDelegation` option changed from `{ berkeley?: boolean }` to `{ legacy?: boolean }`(inverted logic)
  - Migration: `{ berkeley: true }` becomes `{ legacy: false }`
  - This is a breaking change, but it will be handled alongside other breaking changes when the hard fork is applied
Chose `legacy` over `mesa` because the flag controls V1 vs current hashing behavior, not anything network-specific; this is consistent with the existing `sign-legacy.ts` naming and avoids needing another rename for the next network.
  
### TODO

- Verify Mesa node and archive endpoints are still up-to-date at merge time (they may change before the hard fork)

### Notes

- Some examples require pre-funded keys to run. I've manually funded the hardcoded keys used in the examples have via the Mina faucet.
  - **Note**: each key was funded by `10` Mina which might be drained in the future and block CI.
  
- `Mina.faucet()` is currently non-functional: the faucet endpoint (faucet.minaprotocol.com) is deployed behind Vercel, which intercepts programmatic fetch requests from Node.js with a `429` status and returns an HTML security checkpoint page instead of JSON. The o1js client code is correct; this is a server-side issue where Vercel's bot protection blocks automated/non-browser traffic before the request ever reaches the faucet API.

- Some examples like the `voting/run-mesa.ts` example generate random keys and rely on `Mina.faucet()` at runtime, so it will not work until the faucet issue above is resolved.

### DIFF
--- a/src/lib/mina/v1/mina.ts
+++ b/src/lib/mina/v1/mina.ts
@@ -463,7 +463,7 @@ async function waitForFunding(address: string, headers?: HeadersInit): Promise<v
 /**
  * Requests the [testnet faucet](https://faucet.minaprotocol.com/api/v1/faucet) to fund a public key.
  */
-async function faucet(pub: PublicKey, network: string = 'mesa', headers?: HeadersInit) {
+async function faucet(pub: PublicKey, network: string = 'devnet', headers?: HeadersInit) {
   let address = pub.toBase58();
   let response = await fetch('https://faucet.minaprotocol.com/api/v1/faucet', {
     method: 'POST',


src/mina-signer/mina-signer.ts
```diff
@@ -332,11 +332,16 @@ class Client {
...
+   *
+   * **Breaking change:** The `berkeley` option has been replaced with `legacy` (inverted logic).
+   * Migrate by replacing `{ berkeley: true }` with `{ legacy: false }`.
+   *
    * @returns A transaction hash
    */
   hashPayment(
     { data, signature }: SignedLegacy<Json.Payment>,
-    options?: { berkeley?: boolean }
+    options?: { legacy?: boolean }
```
Trivo25: its a little bit confusing to swap the logic around and may be more disruptive than needed. why not just call it `latest` instead of `legacy`? that way we don't have any change in behaviour, it's not called berkeley anymore and it's set up properly for future hardforks

src/examples/fetch-live.ts
```diff
@@ -42,12 +42,12 @@ async function checkActualNetworkConstantsFetching() {
 function configureMinaNetwork() {
   const minaGraphQlEndpoint = useCustomLocalNetwork
     ? 'http://localhost:8080/graphql'
-    : 'https://api.minascan.io/node/devnet/v1/graphql';
+    : 'https://plain-1-graphql.mina-mesa-network.gcp.o1test.net/graphql';
```
Trivo25: this is our testnet mesa node endpoint - not a devnet endpoint - we will very likely shut down the testnet (or at least that endpoint) after mesa is complete. the old URL should still be valid since it's for the public devnet (which will fork as well)

src/examples/fetch.ts
```diff
@@ -1,6 +1,6 @@
-import { fetchAccount, setGraphqlEndpoints, fetchLastBlock, PublicKey, Types } from 'o1js';
+import { PublicKey, Types, fetchAccount, fetchLastBlock, setGraphqlEndpoints } from 'o1js';
 
-setGraphqlEndpoints(['https://api.minascan.io/node/devnet/v1/graphql']);
+setGraphqlEndpoints(['https://plain-1-graphql.mina-mesa-network.gcp.o1test.net/graphql']);
```
Trivo25: same as above - we should take note to change (and deploy the example) right after devnet forked

src/mina-signer/mina-signer.ts
Shigoto-dev19: Makes sense 👍

Shigoto-dev19:
> > > Mina.faucet() is currently non-functional: the faucet endpoint (faucet.minaprotocol.com) is deployed behind Vercel, which intercepts programmatic fetch requests from Node.js with a 429 status and returns an HTML security checkpoint page instead of JSON. The o1js client code is correct; this is a server-side issue where Vercel's bot protection blocks automated/non-browser traffic before the request ever reaches the faucet API.
> > 
> > 
> > > Some examples like the voting/run-mesa.ts example generate random keys and rely on Mina.faucet() at runtime, so it will not work until the faucet issue above is resolved.
> > 
> > 
> > It worked in the past! please create an issue for that - this should be fixed
> 
> Oh no, its not Vercel blocking it - we now require a zk captcha as part of the facuet because bots used to spam the faucet

yes yes, it's about the captcha challenge.

Shigoto-dev19:
> It worked in the past! please create an issue for that - this should be fixed

https://github.com/o1-labs/o1js/issues/2756

Trivo25:
We should be able to just merge this into develop-3.0, as that branch will become our new main

PR #2766, opened 2026-02-09
Native prover backend selection

Trivo25:
This PR focuses on mainly two things:
- building & publishing the native prover artefacts via CI for different architectures 
- implement a programmable way to select the backend (wasm or native) using a meta package @o1js/native which automatically selectes the backend artefact that the user needs via npm

https://github.com/o1-labs/o1js/pull/2766
https://github.com/MinaProtocol/mina/pull/18490

### DIFF
--- a/src/lib/backend.ts
+++ b/src/lib/backend.ts
@@ -4,6 +4,15 @@ type Backend = 'wasm' | 'native';
 
 let locked = false;
 
+// eagerly set the backend preference from env var so that
+// node_backend.js (which reads globalThis directly) picks it up
+if ((globalThis as any).__o1js_backend_preference === undefined) {
+  let env = getBackendFromEnv();
+  if (env !== undefined) {
+    (globalThis as any).__o1js_backend_preference = env;
+  }
+}
+
 /**
  * Set the backend to use for cryptographic operations such as proving.
  *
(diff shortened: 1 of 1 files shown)


src/lib/backend.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (backend !== 'wasm' && backend !== 'native') {
+    throw new Error(`Invalid backend '${backend}'. Must be 'wasm' or 'native'.`);
+  }
+  if (backend === 'native' && typeof window !== 'undefined') {
+    console.error(
+      "No-op: `setBackend()`:The native backend is not available in the browser. Only 'wasm' is supported. Falling back to 'wasm' backend."
+    );
+    (globalThis as any).__o1js_backend_preference = 'wasm';
+  }
+  (globalThis as any).__o1js_backend_preference = backend;
```
Shigoto-dev19: Why not call `await initializeBindings();` at the end instead separately as described in L30?

Trivo25: the issue is not calling `initializeBindings` after `setBackend` but calling `initializeBindings` before `setBackend` because that locks in the backend (the backend is a dependency of `initializeBindings`) - this means that we can call `initializeBindings` as many times as we want after we've set our backend, but not before. in fact, lots of places inside of o1js call `initializeBindings` before we start any heavy lifting (because we do need the backend first) - this happens mostly under the hood unless a developer explicitly has to call `initializeBindings` seperatly

Shigoto-dev19: I see, thanks for the clarification 🙏

Issue #2707, opened 2026-01-09
Debug PoC SRS creation with cache


querolita:
TODO:

Run with custom cache and debug feature to check if the current branch is really reading from cache or if it is actually regenerating it for the first proof, and then that's why the first run is slower than the second proof.

PR #2770, opened 2026-02-10
Enable SRS cacheing on native prover

querolita:
This PR fixes the `prover-keys.ts` logic that was incompatible with the native build. 

In particular, serialization and deserialization calls are different for the Napi build, and some bindings overwrites need to take place to mimic the JSOO route. 

o1js: https://github.com/o1-labs/o1js/pull/2770
mina: https://github.com/MinaProtocol/mina/pull/18449
proof-systems: https://github.com/o1-labs/proof-systems/pull/3488

Closes https://github.com/o1-labs/o1js/issues/2707 and consequently all of https://github.com/o1-labs/o1js/issues/2682

### DIFF
--- a/src/lib/proof-system/cache.ts
+++ b/src/lib/proof-system/cache.ts
@@ -125,7 +125,7 @@ function readCache<T>(
   try {
     let result = cache.read(header);
     if (result === undefined) {
-      if (cache.debug) console.trace(`cache miss: ${header.persistentId}`);
+      if (cache.debug) console.log(`cache miss: ${header.persistentId}`);
       return undefined;
     }
     if (transform === undefined) return result as any as T;
@@ -147,6 +147,15 @@ function writeCache(cache: Cache, header: CacheHeader, value: Uint8Array) {
   }
 }
 
+function isErr(error: unknown, code: string) {
+  return (
+    typeof error === 'object' &&
+    error !== null &&
+    'code' in error &&
+    (error as { code?: unknown }).code === code
+  );
+}
+
 const None: Cache = {
   read() {
     throw Error('not available');
@@ -163,16 +172,28 @@ const FileSystem = (cacheDirectory: string, debug?: boolean): Cache => ({
 
     let headerPath = resolve(cacheDirectory, `${persistentId}.header`);
     let dataPath = resolve(cacheDirectory, persistentId);
+
     // read current uniqueId, return data if it matches
-    let currentId = readFileSync(headerPath, 'utf8');
+    let currentId: string;
+    try {
+      currentId = readFileSync(headerPath, 'utf8');
+    } catch (error) {
+      if (isErr(error, 'ENOENT')) return undefined;
+      throw error;
+    }
     if (currentId !== uniqueId) return undefined;
 
-    if (dataType === 'string') {
-      let string = readFileSync(dataPath, 'utf8');
-      return new TextEncoder().encode(string);
-    } else {
-      let buffer = readFileSync(dataPath);
-      return new Uint8Array(buffer.buffer);
+    try {
+      if (dataType === 'string') {
+        let string = readFileSync(dataPath, 'utf8');
+        return new TextEncoder().encode(string);
+      } else {
+        let buffer = readFileSync(dataPath);
+        return new Uint8Array(buffer.buffer);
+      }
+    } catch (error) {
+      if (isErr(error, 'ENOENT')) return undefined;
+      throw error;
     }
   },
   write({ persistentId, uniqueId, dataType }, data) {


src/bindings/crypto/native/napi-srs.ts
```diff
@@ -255,32 +255,11 @@ function srsPerField(f: 'fp' | 'fq', napi: Napi, conversion: RustConversion<'nat
...
-          let bytes = new TextEncoder().encode(JSON.stringify(comms));
-
-          writeCache(cache, header, bytes);
-        }
-      }
+      // Intentionally removed the old "edge-case cache backfill" block here:
+      // when a commitment existed in-memory but not in cache, we used to re-read /
+      // re-serialize and write the full basis JSON on the proving path.
+      // Keep this disabled for native because it adds large fs + JSON overhead
+      // and regresses proving latency, while correctness does not depend on it.
```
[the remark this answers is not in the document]
Trivo25: btw this is what i meant about when i said "rereading the cache during the first proof"

src/lib/proof-system/cache.ts
```diff
@@ -147,6 +147,15 @@ function writeCache(cache: Cache, header: CacheHeader, value: Uint8Array) {
   }
 }
 
+function isErrno(error: unknown, code: string) {
```
Trivo25: Errno?

src/lib/proof-system/prover-keys.ts
```diff
@@ -28,6 +28,15 @@ enum KeyType {
   WrapVerificationKey,
 }
 
+function isWasmBackend(): boolean {
```
Trivo25: this should probably align with my backend selection implementation where we call `getBackendPreference()` see https://github.com/o1-labs/o1js/blob/1d242c3edec2c818c8b7d69eb1ee991ba2c27290/src/lib/backend.ts

src/lib/proof-system/prover-keys.ts
```diff
@@ -109,10 +119,8 @@ function encodeProverKey(value: SnarkKey): Uint8Array {
     }
     case KeyType.WrapProvingKey: {
       let index = value[1][1];
-      let encoded = wasm.caml_pasta_fq_plonk_index_encode(
-        (wasm as any).prover_index_fq_deserialize((wasm as any).prover_index_fq_serialize(index))
-      );
-      return encoded;
+      if (!isWasmBackend()) return (wasm as any).prover_index_fq_serialize(index);
```
Trivo25: can we fix the seriaulize/deserialize logic instead of doing it manually based on the backend? i worry this becomes too messy at some point

Trivo25: maybe this change helps (its in the cleanup branch)

https://github.com/o1-labs/proof-systems/pull/3447/changes/73838782556085507bb2d22ce93aa278092babf9

https://github.com/o1-labs/proof-systems/pull/3447/changes/6854279c31f6239e0df0cefc813b382997a22ff9

Trivo25 (commented):
The only concern I have here is that we ideally shouldn't change the underlying cache logic and make decisions based on the backend. I'd prefer if we can decouple the cache logic from the backend as much as possible - similar to what we did with the rest of the codebase (where we simply plugged-in the other backend)

src/examples/zkprogram/program.ts
```diff
@@ -1,4 +1,6 @@
-import { Field, ZkProgram, Cache, verify } from 'o1js';
+import { Field, ZkProgram, Cache, verify, setBackend } from 'o1js';
+
+setBackend('wasm');
```
Shigoto-dev19: Isn't wasm backend the default?

src/bindings/crypto/native/napi-srs.ts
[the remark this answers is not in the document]
querolita: I will check again now with both wasm and native to check if that overhead still applies. I will clear the caches in both cases to observe any differences. But I agree that wasm and native should better follow the same logic

src/lib/proof-system/cache.ts
```diff
@@ -161,16 +170,31 @@ const FileSystem = (cacheDirectory: string, debug?: boolean): Cache => ({
   read({ persistentId, uniqueId, dataType }) {
     if (jsEnvironment !== 'node') throw Error('file system not available');
 
-    // read current uniqueId, return data if it matches
-    let currentId = readFileSync(resolve(cacheDirectory, `${persistentId}.header`), 'utf8');
+    let headerPath = resolve(cacheDirectory, `${persistentId}.header`);
```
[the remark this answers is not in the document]
querolita: Okay I will revert it. I was hoping to make the logs cleaner, the trace looked too intimidating. But if that's a good point for a follow up PR then I can do that as well and keep it minimal here.

src/examples/zkprogram/program.ts
querolita: Will remove it, yes, it is the default

Issue #2775, opened 2026-02-17
verify that vks are the same for native prover


Trivo25:
- reuse current framework and vks 
- check native prover vks against the wasm vks

PR #2780, opened 2026-02-19
VK regression checks for native backend

querolita:
This PR extends the checks for the VK regression to check that keys generated with a native backend are not modified with respect to the canonical instance. This implicitly checks that both backends produce the same keys.

Just like before, we do:
```
 npm run regression:dump-vks
 npm run regression:check-vks         
```

Closes https://github.com/o1-labs/o1js/issues/2775

### DIFF
--- a/src/bindings/crypto/bindings/srs.ts
+++ b/src/bindings/crypto/bindings/srs.ts
@@ -1,7 +1,7 @@
 import type { Wasm, RustConversion } from '../bindings.js';
 import { type WasmFpSrs, type WasmFqSrs } from '../../compiled/node_bindings/kimchi_wasm.cjs';
 import { PolyComm } from './kimchi-types.js';
-import { srsCache } from '../cache.js';
+import { srsCache as cache } from '../cache.js';
 import {
   type CacheHeader,
   type Cache,
@@ -26,8 +26,6 @@ function empty(): SrsStore {
 const srsStore = { fp: empty(), fq: empty() };
 
 const CacheReadRegister = new Map<string, boolean>();
-
-let cache: Cache | undefined = srsCache;
 
 const srsVersion = 1;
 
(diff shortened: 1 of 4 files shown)


tests/vk-regression/vk-regression.ts
```diff
@@ -16,7 +17,15 @@ const skipVerificationKeys = false;
 // toggle to override caches
 const forceRecompile = false;
 
-// usage ./run ./tests/vk-regression/vk-regression.ts --bundle --dump ./tests/vk-regression/vk-regression.json
+type Backend = 'wasm' | 'native';
+let backend: Backend = process.env.O1JS_REQUIRE_NATIVE_BINDINGS ? 'native' : 'wasm';
```
Trivo25: I already built in an env variable for this case - `O1JS_BACKEND` handles this logic under the hood. You can then fetch and check the backend via `getBackendPreference()`

scripts/tests/check-vks.sh
```diff
@@ -17,8 +17,13 @@ cd "$ROOT_DIR"
...
 info "Running regression checks..."
-VK_TEST=1 run_cmd ./run ./tests/vk-regression/vk-regression.ts --bundle
-VK_TEST=2 run_cmd ./run ./tests/vk-regression/vk-regression.ts --bundle
+info "Backend: wasm"
+run_cmd env -u O1JS_REQUIRE_NATIVE_BINDINGS VK_TEST=1 ./run ./tests/vk-regression/vk-regression.ts --bundle
+run_cmd env -u O1JS_REQUIRE_NATIVE_BINDINGS VK_TEST=2 ./run ./tests/vk-regression/vk-regression.ts --bundle
+
+info "Backend: native"
+run_cmd env O1JS_REQUIRE_NATIVE_BINDINGS=1 VK_TEST=1 ./run ./tests/vk-regression/vk-regression.ts --bundle
+run_cmd env O1JS_REQUIRE_NATIVE_BINDINGS=1 VK_TEST=2 ./run ./tests/vk-regression/vk-regression.ts --bundle
```
Shigoto-dev19: Nit: It might be better to use the env variable for the backend to run the file separately and not run it for both backends all the time. It would be easier to use locally too

Issue #2782, opened 2026-02-23
Fix `RuntimeTable` for native Kimchi


querolita:
Raised by @Shigoto-dev19 while running tests on `native/cache`

<img width="1452" height="286" alt="Image" src="https://github.com/user-attachments/assets/db06bd36-3e9a-42a5-875f-cc81fc9fac79" />

PR #2786, opened 2026-02-23
Fix `RuntimeTable` native conversion 

querolita:
This creates a wrapper as we previously did with other napi types to be able to create objects from o1js.

Closes https://github.com/o1-labs/o1js/issues/2782


src/bindings/crypto/native/napi-conversion-proof.ts
```diff
@@ -189,7 +186,7 @@ function proofConversionPerField(
   }
 
   function runtimeTableToRust([, id, data]: RuntimeTable): NapiRuntimeTable {
-    return new RuntimeTable(id, core.vectorToRust(data));
```
Shigoto-dev19: Here seems to be the fix for `RuntimeTable is not a constructor`

PR #2800, opened 2026-03-11
only assertAlmostReduced on scalar s in `EddsaSignature.check`

Trivo25:
`check()` was calling `Scalar.assertAlmostReduced(signature.R, signature.s)` which asserts both `r` and `s` against the scalar field modulus. However, `r` is an encoded curve point and not a scalar, its top limb can reach 2^80 due to the sign bit, while the scalar modulus top limb is only ~2^76. this caused `weakBound` inside `assertAlmostReduced` to overflow 88 bits, failing the RangeCheck0 constraint. We now only assert `s` against the scalar modulus. `r` is already covered by the multiRangeCheck on the line above


querolita (approved):
Makes sense that the vk regression checks fail after this

PR #2822, opened 2026-03-24
Add V1 AGENT.md and AGENT_LOG.md

Trivo25:


querolita:
@0x471 Could be a good idea, but because different teams use different agents I believe it's a good approach to centralize the context. Then, as we progress and keep adding knowledge to that file, we can consider if it makes sense to make a `CLAUDE.md` one.

PR #2860, opened 2026-04-27
new mina commit (temp)

Trivo25:


querolita:
I assume this is to test the dry run?

querolita (approved):

Trivo25:
> I assume this is to test the dry run?

yes i can actually close it, i only needed it to trigger the native builds

PR #2866, opened 2026-05-04
fix(srs): wrap getLagrangeBasis in try-catch for web worker compatibility

### LINKED ISSUE #2865: getLagrangeBasis crashes in Web Workers when writable Cache is used
- `caml_fp_srs_get_lagrange_basis` is disabled in the web worker spec (returns `WasmVector` which can't pass through the `SharedArrayBuffer` channel)
- When a writable `Cache` is used in the browser worker context, the call crashes
- Fix: wrap the two call sites in try-catch with fallback to `lagrangeCommitment()` for in-memory computation

mellowcroc:
### Fixed

- Fixed `getLagrangeBasis` crash in web workers when using a writable `Cache`. The function is disabled in the web worker spec (returns `WasmVector` which can't pass through SharedArrayBuffer), so the two call sites now catch the error and fall back to `lagrangeCommitment()` for in-memory computation. Node.js callers are unaffected. Fixes #2865.

### DIFF
--- a/src/bindings/crypto/bindings/srs.ts
+++ b/src/bindings/crypto/bindings/srs.ts
@@ -164,13 +164,18 @@ function srsPerField(f: 'fp' | 'fq', wasm: Wasm, conversion: RustConversion<'was
           if (didRead !== true) {
             // not in cache
             if (cache.canWrite) {
-              // TODO: this code path will throw on the web since `caml_${f}_srs_get_lagrange_basis` is not properly implemented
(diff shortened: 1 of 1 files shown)


mitschabaude:
@Trivo25

Trivo25:
Thanks 🙏 Do you mind adding a quick changelog entry? @mellowcroc 
Also, this is targeting develop-3.0 aka the Mesa hardfork - do you need this as part of the current mainnet or the upcoming HF?

mellowcroc:
We need it for the upcoming hardfork. :)

PR #2869, opened 2026-05-05
Adding zkAppCommandCommitment getters to allow for external signing

### LINKED ISSUE #2868: zkAppCommand commitments are not exported
There is no public function in `Client` to get the zkAppCommand commitment. Currently, these getters are internal functions used in the signing procedure. To allow for external signing (e.g. Ledger), there should be a way to get the commitment only.

Fix: add relevant getters to the exported `Client`

graikos:
Fixes #2868

- Commitments for zkAppCommand are needed for external signing (e.g. Ledger), and currently they are internal to the signing functions
- Fix: Exposes two methods in `Client` to get the zkAppCommand commitments. One for "building" flow, with min. fee checking and validation, and one `FromJSON` to be used with an already built TX (no checks).


Trivo25:
Thank you 🙏 @graikos this is targeting develop-3.0 - do you need this as part of the current mainnet compatible mina-signer or the upcoming HF?

graikos:
We ran into it while developing https://github.com/zksecurity/mina-guard, the target is indeed Mesa, not current, hence the develop-3.0 target. Similarly for #2866.

PR #2885, opened 2026-06-10
CI job for cross backend proving and verifying for chunked proofs

querolita:
This checks that we can generate chunked proofs with native backend and verify them on wasm backend.


querolita:
Closing in favor of https://github.com/o1-labs/o1js/pull/2889

PR #2886, opened 2026-06-10
Standalone verification of chunked proofs

querolita:
Adds standalone verification support for chunked proofs by passing the circuit chunk count from o1js into the Pickles verifier.  This allows verifying a chunked proof together with verification keys without needing to compile the ZkProgram again.

mina side:


querolita:
Closing in favor of https://github.com/o1-labs/o1js/pull/2890

The current PR is merging against a parent branch that is proposing a helper function for serde without requiring a HF. But we are moving away from that approach for now in favor of an actual fix for the next HF.

PR #2891, opened 2026-06-18
Reject invalid Base64 padding in Bytes decoding

haoshengzhen:
## Summary

This PR makes `Bytes.base64Decode()` reject Base64 inputs that contain `=` padding before the end of the input.

Previously, `base64DecodeLookup()` accepted `=` as a valid character anywhere in the input, while the `base64Decode()` docs warned that padding in the middle could cause unexpected decoding results. The decoder now enforces that padding can only appear in the final two positions and that padding is contiguous once it starts.

## Changes

- Add Base64 padding validation before decoding.
- Add a unit test for padding before the end of the input.
- Update the changelog under `Unreleased / Fixed`.

## Verification

- `npm run format:check -- CHANGELOG.md src/lib/provable/bytes.ts src/lib/provable/test/base64.unit-test.ts`
- `npx prettier --config .prettierrc.md.cjs --check CHANGELOG.md`
- `git diff --check`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ This project adheres to
 - Removed `@internal` JSDoc tags from public APIs so they appear in generated
   documentation. https://github.com/o1-labs/o1js/pull/2881
 
+### Fixed
+
+- Reject Base64 inputs with padding characters before the end of the string.
+
 ### Removed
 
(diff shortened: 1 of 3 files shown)


haoshengzhen:
@martyall @Trivo25 Hi, Could you please review this PR at your convenience? Thank you very much.

Issue #2870, opened 2026-05-05
Second prove() hangs indefinitely in web worker when FinalizationRegistry is active


mellowcroc:
## Summary

When running `SmartContract.compile()` → `prove()` → `prove()` in a web worker with `FinalizationRegistry` active and the compile cache populated, the second `prove()` call hangs indefinitely. The first `prove()` succeeds. Fresh compilation (cache miss) does not exhibit the hang. Replacing `FinalizationRegistry` with a no-op before o1js loads avoids it.

## Environment

- o1js branch: `mesa-srs-fix` (commit e0a1022)
- Runtime: web worker with `SharedArrayBuffer` (COOP/COEP headers)
- Browser: Chrome 130+
- `weak-refs` feature: **enabled** in the wasm-bindgen build (confirms FinalizationRegistry is active)

## Root cause analysis

The web worker's `overrideBindings()` in `web-backend.js` proxies WASM calls to a dedicated worker thread. Return values are wrapped on the main thread via `Class.__wrap(ptr)`, which registers them with wasm-bindgen's per-class `FinalizationRegistry`. There are 114 such registries in `plonk_wasm.js`.

The fundamental problem: **multiple JS wrappers are created for the same underlying WASM pointer**. Each wrapper independently registers with a FinalizationRegistry. When GC collects any one of them, the destructor frees the WASM allocation — even though other wrappers (and the Rust prover) still hold the same pointer.

This manifests specifically with the compile cache because `decodeProverKey` calls WASM decode functions (e.g. `caml_pasta_fp_plonk_index_decode`) which go through `overrideBindings`, producing registered wrappers. Fresh compilation keeps everything on the WASM/OCaml side where JS GC can't reach it.

Additionally, the worker calls `__destroy_into_raw()` before sending pointers back to the main thread — transferring ownership. For singleton resources like the SRS, this means the worker gives up ownership each time, but the same allocation is returned on subsequent calls. The main thread creates a new wrapper each time, and when any previous wrapper is GC'd, the underlying singleton is freed.

## Reproduction

1. Load app in web worker with COOP/COEP (FinalizationRegistry shim disabled)
2. `compile()` with warm cache (prover keys loaded via `decodeProverKey`)
3. `prove()` #1 succeeds
4. `prove()` #2 hangs indefinitely

Without cache (fresh compile), both proves succeed.

## Working workaround

Replace `FinalizationRegistry` with a no-op before o1js loads:

```ts
[...]

PR #2897, opened 2026-07-21
fix repeated web-worker proof hangs after finalization

Trivo25:
- Add explicit borrow/move/transfer semantics for wasm
- Replace unsafe pointer polling with atomic SharedArrayBuffer RPC
- Propagate worker errors to the main thread
- Add ABI ownership tests and a two-proof Chromium finalization regression

closes https://github.com/o1-labs/o1js/issues/2870


src/bindings/js/web/worker-rpc.js
```diff
@@ -0,0 +1,229 @@
...
+  Atomics.notify(header, 0);
+}
+
+function waitForWorkerRpcResult(control) {
+  let header = new Int32Array(control, 0, 4);
+  let pause = typeof Atomics.pause === 'function' ? Atomics.pause : () => {};
+  let state = Atomics.load(header, 0);
+  while (state === 0 || state === 1) {
+    pause();
+    state = Atomics.load(header, 0);
```
querolita, line 96: Could it be the case that the worker ever crashed for some reason after being ready and that froze the main flow because nothing else would change the state value afterwards?

querolita (approved):

src/bindings/js/web/worker-rpc.js
Trivo25: are you asking if this _Was_ the case or if this _could be the case in the future_?

Trivo25: The hang fixed in this PR is from bad wasm ownership/finalizers not from missing crash signals

PR #2912, opened 2026-09-08
[ci-only] o1js against mina 01dfe66 -- do not merge

martyall:
**This is a CI probe, not a change to o1js.** It exists only to run o1js CI
against a specific revision of Mina. It is never merged.

The diff is the `src/mina` submodule pointer plus the bullseye apt snapshot
pin for the native build container.

| | |
|---|---|
| Mina commit | `01dfe66171e878f63eb486749e33243b54e61a6f` |
| Mina branch | `martyall/proof-serialization-in-place`, rebased onto `compatible` (MinaProtocol/mina#19373), plus the duplicate-safe nix squash fix 228c409a16 that `main`'s pin already carries |
| proof-systems, inherited through it | `5464fd427bc682287d3ac87f6e3af1954788e37f` (`martyall/fix-3627-compatible`: the three o1-labs/proof-systems#3628 commits replayed onto compatible's pin ab84160fa2) |
| previous `src/mina` pointer on `main` | `9f758cb41cb87fc2719d9b19499ac1d6c4597832` |

proof-systems is nested inside Mina and is not touched directly; moving the
`src/mina` pointer pins it too. Workflows check out `submodules: recursive`,
so CI builds and tests against exactly the revisions above.

https://claude.ai/code/session_016yA7z2fNMo7hfhtuwYWM4R


glyh:
Two things about this probe, from reviewing the Mina side ([MinaProtocol/mina#19373](https://github.com/MinaProtocol/mina/pull/19373)).

**It's pinned to a revision that predates the rename, so it isn't testing the current glue.** The pins here are Mina `e5c5c9c` and proof-systems `de6f1b5f`. Since then the stubs were renamed: `caml_pasta_{fp,fq}_to_bytes_into` / `of_bytes_from` and `caml_bigint_256_*` became `*_blit_to_bigstring` and `*_of_bigstring`, across the Rust, the generated `pasta_bindings.ml`, and the JS in `js/bindings/{field,bigint256}.js`. The current heads are Mina `a87d777` and proof-systems `78a339f`. As pinned, this probe exercises primitive names that no longer exist on either branch.

**It also fails before reaching any of that.** `Prepare (Wasm Backend)` and `upload bindings artifact` both die in the same derivation:

```
source root is source/lib/crypto/proof-systems/poseidon/export_test_vectors
Running phase: patchPhase
substituteStream() in derivation export_test_vectors: ERROR: pattern o1-utils.workspace\ =\ true doesn't match anything in file 'Cargo.toml'
```

o1js's expression patches `poseidon/export_test_vectors/Cargo.toml` expecting `o1-utils.workspace = true`, and that line isn't in the newer proof-systems tree. Worth being clear that this is unrelated to the serialization change: the Mina PR moves the proof-systems submodule forward by 131 commits, of which only the last one is the in-place codec, and the other 130 are upstream `master` (PRs #3569 through #3608). This is collateral from that bump, so re-pointing the probe at the current heads will hit the same wall until either the bump is trimmed to the codec commit alone or o1js's expression is updated for the newer `Cargo.toml`.

Flagging because the JS glue is the one part of that PR with no test coverage anywhere, and this probe is what was supposed to cover it. Happy to help with either fix.

martyall:
Both fixed in c44ff8c1e.

**Pin.** `src/mina` now points at `245589b964`, the current head of `martyall/proof-serialization-in-place`, which carries proof-systems `78a339ff3a` through it. That head also includes the glue change you asked for on MinaProtocol/mina#19373: the six functions now go through `caml_bigstring_blit_bytes_to_ba` / `caml_bigstring_blit_ba_to_bytes` instead of the per-byte loops. I checked the jsoo 6.2.0 runtime we build with and your reading of `caml_bytes_unsafe_set` is right: on a fresh `caml_create_bytes(32)` it takes the append branch for every index in order, so the old reader was 32 string concatenations per element.

**Build.** `flake.nix` now uses `substituteInPlace --replace-quiet` for the `export_test_vectors` patch. The newer proof-systems `Cargo.toml` already has `o1-utils = { workspace = true, features = ["std"] }`, so the substitution is a no-op there and still applies on the older pin used by `main`. That is the smaller of the two fixes; trimming the submodule bump to the codec commit alone is still the right thing for the Mina PR and is tracked there.

**Coverage.** The Mina head also adds a round-trip through `blit_to_bigstring` / `of_bigstring` to the jsoo bindings test (`kimchi_bindings/js/test/bindings_js_test.ml`) for `Bigint_256`, `Fp` and `Fq`, at a nonzero offset into a sentinel-filled bigstring. Mina CI does not run that nodejs target, so this probe's `Prepare (Wasm Backend)` and `Prepare (Native Backend)` are the first build of the current glue anywhere. They are running now.

One caveat on what this probe can show: the o1js perf regression suite times compile, prove and verify at a 20% tolerance and none of its examples serialize proofs on the timed path, so it will not see a serialization-level change in either direction. Parity with `develop` is by construction (same two primitives `Of_binable` compiled to), not measured here.

